PR #3, opened 2020-04-20
Added changes to build.rs to allow ENV variables for PG versions

jamessewell:
eg:

```
PG10_INCLUDE_PATH=...
PG11_INCLUDE_PATH=...
PG12_INCLUDE_PATH=...
```

This will overide building from a local GIT checkout.

This is pretty nasty code, it works but the emphasis was on not changing the existing stuff much to get it working, not refactoring nicely.


zombodb:
hey, I'm re-implementing this literally right now.  It's clear this has to be done ASAP.  So I'm closing this and will push out a commit in a little bit

PR #5, opened 2020-04-20
Ability to launch a BGWorker

jamessewell:
This needs documentation, but I ran out of time. I thought I'd send it
over so you can have a look rather than delay till it's done.

A demo is in ./exmaples/bgworkerdemo. For some reason you can only build
it from ./pgx/pgx, not from ./pgx when using --examples)

There are some missing pieces (get_ functions etc..), but it works.


eeeebbbbrrrr:
I suck at git, so ended up making a new PR, that's been merged with HEAD, and has some fixes for pg10/11/12 support:  https://github.com/zombodb/pgx/pull/26

closing this one.

PR #32, opened 2020-08-06
Fix to stop ctrl-c passing through cargo pgx run

jamessewell:
Pre-fix this would start a Postgres shutdown - potentially a bug with
the way pg_ctl does daemonization of postgres?

This uses some unsafe code - so should be a workaround only.

### DIFF
--- a/cargo-pgx/Cargo.toml
+++ b/cargo-pgx/Cargo.toml
@@ -14,7 +14,6 @@ readme = "README.md"
 clap = { version = "2.33.1", features = [ "yaml" ] }
 colored = "2.0.0"
 env_proxy = "0.4.1"
-exec = "0.3.1"
 num_cpus = "1.13.0"
 pgx-utils = { path = "../pgx-utils", version = "^0.0.12"}
 proc-macro2 = { version = "1.0.19", features = [ "span-locations" ] }
--- a/cargo-pgx/Cargo.toml
+++ b/cargo-pgx/Cargo.toml
@@ -24,3 +24,4 @@ regex = "1.3.9"
 rttp_client = { version = "0.1.0", features = ["tls-native"] }
 syn = { version = "1.0.36", features = [ "extra-traits", "full", "fold", "parsing" ] }
 unescape = "0.1.0"
+fork = "0.1.17"
--- a/cargo-pgx/src/commands/run.rs
+++ b/cargo-pgx/src/commands/run.rs
@@ -6,6 +6,8 @@ use crate::commands::start::start_postgres;
 use crate::commands::stop::stop_postgres;
 use colored::Colorize;
 use pgx_utils::{createdb, get_pg_config, get_psql_path, BASE_POSTGRES_PORT_NO};
+use std::process::Command;
+use std::os::unix::process::CommandExt;
 
 pub(crate) fn run_psql(major_version: u16, dbname: &str, is_release: bool) {
     let pg_config = get_pg_config(major_version);
@@ -39,7 +41,7 @@ pub(crate) fn run_psql(major_version: u16, dbname: &str, is_release: bool) {
 }
 
 fn exec_psql(major_version: u16, dbname: &str) {
-    let mut command = exec::Command::new(get_psql_path(major_version));
+    let mut command = Command::new(get_psql_path(major_version));
     command
         .arg("-h")
         .arg("localhost")
--- a/cargo-pgx/src/commands/start.rs
+++ b/cargo-pgx/src/commands/start.rs
@@ -32,9 +32,11 @@ pub(crate) fn start_postgres(major_version: u16) {
         port.to_string().bold().cyan()
     );
     let mut command = std::process::Command::new(format!("{}/pg_ctl", bindir.display()));
-    // This is unsafe to stop ctrl-c being passed through from psql when using cargo pgx run
-    // without the pre_exec call this would terminate PostgreSQL. This is a workaround until
-    // a better option can be found
+    // Unsafe block is for the pre_exec setsid call below
+    //
+    // This is to work around a bug in PG10 + PG11 which don't call setsid in pg_ctl
+    // This means that when cargo pgx run dumps a user into psql, pushing ctrl-c will abort
+    // the postgres server started by pgx
     unsafe {
         command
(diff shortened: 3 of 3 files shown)


zombodb:
I don't quite understand what this is trying to fix?

jamessewell:
It actually happened to you in one of your demos - if you run cargo pgx run
and then hit ctrl-c from the psql prompt that signal ends up being sent to
the postgres server (which was started by pg_ctl and setsid’ed but not
detached from your tty).

The does a fast shutdown of Postgres.

eeeebbbbrrrr:
So I fixed that already by using execvp() to spawn psql:  https://github.com/zombodb/pgx/commit/d717e5fb4ba4964e0dd8838bac1e567d2ba5bc89

Are you not current with git HEAD?

jamessewell:
I am, SIGTERM still filters through to the postmaster though.

eeeebbbbrrrr:
I hate to be pedantic, but I need a recreate for this. Like what keys are you pressing and when?

jamessewell:
Ctrl-c from psql which has been dropped into from cargo pgx run

eeeebbbbrrrr:
Yeah. I fixed that.  If it’s not fixed for you then we need to understand why. I use MacOS

jamessewell:
I’m on nixos and centos. Can you try on your Ubuntu box?

The issue is actually in pg_ctl, it leaves the postmaster attached to a
tty. I’m working on a patch for that also.

It’s interesting that the new execvp call makes a difference on MacOS, it
must be breaking that link (incorrectly I think).

eeeebbbbrrrr:
I honestly don’t understand what you’re getting at.

eeeebbbbrrrr:
I honestly don’t understand what you’re getting at.

jamessewell:
This is what I'm seeing (the shutdown was at the time I pressed ctrl-c)

```
[nix-shell:~/rust/pgxsharedmem]$ cargo pgx run pg11
building extension with features `pg11`
"cargo" "build" "--features" "pg11" "--no-default-features"
   Compiling pgxsharedmem v0.0.0 (/data/psql/rust/pgxsharedmem)
    Finished dev [unoptimized + debuginfo] target(s) in 3.10s

installing extension
     Copying control file to `/data/pgx/11.8/pgx-install/share/postgresql/extension/pgxsharedmem.control`
     Copying shared library to `/data/pgx/11.8/pgx-install/lib/postgresql/pgxsharedmem.so`
     Writing extension schema to `/data/pgx/11.8/pgx-install/share/postgresql/extension/pgxsharedmem--1.0.sql`
    Finished installing pgxsharedmem
    Starting Postgres v11 on port 28811
    Re-using existing database pgxsharedmem
psql (11.8)
Type "help" for help.

pgxsharedmem=# select now();
              now
-------------------------------
 2020-08-07 00:55:36.795683+00
(1 row)

pgxsharedmem=# ^C
pgxsharedmem=# select 1;
FATAL:  terminating connection due to administrator command
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
!>
```

```
2020-08-07 00:55:12.671 UTC [31540] LOG:  listening on IPv6 address "::1", port 28811
2020-08-07 00:55:12.672 UTC [31540] LOG:  listening on IPv4 address "127.0.0.1", port 28811
2020-08-07 00:55:12.674 UTC [31540] LOG:  listening on Unix socket "/data/pgx/.s.PGSQL.28811"
2020-08-07 00:55:12.700 UTC [31541] LOG:  database system was shut down at 2020-08-07 00:54:12 UTC
2020-08-07 00:55:12.706 UTC [31540] LOG:  database system is ready to accept connections
2020-08-07 00:55:45.827 UTC [31540] LOG:  received fast shutdown request
2020-08-07 00:55:45.829 UTC [31540] LOG:  aborting any active transactions
2020-08-07 00:55:45.829 UTC [31552] FATAL:  terminating connection due to administrator command
2020-08-07 00:55:45.830 UTC [31540] LOG:  background worker "logical replication launcher" (PID 31547) exited with exit code 1
2020-08-07 00:55:45.831 UTC [31542] LOG:  shutting down
2020-08-07 00:55:45.845 UTC [31540] LOG:  database system is shut down
```

jamessewell:
OK so it effects PG10 + PG11 (but not 12)

Relevant fix in PG land:

```
commit bb24439cefead34f195c78f400982f3834593df6
Author: Heikki Linnakangas <heikki.linnakangas@iki.fi>
Date:   Mon Jan 14 14:50:58 2019 +0200

    Detach postmaster process from pg_ctl's session at server startup.

    pg_ctl is supposed to daemonize the postmaster process, so that it's not
    affected by signals to the launching process group.  Before this patch, if
    you had a shell script that used "pg_ctl start", and you interrupted the
    shell script after postmaster had been launched, postmaster was also
    killed.  To fix, call setsid() after forking the postmaster process.

    Long time ago, we had a 'silent_mode' option, which daemonized the
    postmaster process by calling setsid(), but that was removed back in 2011
    (commit f7ea6beaf4).  We discussed bringing that back in some form, but
    pg_ctl is the documented way of launching postmaster to the background, so
    putting the setsid() call in pg_ctl itself seems appropriate.

    Just putting postmaster in a separate session would change the behavior
    when you interrupt "pg_ctl -w start", e.g. with CTRL-C, while it's waiting
    for postmaster to start.  The historical behavior has been that
    interrupting pg_ctl aborts the server launch, which is handy if the server
    is stuck in recovery, for example, and won't fully start up.  To keep that
    behavior, install a signal handler in pg_ctl, to explicitly kill
    postmaster, if pg_ctl is interrupted while it's waiting for the server to
    start up.  This isn't 100% watertight, there is a small window after
    forking the postmaster process, where the signal handler doesn't know the
    postmaster's PID yet, but seems good enough.

    Arguably this is a long-standing bug, but I refrained from back-batching,
    out of fear of breaking someone's scripts that depended on the old
    behavior.
```

eeeebbbbrrrr:
LGTM. Feel free to merge. I think you have the commit bit 

And thanks for going through this.

eeeebbbbrrrr:
You clicked that merge button like a pro!  ❤️

PR #33, opened 2020-08-07
Remove exec crate

jamessewell:
exec crate is deprecated (although the guy hasn't mentioned that in the
README, see https://github.com/faradayio/exec-rs/pull/4)

We can do this with stdlib for POSIX systems now (and Windows never
worked anyway)


jamessewell:
Work merged into https://github.com/zombodb/pgx/pull/32

PR #34, opened 2020-08-12
PgLwLock implementation

jamessewell:
This is the start of the shmem story


pgx/src/shmem.rs
```diff
@@ -0,0 +1,124 @@
...
+        unsafe { (*self.inner.get()).as_ref().unwrap().share() }
+    }
+
+    pub fn exclusive(&self) -> PgLwLockExclusiveGuard<T> {
+        unsafe { (*self.inner.get()).as_mut().unwrap().exclusive() }
+    }
+
+    // To use a lock it has to be attached to an allocated Postgres LWLock, which
+    // we look up by name
+    pub fn attach(&self, lock: String, value: T) -> Result<(), T> {
```
eeeebbbbrrrr, line 26: why not just make this the `new` constructor instead?

eeeebbbbrrrr:
I'd like to see doc comments for all the `pub` stuff here.  Even if just copy-pasta from the Postgres sources

eeeebbbbrrrr:
also, a few `#[pg_test]` functions in the `./pgx/pgx-tests/src/tests/` directory would be nice too.  Even if they're just ensuring the various `PgLwLock` methods don't segfault, that'd be better than nothing.  I don't think the pg_test framework is capable enough to do cross-process testing, but maybe you can use the `pg_locks` view to see if a LWLock has been acquired (do they show up there?)?

pgx/src/shmem.rs
jamessewell: Generally (at least my usage so far) the lock gets created (but not attached) as a global, then attached later on.

let me think about this a bit more

pgx/src/shmem.rs
```diff
@@ -0,0 +1,159 @@
...
+/// This lock can not be poisoned from Rust. Panic and Abort are handled by
+/// PostgreSQL cleanly.
+
+pub struct PgLwLock<T> {
+    inner: UnsafeCell<Option<PgLwLockInner<T>>>,
+}
+
+impl<T> PgLwLock<T> {
+    /// Create a new lock for T by attaching a LWLock, which is looked up by name
+    pub fn new(lock: String, value: T) -> Self {
```
zombodb, line 29: Any particular reason why the lock name throughout is a `String` and not a `&str`?  

Maybe it should even be `&'static str`?

jamessewell:
Look it should probably be a cstr - then that removes an Err case

eeeebbbbrrrr:
That make sense if we want to expose CStr to users?  I guess I lean towards &’static str tho. 

Is it expected (or desired) for lock names to be dynamically generated or are they typically just static values anyways?

jamessewell:
Typically static

zombodb:
This is really great. Awesome work. I’ll get it merged soon. Thinking about what some docs should say

eeeebbbbrrrr:
wow, what just happened?  I merged this.  That's what!

Issue #16, opened 2020-07-16
I am having trouble compiling the hello example


pbrumm:
I am using 20.04 ubuntu with stock rustc.    I have set my path for cargo and have pg_config in path.  

```
[12.3] [stdout] make[4]: Leaving directory '/home/pbrumm/pg_bitquery/target/postgresql-12.3/src/common'
[12.3] [stdout] make[3]: Leaving directory '/home/pbrumm/pg_bitquery/target/postgresql-12.3/src/interfaces/libpq'
[12.3] [stdout] make[2]: Leaving directory '/home/pbrumm/pg_bitquery/target/postgresql-12.3/src/test/isolation'
[12.3] [stdout] make[1]: Leaving directory '/home/pbrumm/pg_bitquery/target/postgresql-12.3/src'
[12.3] [stderr] + cd /home/pbrumm/pg_bitquery/target/postgresql-12.3
[12.3] [stderr] + '[' '!' -f config.status ']'
[12.3] [stderr] + ./configure --prefix=/home/pbrumm/pg_bitquery/target/postgresql-12.3/pgx-install --without-readline --without-zlib -with-pgport=8812
[12.3] [stderr] + '[' '!' -f /home/pbrumm/pg_bitquery/target/postgresql-12.3/pgx-install/bin/postgres ']'
[12.3] [stderr] + make -j 5
[12.3] [stderr] gcc: fatal error: cannot read spec file ‘./specs’: Is a directory
[12.3] [stderr] compilation terminated.
[12.3] [stderr] make[2]: *** [<builtin>: specparse.o] Error 1
[12.3] [stderr] make[2]: *** Waiting for unfinished jobs....
[12.3] [stderr] gcc: fatal error: cannot read spec file ‘./specs’: Is a directory
[12.3] [stderr] compilation terminated.
[12.3] [stderr] make[2]: *** [<builtin>: isolationtester.o] Error 1
[12.3] [stderr] gcc: fatal error: cannot read spec file ‘./specs’: Is a directory
[12.3] [stderr] compilation terminated.
[12.3] [stderr] make[2]: *** [<builtin>: isolation_main.o] Error 1
[12.3] [stderr] make[1]: *** [Makefile:42: all-test/isolation-recurse] Error 2
[12.3] [stderr] make: *** [GNUmakefile:11: all-src-recurse] Error 2
[12.3] [stderr] + exit 1
[12.3] /----------------------------------------

[...]

PR #43, opened 2020-09-07
security: only print friendly env vars

ethanpailes:
People often store secrets in the environment, and likely would
prefer to avoid exposing them in CI logs and the like.

This PR is just meant as a proposal. I opened a PR rather than an
issue because the diff is so tiny, and I would have included it in the
issue description anyway. I could see the pgx project taking the stance
that people should implement some sort of secret santization for any
logs, so this is solving the issue in the wrong place. Personally, I think
a little defense in depth is worthwhile.

### DIFF
--- a/pgx-pg-sys/build.rs
+++ b/pgx-pg-sys/build.rs
@@ -47,9 +47,9 @@ impl bindgen::callbacks::ParseCallbacks for IgnoredMacros {
 }
 
 fn main() -> Result<(), std::io::Error> {
-    // dump parts of the environment
-    for (k, v) in std::env::vars() {
-        if k.starts_with("CARGO") {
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr:
I'm not really down with this.  For example, we would have had a really hard time resolving issue #16 without this.

Perhaps if you want to put that behind some other envvar, then maybe that'd be okay.  Somehting like

```rust
if std::env::var("PGX_BUILD_VERBOSE").or_default() == "true" {
   ...
}
```

When something goes wrong during build.rs, my experience is that it's rarely related to a `CARGO_*` environment variable.  It's always something funky in the user shell.

ethanpailes:
Fair enough, I've updated the patch.

PR #48, opened 2020-10-03
run bindgen on parser.h

ethanpailes:
This patch provides access to to the routines defined in
parser.h via bindgen. I'm interested in the parser.h header
because I want to be able to invoke raw_parser to get a
parse tree that I can poke at.


eeeebbbbrrrr:
Please retarget this to the “develop” branch.

ethanpailes:
@eeeebbbbrrrr you might want to change the default branch so that the PRs that github automatically generates go the right place.

eeeebbbbrrrr:
Doesn't that also change the branch that's shown on the landing page of the repo?  I want that to always be master

ethanpailes:
I would assume so.

PR #49, opened 2020-10-03
run bindgen on parser.h

ethanpailes:
This patch provides access to to the routines defined in
parser.h via bindgen. I'm interested in the parser.h header
because I want to be able to invoke raw_parser to get a
parse tree that I can poke at.


eeeebbbbrrrr:
LGTM.  Merging now.  Will release whenever the next version of pgx is ready.

Thanks!

ethanpailes:
Thanks!

PR #51, opened 2020-10-04
refactor: build.rs - avoid io and lexing

ethanpailes:
This patch started off as an effort to avoid parsing
the same things multiple times, but that turned out to
be a little tricky. Instead we settle for elimitating
the IO associated with the multiple parses by keeping
the intermediary code in memory as `TokenStream2`s.
This also eliminates lexing a second time (though that's
not likely very useful because lexing is usually an order
of magnitude faster than parsing).

The other point of this patch is the creation of the
new `rewrite_items` routine. Right now this routine is
a pretty useless abstraction, but going forward it should
serve as a good place to keep a pipeline of transformations.
I'm interested in this so that I can inject `Display` instances
for all the types which subclass (according to Postgres' internal
object system) `Node`.

Possible issue: When I do a clean build even without my changes
there are a bunch of changes to the `generated-bindings` tree. I think
this is likely due to the fact that I'm running Debian and from what I
can gather from discord chats y'all are using macs? If so, I can finalize
this PR on my mac to see if that will help things.

### DIFF
--- a/pgx-pg-sys/build.rs
+++ b/pgx-pg-sys/build.rs
@@ -10,7 +10,8 @@ use std::collections::HashSet;
 use std::path::PathBuf;
 use std::process::{Command, Output};
 use std::sync::Mutex;
-use syn::export::{ToTokens, TokenStream2};
+use std::error::Error;
+use syn::export::TokenStream2;
 use syn::Item;
 
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr:
> When I do a clean build even without my changes there are a bunch of changes to the generated-bindings tree.

`build.rs` doesn't do any kind of change detection to short-circuit doing all the things, so whenever cargo thinks build.rs needs to run, it just does all the things.  I don't believe this is a platform specific problem.

eeeebbbbrrrr:
In general, this lgtm.  merging now.  Excited about where this is going...

PR #55, opened 2020-10-07
Fix Minor Typo

J0:
This isn't intended to be one of those hacktoberfest spam PRs -- someone just noticed it after reviewing my PR in which I used `pgx`  and I just decided I would fix it.


zombodb:
Thanks!  Typeos are important too!

PR #56, opened 2020-10-21
Add skipgit option to pgx new command.

rustprooflabs:
The project I have already started converting to pgx and the next one on my list both are already tracked in a git repo.  It would be helpful to have the ability to skip the automatic creation of a git repo in the `pgx new` command.

The original functionality to create a git repo and add the files automatically is retained.  Adding the new `skipgit` option does not create the git repo and does not use `git add`.  It does create the `.gitignore` file as that's still a helpful component in the use case of an existing repo.

```
cargo run pgx new test2 skipgit
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `/home/ryanlambert/git/pgx/target/debug/cargo-pgx pgx new test2 skipgit`
```

Creates the following structure:

```
$ ls -alh test2
total 32K
drwxr-xr-x 5 ryanlambert ryanlambert 4.0K Oct 21 14:52 .
drwxr-xr-x 5 ryanlambert ryanlambert 4.0K Oct 21 14:52 ..
drwxr-xr-x 2 ryanlambert ryanlambert 4.0K Oct 21 14:52 .cargo
-rw-r--r-- 1 ryanlambert ryanlambert  401 Oct 21 14:52 Cargo.toml
-rw-r--r-- 1 ryanlambert ryanlambert   73 Oct 21 14:52 .gitignore
drwxr-xr-x 2 ryanlambert ryanlambert 4.0K Oct 21 14:52 sql
drwxr-xr-x 2 ryanlambert ryanlambert 4.0K Oct 21 14:52 src
-rw-r--r-- 1 ryanlambert ryanlambert  131 Oct 21 14:52 test2.control
```

### DIFF
--- a/cargo-pgx/src/cli.yml
+++ b/cargo-pgx/src/cli.yml
@@ -64,8 +64,6 @@ subcommands:
                     short: bgw
                     long: bgworker
                     help: create a background worker template
-                - skipgit:
-                    help: skips git init for existing projects (default is to run git init)
           - install:
               about: install the extension from the current crate to the Postgres specified by whatever "pg_config" is currently on your $PATH
               args:
--- a/cargo-pgx/src/cli.yml
+++ b/cargo-pgx/src/cli.yml
@@ -64,6 +64,8 @@ subcommands:
                     short: bgw
                     long: bgworker
                     help: create a background worker template
+                - skipgit:
+                    help: skips git init for existing projects (default is to run git init)
           - install:
               about: install the extension from the current crate to the Postgres specified by whatever "pg_config" is currently on your $PATH
               args:
--- a/cargo-pgx/src/commands/new.rs
+++ b/cargo-pgx/src/commands/new.rs
@@ -3,13 +3,11 @@
 
 use std::io::Write;
 use std::path::PathBuf;
-use std::process::Command;
 
 pub(crate) fn create_crate_template(
     path: PathBuf,
     name: &str,
     is_bgworker: bool,
-    skip_git: bool,
 ) -> Result<(), std::io::Error> {
     create_directory_structure(&path)?;
     create_control_file(&path, name)?;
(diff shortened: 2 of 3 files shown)


zombodb:
Can we also have it error and exit if it detects an existing git repo and skipgit isn’t specified?  As it stands, it’s kinda destructive right now.

rustprooflabs:
Good idea. I'll look at how to approach that later tonight or tomorrow.

zombodb:
Probably have git_init() look for a .git directory and then just do whatever cargo-pgx does to exit with an error.

eeeebbbbrrrr:
hey, could you re-target this PR against the `develop` branch, please?  I want the master branch to represent the currently-released version.  thanks!

ethanpailes:
For what it's worth I find it surprising when a project template generator makes a specific choice about the version control system to use, so I think it might be best to make this option turned on by default or even rip out the the code to create a git repo entirely. I think most people up to writing a postgres extension know how to `git init`.

eeeebbbbrrrr:
I can be convinced of that, but I did this so that a proper `.gitignore` is created.  All the `*.generated.sql` files don't need to be under source control.

Guess we could still just generate that file and like you said, remove all the "git init" stuff entirely.

ethanpailes:
> I can be convinced of that, but I did this so that a proper .gitignore is created. All the *.generated.sql files don't need to be under source control.

Oh yeah, generating a `.gitignore` is a great idea.

I try to maintain good `.gitignore` files even in projects using other vcs so tools like ripgrep and `fd` can pick them up.

eeeebbbbrrrr:
Then lets just do that.  I'm fine with it.  @rustprooflabs, are you?  Would solve a lot of drama and confusion here.

rustprooflabs:
Sounds good to me!

rustprooflabs:
The git init and git add steps have been removed entirely.

eeeebbbbrrrr:
awesome, thanks.  I'll update `cargo-pgx/README.md` to indicate that it doesn't do this anymore.

PR #65, opened 2020-11-27
Trigger example

eeeebbbbrrrr:
a little example on how to create a trigger function.  needs a readme, and probably an example of changing a returned heap tuple value


Hoverbear:
Rebasing this to merge.

Hoverbear (approved):

PR #70, opened 2020-12-31
provide a bit more detail in getting started guide + refine requirements

krilor:
Thanks for making pgx!
I was following the get started guide today, and updated the README as I went though it.

Wrt. system requirements:

- I removed the explicit dep on libreadline and zlib1g as they are part of the postgress build dependencies (now linked).
- I also stumbled upon a build error on pgx-pg-sys similar to [this](https://www.unadulterated-faff.com/articles/2020/02/04/fixing-rust-compilation-issues-caused-by-missing-packages-part-2.html), so I updated the specific dependencies for that as well. Was thinking about including some info like this in the [clang-sys crate](https://github.com/KyleMayes/clang-sys#dependencies), but I'm not really sure about the actual dependencies here.


eeeebbbbrrrr:
Thanks for this.  It's much appreciated.  It can be difficult to work out some of the dependencies when you're the one that made the thing in the first place!

merging now.

PR #78, opened 2021-01-21
Use 'cargo metadata' to determine the target directory

JLockerman:
This allows 'cargo pgx' to work within workspaces, and ensures that pgx uses the same target directory as the rest of cargo. It will still obey env overrides such as `CARGO_TARGET_DIR` and _should_ obey `build.target-dir` from `cargo.config`, though I have not tested that. Since `cargo pgx` does not have the `--target-dir` option, this doesn't try to forward on to cargo metadata.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1075,6 +1075,7 @@ dependencies = [
  "regex",
  "serde",
  "serde_derive",
+ "serde_json",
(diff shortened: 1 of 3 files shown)


eeeebbbbrrrr:
Thanks a bunch!

PR #85, opened 2021-01-29
feat: update pg versions to latest

sebastianwebber:
this commit updates the postgres version from 10~13 to the latest
available:

- 13.1
- 12.5
- 11.10
- 10.15

related with https://github.com/zombodb/pgx/pull/83

Signed-off-by: Sebastian Webber <sebastian@swebber.me>

### DIFF
--- a/pgx-utils/src/pg_config.rs
+++ b/pgx-utils/src/pg_config.rs
@@ -279,9 +279,9 @@ impl Pgx {
                 PgConfig {
                     version: Some(PgVersion {
                         major_version: 10,
-                        minor_version: 14,
+                        minor_version: 15,
                         url: Url::parse(
-                            "https://ftp.postgresql.org/pub/source/v10.14/postgresql-10.14.tar.bz2",
+                            "https://ftp.postgresql.org/pub/source/v10.15/postgresql-10.15.tar.bz2",
                         )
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr:
Thanks for doing that!  Once the tests all pass I'll merge this and publish a new release this weekend.

eeeebbbbrrrr:
Ha!  So check this out:  `curl -i https://www.postgresql.org/versions.rss`

Would you want to take a stab at teaching cargo-pgx how to use that and parse out the version numbers?  Then we'd never need to deal with this again.

It might also be cool if the various `cargo-pgx` subcommands could emit a little notice letting you know that there's newer versions available and to re-run `cargo pgx init`.

If you're not interested in doing that, that's totally understandable.  I'll probably do it this weekend if you don't have time to mess with it.

eeeebbbbrrrr:
I've pushed changes to teach `cargo pgx init` how to just automatically discover the latest version numbers for the Postgres versions we support (https://github.com/zombodb/pgx/commit/80609a4088097863db350f85a88e11cdcba97306).

As such, I'm going to close this PR, but nonetheless, thanks for your efforts!  It was enough to inspire me to do more work!

I'll get a new version published shortly.

PR #107, opened 2021-03-18
included access/gin/h & access/gist.h

Horusiath:
This PR includes definitions of [gin.h](https://github.com/postgres/postgres/blob/master/src/include/access/gin.h) and [gist.h](https://github.com/postgres/postgres/blob/master/src/include/access/gist.h), both of which have necessary definitions to start work over supporting custom GiST and GIN indexes.

I've regenerated rust defs from these and they seemed to work, but I didn't include the regenerated files, as I've noticed that they were changing the paths to in source code to relative to my local machine.

Are there any other steps necessary to make this PR complete?

### DIFF
--- a/pgx-pg-sys/include/pg10.h
+++ b/pgx-pg-sys/include/pg10.h
@@ -9,6 +9,8 @@
 
 #include "access/amapi.h"
 #include "access/genam.h"
+#include "access/gin.h"
+#include "access/gist.h"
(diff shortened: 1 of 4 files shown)


eeeebbbbrrrr:
This looks just fine.  It's all you need to add more Postgres headers.

I'll regenerate the rust bindings and push those too.  Those only exist in source control to appease docs.rs

PR #117, opened 2021-04-08
Allow overriding function names and returning schema-qualified type

JLockerman:
This PR contains two commits.

The first one adds the ability to override the SQL name of a function defined using `pg_extern`. Though Postgres allows multiple functions to have the same name, we currently cannot define such function families using pgx because the externally-generated symbol created for the rust function must be globally unique. This commit adds a field to the `#[pg_extern]` macro, `#[pg_extern(name = "foo")]` allowing users to define a different SQL name for a function than the rust name. This enables the creation of multiple functions that have the same SQL name, and different rust names, e.g. if you have pgx functions like

```rust
#[pg_extern(name = "plus2")]
pub fn plus2_int(i: i32) -> i32 { i + 2 }

#[pg_extern(name = "plus2")]
pub fn plus2_bigint(i: i64) -> i32 { i + 2 }
```

both of them will be callable like

```SQL
SELECT plus2(foo)
```

For symmetry, the commit adds a similar override for schema, so `#[pg_extern(schema="foo", name="bar")]` will create a function `foo.bar()`.

The second commit allows functions to return schema-qualified SQL types by using module-qualified rust types. e.g. the function like
```rust
#[pg_extern]
pub fn foo() -> myschema::mytype;
```
will generate SQL like
```SQL
CREATE OR REPLACE FUNCTION foo() RETURNS myschema.mytype ...
```

(the PR contains both commits because the second one depends on the first, I remove the second one and open a new PR later if needed)

### DIFF
--- a/cargo-pgx/src/commands/schema.rs
+++ b/cargo-pgx/src/commands/schema.rs
@@ -390,7 +390,7 @@ fn walk_items(
                     type_name
                 ));
                 operator_sql.push(format!(
-                    "CREATE OPERATOR CLASS {type_name}_hash_ops DEFAULT FOR TYPE {type_name} USING hash FAMILY {type_name}_hash_ops AS 
+                    "CREATE OPERATOR CLASS {type_name}_hash_ops DEFAULT FOR TYPE {type_name} USING hash FAMILY {type_name}_hash_ops AS
                         OPERATOR    1   =  ({type_name}, {type_name}),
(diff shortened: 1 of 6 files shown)


Hoverbear (approved):

eeeebbbbrrrr:
LGTM.

Hoverbear:
Thanks for this PR! :)

PR #118, opened 2021-04-08
Improve handling of feature flags

JLockerman:
This PR enables pgx to use cargo feature flags. It has two components:

1. It adds a `--features` command line option to all `cargo-pgx` subcommands that build code or SQL. This flag is analogous to the corresponding `cargo` flag: the features passed in to this flag are passed to the internal cargo commands when building the library. e.g. `cargo pgx run pg13 --features foo` will run `"cargo" "build" "--no-default-features" "--features" "pg13 foo"`
2. It teaches the script generator to obey a subset of `#[cfg(...)]` directives; it skips any object following a `#[cfg(features="foo")]` where the feature `foo` isn't passed in to the `--features` flag. (it also understands `any(...)` and `all(...)`).

The combination of these two commits allows users to enable and disable SQL functionality using cargo features.


eeeebbbbrrrr:
Can you take a look at the test that failed?

eeeebbbbrrrr:
Thanks @JLockerman!  We'll take care of these, but there's two little changes before we merge:

 - centralize the duplicated code in `main.rs` related to pulling "features" string from the argument set
 - update the `--help` output in the various places in `cargo-pgx`' `README.md`

PR #119, opened 2021-04-16
fix(cargo-pgx): on init, lazy fetch version list

Hoverbear:
While working on https://github.com/Hoverbear/pgx/tree/nix-flake, I determined that `cargo-pgx pgx init` would reach to network even if passed all `pg_config` versions it might want.

This patch defers the network call to fetch versions until it actually needs to be done.

This means a call like 

```bash
cargo-pgx pgx init --pg10 postgresql-10.16/bin/pg_config --pg11 postgresql-11.11/bin/pg_config --pg12 postgresql-12.6/bin/pg_config --pg13 postgresql-13.2/bin/pg_config
```

Won't necessarily reach for network.

### DIFF
--- a/cargo-pgx/src/main.rs
+++ b/cargo-pgx/src/main.rs
@@ -58,20 +58,22 @@ fn do_it() -> std::result::Result<(), std::io::Error> {
                     init_pgx(&Pgx::default(SUPPORTED_MAJOR_VERSIONS)?)
                 } else {
                     // user specified arguments, so we'll only install those versions of Postgres
-                    let default_pgx = Pgx::default(SUPPORTED_MAJOR_VERSIONS)?;
+                    let mut default_pgx = None;
                     let mut pgx = Pgx::new();
 
-                    for pg_config in versions.into_iter().map(|(pgver, pg_config_path)| {
-                        if pg_config_path == "download" {
-                            default_pgx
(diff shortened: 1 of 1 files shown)


cargo-pgx/src/main.rs
```diff
@@ -58,20 +58,22 @@ fn do_it() -> std::result::Result<(), std::io::Error> {
...
-                    let default_pgx = Pgx::default(SUPPORTED_MAJOR_VERSIONS)?;
+                    let mut default_pgx = None;
                     let mut pgx = Pgx::new();
 
-                    for pg_config in versions.into_iter().map(|(pgver, pg_config_path)| {
-                        if pg_config_path == "download" {
-                            default_pgx
+                    for (pgver, pg_config_path) in versions {
+                        let config = if pg_config_path == "download" {
+                            if default_pgx.is_none() {
```
eeeebbbbrrrr, line 66: Wanna do .get_or_insert() here instead?

Hoverbear: I initially did that too! However it ends up leaving the code more complicated, since you need to deal with the error from `Pgx::default(...)?`.

eeeebbbbrrrr: Of course.  LGTM as-is.

PR #123, opened 2021-04-21
feat: `cargo-pgx init` does initdb

Hoverbear:
While working on extensions which include GUC options, particularly mandatory ones (such as those in `plrust`), users will want to edit the contents of their `.pgx/data-$VERSION` after an `initdb` has been run.

This PR teaches `cargo-pgx pgx init` to initialize the database, and safely ensures `cargo-pgx pgx run` still does so.

* Migrate `cargo-pgx::commands::run::initdb` to `cargo-pgx::commands::init::initdb`.
* Call `cargo-pgx::commands::init::iinitdb` in `cargo-pgx pgx init`.
* Include backwards compatibility safety, still call `initdb` in `cargo-pgx pgx run`

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,9 +139,9 @@ Next, `pgx` needs to be initialized.  You only need to do this once.
 $ cargo pgx init
 ```
 
-The `init` command downloads Postgres versions v10, v11, v12, v13 and compiles them to `~/.pgx/`.  These installations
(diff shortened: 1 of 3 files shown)


eeeebbbbrrrr:
This is a good change. Before we merge let’s update cargo-pgx’ README.md to note this behavior.

PR #126, opened 2021-04-23
chore: resolve pre-Rust 2021 warnings

Hoverbear:
Resolve 6,756 warnings related to using `panic!()` on things other than literals.

Some of the errors raised here are not `impl Display`, it may be that we wish to do that.


eeeebbbbrrrr:
LGTM. Thanks. That’s a lot of warnings!

eeeebbbbrrrr:
LGTM. Thanks. That’s a lot of warnings!

PR #127, opened 2021-04-23
feat: add aarch64 support

Hoverbear:
Adds aarch64 support to PGX by using [`std::os::raw::c_char`](https://doc.rust-lang.org/std/os/raw/type.c_char.html) instead of `i8` or `u8`s. 

In theory, this may add support for other architectures as well.


eeeebbbbrrrr:
I like the changes here.  And FWIW it instantly clears up my mental confusion about c_char.

PR #136, opened 2021-05-03
feat: --pg_config in cargo-pgx package/install

Hoverbear:
Allows users to set `--pg_config` during the `cargo-pgx pgx install` and `cargo-pgx pgx package` commands.

Also slips the crate output into the nix checks, so I don't accidentally push bad code as much.


Hoverbear:
I've been testing this with pl/Rust and it seems to be working fine. :)

eeeebbbbrrrr (commented):
LGTM.  Ship it!

PR #30, opened 2020-08-03
Change to move PG socket dir to pgx_home

jamessewell:
Some OSes don't allow normal users to write to /run/ (which some
Postgres distros use for socketdir)

PR #139, opened 2021-05-05
chore: pgx-test create sockets in PGX_HOME

Hoverbear:
Following from #30, we can do this in `cargo-pgx pgx test` too!


eeeebbbbrrrr (approved):
Lgtm

PR #159, opened 2021-06-01
Hide thread detection implementation details in guard.rs

timClicks:
Creates a function that hides info about how the main thread is detected. This simplifies the downstream API (no more clunky calls to `get().is_none()`).

Drive by:

- change the name of the `IS_MAIN_THREAD` global variable to `MAIN_THREAD` because it's not being used in a Boolean context
- whitespace (gotta keep that code dapper)


timClicks:
One thing I'm concerned with is returning a blind Boolean. An approach with greater type safety would be returning a dedicated enum, but that felt like overkill here.

Hoverbear:
@timClicks I like this idea, however the tests failed and the logs are expired. Could you try to clear this up for a review?

Hoverbear:
Merged with develop so we can get new CI logs

pgx-pg-sys/src/submodules/guard.rs
```diff
@@ -67,24 +66,33 @@ fn take_panic_location() -> PanicLocation {
 // via pg_module_magic!() this gets set to Some(()) for the "main" thread, and remains at None
 // for all other threads.
 #[cfg(debug_assertions)]
-thread_local! { pub(crate) static IS_MAIN_THREAD: once_cell::sync::OnceCell<()> = once_cell::sync::OnceCell::new() }
+thread_local! { pub(crate) static MAIN_THREAD: once_cell::sync::OnceCell<()> = once_cell::sync::OnceCell::new() }
+
+/// Return true when called from the same thread that executed [pg_module_magic!()]
```
Hoverbear, line 71:
```suggestion
/// Return true when called from the same thread that executed [`pg_module_magic!()`]
```

timClicks:
Sorry for letting this languish @Hoverbear.  I've lost a lot of the original context, but I can probably pick this PR back up if you would like.

Hoverbear:
@timClicks It's up to you. :) The code doesn't build at all (unclosed delimiters) so I think this was actually a draft anyways.

pgx-pg-sys/src/submodules/guard.rs
timClicks: The square brackets create a link in the rendered doc.

timClicks: Although this doesn't actually work - `pgx::pg_module_magic` isn't in local scope

timClicks:
Closing without merging as the concurrency checks are now failing. Can re-open if I think of a reliable way to implement this.

PR #168, opened 2021-07-20
Don't look up the type oid for types that never use it

JLockerman:
Getting function argument type oids can be surprisingly expensive; one Promscale benchmark measures it as taking over 10% of runtime (@cevian should have more info). Fortunately, most types do not actually make use of this information, and the `getarg_type()` call can be optimized out if this fact is known. This commit adds an opt-in method for declaring that a type does not make use of the `typeoid` in `from_datum()`, and updates `pg_getarg()` to make use of this information.

This commit makes the following changes:
 1. It adds a `const` boolean to `FromDatum` called `NEEDS_TYPID`. This boolean is defaulted to `true`. Types that never use the `typeoid` parameter to `from_datum()` can override this to `false`.
 2. `pg_getarg()` updated to look at `T::NEEDS_TYPID` before calling `get_getarg_type()`. If `T::NEEDS_TYPID == true` it calls `get_getarg_type()` like it currently does for all types; if it is `false` it instead passes `InvalidOid` as the (unused) `typeoid` argument to `from_datum()`.
 3. It edits a number of built in types to make use of this optimization.

Since this feature is both opt-in and backwards compatible, it should not break any downstream users.

### DIFF
--- a/pgx/src/datum/date.rs
+++ b/pgx/src/datum/date.rs
@@ -7,6 +7,7 @@ use std::ops::{Deref, DerefMut};
 #[derive(Debug)]
 pub struct Date(time::Date);
 impl FromDatum for Date {
+    const NEEDS_TYPID: bool = false;
     #[inline]
     unsafe fn from_datum(datum: pg_sys::Datum, is_null: bool, _typoid: u32) -> Option<Date> {
         if is_null {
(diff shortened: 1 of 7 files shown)


Hoverbear (approved):
Thanks for this! :)

PR #169, opened 2021-07-20
Don't run all tests in cargo pgx test

JLockerman:
In a workspace this will run non-extension tests as well which is inconsistent with `cargo test`. The old behavior can be recovered with the new `--workspace` flag.

### DIFF
--- a/cargo-pgx/src/commands/test.rs
+++ b/cargo-pgx/src/commands/test.rs
@@ -9,6 +9,7 @@ use std::process::{Command, Stdio};
 pub fn test_extension(
     pg_config: &PgConfig,
     is_release: bool,
+    test_workspace: bool,
     additional_features: Vec<&str>,
 ) -> Result<(), std::io::Error> {
     let major_version = pg_config.major_version()?;
@@ -23,7 +24,6 @@ pub fn test_extension(
         .stdout(Stdio::inherit())
         .stderr(Stdio::inherit())
         .arg("test")
-        .arg("--all")
         .arg("--features")
         .arg(features)
         .arg("--no-default-features")
@@ -35,6 +35,10 @@ pub fn test_extension(
 
     if is_release {
         command.arg("--release");
+    }
+
+    if test_workspace {
+        command.arg("all");
     }
 
     eprintln!("{:?}", command);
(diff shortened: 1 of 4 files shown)


cargo-pgx/src/commands/test.rs
```diff
@@ -37,6 +37,10 @@ pub fn test_extension(
         command.arg("--release");
     }
 
+    if test_workspace {
+        command.arg("all");
```
Hoverbear, line 41: Based on the help:

```rust
❯ cargo test --help
cargo-test 
Execute all unit and integration tests and build examples of a local package

USAGE:
    cargo test [OPTIONS] [TESTNAME] [-- <args>...]

OPTIONS:
    -q, --quiet                      Display one character per test instead of one line
        --lib                        Test only this package's library unit tests
        --bin <NAME>...              Test only the specified binary
        --bins                       Test all binaries
        --example <NAME>...          Test only the specified example
        --examples                   Test all examples
        --test <NAME>...             Test only the specified test target
        --tests                      Test all tests
        --bench <NAME>...            Test only the specified bench target
        --benches                    Test all benches
        --all-targets                Test all targets
        --doc                        Test only this library's documentation
        --no-run                     Compile, but don't run tests
        --no-fail-fast               Run all tests regardless of failure
    -p, --package <SPEC>...          Package to run tests for
        --all                        Alias for --workspace (deprecated)
```

This should be `--workspace` not `all`.

JLockerman: I... read the `help` message backwards (thought it'd said that `workspace` was deprecated and replaced by `all` 😞)

JLockerman: @Hoverbear fixed. It left the interior cargo command calling `--all` for now, to keep compatibility with old cargo versions; I can change it also if that's what you'd perfer.

Hoverbear: That works for me! Did you try this? It seems `--all` is correct, where you have `all` here. (Check line 27ish to see other examples?)

Hoverbear:
```suggestion
        command.arg("--all");
```

PR #172, opened 2021-08-03
Added support for uuid PostgreSQL type

kpumuk:
This PR brings `uuid` PostgreSQL type to pgx. Internally, it is represented as a 16 bytes buffer in Postgres, and so I tried to mimic it in Rust.

I chose to define an internal `Uuid` type instead of exposing a similar type from `uuid` crate. This and using `[u8; 16]` everywhere should provide a flexible enough interface for anybody who wants to use both libraries together.

### DIFF
--- a/pgx-tests/src/tests/uuid_tests.rs
+++ b/pgx-tests/src/tests/uuid_tests.rs
@@ -1,0 +1,45 @@
+use pgx::*;
+
+pub const TEST_UUID_V4: UuidBytes = [
+    0x12, 0x3e, 0x45, 0x67, 0xe8, 0x9b, 0x12, 0xd3, 0xa4, 0x56, 0x42, 0x66, 0x14, 0x17, 0x40, 0x00,
+];
+
+#[pg_extern]
+fn accept_uuid(uuid: Uuid) -> Uuid {
+    uuid
+}
+
+#[pg_extern]
+fn return_uuid() -> Uuid {
+    Uuid::from_bytes(TEST_UUID_V4)
+}
+
+#[cfg(any(test, feature = "pg_test"))]
+mod tests {
+    #[allow(unused_imports)]
+    use crate as pgx_tests;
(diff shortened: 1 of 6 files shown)


pgx/src/datum/uuid.rs
```diff
@@ -0,0 +1,116 @@
...
+            ))?;
+        }
+
+        let mut bytes = [0; UUID_BYTES_LEN];
+        bytes.copy_from_slice(b);
+        Ok(Uuid::from_bytes(bytes))
+    }
+
+    fn format(&self, f: &mut std::fmt::Formatter<'_>, case: UuidFormatCase) -> std::fmt::Result {
+        let hyphenated = f.sign_minus();
```
Hoverbear, line 72: Such a good trick. :)

pgx-tests/src/tests/uuid_tests.rs
```diff
@@ -0,0 +1,74 @@
...
+    fn test_return_uuid() {
+        let result = Spi::get_one::<bool>(
+            "SELECT return_uuid() = '123e4567-e89b-12d3-a456-426614174000'::uuid;",
+        )
+        .expect("SPI result was null");
+        assert!(result)
+    }
+
+    #[pg_test]
+    fn test_parse_uuid_v4() {
```
Hoverbear, line 69: Think we could add a `#[should_fail]`?

kpumuk: PostgreSQL validates UUID, so it is kind of an impossible situation.

```sql
SELECT 'hello'::uuid;
--- ERROR:  invalid input syntax for type uuid: "hello"
--- LINE 1: SELECT 'hello'::uuid;
```

Hoverbear: Oh, even better! Thanks!

Hoverbear (approved):

PR #177, opened 2021-08-19
Use CASCADE while creating extension

nikkhils:
Extensions might have dependencies on other extensions. If there are
no dependencies even then using "CASCADE" does not produce any side
effects. So add it wholesale without adding too much smarts.


eeeebbbbrrrr:
LGTM. We’ll get it merged soon. Thanks!

eeeebbbbrrrr:
released: https://github.com/zombodb/pgx/releases/tag/v0.1.22

Issue #102, opened 2021-03-03
How to define an aggregate function?


Horusiath:
Hi. Before I begin - thanks for this great project!

I'm trying to build an aggregate function for my custom data type as described in [Postgres documentation](https://www.postgresql.org/docs/10/xaggr.html). However I couldn't find a programmatic way of declaring such function using pgx API. It doesn't seem to be possible from Rust level - only from custom sql load scripts, or did I miss something?

eeeebbbbrrrr:
Hey!  I haven't sat down to implement aggregate functions myself, but it should be doable.  You'd define the various init/collect/finalize functions you need as standard `[pg_extern]` functions, and then use the `extension_sql!` macro to handwrite the `CREATE AGGREGATE` statement.

Also, it's undocumented right now, but pgx has `pgx::pg_func_extra()` which lets you setup the "fn_extra" pointer.  You do need the `pg_sys::FunctionCallInfo` pointer, but that's easy to get -- just declare that as your **last** argument of any `#[pg_extern]` function.

I'd like to see better organization around this as time goes one.  I think pgx can make the story around custom aggregate functions very compelling.

laysakura:
Here shows an example on how to create `MYAVG` aggregate:
https://github.com/laysakura/pgx-kungfu/blob/ebc658b819732fb1ea581a5cd4aecaad474bfbc8/my_extension/src/lib.rs

I just followed the above advise:

> You'd define the various init/collect/finalize functions you need as standard [pg_extern] functions, and then use the extension_sql! macro to handwrite the CREATE AGGREGATE statement.

Hoverbear:
It seems this is possible and we should create an example demonstrating this for PGX proper!

PR #187, opened 2021-09-03
feat: add aggregate example

Hoverbear:
Adds a aggregate example as demonstrated by @laysakura in #102 . Thanks!!!

### DIFF
--- a/pgx-examples/aggregate/.gitignore
+++ b/pgx-examples/aggregate/.gitignore
@@ -1,0 +1,7 @@
+.DS_Store
+.idea/
+/target
+*.iml
+**/*.rs.bk
+Cargo.lock
+sql/*.generated.sql
--- a/pgx-examples/aggregate/aggregate.control
+++ b/pgx-examples/aggregate/aggregate.control
@@ -1,0 +1,5 @@
+comment = 'aggregate:  Created by pgx'
+default_version = '1.0'
+module_pathname = '$libdir/aggregate'
+relocatable = false
+superuser = false
--- a/pgx-examples/aggregate/src/lib.rs
+++ b/pgx-examples/aggregate/src/lib.rs
@@ -1,26 +1,59 @@
 use pgx::*;
+use std::{
+    str::FromStr,
+    ffi::CStr,
+};
 
 pg_module_magic!();
 
 
 use serde::{Deserialize, Serialize};
 
-#[derive(Serialize, Deserialize, PostgresType)]
+
+#[derive(Copy, Clone, PostgresType)]
+#[pgvarlena_inoutfuncs]
 pub struct IntegerAvgState {
     sum: i32,
     n: i32,
 }
+
+impl PgVarlenaInOutFuncs for IntegerAvgState {
+    fn input(input: &CStr) -> PgVarlena<Self> {
+        let mut result = PgVarlena::<Self>::new();
+
+        let mut split = input.to_bytes().split(|b| *b == b',');
+        let sum = split.next().map(|value| 
+            i32::from_str(unsafe { std::str::from_utf8_unchecked(value) }).expect("invalid i32")
+        ).expect("expected sum");
+        let n = split.next().map(|value| 
+            i32::from_str(unsafe { std::str::from_utf8_unchecked(value) }).expect("invalid i32")
+        ).expect("expected n");
+
+        result.sum = sum;
+        result.n = n;
+
+        result
+    }
+    fn output(&self, buffer: &mut StringInfo) {
+        buffer.push_str(&self.sum.to_string());
+        buffer.push(',');
+        buffer.push_str(&self.n.to_string());
+    }
+}
+
+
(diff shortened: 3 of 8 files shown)


Hoverbear:
Based on some feedback from @eeeebbbbrrrr I migrated this to using `PgVarlena<T>`, this should yield better performance.

pgx-examples/aggregate/.gitignore
```diff
@@ -0,0 +1,7 @@
+.DS_Store
+.idea/
+/target
+*.iml
```
timClicks, line 4: out of interest, what's an `.iml` file?

pgx-examples/aggregate/.gitignore
```diff
@@ -0,0 +1,7 @@
+.DS_Store
+.idea/
```
timClicks, line 2:
```suggestion
.idea/
.vscode/
```

pgx-examples/aggregate/aggregate.control
```diff
@@ -0,0 +1,5 @@
+comment = 'aggregate:  Created by pgx'
```
timClicks, line 1: Thought: perhaps we should have a URL that people can visit so they know what pgx is i they encounter it in their logs?

pgx-examples/aggregate/src/lib.rs
```diff
@@ -0,0 +1,122 @@
+use pgx::*;
+use std::{
+    str::FromStr,
+    ffi::CStr,
+};
+
+pg_module_magic!();
+
+
+use serde::{Deserialize, Serialize};
```
timClicks, line 10: Is this where we want this import?

pgx-examples/aggregate/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+    n: i32,
+}
+
+impl PgVarlenaInOutFuncs for IntegerAvgState {
+    fn input(input: &CStr) -> PgVarlena<Self> {
+        let mut result = PgVarlena::<Self>::new();
+
+        let mut split = input.to_bytes().split(|b| *b == b',');
+        let sum = split.next().map(|value| 
+            i32::from_str(unsafe { std::str::from_utf8_unchecked(value) }).expect("invalid i32")
```
timClicks, line 26: In the error message, should we match the Postgres type name?

pgx-examples/aggregate/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+
+        result.sum = sum;
+        result.n = n;
+
+        result
+    }
+    fn output(&self, buffer: &mut StringInfo) {
+        buffer.push_str(&self.sum.to_string());
+        buffer.push(',');
+        buffer.push_str(&self.n.to_string());
```
timClicks, line 40: Personal preference only. Slightly easier to skim read.

```suggestion
        buffer.push_str(&format!("{},{}"), self.sum, self.n);
```

timClicks (approved):
This is really great Ana. I have a couple of questions and a suggestion or two, for my own understanding mostly, but nothing that would block merging. Well done.

pgx-examples/aggregate/.gitignore
Hoverbear: I dunno!

pgx-examples/aggregate/.gitignore
Hoverbear: This is from our standard template, if you'd like to change that, please open an issue.

pgx-examples/aggregate/aggregate.control
Hoverbear: I suggest opening an issue if you feel passionate about this in a general case! I think most users would set this to what their `Cargo.toml` description is!

pgx-examples/aggregate/src/lib.rs
Hoverbear: I keep forgetting we don't run `rustfmt` on CI...

pgx-examples/aggregate/src/lib.rs
Hoverbear: I'm not sure this is important, it's an fairly small example. :)

pgx-examples/aggregate/.gitignore
eeeebbbbrrrr: It’s a CLion project file

pgx-examples/aggregate/src/lib.rs
eeeebbbbrrrr: That would be doing allocations in what might be a hot path for Postgres. The idea of StringBuffer is to save that.

timClicks: One of the reasons I considered using format was that it would only involve 1 allocation rather than 2 (each call to `.to_string()`).

eeeebbbbrrrr: Fair. Push it

Issue #202, opened 2021-09-21
(develop) `cargo pgx new` not generates `pgx-linker-script.sh` in it's template


comdiv:
So after `cargo pgx new my_ext` 
Building falling with linker `/home/comdiv/code/hw/./.cargo/pgx-linker-script.sh` not found

comdiv:
Have added manually from pgx repository itself:
```
#! /usr/bin/env bash
# Auto-generated by pgx. You may edit this, or delete it to have a new one created.

if [[ $CARGO_BIN_NAME == "sql-generator" ]]; then
    gcc -Wl,-undefined,dynamic_lookup,-dynamic-list=$CARGO_MANIFEST_DIR/.cargo/pgx-dynamic-list.txt $@
else
    gcc -Wl,-undefined,dynamic_lookup $@
fi
```

Hoverbear:
Thanks for this! We spoke on Discord and you have it sorted. :)

Hoverbear:
This is fixed in #197 which is merged into `develop`. :)

PR #197, opened 2021-09-20
Enable new SQL generator on Mac

Hoverbear:
We whoops'd in #165 and the PR didn't quite work as expected on Mac!

This PR fixes it by using Mac specific flags as determined by @eeeebbbbrrrr .

PR #205, opened 2021-09-22
(cargo-pgx/new) add required pgx-linker-script.sh, pgx-dynamic-list.sh to .cargo dir (#202)

comdiv:
Have fixed insufficient files for `cargo pgx new` command due to new sql-generator


Hoverbear:
Thanks!!!! We merged #197 that also did this. :) Really appreciate the PR.

Issue #196, opened 2021-09-19
Public API of pgx exposes UB to safe code


dtolnay:
For example, the following safe program leads to pgx performing Undefined Behavior:

```rust
// [dependencies]
// pgx = { version = "0.1.22", features = ["pg13"] }

fn main() {
    pgx::itemptr::item_pointer_is_valid(0xdeadbeefusize as _);
}
```

```console
Segmentation fault (core dumped)
```

In general it is unsound for a safe API to result in UB.

eeeebbbbrrrr:
There's probably a number of issues like this across pgx.  We should definitely declare this functions (and its callers) as `unsafe`.  That particular function is meant to mimic the Postgres `ItemPointerIsValid` macro:

```C
/*
 * ItemPointerIsValid
 *              True iff the disk item pointer is not NULL.
 */
#define ItemPointerIsValid(pointer) \
        ((bool) (PointerIsValid(pointer) && ((pointer)->ip_posid != 0)))
```

Giving it a bad pointer like `0xdeadbeef` isn't a thing we can trap (afaik?), so all I can see is to mark it `unsafe`.

Hoverbear:
Thanks so much for reporting this @dtolnay, I've fixed this specific case and will keep my eyes out for more. :)

PR #210, opened 2021-09-23
chore: reduce UB in public API

Hoverbear:
Resolves the example from #196 . There are likely more though!


Hoverbear:
Hm, this should have been `fix`.

PR #227, opened 2021-09-25
fix: Remove circular dependency from pgx-macro dev-deps

Hoverbear:
In #165 we added some doctests to `pgx-macros`.

It turns out that was quite foolish as they depended on `pgx`, creating a circular dependency which hits https://github.com/rust-lang/cargo/issues/4242, meaning we cannot publish.

This removes that.

I guess we will need to cut a `0.2.0-beta.1` since https://crates.io/crates/pgx-utils/0.2.0-beta.0 was released before this could be caught.

Unfortunately I have not found a way to determine if this will be an issue before publishing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -254,7 +254,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-pgx"
-version = "0.2.0-beta.0"
+version = "0.2.0-beta.1"
 dependencies = [
  "clap",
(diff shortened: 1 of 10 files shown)


Hoverbear:
Including a `0.2.0-beta.1` bump as I expect this to be the only PR of `0.2.0-beta.0` to `0.2.0-beta.1`.

PR #230, opened 2021-09-28
First Class Aggregate Support

### LINKED ISSUE #194: First class aggregate support
Inspired by #102 .

Users wishing to create [aggregate functions](https://www.postgresql.org/docs/current/functions-aggregate.html) with `pgx` can get **most** of the way there without a custom SQL block.

What they do need a SQL block for, is something like:

```sql
CREATE AGGREGATE DEMOAVG (integer) (
    sfunc = integer_avg_state_func,
    stype = IntegerAvgState,
    finalfunc = integer_avg_final_func,
    initcond = '0,0'
);
```

As seen in the [`pgx-examples/aggregate/src/lib.rs`](https://github.com/zombodb/pgx/blob/develop/pgx-examples/aggregate/src/lib.rs#L70-L80) file.

The syntax for [`CREATE AGGREGATE`](https://www.postgresql.org/docs/current/sql-createaggregate.html) notably includes:

* The argument of the aggregate (`integer` here).
    * We can have the use specify this as a Rust type, or string of SQL representing a type.
    * The Postgres docs state to pass `*` here to accept none, we could copy this.
* The `initcond` allows users to set an initial condition (state).
    * We could support via primitive or a `const fn`.
* `sfunc` and `finalfunc` (optional) are references to the related functions.
    * We will want to let users specify Rust functions or a string pointing to a SQL function.
* `PARALLEL = { SAFE | RESTRICTED | UNSAFE }` similar to our extern function support.

There are several other options, and while it is not required that we support all of them, we should consider which others to support, and design the code to be able to adapt to new options easily.

# Syntax

We could support this via a macro in a direct, simple form:

```rust
#[derive(Copy, Clone)]
#[pgvarlena_inoutfuncs]
#[postgres_aggregate(
    sfunc = integer_avg_state_func, // or `"my_sql_state_func"`
    finalfunc = integer_avg_final_func, // or `"my_sql_final_func"`
    initcond = '0,0',
    arg_data_type = i32, // or `"integer"`
)]
pub struct IntegerAvgState {
    sum: i32,
    n: i32,
}
```

Since #165 allows us to include the results of things like `const fn`s or `const`s in our SQL, we could also support like:

```rust
initcond = my_const_fn()
```

Or:

```rust
initcond = MY_CONST_VALUE
```

We could also consider providing a trait, which, if `postgres_aggregate` was used without any options, could look something like this:

```rust
#[derive(Copy, Clone)]
#[pgvarlena_inoutfuncs]
#[postgres_aggregate]
pub struct IntegerAvgState {
    sum: i32,
    n: i32,
}

impl PostgresAggregate for IntegerAvgState {
    const INITIAL_CONDITION: Self = Self { sum: 0, n: 0 };
    type ARG_DATA_TYPE: i32;
    fn state_function(&self, v: Self::ARG_DATA_TYPE) -> Self {
        todo!()
    }
}
```

However the question of how to handle optional values like `finalfunc` gets murky. The macro-attributes are more flexible.

@eeeebbbbrrrr and I discussed using a proc macro on an impl block, like so:

```rust
trait PostgresAggregate where Self: Sized {
    type Arg;
    type Finalize;
    type MovingState;

    fn state(&self, v: Self::Arg) -> Self;

    fn finalize(&self) -> Self::Finalize {
        unimplemented!("pgx stub, define in impls")
    }

    fn combine(&self, other: Self) -> Self {
        unimplemented!("pgx stub, define in impls")
    }
    
    fn serial(&self) -> Vec<u8> {
        unimplemented!("pgx stub, define in impls")
    }

    fn deserial(&self, buf: Vec<u8>, internal: Internal<Self>) -> Internal<Self> {
        unimplemented!("pgx stub, define in impls")
    }

    fn moving_state(mstate: Self::MovingState, v: Self::Arg) -> Self::MovingState {
        unimplemented!("pgx stub, define in impls")
    }

    fn moving_finalize(mstate: Self::MovingState) -> Self::Finalize {
        unimplemented!("pgx stub, define in impls")
    }

}

#[derive(Copy, Clone, PostgresType)]
#[pgvarlena_inoutfuncs]
pub struct IntegerAvgState {
    sum: i32,
    n: i32,
}

#[pg_aggregate(
  parallel = Safe,
  initial_condition = "1,2",
  moving_initial_condition = "1,2",
  sort_operator = "example",
)]
impl PostgresAggregate for IntegerAvgState {
    type Arg = i32;
    type Finalize = i32;
    type MovingState = ();

    fn state(&self, v: Self::Arg) -> Self {
        todo!()
    }

    //#[pgx(finalfunc_modify = READ_ONLY)]
    fn finalize(&self) -> Self::Finalize {
        self.sum / self.n
    }
}
```

This seems like a nice way to "feel" rusty while also being able to tell if the user specified optional things like `finalfunc`.

Since we are using a proc macro here, we can intelligently add items like `finalize` with stubs if they are not present.

Hoverbear:
Closes #194.

Implements Aggregate support in `pgx`.

Creating an aggregate looks roughly like:

```rust

#[derive(Copy, Clone, PostgresType, Serialize, Deserialize)]
#[pgvarlena_inoutfuncs]
pub struct IntegerAvgState {
    sum: i32,
    n: i32,
}

impl PgVarlenaInOutFuncs for IntegerAvgState {
    fn input(input: &CStr) -> PgVarlena<Self> {
        let mut result = PgVarlena::<Self>::new();

        let mut split = input.to_bytes().split(|b| *b == b',');
        let sum = split
            .next()
            .map(|value| {
                i32::from_str(unsafe { std::str::from_utf8_unchecked(value) }).expect("invalid i32")
            })
            .expect("expected sum");
        let n = split
            .next()
            .map(|value| {
                i32::from_str(unsafe { std::str::from_utf8_unchecked(value) }).expect("invalid i32")
            })
            .expect("expected n");

        result.sum = sum;
        result.n = n;

        result
    }
    fn output(&self, buffer: &mut StringInfo) {
        buffer.push_str(&format!("{},{}", self.sum, self.n));
    }
}

#[pg_aggregate]
impl Aggregate for IntegerAvgState {
    type State = PgVarlena<Self>;
    type Args = i32;
    const NAME: &'static str = "DEMOAVG";

    const INITIAL_CONDITION: Option<&'static str> = Some("0,0");

    fn state(mut current: Self::State, arg: Self::Args) -> Self::State {
        current.sum += arg;
        current.n += 1;
        current
    }

    // You can skip all these:
    type Finalize = i32;
    // type OrderBy = i32;
    // type MovingState = i32;

    // const PARALLEL: Option<ParallelOption> = Some(ParallelOption::Safe);
    // const FINALIZE_MODIFY: Option<FinalizeModify> = Some(FinalizeModify::ReadWrite);
    // const MOVING_FINALIZE_MODIFY: Option<FinalizeModify> = Some(FinalizeModify::ReadWrite);

    // const SORT_OPERATOR: Option<&'static str> = Some("sortop");
    // const MOVING_INITIAL_CONDITION: Option<&'static str> = Some("1,1");
    // const HYPOTHETICAL: bool = true;

    // You can skip all these:
    fn finalize(current: Self::State) -> Self::Finalize {
        current.sum / current.n
    }

    // fn combine(current: Self::State, _other: Self::State) -> Self::State {
    //     unimplemented!("pgx stub, define in impls")
    // }

    // fn moving_state(_mstate: Self::MovingState, _v: Self::Args) -> Self::MovingState {
    //     unimplemented!("pgx stub, define in impls")
    // }

    // fn moving_state_inverse(_mstate: Self::MovingState, _v: Self::Args) -> Self::MovingState {
    //     unimplemented!("pgx stub, define in impls")
    // }

    // fn moving_finalize(_mstate: Self::MovingState) -> Self::Finalize {
    //     unimplemented!("pgx stub, define in impls")
    // }
}

impl Default for IntegerAvgState {
    fn default() -> Self {
        Self { sum: 0, n: 0 }
    }
}
```


JamesGuthrie:
I looked at this change again, this time with an aggregate which uses the `serial` and `deserial` functions, and whose `State` is defined as `Option<Self>`. With the following changes, I'm able to get my example to compile:

```diff
diff --git a/pgx-utils/src/sql_entity_graph/pg_aggregate/mod.rs b/pgx-utils/src/sql_entity_graph/pg_aggregate/mod.rs
index 51d1e3bb..6f3fb368 100644
--- a/pgx-utils/src/sql_entity_graph/pg_aggregate/mod.rs
+++ b/pgx-utils/src/sql_entity_graph/pg_aggregate/mod.rs
@@ -271,8 +271,8 @@ impl PgAggregate {
             pg_externs.push(parse_quote! {
                 #[allow(non_snake_case)]
                 #[pg_extern]
-                fn #fn_name(this: #target_path) -> Vec<u8> {
-                    this.serial()
+                fn #fn_name(this: #type_state_without_self) -> Vec<u8> {
+                    <#target_path as pgx::Aggregate>::serial(this)
                 }
             });
             Some(fn_name)
@@ -295,13 +295,13 @@ impl PgAggregate {
                 #[allow(non_snake_case)]
                 #[pg_extern]
                 fn #fn_name(this: #type_state_without_self, buf: Vec<u8>, internal: pgx::PgBox<#type_state_without_self>) -> pgx::PgBox<#type_state_without_self> {
-                    this.deserial(buf, internal)
+                    <#target_path as pgx::Aggregate>::deserial(this, buf, internal)
                 }
             });
             Some(fn_name)
         } else {
             item_impl.items.push(parse_quote! {
-                fn deserial(current: #type_state_without_self, _buf: Vec<u8>, _internal: pgx::PgBox<Self>) -> pgx::PgBox<Self> {
+                fn deserial(current: #type_state_without_self, _buf: Vec<u8>, _internal: pgx::PgBox<#type_state_without_self>) -> pgx::PgBox<#type_state_without_self> {
                     unimplemented!("Call to deserial on an aggregate which does not support it.")
                 }
             });
@@ -778,7 +778,7 @@ mod tests {
                     todo!()
                 }
 
-                fn deserial(current: Self::State, _buf: Vec<u8>, _internal: PgBox<Self>) -> PgBox<Self> {
+                fn deserial(current: Self::State, _buf: Vec<u8>, _internal: PgBox<Self::State>) -> PgBox<Self::State> {
                     todo!()
                 }
 
diff --git a/pgx/src/aggregate.rs b/pgx/src/aggregate.rs
index ebe94de9..42169e1f 100644
--- a/pgx/src/aggregate.rs
+++ b/pgx/src/aggregate.rs
@@ -157,7 +157,7 @@ where
     fn serial(current: Self::State) -> Vec<u8>;
 
     /// **Optional:** This function can be skipped, `#[pg_aggregate]` will create a stub.
-    fn deserial(current: Self::State, _buf: Vec<u8>, _internal: PgBox<Self>) -> PgBox<Self>;
+    fn deserial(current: Self::State, _buf: Vec<u8>, _internal: PgBox<Self::State>) -> PgBox<Self::State>;
 
     /// **Optional:** This function can be skipped, `#[pg_aggregate]` will create a stub.
     fn moving_state(_mstate: Self::MovingState, _v: Self::Args) -> Self::MovingState;
```

But I'm running into the following issue when running the `sql-generator`:

```
   0: Failed to map argument `internal` type `pgx::pgbox::PgBox<core::option::Option<_MY_TYPE_HERE_>>` to SQL type while building function `my_aggregate_deserial`.
```

Hoverbear:
Thanks for catching this @JamesGuthrie !

Hoverbear:
I think that should do it, @JamesGuthrie . Of note I had to create a mapping for `PgBox<Option<T>>` types in the generated code.

JamesGuthrie:
Awesome, that seems to have fixed it for our aggregates! Now all of our aggregates are using this, and tests are passing.

For us the final piece would be handling the allocation in the aggregate's memory context.

Some high-level comments:

- In the end we didn't need to (or couldn't) use `serial` and `deserial`. Serialization functions can only be defined if the state type is `INTERNAL`. I haven't tried defining Self::State as Internal - is it intentional that this should be possible?
- On that note: I'm not sure what the accepted convention is on when to use `INTERNAL` vs a PG type. For aggregates I think the conservative approach is to define most functions as taking `INTERNAL`.
- The args-as-tuple approach does make some things more obscure, for example the signature of our aggregate is now:
    ```SQL
    CREATE OR REPLACE FUNCTION _prom_ext."vector_selector_state"(
        "this" _prom_ext.VectorSelector, /* core::option::Option<promscale::aggregates::vector_selector::VectorSelector> */
        "arg_one" TimestampTz, /* promscale::raw::TimestampTz */
        "arg_two" TimestampTz, /* promscale::raw::TimestampTz */
        "arg_three" bigint, /* i64 */
        "arg_four" bigint, /* i64 */
        "arg_five" TimestampTz, /* promscale::raw::TimestampTz */
        "arg_six" double precision /* f64 */
    ) RETURNS _prom_ext.VectorSelector /* core::option::Option<promscale::aggregates::vector_selector::VectorSelector> */
    LANGUAGE c /* Rust */
    AS 'MODULE_PATHNAME', 'vector_selector_state_wrapper';
    ```
    `arg_one` and `arg_two` are not very descriptive.
- (I'm not sure how relevant this is in the context of aggregates): there's no way to declare the functions as `IMMUTABLE`

Hoverbear:
I am still poking at the memory context stuff. :) It's a new topic for me. We definitely want that.

> The args-as-tuple approach does make some things more obscure, for example the signature of our aggregate is now: 

I agree with this. It's ugly. :( I was thinking about making `name!()` work here, but we'd still need those fallback names... However it's the only way we can really handle variadic args in Rust.

If you look in https://www.postgresql.org/docs/current/sql-createaggregate.html, `argname` is only for documentation purposes, so we could probably even plain omit those names from the generated SQL.

> (I'm not sure how relevant this is in the context of aggregates): there's no way to declare the functions as IMMUTABLE

Oh!!!! You're right! We should add a way to pass those pgextern args in...

JamesGuthrie:
> Oh!!!! You're right! We should add a way to pass those pgextern args in...

For this specific case I'm not sure that it's necessary: https://stackoverflow.com/questions/70784048/how-does-postgres-optimise-immutable-c-lang-aggregate-functions/70784425#70784425

JamesGuthrie:
> I agree with this. It's ugly. :( I was thinking about making `name!()` work here, but we'd still need those fallback names... However it's the only way we can really handle variadic args in Rust. We could maybe omit it: https://www.postgresql.org/docs/current/sql-createfunction.html

I think that `name!()` support would be great! I'm mostly concerned with the generated SQL not being nice to read, so being able to make that nicer through `name!()` would be great.

I've updated all of our aggregates to the latest changes in this branch. With the addition of the `fcinfo` parameter, the generated state transition function now gets caught by `clippy::too_many_arguments`. Perhaps we need this:

```diff
diff --git a/pgx-utils/src/sql_entity_graph/pg_aggregate/mod.rs b/pgx-utils/src/sql_entity_graph/pg_aggregate/mod.rs
index 8b8c0613..a83558b8 100644
--- a/pgx-utils/src/sql_entity_graph/pg_aggregate/mod.rs
+++ b/pgx-utils/src/sql_entity_graph/pg_aggregate/mod.rs
@@ -203,6 +203,7 @@ impl PgAggregate {
 
             pg_externs.push(parse_quote! {
                 #[allow(non_snake_case)]
+                #[allow(clippy::too_many_arguments)]
                 #[pg_extern]
                 fn #fn_name(this: #type_state_without_self, #(#args_with_names),*, fcinfo: pgx::pg_sys::FunctionCallInfo) -> #type_state_without_self {
                     <#target_path as pgx::Aggregate>::in_memory_context(
@@ -345,6 +346,7 @@ impl PgAggregate {
                 .map(|name| Ident::new(name, fn_state.span()));
             pg_externs.push(parse_quote! {
                 #[allow(non_snake_case)]
+                #[allow(clippy::too_many_arguments)]
                 #[pg_extern]
                 fn #fn_name(
                     mstate: <#target_path as pgx::Aggregate>::MovingState,
```

Hoverbear:
And that should make `name!()` work on args.

JamesGuthrie:
> And that should make `name!()` work on args.

🎉  perfect, I'm using that without any issues!

Hoverbear:
Alright, let's merge this, I feel like this is a fairly complete initial implementation.

PR #274, opened 2021-11-01
`PgBox` rewrite

eeeebbbbrrrr:
This PR rewrites the internals of `PgBox` such that it's now the size of a pointer, instead of twice the size of pointer.

It introduces some zero-sized types (ZSTs) to indicate if the pointer owned by a `PgBox` instance was allocated by Rust or by Postgres, which might cause compilation errors for existing users.  Its default understanding is `AllocatedByPostgres`.


pgx/src/pgbox.rs
```diff
@@ -80,46 +82,62 @@ use std::ops::{Deref, DerefMut};
...
-                self.owner_string()
-            )),
-            None => f.write_str(&format!(
-                "PgBox<{}> (ptr=NULL, owner={:?})",
-                std::any::type_name::<T>(),
-                self.owner_string()
-            )),
+impl<T> WhoAllocated<T> for AllocatedByPostgres {
+    /// Doesn't do anything
+    #[inline]
```
timClicks, line 99: Unnecessary, I think

pgx/src/pgbox.rs
```diff
@@ -80,46 +82,62 @@ use std::ops::{Deref, DerefMut};
...
+pub struct PgBox<T, AllocatedBy: WhoAllocated<T> = AllocatedByPostgres> {
+    ptr: Option<NonNull<T>>,
+    __marker: PhantomData<AllocatedBy>,
 }
 
-struct Inner<T> {
-    ptr: Option<*mut T>,
-    allocated_by_pg: bool,
+pub trait WhoAllocated<T> {
+    fn free(ptr: *mut T);
```
timClicks, line 92: Interesting that `free()` doesn't take `self`

pgx/src/pgbox.rs
```diff
@@ -229,54 +254,25 @@ impl<T> PgBox<T> {
...
+            ptr: None,
+            __marker: PhantomData,
         }
     }
 
     /// Are we boxing a NULL?
+    #[inline]
     pub fn is_null(&self) -> bool {
-        self.inner.ptr.is_none()
+        self.ptr.is_none()
```
timClicks, line 268: Should be impossible to have a null pointer within the `Some` that we've created, but ...

```suggestion
        match self.ptr {
            Some(ptr) => ptr.is_null(),
            None => true,
        }
```

timClicks (approved):
Impressive work

pgx-examples/triggers/src/lib.rs
```diff
@@ -69,7 +69,9 @@ mod tests {
 
     #[pg_test]
     fn test_insert() {
-        Spi::run(r#"INSERT INTO test (title, description, payload) VALUES ('a different title', 'a different description', '{"key": "value"}')"#);
+        Spi::run(
+            r#"INSERT INTO test (title, description, payload) VALUES ('a different title', 'a different description', '{"key": "value"}')"#,
```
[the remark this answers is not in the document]
eeeebbbbrrrr: This isn't part of the changes.  It just got picked up by rustfmt.

pgx/src/pgbox.rs
eeeebbbbrrrr: Not exactly:  https://github.com/zombodb/pgx/blob/2b62ea2ee5905ebf1c638beccecda7dbc5ccc1a9/pgx/src/pgbox.rs#L258

pgx/src/pgbox.rs
eeeebbbbrrrr: These are, I guess you'd call them, "associated types"?  So there's no instance of them anywhere to reference as `self`.  This is how I was able to reduce this thing to the size of a pointer.

pgx/src/pgbox.rs
eeeebbbbrrrr: The impl itself?  No, it's necessary as the `Drop` impl does `AllocatedBy::free(...)`:  https://github.com/zombodb/pgx/blob/2b62ea2ee5905ebf1c638beccecda7dbc5ccc1a9/pgx/src/pgbox.rs#L332-L338

timClicks: I was referring to the inline

PR #275, opened 2021-11-02
Internal type support

eeeebbbbrrrr:
This drastically improves the usability of the `Internal` Postgres type, and also gets it working with the new sql generator.


pgx/src/datum/internal.rs
```diff
@@ -1,23 +1,93 @@
...
+        match self.0 {
+            None => None,
+            Some(datum) => (datum.get() as *mut T).as_mut(),
+        }
+    }
+}
+
+impl FromDatum for Internal {
+    #[inline]
+    unsafe fn from_datum(datum: pg_sys::Datum, is_null: bool, _: pg_sys::Oid) -> Option<Internal> {
```
timClicks, line 72: Suggest adding a note that this function can panic

```suggestion
    /// Converts this `Internal` into a `Datum`
    ///
    /// ## Panics
    ///
    /// Will panic if the `is_null` argument is false, but the actual value is `0`.
    #[inline]
    unsafe fn from_datum(datum: pg_sys::Datum, is_null: bool, _: pg_sys::Oid) -> Option<Internal> {
```

pgx/src/datum/sql_entity_graph/extension_sql.rs
```diff
@@ -167,79 +145,34 @@ impl Display for SqlDeclaredEntity {
...
+            option_vec_option: format!("Option<Vec<Option<{}>>", name),
+            array: format!("Array<{}>", name),
+            option_array: format!("Option<{}>", name),
+            varlena: format!("Varlena<{}>", name),
+            pg_box: vec![
+                format!("pgx::pgbox::PgBox<{}>", name),
+                format!("pgx::pgbox::PgBox<{}, pgx::pgbox::AllocatedByRust>", name),
+                format!(
+                    "pgx::pgbox::PgBox<{}, pgx::pgbox::AllocatedByPostgres>",
+                    name
```
timClicks, line 168: Suggest using `std::any::type_name` rather than hard-coding these strings. Will become brittle if things ever are refactored.

timClicks (approved):

pgx/src/datum/sql_entity_graph/extension_sql.rs
eeeebbbbrrrr: we can't here as the name itself is dynamically generated.  :(  But you're right -- we'll get tripped up by this at some point.  We do have tests that use `PgBox` so they'll at least light up when we move things around.

PR #278, opened 2021-11-05
Improve fuzzy matching

JLockerman:
While doing fuzzy matching on an optional `extension_sql!` declared type, the schema generator will look for `core::option::Option<Foo>` while the `SqlDeclared` will store `Option<Foo>`. This commit alters the fuzzy matcher to search for both types so that such types can be used.

Fixes issue https://github.com/zombodb/pgx/issues/277.


JLockerman:
Note that this doesn't handle `Option<PgBox<Foo>>`. A more-principled solution might be to walk the type stripping `Option`s, `PgBox`s, and `Vec`s until the base type is reached and then only comparing that. I didn't attempt this as it's a more invasive change, and I'm not sure how many levels of removal are correct.

PR #279, opened 2021-11-07
jcrist1 generic tuples

jcrist1:
This adds a very generic typed tuple deserialisation to SpiHeapTupleData. This depends on a crate I've created that allows induction of methods on tuples on the length of the tuple called tuple-traits. This is more of a proof of concept, and I'd like to solicit feedback on it. 
I'm a little afraid that this might be inefficient, but optimistic that the compiler should optimise all of the extra stuff out, but I haven't profiled it yet.

It adds a method 
```rust
pub fn get_tuple<(A1, A2, ...., An)>(&self) -> (Option<A1>, ..., Option<An>)
```
to the SpiHeapTupleData struct, that will try to get each type from that position.

I implemented a simple test, which is probably the best way of groking how it works.

I think this would be a cool way to get typed entries from Spi, but definitely need some feedback on this and the new dependency that this adds: [tuple_tricks](https://github.com/jcrist1/tuple_tricks)

### DIFF
--- a/pgx-tests/src/tests/spi_tests.rs
+++ b/pgx-tests/src/tests/spi_tests.rs
@@ -72,6 +72,29 @@ mod tests {
     }
 
     #[pg_test]
+    fn test_get_tuple() {
+        Spi::execute(|client| {
+            let res = client
+                .select(
+                    "SELECT 42::bigint, 'hello', null, 23, 3.141::real",
+                    None,
+                    None,
+                )
+                .next()
+                .expect("Should be at least one entry but none returned!")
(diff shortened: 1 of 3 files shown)


Hoverbear (commented):
This is really cool! Since `pgx` gets published to crates.io, we can't merge `git` deps. If you get around to publishing this crate I'd love to revisit this PR. :)

jcrist1:
Cool! I’ll push the library to crates.io, but close this PR so it doesn’t clog things up on your end, and open a new one when I’ve got this cleaned up.

PR #286, opened 2021-11-10
Add `cargo pgx init --no-restart`

JLockerman:
When installing a pgx extension on an already running system, it can be useful to point `cargo pgx` at that system's `pg_config` in order to ensure that the extension is being compiled against the correct version. Right now `cargo pgx init --pgXX` restarts the database, so the only way to do this is by manually writing the `config.toml`. This commit adds a flag to `cargo pgx init` `--no-restart` that suppresses the default behavior, and allows `cargo pgx` to build against an already running system.

### DIFF
--- a/cargo-pgx/src/cli.yml
+++ b/cargo-pgx/src/cli.yml
@@ -41,6 +41,9 @@ subcommands:
                     takes_value: true
                     required: false
                     help: if installed locally, the path to PG14's 'pg_config' tool, or 'download' to have pgx download/compile/install it
+                - no-restart:
+                    long: no-restart
+                    help: do not stop and initialize the Postgres instance(s) that pgx will point at.
           - start:
               about: start a pgx-managed Postgres instance
               args:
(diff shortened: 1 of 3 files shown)


JLockerman:
Here's a PR based on an interesting workflow I had to run recently. I'm not sure this is the best way to it, but it was simple enough that I figured I'd open a PR to start the conversation.

eeeebbbbrrrr:
You're saying if you do `cargo pgx init --pg12=/path/to/pg12/bin/pg_config` you'd want to avoid (re)starting the database instance already installed?

I'm a little confused as we should only be attempting to stop/start the postmaster instance that pgx manages, running against the data dir in `~/.pgx/`, which I don't understand why that'd be bad.  I suppose this could be bugged and we're accidentally trying the installed postgres?

JLockerman:
I _thought_ that was what was happening, but I could have messed something up in my config, going to need to look at it in more detail...

PR #289, opened 2021-11-10
Add a few more useful headers

JLockerman:
`"utils/datetime.h"` for access to postgres's time parsing functions
`"mb/pg_wchar.h"` for access to functions that allow us to work with non-utf8 database encodings.

I haven't tried to regenerate the cached versions of the pg-sys files as I am not sure how to do that.


eeeebbbbrrrr:
This is included in v0.2.4

PR #314, opened 2021-11-26
nix template tweaks

bbigras:
The template builds if we disable the tests `doCheck = false` in `default.nix`.


bbigras:
firendly ping @eeeebbbbrrrr

PR #333, opened 2021-12-15
Remove old .so before copying new one into place

### LINKED ISSUE #328: Problems running recompiled extension on M1/Monterey combo
I seem to be running into an issue in the M1/Monterey combo which appears to be a bug in macOS.

The short version is: When I change the contents of my extension and recompile, `dylib` refuses to load it and fails with `Code Signature Invalid`. This appears to have something to do with the file being "copied over" from `/target...` to the `.pgx` directory. I have opened an issue on the apple developer forums which goes into a fair amount of detail: https://developer.apple.com/forums/thread/696460

With the following (awfully hacky) patch (which just removes the `.so` before copying it in), I am able to work around the issue that I'm seeing:

```diff
diff --git a/cargo-pgx/src/commands/install.rs b/cargo-pgx/src/commands/install.rs
index c4d0356..fd56b98 100644
--- a/cargo-pgx/src/commands/install.rs
+++ b/cargo-pgx/src/commands/install.rs
@@ -45,6 +45,12 @@ pub(crate) fn install_extension(
         let mut dest = base_directory.clone();
         dest.push(&pkgdir);
         dest.push(format!("{}.so", extname));
+        if dest.exists() {
+            handle_result!(
+                std::fs::remove_file(&dest),
+                format!("unable to remove existing file {}", dest.display())
+            )
+        }
         copy_file(&shlibpath, &dest, "shared library", false);
     }
 
```

@Hoverbear I'm curious to find out if you run into the same issue when you upgrade to Monterey.

JamesGuthrie:
This is a workaround for an issue on new M1 Macs, as outlined in [1].

The basic rundown is: Apple has recently added mandatory signing of
binaries and libraries, with associated automatic signing of
locally-compiled artifacts.

The signature verification uses a static in-kernel inode-based signature
cache. Once cached, the signature associated with an inode is not
updated.

This means that by "copying over" the shared library, the kernel
compares the outdated signature (from its static cache) to the new
shared library contents.

This workaround removes the existing .so, ensuring that a new inode is
used for the newly-compiled extension .so.

Fixes #328

[1]: https://developer.apple.com/documentation/security/updating_mac_software

### DIFF
--- a/cargo-pgx/src/commands/install.rs
+++ b/cargo-pgx/src/commands/install.rs
@@ -45,6 +45,18 @@ pub(crate) fn install_extension(
         let mut dest = base_directory.clone();
         dest.push(&pkgdir);
         dest.push(format!("{}.so", extname));
+
+        if cfg!(target_os = "macos") {
+            // Remove the existing .so if present. This is a workaround for an
+            // issue highlighted by the following apple documentation:
+            // https://developer.apple.com/documentation/security/updating_mac_software
+            if dest.exists() {
+                handle_result!(
+                std::fs::remove_file(&dest),
+                format!("unable to remove existing file {}", dest.display())
+            )
+            }
+        }
         copy_file(&shlibpath, &dest, "shared library", false);
     }
 


Hoverbear (approved):
This seems to resolve the issues I was facing!

cargo-pgx/src/commands/install.rs
```diff
@@ -45,6 +45,18 @@ pub(crate) fn install_extension(
...
         dest.push(&pkgdir);
         dest.push(format!("{}.so", extname));
+
+        if cfg!(target_os = "macos") {
+            // Remove the existing .so if present. This is a workaround for an
+            // issue highlighted by the following apple documentation:
+            // https://developer.apple.com/documentation/security/updating_mac_software
+            if dest.exists() {
+                handle_result!(
+                std::fs::remove_file(&dest),
```
Hoverbear, line 55: Rustfmt will clean this indentation up later.

Issue #340, opened 2021-12-21
Schema generation empty on alpine


JamesGuthrie:
I'm hitting an issue where the generated schema file is empty on alpine linux.

The most minimal case that I have for reproduction at the moment is the following `Dockerfile`

```Dockerfile
FROM timescale/timescaledb:2.5.0-pg14

RUN apk add --no-cache --virtual .build-deps \
        coreutils \
        dpkg-dev dpkg \
        gcc \
        libc-dev \
        make \
        util-linux-dev \
        clang \
        llvm \
        git \
        llvm-dev clang-libs \
        dpkg-dev dpkg \
        gcc \
        libc-dev \
        libedit-dev \
        libxml2-dev \
        libxslt-dev \
        linux-headers \
        clang g++ \
        make \
        openssl-dev \
        util-linux-dev \
        zlib-dev \
        icu-dev \
        rustup

USER postgres

ENV PATH=/usr/local/cargo/bin:$PATH \
    RUST_VERSION=1.57.0 \
    # prevents some issues with openssl
    RUSTFLAGS="-C target-feature=-crt-static"

RUN set -eux; \
    apkArch="$(apk --print-arch)"; \
    case "$apkArch" in \
        x86_64) rustArch='x86_64-unknown-linux-musl'; ;; \
        aarch64) rustArch='aarch64-unknown-linux-musl'; ;; \
        *) echo >&2 "unsupported architecture: $apkArch"; exit 1 ;; \
    esac; \
    rustup-init -y --no-modify-path --profile minimal  --component rustfmt --default-toolchain $RUST_VERSION --default-host ${rustArch};

RUN source $HOME/.cargo/env && cargo install cargo-pgx; \
    cargo pgx init --pg14 /usr/local/bin/pg_config

RUN source $HOME/.cargo/env && cd && \
    cargo pgx new test && \
    cd test && \
    cargo pgx schema -f pg14

```

Creating that Dockerfile and then running:

```bash
docker build --platform=linux/arm64/v8 . -t pgx-empty-schema-repro

docker run -ti pgx-empty-schema-repro bash -c 'cat $HOME/test/sql/test-1.0.sql'
```

Hoverbear:
This is quite curious!

It seems that while the generated `libtest.so` has the required symbols:

```bash
0ab4bd9d41d7:~/test$ objdump -TC target/debug/libtest.so | grep __pgx_
0000000000016f00 g    DF .text  000000000000012b  Base        __pgx_marker
[...]

PR #359, opened 2022-01-03
Allow compilation on OpenBSD

ledeuns:
Two changes to allow to build pgx on OpenBSD :
- pgx requires GNU `make`. Base `make` is BSD `make` so we need to pass `gmake` (via an envvar)
- OpenBSD has `sigsetjmp`

### DIFF
--- a/pgx-pg-sys/build.rs
+++ b/pgx-pg-sys/build.rs
@@ -559,8 +559,9 @@ fn build_shim_for_version(
         .unwrap();
     }
 
+    let make = env!("MAKE").unwrap_or("make".to_string());
     let rc = run_command(
-        Command::new("make")
+        Command::new(make)
             .arg("clean")
             .arg(&format!("libpgx-cshim-{}.a", major_version))
             .env("PG_TARGET_VERSION", format!("{}", major_version))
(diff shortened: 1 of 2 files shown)


pgx-pg-sys/build.rs
```diff
@@ -559,8 +559,9 @@ fn build_shim_for_version(
         .unwrap();
     }
 
+    let make = env!("MAKE").unwrap_or("make".to_string());
```
Hoverbear, line 562: In this case `option_env!` is preferable, if `MAKE` is not set with `env!` the build will fail.

Docs: [`env`](https://doc.rust-lang.org/std/macro.env.html) vs [`option_env`](https://doc.rust-lang.org/std/macro.option_env.html)

ledeuns: Thank you for the pointer

JamesGuthrie:
FYI If you're planning on using schema generation on OpenBSD then you're probably going to need to add it as a target to https://github.com/zombodb/pgx/blob/master/cargo-pgx/src/templates/cargo_config. See #362 and #340 for more context.

eeeebbbbrrrr:
We're gonna ship this but don't have a way to test.

ledeuns:
I can confirm it works on OpenBSD 7.1. Thank you !

eeeebbbbrrrr:
Amazing! Thanks a bunch for testing and of course this PR. I’m glad we made the changes that accidentally got it working!

PR #370, opened 2022-01-10
Extern cost

### LINKED ISSUE #369: `#[pg_extern]` needs to support the `COST` property
for example:

```
  |
5 | #[pg_extern(immutable, parallel_safe, cost = "1000000.0")]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | |
  | expected expression
  | while parsing this struct
  |
  = note: this error originates in the attribute macro `pg_extern` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Specifically, it's the `COST` property here:  https://www.postgresql.org/docs/current/sql-createfunction.html

(as an aside, we ought to support all the options, but I need COST sooner rather than later.

Hoverbear:
Fixes #369 .

Also supports `cost = 1000` and `cost = 10000.6`

### DIFF
--- a/pgx-utils/src/lib.rs
+++ b/pgx-utils/src/lib.rs
@@ -197,6 +197,7 @@ pub enum ExternArgs {
     Error(String),
     Schema(String),
     Name(String),
+    Cost(String),
     Requires(Vec<PositioningRef>),
 }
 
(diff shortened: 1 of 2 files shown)


eeeebbbbrrrr:
LGTM.  Merging now.

PR #385, opened 2022-01-18
support a pgx extension compiling with no_std

eeeebbbbrrrr:
This plumbs through changes to allow a pgx extension to declare itself `#![no_std]` and still have it to compile correctly.

Use of `extern crate alloc`, however, is **required**.

There's an example in `pgx-examples/nostd` that demonstrates that all the things we officially support at least compile.

A release note:

 - Code that implements the `pgx::InOutFuncs` or `pgx:PgVarlenaInOutFuncs` traits to provide custom input/output functions for a custom `#[derive(PostgresType)]` will need to update the signatures to reference `pgx::cstr_core::CStr` instead of `std::ffi::CStr`.


pgx-examples/nostd/sql/nostd-1.0.sql
```diff
@@ -0,0 +1,5 @@
+/* 
+This file is auto generated by pgx.
+
+The ordering of items is not stable, it is driven by a dependency graph.
+*/
```
Hoverbear, line 5: I think this is an old sql file. We can delete this.

pgx-macros/src/lib.rs
```diff
@@ -290,12 +290,12 @@ extension_sql!(r#"\
 );
 
 #[pg_extern(immutable)]
-fn complex_in(input: &std::ffi::CStr) -> PgBox<Complex> {
+fn complex_in(input: &pgx::cstr_core::CStr) -> PgBox<Complex> {
```
Hoverbear, line 293: Hmm it'd be nice if we could provide like a `from` impl for this in `std` code so that users could just `.into()` it.

pgx/src/lib.rs
```diff
@@ -284,29 +287,28 @@ macro_rules! pg_magic_func {
...
             #[cfg(any(feature = "pg10", feature = "pg11", feature = "pg12"))]
             const MY_MAGIC: pgx::pg_sys::Pg_magic_struct = pgx::pg_sys::Pg_magic_struct {
-                len: size_of::<pgx::pg_sys::Pg_magic_struct>() as c_int,
-                version: pgx::pg_sys::PG_VERSION_NUM as c_int / 100,
-                funcmaxargs: pgx::pg_sys::FUNC_MAX_ARGS as c_int,
-                indexmaxkeys: pgx::pg_sys::INDEX_MAX_KEYS as c_int,
-                namedatalen: pgx::pg_sys::NAMEDATALEN as c_int,
-                float4byval: pgx::pg_sys::USE_FLOAT4_BYVAL as c_int,
-                float8byval: pgx::pg_sys::USE_FLOAT8_BYVAL as c_int,
+                len: size_of::<pgx::pg_sys::Pg_magic_struct>() as i32,
```
Hoverbear, line 295: This is mildly dangerous for portability... according to https://doc.rust-lang.org/std/os/raw/type.c_int.html, this is almost always an i32, but it's possible it is not... Ref: https://github.com/rust-lang/rfcs/issues/3012

Hoverbear: I *think* this will be fine as long as we don't end up with like `avr` or `msp430` users.

Hoverbear (approved):

PR #386, opened 2022-01-18
Allow pgx-pg-sys docs to build on docs.rs

Hoverbear:
This should work according to my tests running docs.rs locally. 

Docs: https://docs.rs/about/metadata


eeeebbbbrrrr:
ha, is this really it?

Hoverbear:
I think so!?

image

PR #389, opened 2022-01-19
Validate cargo-pgx and pgx crates are compatible using semver

Hoverbear:
This PR cleans up some of our feature handling for `cargo-pgx` to use the same API that `cargo` itself uses (with `--no-default-features` etc) and uses that new API to validate that the `pgx` crates appearing in the metadata are indeed semver compatible with `cargo-pgx`. It warns if they are not.

This is a safety step for when we make breaking changes in the future. It is not itself a breaking change.

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -253,6 +253,7 @@ version = "0.2.6"
 dependencies = [
  "cargo_metadata",
  "clap 3.0.10",
+ "clap-cargo",
  "color-eyre",
  "colored",
  "env_proxy",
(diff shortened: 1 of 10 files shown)


eeeebbbbrrrr:
I’d like to see an error if major is zero and minors differ. 

Otherwise if major is non-zero then we only error if major mismatches. 

We’re gonna be at zero for a bit longer but I want a way to indicate to users that there’s some breaking change that’s gonna force them to synchronize both sides. I know it’s not exactly semver but I think it’s justifiable.

Hoverbear:
The current strategy uses "Compatible" updates with Semver: https://docs.rs/semver/1.0.4/semver/enum.Op.html#opcaretcompatible-updates

It looks like what you're describing is "patch" update. :) So I'll set that and raise this to an error instead of a warning.

PR #400, opened 2022-01-25
Add type mapping for 'pg_sys::TimestampTz' to 'timestamp with time zone'

JamesGuthrie:
This allows the `pg_sys::TimestampTz' type to be used in function
signatures, providing zero-cost access to the underlying i64 timestamp
of a 'timestamp with time zone' field.


Hoverbear (approved):

Hoverbear:
Thanks!

PR #402, opened 2022-01-25
Make the pgversion optional when set in features.

### LINKED ISSUE #388: `cargo-pgx` should only require `pg$VERSION` if there is no default `pg$VERSION` feature.
Right now it's difficult for users to add features to their extensions.

This is because `cargo-pgx` always passes `--no-default-features`. 

> We pass `--no-default-features` so that we can pass a specific `pg$VERSION` feature if the package specifies a default. (Cargo doesn't let us negate features)

In extensions, such as our examples, users are shown they can set features like so:

https://github.com/zombodb/pgx/blob/733f5ed249cb6a1c85b59ef5784f93ed2a6c287b/pgx-examples/strings/Cargo.toml#L13

This works great, but if a user added a feature (like an optional `no_std` feature ala #385) this feature would *also* get ignored, leading to them needing to use `cargo` instead of `cargo pgx` directly, harming the user experience.

Notably, this issue is twofold:

* We should allow commands like `cargo pgx run` and `cargo pgx test` to use the default `pg$VERSION` feature to detect their default target.
* We should parse the `cargo metadata` output to determine the default features of a package and include those in our `cargo` invocations whenever the user **does** pass a `pg$VERSION` arg.

Hoverbear:
Fixes #388 .


Hoverbear:
I need to update the docs!

pgx-tests/src/framework.rs
```diff
@@ -236,25 +236,24 @@ fn install_extension() -> eyre::Result<()> {
...
+
     let mut command = Command::new("cargo");
     command
         .arg("pgx")
         .arg("install")
+        .arg("--pg-config")
+        .arg(pg_config.path().ok_or(eyre!("No pg_config found"))?)
+        .arg("--features")
+        .arg(format!(
+            "pg{} pg_test",
```
Hoverbear, line 251: What about the other features the user passed??? 🤔

Hoverbear:
Okay I think this is finally ready. Sheesh.

PR #409, opened 2022-01-28
Correctly transform control file version number in sql entity graph

JamesGuthrie:
Commit f00c335a7 added the ability to use '@CARGO_VERSION@' as a
placeholder for 'default_version' in the extension's .control file.

That change missed the fact that the contents of the .control file are
loaded into the root node of the sql entity graph. This means that when
using the placeholder, the default_version in the sql entity graph's
control file is '@CARGO_VERSION@'.

This change replaces the placeholder with the package version when
executing the compiled sql-generator.

### DIFF
--- a/cargo-pgx/src/command/install.rs
+++ b/cargo-pgx/src/command/install.rs
@@ -326,10 +326,9 @@ pub(crate) fn get_version() -> eyre::Result<String> {
     match get_property("default_version")? {
         Some(v) => {
             if v == "@CARGO_VERSION@" {
-                let metadata = MetadataCommand::new()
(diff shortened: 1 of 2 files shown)


Hoverbear:
Oh!!! Good catch!

Hoverbear (approved):

JamesGuthrie:
@Hoverbear as discussed on Discord, I've updated this PR to compile the `CARGO_PKG_VERSION` into the `sql-generator` binary, and use that to replace `@CARGO_VERSION@` with.

Hoverbear:
Thanks for this!

PR #377, opened 2022-01-14
POC of hook

Hoverbear:
POC of the hook addition @JamesGuthrie and I talked about.

PR #410, opened 2022-01-28
feat: extend sql generation via to_sql attribute

bitwalker:
/cc @Hoverbear @JamesGuthrie 

This PR was spawned from some needs we have over at Timescale in our Promscale extension. I looked into #377 as an option, but found that due to expectations the `ToSql` trait has about the entity graph, trying to manipulate entities at that stage is delicate at best. For example, you can't conditionally elide certain entities from SQL generation by removing them from the graph, because things like types are expected to exist in the graph during lowering. To make it work, you'd essentially need to reconstruct the graph within the hook to join the incoming and outgoing neighbors of those entities so that the dependency ordering is preserved without those entities being present in the output. It feels like that is exposing too much of the pgx internals implicitly, in a way that can break non-obviously and will always be fragile to internal changes.

So as an alternative, this PR introduces an approach that I think you'll like, and is less invasive while still providing quite a bit of flexibility around how entities get lowered to SQL. It works great for our needs, and others that are in a similar boat, so I'm hoping it is something you all are open to merging!

---

This commit introduces a new `#[to_sql]` attribute that allows for
customizing the behavior of SQL generation on certain entities in two
different ways:

1. `#[to_sql(false)]` disables generation of the decorated item's SQL
2. `#[to_sql(path::to::function)]` delegates responsibility for
   generating SQL to the named function

In the latter case, the function has almost the same signature as the
`to_sql` function of the built-in `ToSql` trait, except the function
receives a reference to a `SqlGraphEntity` in place of `&self`. This
allows for extending any of the `SqlGraphEntity` variants using a single
function, as trying to use specific typed functions for each entity type
was deemed overly complex. This also works well with the way `ToSql` is
invoked today, which starts by calling the implementation on a value of
type `SqlGraphEntity`, so we can perform all the checks in a single
place.

As an aside, these custom callbacks can still delegate to the built-in
`ToSql` trait if desired. For example, if you wish to write the
generated SQL for specific entities to a different file. One thing that might make
this even more powerful is if the lowering is by value (and also receives a mutable
reference to the `PgxSql` context), as this would allow custom functions to
modify certain attributes of the entities during lowering (as a simple exercise,
consider how one would modify the naming scheme used by the generator without
breaking dependents of the lowered entity - currently that isn't possible without taking
over almost all of the lowering yourself). Not super important I don't think, but something
to mull over.

The motivation for this is that in some edge cases, it is desirable to
elide or modify the SQL being generated. In our case specifically, we
need to chop up the generated SQL so that we can split out non-idempotent
operations separately from idempotent operations in order to properly
manage extension upgrades. Rather than lose the facilities pgx provides,
the `#[to_sql]` attribute allows us to make ad-hoc adjustments to what,
when and where things get generated without needing any upstream support
for those details.

### DIFF
--- a/pgx-macros/src/lib.rs
+++ b/pgx-macros/src/lib.rs
@@ -402,6 +402,7 @@ Optionally accepts the following attributes:
 * `parallel_unsafe`: Corresponds to [`PARALLEL UNSAFE`](https://www.postgresql.org/docs/current/sql-createfunction.html).
 * `parallel_restricted`: Corresponds to [`PARALLEL RESTRICTED`](https://www.postgresql.org/docs/current/sql-createfunction.html).
 * `no_guard`: Do not use `#[pg_guard]` with the function.
+* `sql`: Same arguments as [`#[pgx(sql = ..)]`](macro@pgx).
 
 Functions can accept and return any type which `pgx` supports. `pgx` supports many PostgreSQL types by default.
 New types can be defined via [`macro@PostgresType`] or [`macro@PostgresEnum`].
@@ -597,7 +598,7 @@ enum DogNames {
 ```
 
 */
-#[proc_macro_derive(PostgresEnum, attributes(requires))]
+#[proc_macro_derive(PostgresEnum, attributes(requires, pgx))]
 pub fn postgres_enum(input: TokenStream) -> TokenStream {
     let ast = parse_macro_input!(input as syn::DeriveInput);
 
@@ -683,9 +684,12 @@ Optionally accepts the following attributes:
 
 * `inoutfuncs(some_in_fn, some_out_fn)`: Define custom in/out functions for the type.
 * `pgvarlena_inoutfuncs(some_in_fn, some_out_fn)`: Define custom in/out functions for the `PgVarlena` of this type.
-
+* `sql`: Same arguments as [`#[pgx(sql = ..)]`](macro@pgx).
 */
-#[proc_macro_derive(PostgresType, attributes(inoutfuncs, pgvarlena_inoutfuncs, requires))]
+#[proc_macro_derive(
+    PostgresType,
+    attributes(inoutfuncs, pgvarlena_inoutfuncs, requires, pgx)
+)]
 pub fn postgres_type(input: TokenStream) -> TokenStream {
     let ast = parse_macro_input!(input as syn::DeriveInput);
 
@@ -911,12 +915,16 @@ enum DogNames {
     Brandy,
 }
 ```
+Optionally accepts the following attributes:
 
+* `sql`: Same arguments as [`#[pgx(sql = ..)]`](macro@pgx).
 */
-#[proc_macro_derive(PostgresEq)]
+#[proc_macro_derive(PostgresEq, attributes(pgx))]
 pub fn postgres_eq(input: TokenStream) -> TokenStream {
     let ast = parse_macro_input!(input as syn::DeriveInput);
-    impl_postgres_eq(ast).into()
+    impl_postgres_eq(ast)
+        .unwrap_or_else(syn::Error::into_compile_error)
+        .into()
 }
 
 /**
@@ -935,12 +943,16 @@ enum DogNames {
     Brandy,
 }
 ```
+Optionally accepts the following attributes:
 
+* `sql`: Same arguments as [`#[pgx(sql = ..)]`](macro@pgx).
 */
-#[proc_macro_derive(PostgresOrd)]
+#[proc_macro_derive(PostgresOrd, attributes(pgx))]
 pub fn postgres_ord(input: TokenStream) -> TokenStream {
     let ast = parse_macro_input!(input as syn::DeriveInput);
-    impl_postgres_ord(ast).into()
+    impl_postgres_ord(ast)
+        .unwrap_or_else(syn::Error::into_compile_error)
+        .into()
 }
 
 /**
@@ -956,12 +968,16 @@ enum DogNames {
     Brandy,
 }
 ```
+Optionally accepts the following attributes:
 
+* `sql`: Same arguments as [`#[pgx(sql = ..)]`](macro@pgx).
 */
-#[proc_macro_derive(PostgresHash)]
+#[proc_macro_derive(PostgresHash, attributes(pgx))]
 pub fn postgres_hash(input: TokenStream) -> TokenStream {
     let ast = parse_macro_input!(input as syn::DeriveInput);
-    impl_postgres_hash(ast).into()
+    impl_postgres_hash(ast)
+        .unwrap_or_else(syn::Error::into_compile_error)
+        .into()
 }
 
 /**
@@ -991,15 +1007,26 @@ pub fn pg_aggregate(_attr: TokenStream, item: TokenStream) -> TokenStream {
 }
 
 /**
-An inner attribute for [`#[pg_aggregate]`](macro@pg_aggregate).
+A helper attribute for various contexts.
+
+## Usage with [`#[pg_aggregate]`](macro@pg_aggregate).
 
 It can be decorated on functions inside a [`#[pg_aggregate]`](macro@pg_aggregate) implementation.
 In this position, it takes the same args as [`#[pg_extern]`](macro@pg_extern), and those args have the same effect.
 
-Used outside of a [`#[pg_aggregate]`](macro@pg_aggregate), this does nothing.
+## Usage for configuring SQL generation
+
+This attribute can be used to control the behavior of the SQL generator on a decorated item,
+e.g. `#[pgx(sql = false)]`
+
+Currently `sql` can be provided one of the following:
+
+* Disable SQL generation with `#[pgx(sql = false)]`
+* Call custom SQL generator function with `#[pgx(sql = path::to_function)]`
+* Render a specific fragment of SQL with a string `#[pgx(sql = "CREATE OR REPLACE FUNCTION ...")]`
+
 */
 #[proc_macro_attribute]
 pub fn pgx(_attr: TokenStream, item: TokenStream) -> TokenStream {
     item
 }
-
--- a/pgx-tests/src/tests/schema_tests.rs
+++ b/pgx-tests/src/tests/schema_tests.rs
@@ -4,14 +4,66 @@ use pgx::*;
 
 #[pgx::pg_schema]
 mod test_schema {
+    use pgx::datum::sql_entity_graph::{PgxSql, SqlGraphEntity};
     use pgx::*;
     use serde::{Deserialize, Serialize};
 
     #[pg_extern]
     fn func_in_diff_schema() {}
 
+    #[pg_extern(sql = false)]
+    fn func_elided_from_schema() {}
+
+    #[pg_extern(sql = generate_function)]
+    fn func_generated_with_custom_sql() {}
+
     #[derive(Debug, PostgresType, Serialize, Deserialize)]
     pub struct TestType(pub u64);
+
+    #[derive(Debug, PostgresType, Serialize, Deserialize)]
+    #[pgx(sql = false)]
+    pub struct ElidedType(pub u64);
+
+    #[derive(Debug, PostgresType, Serialize, Deserialize)]
+    #[pgx(sql = generate_type)]
+    pub struct OtherType(pub u64);
+
+    #[derive(Debug, PostgresType, Serialize, Deserialize)]
+    #[pgx(sql = "CREATE TYPE test_schema.ManuallyRenderedType;")]
+    pub struct OverriddenType(pub u64);
+
+    fn generate_function(
+        entity: &SqlGraphEntity,
+        _context: &PgxSql,
+    ) -> Result<String, Box<dyn std::error::Error + Send + Sync + 'static>> {
+        if let SqlGraphEntity::Function(ref func) = entity {
+            Ok(format!("\
+                CREATE OR REPLACE FUNCTION test_schema.\"func_generated_with_custom_name\"() RETURNS void\n\
+                LANGUAGE c /* Rust */\n\
+                AS 'MODULE_PATHNAME', '{unaliased_name}_wrapper';\
+                ",
+                unaliased_name = func.unaliased_name,
+            ))
+        } else {
+            panic!("expected extern function entity, got {:?}", entity);
+        }
+    }
+
+    fn generate_type(
+        entity: &SqlGraphEntity,
+        _context: &PgxSql,
+    ) -> Result<String, Box<dyn std::error::Error + Send + Sync + 'static>> {
+        if let SqlGraphEntity::Type(ref ty) = entity {
+            Ok(format!(
+                "\n\
+                CREATE TYPE test_schema.Custom{name};\
+                ",
+                name = ty.name,
+            ))
+        } else {
+            panic!("expected type entity, got {:?}", entity);
+        }
+    }
 }
 
 #[pg_extern(schema = "test_schema")]
@@ -43,5 +95,65 @@ mod tests {
     #[pg_test]
     fn test_type_in_different_schema() {
         Spi::run("SELECT type_in_diff_schema();");
+    }
+
+    #[pg_test]
+    fn elided_extern_is_elided() {
+        // Validate that a function we know exists, exists
+        let result: bool = Spi::get_one(
+            "SELECT exists(SELECT 1 FROM pg_proc WHERE proname = 'func_in_diff_schema');",
+        )
+        .expect("expected result");
+        assert_eq!(result, true);
+
+        // Validate that the function we expect not to exist, doesn't
+        let result: bool = Spi::get_one(
+            "SELECT exists(SELECT 1 FROM pg_proc WHERE proname = 'func_elided_from_schema');",
+        )
+        .expect("expected result");
+        assert_eq!(result, false);
+    }
+
+    #[pg_test]
+    fn elided_type_is_elided() {
+        // Validate that a type we know exists, exists
+        let result: bool =
+            Spi::get_one("SELECT exists(SELECT 1 FROM pg_type WHERE typname = 'testtype');")
+                .expect("expected result");
+        assert_eq!(result, true);
+
+        // Validate that the type we expect not to exist, doesn't
+        let result: bool =
+            Spi::get_one("SELECT exists(SELECT 1 FROM pg_type WHERE typname = 'elidedtype');")
+                .expect("expected result");
+        assert_eq!(result, false);
+    }
+
+    #[pg_test]
+    fn custom_to_sql_extern() {
+        // Validate that the function we generated has the modifications we expect
+        let result: bool = Spi::get_one("SELECT exists(SELECT 1 FROM pg_proc WHERE proname = 'func_generated_with_custom_name');").expect("expected result");
+        assert_eq!(result, true);
+
+        Spi::run("SELECT test_schema.func_generated_with_custom_name();");
+    }
+
+    #[pg_test]
+    fn custom_to_sql_type() {
+        // Validate that the type we generated has the expected modifications
+        let result: bool =
+            Spi::get_one("SELECT exists(SELECT 1 FROM pg_type WHERE typname = 'customothertype');")
+                .expect("expected result");
+        assert_eq!(result, true);
+    }
+
+    #[pg_test]
+    fn custom_handwritten_to_sql_type() {
+        // Validate that the SQL we provided was used
+        let result: bool = Spi::get_one(
+            "SELECT exists(SELECT 1 FROM pg_type WHERE typname = 'manuallyrenderedtype');",
+        )
+        .expect("expected result");
+        assert_eq!(result, true);
     }
 }
--- a/pgx-utils/src/sql_entity_graph/pg_extern/attribute.rs
+++ b/pgx-utils/src/sql_entity_graph/pg_extern/attribute.rs
@@ -1,4 +1,4 @@
-use crate::sql_entity_graph::PositioningRef;
+use crate::sql_entity_graph::{PositioningRef, ToSqlConfig};
 use proc_macro2::{Span, TokenStream as TokenStream2};
 use quote::{quote, ToTokens, TokenStreamExt};
 use syn::{
@@ -6,29 +6,6 @@ use syn::{
     punctuated::Punctuated,
     Token,
 };
-
-#[derive(Debug, Clone)]
-pub struct PgxAttributes {
-    pub attrs: Punctuated<Attribute, Token![,]>,
-}
-
-impl Parse for PgxAttributes {
-    fn parse(input: ParseStream) -> Result<Self, syn::Error> {
-        Ok(Self {
-            attrs: input.parse_terminated(Attribute::parse)?,
-        })
-    }
-}
-
-impl ToTokens for PgxAttributes {
-    fn to_tokens(&self, tokens: &mut TokenStream2) {
-        let attrs = &self.attrs;
-        let quoted = quote! {
-            vec![#attrs]
-        };
-        tokens.append_all(quoted);
-    }
-}
 
 #[derive(Debug, Clone, Hash, Eq, PartialEq)]
 pub enum Attribute {
@@ -46,6 +23,7 @@ pub enum Attribute {
     Name(syn::LitStr),
     Cost(syn::Expr),
     Requires(Punctuated<PositioningRef, Token![,]>),
+    Sql(ToSqlConfig),
 }
 
 impl ToTokens for Attribute {
@@ -84,6 +62,10 @@ impl ToTokens for Attribute {
                     .map(|x| x.to_token_stream())
                     .collect::<Vec<_>>();
                 quote! { pgx::datum::sql_entity_graph::ExternArgs::Requires(vec![#(#items_iter),*],) }
+            }
+            // This attribute is handled separately
+            Attribute::Sql(_) => {
+                return;
             }
         };
         tokens.append_all(quoted);
@@ -128,6 +110,23 @@ impl Parse for Attribute {
                 let content;
                 let _bracket = syn::bracketed!(content in input);
                 Self::Requires(content.parse_terminated(PositioningRef::parse)?)
+            }
+            "sql" => {
+                use crate::sql_entity_graph::ArgValue;
+                use syn::Lit;
+
+                let _eq: Token![=] = input.parse()?;
+                match input.parse::<ArgValue>()? {
+                    ArgValue::Path(p) => Self::Sql(ToSqlConfig::from(p)),
+                    ArgValue::Lit(Lit::Bool(b)) => Self::Sql(ToSqlConfig::from(b.value)),
+                    ArgValue::Lit(Lit::Str(s)) => Self::Sql(ToSqlConfig::from(s)),
+                    ArgValue::Lit(other) => {
+                        return Err(syn::Error::new(
+                            other.span(),
+                            "expected boolean, path, or string literal",
+                        ))
+                    }
+                }
             }
             _ => return Err(syn::Error::new(Span::call_site(), "Invalid option")),
         };
--- a/pgx-utils/src/sql_entity_graph/pg_extern/mod.rs
+++ b/pgx-utils/src/sql_entity_graph/pg_extern/mod.rs
@@ -5,19 +5,22 @@ mod returning;
 mod search_path;
 
 pub use argument::Argument;
-use attribute::{Attribute, PgxAttributes};
+use attribute::Attribute;
 pub use operator::PgOperator;
 use operator::{PgxOperatorAttributeWithIdent, PgxOperatorOpName};
-use returning::Returning;
 pub(crate) use returning::NameMacro;
+use returning::Returning;
 use search_path::SearchPathList;
 
 use eyre::WrapErr;
 use proc_macro2::{Ident, Span, TokenStream as TokenStream2};
 use quote::{quote, ToTokens, TokenStreamExt};
 use std::convert::TryFrom;
-use syn::parse::{Parse, ParseStream};
-use syn::Meta;
+use syn::parse::{Parse, ParseStream, Parser};
+use syn::punctuated::Punctuated;
+use syn::{Meta, Token};
+
+use crate::sql_entity_graph::ToSqlConfig;
 
 /// A parsed `#[pg_extern]` item.
 ///
@@ -42,42 +45,35 @@ use syn::Meta;
 /// ```
 #[derive(Debug, Clone)]
 pub struct PgExtern {
-    attrs: Option<PgxAttributes>,
-    attr_tokens: proc_macro2::TokenStream,
+    attrs: Vec<Attribute>,
     func: syn::ItemFn,
+    to_sql_config: ToSqlConfig,
 }
 
 impl PgExtern {
     fn name(&self) -> String {
         self.attrs
-            .as_ref()
-            .and_then(|a| {
-                a.attrs.iter().find_map(|candidate| match candidate {
-                    Attribute::Name(name) => Some(name.value()),
-                    _ => None,
-                })
+            .iter()
+            .find_map(|a| match a {
+                Attribute::Name(name) => Some(name.value()),
+                _ => None,
             })
             .unwrap_or_else(|| self.func.sig.ident.to_string())
     }
 
     fn schema(&self) -> Option<String> {
-        self.attrs.as_ref().and_then(|a| {
-            a.attrs.iter().find_map(|candidate| match candidate {
-                Attribute::Schema(name) => Some(name.value()),
-                _ => None,
-            })
+        self.attrs.iter().find_map(|a| match a {
+            Attribute::Schema(name) => Some(name.value()),
+            _ => None,
         })
     }
 
-    fn extern_attrs(&self) -> Option<&PgxAttributes> {
-        self.attrs.as_ref()
+    pub fn extern_attrs(&self) -> &[Attribute] {
+        self.attrs.as_slice()
     }
 
-    pub fn extern_attr_tokens(&self) -> &proc_macro2::TokenStream {
-        &self.attr_tokens
-    }
-
-    fn overridden(&self) -> Option<String> {
+    fn overridden(&self) -> Option<syn::LitStr> {
+        let mut span = None;
         let mut retval = None;
         let mut in_commented_sql_block = false;
         for attr in &self.func.attrs {
@@ -88,7 +84,8 @@ impl PgExtern {
                         Meta::Path(_) | Meta::List(_) => continue,
                         Meta::NameValue(mnv) => mnv,
                     };
-                    if let syn::Lit::Str(inner) = content.lit {
+                    if let syn::Lit::Str(ref inner) = content.lit {
+                        span.get_or_insert(content.lit.span());
                         if !in_commented_sql_block && inner.value().trim() == "```pgxsql" {
                             in_commented_sql_block = true;
                         } else if in_commented_sql_block && inner.value().trim() == "```" {
@@ -105,7 +102,7 @@ impl PgExtern {
                 }
             }
         }
-        retval
+        retval.map(|s| syn::LitStr::new(s.as_ref(), span.unwrap()))
     }
 
     fn operator(&self) -> Option<PgOperator> {
@@ -191,12 +188,27 @@ impl PgExtern {
     }
 
     pub fn new(attr: TokenStream2, item: TokenStream2) -> Result<Self, syn::Error> {
-        let attrs = syn::parse2::<PgxAttributes>(attr.clone()).ok();
+        let mut attrs = Vec::new();
+        let mut to_sql_config: Option<ToSqlConfig> = None;
+
+        let parser = Punctuated::<Attribute, Token![,]>::parse_terminated;
+        let punctuated_attrs = parser.parse2(attr)?;
+        for pair in punctuated_attrs.into_pairs() {
+            match pair.into_value() {
+                Attribute::Sql(config) => {
+                    to_sql_config.get_or_insert(config);
+                }
+                attr => {
+                    attrs.push(attr);
+                }
+            }
+        }
+
         let func = syn::parse2::<syn::ItemFn>(item)?;
         Ok(Self {
-            attrs: attrs,
-            attr_tokens: attr,
-            func: func,
+            attrs,
+            func,
+            to_sql_config: to_sql_config.unwrap_or_default(),
         })
     }
 }
@@ -207,7 +219,7 @@ impl ToTokens for PgExtern {
         let name = self.name();
         let schema = self.schema();
         let schema_iter = schema.iter();
-        let extern_attrs = self.extern_attrs();
+        let extern_attrs = self.attrs.iter().collect::<Punctuated<_, Token![,]>>();
         let search_path = self.search_path().into_iter();
         let inputs = self.inputs().unwrap();
         let returns = match self.returns() {
@@ -221,7 +233,14 @@ impl ToTokens for PgExtern {
             }
         };
         let operator = self.operator().into_iter();
-        let overridden = self.overridden().into_iter();
+        let to_sql_config = match self.overridden() {
+            None => self.to_sql_config.clone(),
+            Some(content) => {
+                let mut config = self.to_sql_config.clone();
+                config.content = Some(content);
+                config
+            }
+        };
 
         let sql_graph_entity_fn_name =
             syn::Ident::new(&format!("__pgx_internals_fn_{}", ident), Span::call_site());
@@ -237,12 +256,12 @@ impl ToTokens for PgExtern {
                     line: line!(),
                     module_path: core::module_path!(),
                     full_path: concat!(core::module_path!(), "::", stringify!(#ident)),
-                    extern_attrs: #extern_attrs,
+                    extern_attrs: vec![#extern_attrs],
                     search_path: None#( .unwrap_or(Some(vec![#search_path])) )*,
                     fn_args: vec![#(#inputs),*],
                     fn_return: #returns,
                     operator: None#( .unwrap_or(Some(#operator)) )*,
-                    overridden: None#( .unwrap_or(Some(#overridden)) )*,
+                    to_sql_config: #to_sql_config,
                 };
                 pgx::datum::sql_entity_graph::SqlGraphEntity::Function(submission)
             }
@@ -253,13 +272,27 @@ impl ToTokens for PgExtern {
 
 impl Parse for PgExtern {
     fn parse(input: ParseStream) -> Result<Self, syn::Error> {
-        let attrs: Option<PgxAttributes> = input.parse().ok();
-        let func = input.parse()?;
-        let attr_tokens: proc_macro2::TokenStream = attrs.clone().into_token_stream();
+        let mut attrs = Vec::new();
+        let mut to_sql_config: Option<ToSqlConfig> = None;
+
+        let parser = Punctuated::<Attribute, Token![,]>::parse_terminated;
+        let punctuated_attrs = input.call(parser).ok().unwrap_or_default();
+        for pair in punctuated_attrs.into_pairs() {
+            match pair.into_value() {
+                Attribute::Sql(config) => {
+                    to_sql_config.get_or_insert(config);
+                }
+                attr => {
+                    attrs.push(attr);
+                }
+            }
+        }
+
+        let func: syn::ItemFn = input.parse()?;
         Ok(Self {
             attrs,
-            attr_tokens,
             func,
+            to_sql_config: to_sql_config.unwrap_or_default(),
         })
     }
 }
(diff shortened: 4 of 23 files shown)


Hoverbear:
Generally a fan of this idea, like the `serde` Derive like strategy here.

@eeeebbbbrrrr and I have just introduced a `#[pgx]` attribute for the aggregates in #230. I think it'd be more consistent to use that. (Thinking like `serde`)

I'll give this a spin. :)

bitwalker:
@Hoverbear Thanks! I can certainly rework this to use the `#[pgx]` attribute instead, makes sense to me. Are you thinking the end result would then be `#[pgx(to_sql(false))]`/`#[pgx(to_sql(path::to::function))]`?

As an aside, I'm not really sure why the one test is failing, but will try to reproduce and track that down today. As far as I can tell it shouldn't be affected by the changes, but it might just be something subtle with the change to the `ToSql` implementation for `SqlGraphEntity` that isn't obvious at first glance.

pgx/src/datum/sql_entity_graph/aggregate.rs
```diff
@@ -301,7 +302,7 @@ impl ToSql for PgAggregateEntity {
             "\n\
                 -- {file}:{line}\n\
                 -- {full_path}\n\
-                CREATE AGGREGATE {schema}{name} ({args}{maybe_order_by})\n\
+                CREATE OR REPLACE AGGREGATE {schema}{name} ({args}{maybe_order_by})\n\
```
Hoverbear, line 305: This seems unrelated?

bitwalker: Ah indeed, didn't intend to add that change in this PR, apologies! I'll skim through and make sure other unrelated changes didn't also get added accidentally.

pgx/src/datum/sql_entity_graph/mod.rs
```diff
@@ -64,6 +64,72 @@ pub trait ToSql {
     fn to_sql(&self, context: &PgxSql) -> eyre::Result<String>;
 }
 
+/// Represents a pointer to a function that can transform a SqlGraphEntity to a SQL string
```
Hoverbear, line 67: Nit: Not actually a pointer. It's just a type alias.

pgx/src/datum/sql_entity_graph/mod.rs
```diff
@@ -64,6 +64,72 @@ pub trait ToSql {
...
+///
+/// This is used to provide a facility for overriding the default SQL generator behavior using
+/// the `#[to_sql(path::to::function)]` attribute in circumstances where the default behavior is
+/// not desirable.
+///
+/// Implementations can invoke `ToSql::to_sql(entity, context)` on the unwrapped SqlGraphEntity
+/// type should they wish to delegate to the default behavior for any reason.
+pub type ToSqlFn = fn(&SqlGraphEntity, &PgxSql) -> eyre::Result<String>;
```
Hoverbear, line 75: I don't really want to infect downstream users with `eyre::Result`. Can we do a `Result<String, impl std::error::Error>` here?

bitwalker: I'll experiment a bit and see, but I'm sure we can use something along those lines, as long as it is convertible to `eyre::Result` internally. For `eyre::Result` that should be anything that implements `std::error::Error + Send + Sync + 'static`

pgx/src/datum/sql_entity_graph/mod.rs
bitwalker: I meant that the type represents a pointer to a function, but you're right, that's a semantic distinction that isn't really relevant to users of pgx. Any suggestions on how best to document its purpose? I suppose in practice this doesn't need to be publically documented per se, since users are not constructing a value of this type anywhere, it is just necessary for functions they intend to be targets of `#[to_sql(path::to::function)]` to have the same signature.

pgx-tests/src/tests/schema_tests.rs
```diff
@@ -4,14 +4,58 @@ use pgx::*;
...
     use serde::{Deserialize, Serialize};
 
     #[pg_extern]
     fn func_in_diff_schema() {}
 
+    #[pg_extern]
+    #[to_sql(false)]
+    fn func_elided_from_schema() {}
+
+    #[pg_extern]
```
Hoverbear, line 18: I think `#[pg_extern(to_sql = false)]` is a bit more idiomatic here.

pgx-tests/src/tests/schema_tests.rs
```diff
@@ -4,14 +4,58 @@ use pgx::*;
...
+    #[to_sql(generate_function)]
+    fn func_generated_with_custom_sql() {}
+
     #[derive(Debug, PostgresType, Serialize, Deserialize)]
     pub struct TestType(pub u64);
+
+    #[derive(Debug, PostgresType, Serialize, Deserialize)]
+    #[to_sql(false)]
```
Hoverbear, line 26: As mentioned in a comment, `#[pgx(to_sql = false)]` would be be more matching how we see `serde` and similar crate do it.

pgx/src/datum/sql_entity_graph/mod.rs
```diff
@@ -64,6 +64,72 @@ pub trait ToSql {
...
+    }
+}
+impl std::cmp::Eq for ToSqlConfig {}
+impl std::hash::Hash for ToSqlConfig {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.enabled.hash(state);
+        self.callback.map(|cb| std::ptr::addr_of!(cb)).hash(state);
+    }
+}
+impl std::fmt::Debug for ToSqlConfig {
```
Hoverbear, line 115: This couldn't be derived? Interesting.

pgx/src/datum/sql_entity_graph/mod.rs
```diff
@@ -64,6 +64,72 @@ pub trait ToSql {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        let callback = self.callback.map(|cb| std::ptr::addr_of!(cb));
+        f.debug_struct("ToSqlConfig")
+            .field("enabled", &self.enabled)
+            .field("callback", &format_args!("{:?}", &callback))
+            .finish()
+    }
+}
+impl Default for ToSqlConfig {
```
Hoverbear, line 124: You can `#[derive(Debug)]` this. :)

pgx/src/datum/sql_entity_graph/mod.rs
```diff
@@ -64,6 +64,72 @@ pub trait ToSql {
...
+/// The default configuration has `enabled` set to `true`, and `callback` to `None`, which indicates
+/// that the default SQL generation behavior will be used. These are intended to be mutually exclusive
+/// options, so `callback` should only be set if generation is enabled.
+///
+/// When `enabled` is false, no SQL is generated for the item being configured.
+///
+/// When `callback` has a value, the corresponding `ToSql` implementation should invoke the
+/// callback instead of performing their default behavior.
+#[derive(Clone)]
+pub struct ToSqlConfig {
```
Hoverbear, line 92:
```suggestion
pub struct ToSqlConfigEntity {
```

Otherwise we see duplicated docs:
image

pgx/src/datum/sql_entity_graph/sql_graph_entity.rs
```diff
@@ -110,13 +113,40 @@ impl ToSql for SqlGraphEntity {
                 }
             }) {
                 Ok(String::default())
-            } else { item.to_sql(context) },
-            SqlGraphEntity::Type(item) => item.to_sql(context),
+            } else {
+                item.to_sql(context)
+            }
+            SqlGraphEntity::Type(item) if !item.to_sql_config.enabled => Ok(String::default()),
+            SqlGraphEntity::Type(item) => if let Some(callback) = item.to_sql_config.callback {
```
Hoverbear, line 120: I think this decision could be inside `item.to-sql()`? It seems since the config is local to the entity it should decide to have SQL output or not?

pgx/src/datum/sql_entity_graph/aggregate.rs
Hoverbear: I didn't see any! I was trying to figure out why that test was failing. :)

pgx/src/datum/sql_entity_graph/mod.rs
Hoverbear: How about like "The signature of a function that can transform ..."?

pgx/src/datum/sql_entity_graph/mod.rs
Hoverbear: Yeah, that'd be good. That means that users won't necessarily need to use eyre. :)

Hoverbear:
I'm really excited about the direction this PR takes us, as well as the PR. :)

pgx/src/datum/sql_entity_graph/aggregate.rs
Hoverbear: `CREATE OR REPLACE AGGREGATE` is not valid in pg10 :( https://www.postgresql.org/docs/10/sql-createaggregate.html

pgx/src/datum/sql_entity_graph/mod.rs
bitwalker: Good catch, not sure why I didn't do that here!

pgx/src/datum/sql_entity_graph/mod.rs
bitwalker: Yeah, basically rustc complains that it can't derive them because in order for the derivation to be valid, it would have to be implemented for all possible lifetimes that the references in the argument list could be valid for, e.g. you get errors like this:

```
error: implementation of `Debug` is not general enough
  --> pgx/src/datum/sql_entity_graph/mod.rs:94:5
   |
91 | #[derive(Debug, Clone)]
   |          ----- in this derive macro expansion
...
94 |     pub callback: Option<ToSqlFn>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ implementation of `Debug` is not general enough
   |
   = note: `Debug` would have to be implemented for the type `for<'r, 's> fn(&'r SqlGraphEntity, &'s PgxSql) -> std::result::Result<std::string::String, ErrReport>`
   = note: ...but `Debug` is actually implemented for the type `fn(&'0 SqlGraphEntity, &PgxSql) -> std::result::Result<std::string::String, ErrReport>`, for some specific lifetime `'0`
   = note: this error originates in the derive macro `Debug` (in Nightly builds, run with -Z macro-backtrace for more info)
```

So the workaround here is just formatting as a raw pointer with no associated lifetime. The other handwritten impls are necessary for similar reasons (either there is no derivable impl, or the derivation has to be done by hand because of the lifetime issue).

pgx/src/datum/sql_entity_graph/sql_graph_entity.rs
bitwalker: Yeah I originally lifted the checks into this impl since everything goes through it anyway, and it allows us to avoid creating an extra `SqlGraphEntity` value in the item impl, but thinking on it further, I suspect the optimizer can do well enough on this so as to make it a moot point. I'll move this work into the individual impls.

bitwalker: Digging a bit further, there was a little more to my choice here that I forgot, and I'd be interested in your take on what the best approach is.

By delegating all the work to the `ToSql` impl of `SqlGraphEntity`, we can just pass a reference to the current entity to the callback and we're off to the races. If we instead want to delegate the work to the individual `ToSql` impls, things get complicated because we need to wrap the items in the `SqlGraphEntity` enum for genericity (i.e. the ability to handle various entity types with a single callback), but since the `ToSql` impls themselves operate on a reference, we can't get a value of `SqlGraphEntity` without cloning the item, imposing quite a bit of additional overhead.

I think in principle it would make the most sense for the impls to have the responsibility, but given the tradeoffs, leaving control in the hands of the one impl (that of `SqlGraphEntity`) is the better choice. I'm open to a different approach though, just curious what your take is on it.

Hoverbear:
I was thinking more about this PR and we have a `pgx_sql` functionality associated with `pg_extern`. I think we could converge it's functionality with this.

```rust
/// ```pgx_sql
/// CREATE FUNCTION ...
/// ```
#[pg_extern]
function dooper() {}
```

Could become

```rust
#[pg_extern(sql = "
CREATE FUNCTION ...
")]
function dooper() {}
```

So I wonder if we should change `to_sql` to `sql` in prep for that...

pgx/src/datum/sql_entity_graph/mod.rs
Hoverbear: Whoa, super interesting. Thanks for trying this. :)

pgx/src/datum/sql_entity_graph/mod.rs
Hoverbear: Eyre guides us here! https://docs.rs/eyre/latest/eyre/trait.WrapErr.html#wrapping-types-that-dont-impl-error-eg-str-and-boxdyn-error

This works:

```patch
diff --git a/pgx-tests/src/tests/schema_tests.rs b/pgx-tests/src/tests/schema_tests.rs
index 8c16bc2..0953079 100644
--- a/pgx-tests/src/tests/schema_tests.rs
+++ b/pgx-tests/src/tests/schema_tests.rs
@@ -30,7 +30,7 @@ mod test_schema {
     #[to_sql(generate_type)]
     pub struct OtherType(pub u64);
 
-    fn generate_function(entity: &SqlGraphEntity, _context: &PgxSql) -> eyre::Result<String> {
+    fn generate_function(entity: &SqlGraphEntity, _context: &PgxSql) -> Result<String, Box<dyn std::error::Error + Send + Sync>> {
         if let SqlGraphEntity::Function(ref func) = entity {
             Ok(format!("\
                 CREATE OR REPLACE FUNCTION test_schema.\"func_generated_with_custom_name\"() RETURNS void\n\
@@ -44,7 +44,7 @@ mod test_schema {
         }
     }
 
-    fn generate_type(entity: &SqlGraphEntity, _context: &PgxSql) -> eyre::Result<String> {
+    fn generate_type(entity: &SqlGraphEntity, _context: &PgxSql) -> Result<String, Box<dyn std::error::Error + Send + Sync>> {
         if let SqlGraphEntity::Type(ref ty) = entity {
             Ok(format!(
                 "\n\
diff --git a/pgx/src/datum/sql_entity_graph/mod.rs b/pgx/src/datum/sql_entity_graph/mod.rs
index e86dbf2..d41f0a8 100644
--- a/pgx/src/datum/sql_entity_graph/mod.rs
+++ b/pgx/src/datum/sql_entity_graph/mod.rs
@@ -72,7 +72,7 @@ pub trait ToSql {
 ///
 /// Implementations can invoke `ToSql::to_sql(entity, context)` on the unwrapped SqlGraphEntity
 /// type should they wish to delegate to the default behavior for any reason.
-pub type ToSqlFn = fn(&SqlGraphEntity, &PgxSql) -> eyre::Result<String>;
+pub type ToSqlFn = fn(&SqlGraphEntity, &PgxSql) -> Result<String, Box<dyn std::error::Error + Send + Sync + 'static>>;
 
 /// Represents configuration options for tuning the SQL generator.
 ///
diff --git a/pgx/src/datum/sql_entity_graph/sql_graph_entity.rs b/pgx/src/datum/sql_entity_graph/sql_graph_entity.rs
index 9ea96ab..d85076e 100644
--- a/pgx/src/datum/sql_entity_graph/sql_graph_entity.rs
+++ b/pgx/src/datum/sql_entity_graph/sql_graph_entity.rs
@@ -3,6 +3,7 @@ use super::{
     PostgresEnumEntity, PostgresHashEntity, PostgresOrdEntity, PostgresTypeEntity, SchemaEntity,
     SqlGraphIdentifier, ToSql,
 };
+use eyre::{WrapErr, eyre};
 
 /// An entity corresponding to some SQL required by the extension.
 #[derive(Debug, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
@@ -94,7 +95,7 @@ impl ToSql for SqlGraphEntity {
             },
             SqlGraphEntity::Function(item) if !item.to_sql_config.enabled => Ok(String::default()),
             SqlGraphEntity::Function(item) => if let Some(callback) = item.to_sql_config.callback {
-                callback(self, context)
+                callback(self, context).map_err(|e| eyre!(e)).wrap_err("Failed to run specified `#[pgx(sql = path)] function")
             } else if context.graph.neighbors_undirected(context.externs.get(item).unwrap().clone()).any(|neighbor| {
                 let neighbor_item = &context.graph[neighbor];
                 match neighbor_item {
@@ -118,32 +119,32 @@ impl ToSql for SqlGraphEntity {
             }
             SqlGraphEntity::Type(item) if !item.to_sql_config.enabled => Ok(String::default()),
             SqlGraphEntity::Type(item) => if let Some(callback) = item.to_sql_config.callback {
-                callback(self, context)
+                callback(self, context).map_err(|e| eyre!(e)).wrap_err("Failed to run specified `#[pgx(sql = path)] function")
             } else {
                 item.to_sql(context)
             }
             SqlGraphEntity::BuiltinType(_) => Ok(String::default()),
             SqlGraphEntity::Enum(item) if !item.to_sql_config.enabled => Ok(String::default()),
             SqlGraphEntity::Enum(item) => if let Some(callback) = item.to_sql_config.callback {
-                callback(self, context)
+                callback(self, context).map_err(|e| eyre!(e)).wrap_err("Failed to run specified `#[pgx(sql = path)] function")
             } else {
                 item.to_sql(context)
             }
             SqlGraphEntity::Ord(item) if !item.to_sql_config.enabled => Ok(String::default()),
             SqlGraphEntity::Ord(item) => if let Some(callback) = item.to_sql_config.callback {
-                callback(self, context)
+                callback(self, context).map_err(|e| eyre!(e)).wrap_err("Failed to run specified `#[pgx(sql = path)] function")
             } else {
                 item.to_sql(context)
             }
             SqlGraphEntity::Hash(item) if !item.to_sql_config.enabled => Ok(String::default()),
             SqlGraphEntity::Hash(item) => if let Some(callback) = item.to_sql_config.callback {
-                callback(self, context)
+                callback(self, context).map_err(|e| eyre!(e)).wrap_err("Failed to run specified `#[pgx(sql = path)] function")
             } else {
                 item.to_sql(context)
             }
             SqlGraphEntity::Aggregate(item) if !item.to_sql_config.enabled => Ok(String::default()),
             SqlGraphEntity::Aggregate(item) => if let Some(callback) = item.to_sql_config.callback {
-                callback(self, context)
+                callback(self, context).map_err(|e| eyre!(e)).wrap_err("Failed to run specified `#[pgx(sql = path)] function")
             } else {
                 item.to_sql(context)
             }
```

pgx/src/datum/sql_entity_graph/sql_graph_entity.rs
Hoverbear: I fiddled around and I understand now.

Thinking on it: If a user decorates say `fn doop()`:

```rust
#[pg_extern(sql = doop_sql)]
fn doop() {}

fn doop_sql(entity: &SqlGraphEntity, pgx: &PgxSql) -> Result<String, Box<dyn std::error::Error + Send + Sync + 'static>> {
  // ...
}
```

Here they **know** entity is a `SqlGraphEntity::Function` variant.  You can see this from your examples:

https://github.com/zombodb/pgx/blob/f5e89cfc0aa88e5451c5da63046c4c328c3a3711/pgx-tests/src/tests/schema_tests.rs#L33-L58

On the other hand, I can totally see the use case of having one of these functions being used by multiple things...

I... think your strategy is good.

pgx/src/datum/sql_entity_graph/mod.rs
bitwalker: Ah nice! I thought I tried that, but `WrapErr` was the missing piece of the puzzle :)

bitwalker:
@Hoverbear I've made your recommended changes, and added a WIP commit that merges the `pgxsql` behavior under the `#[pgx(sql)]` handling as `#[pgx(sql = "SQL HERE")]` just to see what that looks like. We can modify that or remove it as you see fit, but once we're happy with it, I'll squash it into the the other changes.

The only thing with the merge that changes behavior, is that previously, `pgxsql` would result in the SQL generator rendering what would have been generated as a comment, followed by the overriding SQL code. Now, that no longer happens, and instead the overridding SQL is assumed to take over complete manual control of SQL generation for the decorated entity, same as the callback form. I'm not sure to what degree that is an issue for backwards compatibility, but we can probably special case the use of `#[pgx(sql = string)]` to behave exactly like `pgxsql` if it is essential to preserve that behavior. In any case, this was just a first stab at it, so I'm expecting some back and forth :)

pgx-macros/src/lib.rs
```diff
@@ -958,10 +970,19 @@ enum DogNames {
 ```
 
 */
-#[proc_macro_derive(PostgresHash)]
+#[proc_macro_derive(PostgresHash, attributes(pgx))]
 pub fn postgres_hash(input: TokenStream) -> TokenStream {
     let ast = parse_macro_input!(input as syn::DeriveInput);
-    impl_postgres_hash(ast).into()
+    let to_sql_config = match sql_entity_graph::ToSqlConfig::from_attributes(ast.attrs.as_slice()) {
```
Hoverbear, line 976: Do you think we could move this into the `impl_postgres_hash` function? I think it'd be more testable. (I do aspire to get these well tested one day...)

pgx-macros/src/lib.rs
```diff
@@ -937,10 +940,19 @@ enum DogNames {
 ```
 
 */
-#[proc_macro_derive(PostgresOrd)]
+#[proc_macro_derive(PostgresOrd, attributes(pgx))]
 pub fn postgres_ord(input: TokenStream) -> TokenStream {
     let ast = parse_macro_input!(input as syn::DeriveInput);
-    impl_postgres_ord(ast).into()
+    let to_sql_config = match sql_entity_graph::ToSqlConfig::from_attributes(ast.attrs.as_slice()) {
```
Hoverbear, line 946: Do you think we could move this into the `impl_postgres_ord` function? I think it'd be more testable. (I do aspire to get these well tested one day...)

pgx-tests/src/tests/schema_tests.rs
```diff
@@ -4,14 +4,68 @@ use pgx::*;
...
     use serde::{Deserialize, Serialize};
 
     #[pg_extern]
     fn func_in_diff_schema() {}
 
+    #[pg_extern]
+    #[pgx(sql = false)]
+    fn func_elided_from_schema() {}
+
+    #[pg_extern]
```
Hoverbear, line 18: This attribute macro already has several attributes, I think `#[pg_extern(sql = ...)]` is a reasonable addition to the mix and lets us avoid having the dreaded double attributes.

pgx/src/datum/sql_entity_graph/mod.rs
```diff
@@ -64,6 +64,110 @@ pub trait ToSql {
...
+/// When `callback` has a value, the corresponding `ToSql` implementation should invoke the
+/// callback instead of performing their default behavior.
+#[derive(Default, Clone)]
+pub struct ToSqlConfigEntity {
+    pub enabled: bool,
+    pub callback: Option<ToSqlFn>,
+    pub content: Option<&'static str>,
+}
+impl ToSqlConfigEntity {
+    /// Given a SqlGraphEntity, this function converts it to SQL based on the current configuration.
```
Hoverbear, line 102: Thank you so much for doc comments. :)

eeeebbbbrrrr:
> I was thinking more about this PR and we have a `pgx_sql` functionality associated with `pg_extern`. I think we could converge it's functionality with this.
> 
> ```rust
> /// ```pgx_sql
> /// CREATE FUNCTION ...
> /// ```
> #[pg_extern]
> function dooper() {}
> ```
> 
> Could become
> 
> ```rust
> #[pg_extern(sql = "
> CREATE FUNCTION ...
> ")]
> function dooper() {}
> ```
> 
> So I wonder if we should change `to_sql` to `sql` in prep for that... I have asked @eeeebbbbrrrr

I'm not jazzed about it b/c then I'll have to make changes to ZomboDB, but I suppose it's better and more consistent with how Rust would want us to do things.

Hoverbear:
@eeeebbbbrrrr I will personally make the changes to ZDB for you if that helps. :)

bitwalker:
@Hoverbear Any recommendations on getting to the bottom of the test failures? I can't seem to get any kind of useful output explaining what's going on either locally or in the CI runs, and I'm not sure why. As best I can tell, something in the test framework is swallowing output, but I haven't been able to find where yet. Have you run into this before?

My latest local run looks like this:

```
❯ RUST_BACKTRACE=1 PATH=$HOME/.pgx/14.1/pgx-install/bin:$PATH cargo test --all --features "pg14 pg_test" --no-default-features
warning: pgx-parent v0.2.6 (/Users/paulschoenfelder/src/github.com/timescale/pgx) ignoring invalid dependency `cargo-pgx` which is missing a lib target
   Compiling libc v0.2.112
<snip>
   Compiling pgx-pg-sys v0.2.6 (/Users/paulschoenfelder/src/github.com/timescale/pgx/pgx-pg-sys)
   Compiling pgx-macros v0.2.6 (/Users/paulschoenfelder/src/github.com/timescale/pgx/pgx-macros)
   Compiling cargo-pgx v0.2.6 (/Users/paulschoenfelder/src/github.com/timescale/pgx/cargo-pgx)
warning: field is never read: `name_tab`
   --> pgx-pg-sys/build.rs:339:5
    |
339 |     name_tab: HashMap<String, usize>,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default

warning: field is never read: `item_offset_tab`
   --> pgx-pg-sys/build.rs:341:5
    |
341 |     item_offset_tab: Vec<Option<usize>>,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: `pgx-pg-sys` (build script) generated 2 warnings
   Compiling pgx v0.2.6 (/Users/paulschoenfelder/src/github.com/timescale/pgx/pgx)
warning: field is never read: `status_code`
  --> pgx/src/spi.rs:61:5
   |
61 |     status_code: SpiOk,
   |     ^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: `pgx` (lib) generated 1 warning
   Compiling pgx-tests v0.2.6 (/Users/paulschoenfelder/src/github.com/timescale/pgx/pgx-tests)
warning: `pgx` (lib test) generated 1 warning (1 duplicate)
   Compiling pgx-parent v0.2.6 (/Users/paulschoenfelder/src/github.com/timescale/pgx)
    Finished test [unoptimized + debuginfo] target(s) in 1m 01s
     Running unittests (target/debug/deps/cargo_pgx-9bf22e45fc25ef4e)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests (target/debug/deps/pgx-896f2e0d717f0a47)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests (target/debug/deps/pgx_macros-58ea9063254925f7)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests (target/debug/deps/pgx_parent-78081190ec63d461)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests (target/debug/deps/pgx_pg_sys-02a95c8bfbc2d18f)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests (target/debug/deps/pgx_tests-1aa90d7500c05cf6)

running 158 tests
building extension with features `pg14 pg_test`
"cargo" "build" "--features" "pg14 pg_test" "--no-default-features"
 <snip>
   Compiling pgx-utils v0.2.6 (/Users/paulschoenfelder/src/github.com/timescale/pgx/pgx-utils)
   Compiling proc-macro-crate v1.1.0
   Compiling pgx-pg-sys v0.2.6 (/Users/paulschoenfelder/src/github.com/timescale/pgx/pgx-pg-sys)
   Compiling pgx-macros v0.2.6 (/Users/paulschoenfelder/src/github.com/timescale/pgx/pgx-macros)
warning: field is never read: `name_tab`
   --> pgx-pg-sys/build.rs:339:5
    |
339 |     name_tab: HashMap<String, usize>,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default

warning: field is never read: `item_offset_tab`
   --> pgx-pg-sys/build.rs:341:5
    |
341 |     item_offset_tab: Vec<Option<usize>>,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: `pgx-pg-sys` (build script) generated 2 warnings
   Compiling pgx v0.2.6 (/Users/paulschoenfelder/src/github.com/timescale/pgx/pgx)
warning: field is never read: `status_code`
  --> pgx/src/spi.rs:61:5
   |
61 |     status_code: SpiOk,
   |     ^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: `pgx` (lib) generated 1 warning
   Compiling pgx-tests v0.2.6 (/Users/paulschoenfelder/src/github.com/timescale/pgx/pgx-tests)
    Finished dev [unoptimized + debuginfo] target(s) in 58.74s

installing extension
     Copying control file to `/Users/paulschoenfelder/.pgx/14.1/pgx-install/share/postgresql/extension/pgx_tests.control`
     Copying shared library to `/Users/paulschoenfelder/.pgx/14.1/pgx-install/lib/postgresql/pgx_tests.so`
 Discovering SQL entities
  Discovered 289 SQL entities: 29 schemas (2 unique), 248 functions, 9 types, 1 enums, 2 sqls, 0 ords, 0 hashes, 0 aggregates
running SQL generator
"/Users/paulschoenfelder/src/github.com/timescale/pgx/target/debug/sql-generator" "--sql" "/Users/paulschoenfelder/.pgx/14.1/pgx-install/share/postgresql/extension/pgx_tests--1.0.sql"
     Copying extension schema file to `/Users/paulschoenfelder/.pgx/14.1/pgx-install/share/postgresql/extension/pgx_tests--1.0.sql`
    Finished installing pgx_tests
The files belonging to this database system will be owned by user "paulschoenfelder".
This user must also own the server process.

The database cluster will be initialized with locale "en_US.UTF-8".
The default database encoding has accordingly been set to "UTF8".
The default text search configuration will be set to "english".

Data page checksums are disabled.

creating directory /Users/paulschoenfelder/src/github.com/timescale/pgx/target/pgx-test-data-14 ... ok
creating subdirectories ... ok
selecting dynamic shared memory implementation ... posix
selecting default max_connections ... 100
selecting default shared_buffers ... 128MB
selecting default time zone ... America/New_York
creating configuration files ... ok
running bootstrap script ... ok
performing post-bootstrap initialization ... test tests::anyarray_tests::tests::pg_test_anyarray_arg has been running for over 60 seconds
test tests::array_tests::tests::pg_test_array_deny_nulls has been running for over 60 seconds
test tests::array_tests::tests::pg_test_count_nulls has been running for over 60 seconds
test tests::array_tests::tests::pg_test_count_true has been running for over 60 seconds
test tests::array_tests::tests::pg_test_count_true_sliced has been running for over 60 seconds
test tests::array_tests::tests::pg_test_optional_array has been running for over 60 seconds
test tests::array_tests::tests::pg_test_optional_array_with_default has been running for over 60 seconds
test tests::array_tests::tests::pg_test_return_text_array has been running for over 60 seconds
test tests::array_tests::tests::pg_test_return_zero_length_vec has been running for over 60 seconds
test tests::array_tests::tests::pg_test_serde_serialize_array has been running for over 60 seconds
ok
syncing data to disk ... ok

initdb: warning: enabling "trust" authentication for local connections
You can change this by editing pg_hba.conf or using the option -A, or
--auth-local and --auth-host, the next time you run initdb.

Success. You can now start the database server using:

    /Users/paulschoenfelder/.pgx/14.1/pgx-install/bin/pg_ctl -D /Users/paulschoenfelder/src/github.com/timescale/pgx/target/pgx-test-data-14 -l logfile start

Stopping Postgres

error: test failed, to rerun pass '-p pgx-tests --lib'
```

Hoverbear:
@bitwalker I have encountered this too, try `cargo pgx run` and then `create extension $EXT` and see if it gives you a better error?

bitwalker:
@Hoverbear Thanks, that helped me find one typo that was causing the extension to fail to load, and after that a re-run seemed to get me further than I had before, which brought me to a failure due to an invalid code block in my doc comments. Fixing _that_ though, brings me back to seemingly where I started, but this time using `cargo pgx run`, etc., works as expected when creating the extension, but trying to run one of the test functions manually from the shell gives me an error like `could not access file "$libdir/pgx_tests": No such file or directory`, which I think makes sense, because the extension dylib is actually versioned in the install directory, i.e. `pgx_tests-1.0.so`; but that's completely unrelated to my changes, so I'm a bit confused as to what might be going on there.

In any case, I've pushed the fixes for the two issues that are related to my changes, and hoping that when run in the CI environment things pass, and that the failures locally are just an environmental thing.

Hoverbear:
@bitwalker You're doing great, Thanks so much for this!

bitwalker:
@Hoverbear I think once this PR gets merged it'll stop hassling you, I hope haha. At least it looks like at least this run is finally going to be green 🤞

Hoverbear:
Yessssss!!! I'll give it a final lookover and get it merged. Thanks. :)

pgx-tests/src/tests/schema_tests.rs
```diff
@@ -4,14 +4,66 @@ use pgx::*;
...
     use pgx::*;
     use serde::{Deserialize, Serialize};
 
     #[pg_extern]
     fn func_in_diff_schema() {}
 
+    #[pg_extern(sql = false)]
+    fn func_elided_from_schema() {}
+
+    #[pg_extern(sql = "generate_function")]
```
Hoverbear, line 17: It should actually be valid to do like `#[pg_extern(sql = generate_function)]` which would remove any ambiguity about if it's SQL or a function...

pgx-utils/src/sql_entity_graph/pg_extern/attribute.rs
```diff
@@ -129,6 +111,21 @@ impl Parse for Attribute {
                 let _bracket = syn::bracketed!(content in input);
                 Self::Requires(content.parse_terminated(PositioningRef::parse)?)
             }
+            "sql" => {
+                use syn::Lit;
+
+                let _eq: Token![=] = input.parse()?;
+                match input.parse::<syn::Lit>()? {
```
Hoverbear, line 118: I imagine this could also handle a `syn::Path`.

pgx-utils/src/sql_entity_graph/to_sql.rs
```diff
@@ -0,0 +1,149 @@
...
+        Self {
+            enabled: true,
+            callback: None,
+            content: None,
+        }
+    }
+}
+
+const INVALID_ATTR_CONTENT: &str =
+    "expected `#[pgx(sql = content)]`, where `content` is a boolean, string, or path to a function";
```
Hoverbear, line 51: You mention a function path here but I don't see code to handle it. :(

pgx-utils/src/sql_entity_graph/pg_extern/mod.rs
```diff
@@ -207,7 +219,7 @@ impl ToTokens for PgExtern {
         let name = self.name();
         let schema = self.schema();
         let schema_iter = schema.iter();
-        let extern_attrs = self.extern_attrs();
+        let extern_attrs = self.attrs.iter().collect::<Punctuated<_, Token![,]>>();
```
Hoverbear, line 222: Clever. :)

pgx-utils/src/sql_entity_graph/to_sql.rs
bitwalker: It's on line 79, we expect the path to be provided as a string

pgx-tests/src/tests/schema_tests.rs
bitwalker: Good point, I think that makes more sense, I'll adjust it accordingly

pgx-utils/src/sql_entity_graph/pg_extern/attribute.rs
bitwalker: When the path was expected as a string, that wasn't the case, but with your other suggested change, this will be one of the affected parses.

pgx-tests/src/tests/schema_tests.rs
bitwalker: @Hoverbear I just realized that the reason I used a string literal here, is that `MetaNameValue` expects the value part of `key = value` to be a `syn::Lit`. Your suggestion isn't supported AFAIK, unless you choose to avoid `Attribute::parse_meta` for parsing the attribute contents

pgx-macros/src/lib.rs
```diff
@@ -402,6 +402,7 @@ Optionally accepts the following attributes:
 * `parallel_unsafe`: Corresponds to [`PARALLEL UNSAFE`](https://www.postgresql.org/docs/current/sql-createfunction.html).
 * `parallel_restricted`: Corresponds to [`PARALLEL RESTRICTED`](https://www.postgresql.org/docs/current/sql-createfunction.html).
 * `no_guard`: Do not use `#[pg_guard]` with the function.
+* `sql`: Same arguments as `#[pgx(sql = ..)]`
```
Hoverbear, line 405:
```suggestion
* `sql`: Same arguments as [`#[pgx(sql = ..)]`](macro@pgx).
```

pgx-macros/src/lib.rs
```diff
@@ -685,7 +686,10 @@ Optionally accepts the following attributes:
 * `pgvarlena_inoutfuncs(some_in_fn, some_out_fn)`: Define custom in/out functions for the `PgVarlena` of this type.
```
Hoverbear, line 687:
```suggestion
* `sql`: Same arguments as [`#[pgx(sql = ..)]`](macro@pgx).
```

pgx-macros/src/lib.rs
```diff
@@ -913,10 +917,12 @@ enum DogNames {
 ```
```
Hoverbear, line 918:
```suggestion
Optionally accepts the following attributes:

* `sql`: Same arguments as [`#[pgx(sql = ..)]`](macro@pgx).
```

pgx-macros/src/lib.rs
```diff
@@ -937,10 +943,12 @@ enum DogNames {
...
```
Hoverbear, line 944:
```suggestion
Optionally accepts the following attributes:

* `sql`: Same arguments as [`#[pgx(sql = ..)]`](macro@pgx).
```

pgx-macros/src/lib.rs
```diff
@@ -958,10 +966,12 @@ enum DogNames {
...
```
Hoverbear, line 967:
```suggestion
Optionally accepts the following attributes:

* `sql`: Same arguments as [`#[pgx(sql = ..)]`](macro@pgx).
```

pgx-macros/src/lib.rs
```diff
@@ -991,15 +1001,26 @@ pub fn pg_aggregate(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
-Used outside of a [`#[pg_aggregate]`](macro@pg_aggregate), this does nothing.
+## Usage for configuring SQL generation
+
+This attribute can be used to control the behavior of the SQL generator on a decorated item,
+e.g. `#[pgx(sql = false)]`
+
+Currently `sql` can be provided one of the following:
+
+* Disable SQL generation with `#[pgx(sql = false)]`
+* Call custom SQL generator function with `#[pgx(sql = "path::to_function")]`
```
Hoverbear, line 1019:
```suggestion
* Call custom SQL generator function with `#[pgx(sql = path::to_function)]`
```

pgx-tests/src/tests/schema_tests.rs
Hoverbear: Hmmm I've seen this before in https://docs.rs/tracing/latest/tracing/attr.instrument.html#examples-1 ...

Hoverbear: I guess part of my concern (and want for that distinction) is I could very well see a bug where this code:

https://github.com/zombodb/pgx/blob/a9d18c59cf0e3932f7cf163cc2ec36318f271ad2/pgx-utils/src/sql_entity_graph/to_sql.rs#L78-L92

Parses some SQL as a Rust Path for some reason, and creates weird behavior.

bitwalker: Yeah in that case, they are abandoning `Attribute::parse_meta` in favor of `Attribute::parse_args`, which allows them to use their own grammer for the attribute contents by manually parsing them. It's more effort but worth it when you need to support something that doesn't fit into the standard structured meta arguments approach. We can do something similar if you think it's worth it.

Hoverbear: I would definitely prefer a lack of ambiguity that if you have time, I know I've already asked a lot of you so if you are pressed on time I can wrap it up.

bitwalker: I should have a commit that implements this ready shortly, just FYI

bitwalker:
Once you're happy with this, just let me know and I'll squash commits, just leaving them broken up while we iterate on things

Hoverbear:
Newest commit [344f685](https://github.com/zombodb/pgx/pull/410/commits/344f685787f83d8d3bc72d3336d7b0ba21ba3f61) looks good to me. :)

bitwalker:
Awesome! Once the tests pass, I'll go ahead and squash then :)

pgx-tests/src/tests/schema_tests.rs
```diff
@@ -4,14 +4,66 @@ use pgx::*;
...
+        entity: &SqlGraphEntity,
+        _context: &PgxSql,
+    ) -> Result<String, Box<dyn std::error::Error + Send + Sync + 'static>> {
+        if let SqlGraphEntity::Function(ref func) = entity {
+            Ok(format!("\
+                CREATE OR REPLACE FUNCTION test_schema.\"func_generated_with_custom_name\"() RETURNS void\n\
+                LANGUAGE c /* Rust */\n\
+                AS 'MODULE_PATHNAME', '{unaliased_name}_wrapper';\
+                ",
+                unaliased_name = func.unaliased_name,
```
Hoverbear, line 45: 😅 I'm gonna need to clean up this API a bit.

Hoverbear:
I think my only remaining nit about this PR is we don't leave an comment anchor in the SQL noting that, for example, the content was overridden or skipped. I think I'll make a follow up PR for this since I've asked enough of you. :)

pgx-tests/src/tests/schema_tests.rs
bitwalker: I had a similar thought when I was thinking through how to expose stuff to those functions haha. In my perfect world, users of this would get access to a hypothetical `SqlBuilder` API that let you replicate this using the builder pattern:

```rust
fn generate_function(entity: &SqlGraphEntity, builder: &mut SqlBuilder, _context: &PgxSql) -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
    if let SqlGraphEntity::Function(ref f) = entity {
        // get_or_create lets you modify existing entities after initial construction, but before the builder is consumed
        // returns a entity builder that holds a mutable reference to the parent builder
        // when the entity is finalized, only then is the entity added to the parent builder
        let mut func = builder.get_or_create_function(f.name);
        func
          .args(&[]) // defaults to no arguments if not set
          .returns("void") // defaults to 'void' if not set
          .replace(true) // indicates the builder should use CREATE OR REPLACE
          .lang(FunctionLanguage::Rust) // lowers to LANGUAGE c /* Rust */
          .as_nif("MODULE_PATHNAME", format!("{}_wrapper", func.unaliased_name)) // indicates the function is a Natively Implemented Function from the given module
          .build()?; // adds the built function to the parent builder
        Ok(())
    } else {
        panic!("expected extern function entity, got {:?}", entity);
    }
}
```

It'd be a big undertaking to cover all possible entities, but could probably start by just covering the cases that pgx itself requires, expose some kind of escape hatch akin to `extension_sql!`, and expand the API with specific entities as needed. Not sure if it is worth it or not, but would be a lot nicer and less error prone to work with!

bitwalker:
@Hoverbear Yeah I think that would be fairly easy to add. I also think it'd be nice to ensure some easy mistakes are caught by pgx when using SQL strings returned from the callback or given as a literal, e.g. that any non-comment lines are terminated with a semicolon. I didn't add any such checks, since I wasn't sure what degree of validation we would want (though I did make an effort to ensure newlines are prepended when rendering the strings to help readability of the final output), but it is almost certainly something that will help avoid unnecessary bug reports from users.

I agree though that it probably makes sense to merge this and make some of those improvements in follow-on PRs just to keep things tighter for review, just let me know what you need from me, and I'll try to make some time :)

Hoverbear:
Let's merge this as it is now, I'm just prepping a follow up for the anchor comments as I had a particular way I wanted to do it. :)

PR #417, opened 2022-02-02
Flush to tar `stdio`

Hoverbear:
While running `cargo pgx init` users may experience errors while passing the downloaded to `tar` via `stdio`, this is because we don't currently flush.


eeeebbbbrrrr:
LGTM.  Thanks, @Hoverbear!

jcdickinson:
The fix works for me, thanks!

PR #419, opened 2022-02-03
Expand pgx::Internal API

Hoverbear:
Expand the `pgx::Internal` API to include some familiar faces from `Option` and improve usability. Check out the tests to see how they work!

### DIFF
--- a/pgx/src/datum/internal.rs
+++ b/pgx/src/datum/internal.rs
@@ -21,6 +21,7 @@ use crate::{pg_sys, FromDatum, IntoDatum, PgMemoryContexts};
 ///
 /// We make no guarantees about what the internal [pg_sys::Datum] actually points to in memory, so
 /// it is your responsibility to ensure that what you're casting it to is really what it is.
+#[derive(Default)]
 pub struct Internal(Option<pg_sys::Datum>);
 
 impl Internal {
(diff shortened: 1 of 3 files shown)


pgx/src/datum/internal.rs
```diff
@@ -34,7 +35,13 @@ impl Internal {
         ))
     }
 
-    /// Return a reference to the memory pointed to by this [Internal], as `Some(&T)`, unless the
+    /// Returns if the internal value is initialized. If false, this is a null pointer.
```
eeeebbbbrrrr, line 38: Returns true if

pgx/src/datum/internal.rs
```diff
@@ -43,10 +50,28 @@ impl Internal {
     /// your responsibility.
     #[inline]
     pub unsafe fn get<T>(&self) -> Option<&T> {
-        self.0.map(|datum| (datum as *const T).as_ref()).flatten()
+        self.0.and_then(|datum| (datum as *const T).as_ref())
```
eeeebbbbrrrr, line 53: Does this change our minim rust version?  Not that I care but if it does we need to document it

eeeebbbbrrrr (dismissed):

pgx/src/datum/internal.rs
Hoverbear: I don't think so, these have been around awhile.

PR #420, opened 2022-02-04
Fix install copy issues

JamesGuthrie:
A couple of erroneous early returns cause copying of artifacts to be
silently skipped.


JamesGuthrie:
I've spent the last couple of days thinking that I'm crazy 😂

Hoverbear (approved):

Hoverbear:
Oh nooooooooo. Glad you found it, thanks!

PR #429, opened 2022-02-09
Update versions rework

Hoverbear:
The existing update-versions script seems to have fallen behind current with `cargo-release`.

After discussion we decided to use `sed`. This does that.


Hoverbear:
Wow I really bungled that one, forgot to update the versions.

PR #433, opened 2022-02-10
Use correctly spanned fcinfo

### LINKED ISSUE #432: Regression: v0.3.0 can't compile #[pg_extern] functions with the `FunctionCallInfo` arg
For example:

```rust
#[pg_extern]
fn foo_func(_fcinfo: pg_sys::FunctionCallInfo) -> PgBox<pg_sys::IndexAmRoutine> {
   todo!()
}
```

results in this error:

```
error[E0425]: cannot find value `_fcinfo` in this scope
  --> foo.rs:35:20
   |
35 | fn foo_func(_fcinfo: pg_sys::FunctionCallInfo) -> PgBox<pg_sys::IndexAmRoutine> {
   |                    ^^^^^^^ help: a local variable with a similar name exists: `fcinfo`
```

Renaming the variable to `fcinfo` compiles, but then, in my case, I get a warning that it's not used.  Which is true and why I declared it with the underscore prefix.

This worked just fine in <v0.3.0

Hoverbear:
Fixes #432 .

Turns out I had the wrong idea for the Function rewriter.

### DIFF
--- a/pgx-macros/src/rewriter.rs
+++ b/pgx-macros/src/rewriter.rs
@@ -723,30 +723,7 @@ impl FunctionSignatureRewriter {
 
         let mut stream = proc_macro2::TokenStream::new();
         let mut i = 0usize;
-        let mut fcinfo_ident = None;
-
-        // Get the fcinfo ident, if it exists.
-        // We do this because we need to get the **right** ident, if it exists, so Rustc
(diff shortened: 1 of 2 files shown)


Hoverbear:
Confirming this fixes the specific reproduction @eeeebbbbrrrr found.

PR #441, opened 2022-02-15
`cargo-pgx` can generate schemas

Hoverbear:
This PR reworks how we do SQL (again). It amounts to a lot of code movement but very little actual changes.

It does however amount to several API breaks.

# What's different?

* `pgx::datum::sql_entity_graph` and `pgx_utils::sql_entity_graph` are merged and present in `pgx_utils::sql_entity_graph` together.
* `pgx_utils` is now exported from `pgx` at `pgx::utils`, this is primarily to expose `pgx_utils::sql_entity_graph` without requiring users have `pgx_utils` in their `Cargo.toml`.
* The `sql-generator` binary, `.cargo/pgx-linker-script.sh` and related `.cargo/config` settings are no longer needed. The old `.cargo/config` from 0.1.2x is back.
* `pgx_graph_magic!()` now expands with `__pgx_source_only_sql_mappings` and `__pgx_typeid_sql_mappings` symbols, allowing `cargo-pgx` to get the appropriate SQL mappings.
* `cargo pgx schema` now:
  1. Uses `--message-format=json` on some `cargo` invocations and captures that output. 
  2. Processes that JSON to find the `OUT_DIR` used by the relevant `pgx_pg_sys` build.
  3. It uses that `OUT_DIR` data to read, generate, then build (with `rustc`) a 'stub' shared object. (This gets cached!)
  4. `dlopen`s that shared object.
  5. `dlopen`s the extension.
  6. Calls the `pgx_utils::sql_entity_graph::PgxSql` builders as the `sql-generator` used to.
* `cargo pgx test` and `cargo pgx install` had several changes to make them more appropriately and cleanly capture and pass build data for this new process.
* `cargo pgx schema`, `cargo pgx install`, and `cargo pgx package` all had outputs changed.
* Most `sql_graph_entity` related functions got a `#[doc(hidden)]` attribute to clean up their docs.
* `RUST_LOG` env vars get correctly passed in `pgx test`/

# UX

Now generating a schema, or packaging/installing/testing an extension looks like so:

image

Users can optionally enable debug or logging still, here's isolated tracing symbols for a `cargo pgx schema` run where stubs did not exist:

image

And now where they do:

image

Because we use `--message-format=json-render-diagnostics` Cargo still outputs nice errors:

image

# Performance

This PR has a fairly dramatic impact on build performance!

Using this script:

```bash
#! /usr/bin/env bash
set -e

git checkout master
cargo install --path cargo-pgx &> /dev/null
pushd pgx-tests/
[...]

PR #455, opened 2022-02-25
Generate pgx-pg-sys bindings based on features

### LINKED ISSUE #452: `pgx-pg-sys` should only generate version-specific bindings
Right now `pgx-pg-sys`'s build.rs consumes a considerable amount of build time.

Here's an example from the `aggregate` example (this is after #441) from `cargo +nightly build -Z timings`

image

The build script consumes about 1/3 of the total build time, and doesn't happen in parallel with other things.

One way to reduce this time is reducing the amount of `bindgen` we do. Another way is reducing the amount of IO that's done.

Currently `pgx-pg-sys` creates bindings for **all** supported Postgres versions:

https://github.com/zombodb/pgx/blob/51f604668a3a8723bfe7e2587212a10c1491e955/pgx-pg-sys/build.rs#L85-L89

The work from #441 revealed that different feature flags create a different `OUT_DIR` in the `target/build/` folder, so a `cargo build --features pg12` and a `cargo build --features pg13` have different `OUT_DIR`s, meaning we only need to bindgen for the specific version flag passed:

image

Doing this would reduce both the amount of `bindgen`'ing and the amount of IO done considerably.

## The docs.rs problem

This is done partially for **docs.rs** reasons, as it doesn't run `cargo pgx init` to create the development postgres installs which the build script depends on.

The build script gets exited early on docs.rs: https://github.com/zombodb/pgx/blob/51f604668a3a8723bfe7e2587212a10c1491e955/pgx-pg-sys/build.rs#L52-L54

While on docs.rs we use the bindings generated in `pgx-pg-sys/src/pg*.rs` instead of generated bindings in the `$OUT_DIR`: https://github.com/zombodb/pgx/blob/51f604668a3a8723bfe7e2587212a10c1491e955/pgx-pg-sys/src/lib.rs#L60-L65

Thus, we still need a way for `pgx-pg-sys` to update **all** the `pgx-pg-sys/src/` bindings before a release.

We currently have a `PGX_PG_SYS_SKIP_BINDING_REWRITE` env var that disables writing to the `pgx-pg-sys/src` dir. This was added to support Nix, as the directory isn't writable while sandboxed.

https://github.com/zombodb/pgx/blob/51f604668a3a8723bfe7e2587212a10c1491e955/pgx-pg-sys/build.rs#L107-L114

I propose removing this and instead having `PGX_PG_SYS_GENERATE_BINDINGS_FOR_RELEASE` which we use like so at the end of https://github.com/zombodb/pgx/blob/master/update-versions.sh:

```bash
PGX_PG_SYS_GENERATE_BINDINGS_FOR_RELEASE=1 cargo build -p pgx-pg-sys
```

Then, in `pgx-pg-sys`, we can detect this flag and update the bindings as needed prior to a release.

> We only need these to update prior to release, since they are only used for docs.rs.

Hoverbear:
This resolves #452 .

Please note these performance measurements are done on a 32 core machine which was not previously CPU bound, and still is not CPU bound, so the gains between "generating all bindings" and "generating one binding" is quite small. On a machine with less cores (or doing other things) these gains would be more noticeable.

Here's with generating all bindings:

image

While doing just one feature:

image

> Notice the dramatic reduction in CPU!

As discussed in #452, this adjusts the `update-versions.sh` script to make it regenerate the bindings (and build all packages to make sure they aren't broken)


Hoverbear:
Here's comparing the CPU load for a ZomboDB build between #441 and this PR (which builds atop it since it's merged to `develop`)

Before:

image

After:

image

eeeebbbbrrrr:
LGTM.  However, please retarget this PR to the `develop` branch.

PR #475, opened 2022-03-08
Workspace support

Hoverbear:
This adds package-id support (`--package`) to most `cargo-pgx` commands, it also supports `--manifest-path`, like `cargo`.


Hoverbear:
We did not end up adding `--workspace` as the only one that made sense was `cargo pgx test --workspace` but that had numerous restrictions and channels since we couldn't support concepts like two extensions having differing pgversions for their feature flags, etc.

Hoverbear:
Turns out we are merging this for 0.4.0!

PR #479, opened 2022-03-14
Add appendinfo.h and regenerate bindings

einarmo:
Seems to have upgraded bindings to 12.10 from 12.9, among other things.

appendinfo.h contains a couple of methods used for foreign data wrappers. In particular `add_row_identity_var` is referenced from the official postgres documentation, and is nice to have here.

The file exists for pg12 and pg13 as well, but it's just the `add_row_identity_var` method first added in pg14 which seem generally useful here. I think only that method is needed for external applications in practice.


Hoverbear:
@einarmo this looks good overall, however it is kind of our policy to include the same things wherever possible.

Could I ask you to add the includes for `pg12` and `pg13` as well?

PR #485, opened 2022-03-16
Update the nix template CI with the one from pl/Rust

Hoverbear:
Steal the one from pl/Rust.


Hoverbear:
I'm not waiting for CI since this isn't tested on CI.

PR #488, opened 2022-03-17
Improve Internal perf

Hoverbear:
Thanks to a report by @jamessewell we were able to find some performance improvements which lead to us doing a bit more allocation than needed in some `Internal` type calls.


pgx/src/datum/into.rs
```diff
@@ -45,6 +45,23 @@ where
     }
 }
 
+impl<T, E> IntoDatum for Result<T, E>
```
eeeebbbbrrrr, line 48: This is clearly an accidental addition.

eeeebbbbrrrr:
Please target this to the `develop` branch.

pgx/src/datum/into.rs
Hoverbear: Can you tell what I was thinking about last night? Ha.

PR #490, opened 2022-03-21
These changes are a first go at a concept of cargo-pgx templates

jamessewell:
They replace the previous single shot bgworker config switch

All files are templated out of the ./src/templates directories

Possible to do:

- check (at compile time) that templates have a MVP list of files (./src/lib.rs, Cargo.toml etc..)
- Build (at compile time) a list of ./src/template directories to include (static atm)


cargo-pgx/Cargo.toml
```diff
@@ -1,41 +1,74 @@
 [package]
-name = "cargo-pgx"
```
Hoverbear, line 2: There seems to be quite a bit of unrelated changes in this file.

cargo-pgx/src/templates/aggregate/src/lib.rs
```diff
@@ -0,0 +1,62 @@
+use pgx::*;
+use std::collections::HashSet;
+
+pg_module_magic!();
+
+#[derive(Copy, Clone, Default, Debug)]
+pub struct {{ camel-case name }};
```
Hoverbear, line 7: It's a bit unfortunate how the template syntax is so similar to Rust :|

Hoverbear:
All in all I'm pretty pleased with this @jamessewell . :) In your commit you mentioned this was a first go, are you planning on making further changes or should I take it for a spin?

eeeebbbbrrrr:
Looks like this has bitrotted quite a bit.  @jamessewell if you wanted to bring this up to modern times I'd be happy to help get your vision over the finish line.  Until then, I'm closing this.

PR #505, opened 2022-03-25
`cargo pgx status` always defaults to `all`

eeeebbbbrrrr:
`cargo pgx status` should report the status of all pgx-managed Postgres versions.  It should not be dependent on what's in `Cargo.toml`.

### DIFF
--- a/cargo-pgx/src/command/status.rs
+++ b/cargo-pgx/src/command/status.rs
@@ -1,7 +1,7 @@
 // Copyright 2020 ZomboDB, LLC <zombodb@gmail.com>. All rights reserved. Use of this source code is
 // governed by the MIT license that can be found in the LICENSE file.
 
-use eyre::{eyre};
+use eyre::eyre;
 use owo_colors::OwoColorize;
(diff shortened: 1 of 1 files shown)


cargo-pgx/src/command/status.rs
```diff
@@ -34,23 +33,7 @@ impl CommandExecute for Status {
     fn execute(self) -> eyre::Result<()> {
         let pgx = Pgx::from_config()?;
 
-        let pg_version = match self.pg_version {
-            Some(s) => s,
-            None => {
```
Hoverbear, line 39: I think default to `"all"` here.

cargo-pgx/src/command/status.rs
```diff
@@ -1,14 +1,13 @@
 // Copyright 2020 ZomboDB, LLC <zombodb@gmail.com>. All rights reserved. Use of this source code is
 // governed by the MIT license that can be found in the LICENSE file.
 
-use eyre::{eyre, WrapErr};
+use eyre::{eyre};
```
Hoverbear, line 4:
```suggestion
use eyre::eyre;
```

PR #510, opened 2022-03-26
Add from timestamp from impls

jamessewell:
pgx::Timestamp from pg_sys::timestamp
pgx::TimestampWithTimeZone from pg_sys::timestamptz

### DIFF
--- a/pgx/src/datum/time_stamp.rs
+++ b/pgx/src/datum/time_stamp.rs
@@ -8,6 +8,13 @@ use time::{format_description::FormatItem, PrimitiveDateTime};
 
 #[derive(Debug)]
 pub struct Timestamp(time::PrimitiveDateTime);
+
+impl From<pg_sys::Timestamp> for Timestamp {
+    fn from(item: pg_sys::Timestamp) -> Self {
+        unsafe { Timestamp::from_datum(item as usize, false, pg_sys::TIMESTAMPOID).unwrap() }
+    }
+}
+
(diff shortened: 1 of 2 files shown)


Hoverbear (approved):

pgx/src/datum/time_stamp.rs
```diff
@@ -8,6 +8,13 @@ use time::{format_description::FormatItem, PrimitiveDateTime};
 
 #[derive(Debug)]
 pub struct Timestamp(time::PrimitiveDateTime);
+
+impl From<pg_sys::Timestamp> for Timestamp {
+    fn from(item: pg_sys::Timestamp) -> Self {
+        unsafe { Timestamp::from_datum(item as usize, false, pg_sys::TIMESTAMPOID).unwrap() }
```
Hoverbear, line 14: Seems these should be `TryFrom` given that you're unwrapping them?

jamessewell: I'm actually not sure about this. Because I'm passing false for `is_null` I can never get a `None` back, so the unwrap is fine.

What I'm not sure about is what a `pg_sys::Timestamp` which actually is null (as opposed to an actual Option) would look like and how you'd get it?

jamessewell: ZomboDB: Yah. That’ll never panic

I'm feeling more confident now!

PR #511, opened 2022-03-26
update copyright headers

eeeebbbbrrrr:
I put this against `develop` but I think we should also merge it into `master`.  No need for a new release, of course.

### DIFF
--- a/.licensure.yml
+++ b/.licensure.yml
@@ -22,7 +22,9 @@ licenses:
       - name: ZomboDB, LLC and Technology Concepts & Design, Inc.
         email: support@tcdi.com
     template: |
-      Copyright 2019-[year] [name of author]. All rights reserved. 
+      Portions Copyright 2019-2021 ZomboDB, LLC. 
+      Portions Copyright 2021-[year] Technology Concepts & Design, Inc. <support@tcdi.com>.
+      All rights reserved. 
       Use of this source code is governed by the [ident] license that can be found in the LICENSE file.
 comments:
   - columns: 80
--- a/.licensure.yml
+++ b/.licensure.yml
@@ -1,14 +1,55 @@
-author: ZomboDB, LLC
-ident: MIT
-email: zombodb@gmail.com
+change_in_place: true
 excludes:
   - Cargo.toml
-  - ".*\\.h"
-  - "\\.cargo/config"
+  - Cargo.lock
+  - LICENSE
+  - pgx-utils/assets/ansi.tmTheme
+  - ".envrc"
+  - ".gitignore"
+  - "flake\\..*"
+  - "logo.*"
+  - ".github/**/*"
+  - ".cargo/config"
   - "cargo-pgx/src/templates/*"
-  - ".github/workflows/*"
   - ".*\\.control"
   - ".*\\.md"
+  - ".*\\.nix"
   - ".*\\.yml"
-  - ".*\\.sh"
-  - load-order.txt
+licenses:
+  - files: any
+    ident: MIT
+    authors:
+      - name: ZomboDB, LLC and Technology Concepts & Design, Inc.
+        email: support@tcdi.com
+    template: |
+      Copyright 2019-[year] [name of author]. All rights reserved. 
+      Use of this source code is governed by the [ident] license that can be found in the LICENSE file.
+comments:
+  - columns: 80
+    extensions:
+      - rs
+      - c
+      - h
+    commenter:
+      type: line
+      comment_char: "//"
+      trailing_lines: 0
+  - columns: 120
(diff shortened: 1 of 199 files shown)


LICENSE
```diff
@@ -1,6 +1,8 @@
 MIT License
 
-Copyright (c) 2020 ZomboDB, LLC
+Portions Copyright 2019-2021 ZomboDB, LLC.
+Portions Copyright 2021-2022 Technology Concepts & Design, Inc. <support@tcdi.com>.
+All rights reserved.
```
Hoverbear, line 5: I think we are MIT, so this isn't true.

nix/templates/default/src/lib.rs
```diff
@@ -1,3 +1,7 @@
+// Portions Copyright 2019-2021 ZomboDB, LLC.  Portions Copyright 2021-2022
+// Technology Concepts & Design, Inc. <support@tcdi.com>.  All rights reserved.  Use of this
```
Hoverbear, line 2:
```suggestion
// Technology Concepts & Design, Inc. <support@tcdi.com>.  Use of this
```

.licensure.yml
```diff
@@ -1,14 +1,57 @@
...
+  - files: any
+    ident: MIT
+    authors:
+      - name: ZomboDB, LLC and Technology Concepts & Design, Inc.
+        email: support@tcdi.com
+    template: |
+      Portions Copyright 2019-2021 ZomboDB, LLC. 
+      Portions Copyright 2021-[year] Technology Concepts & Design, Inc. <support@tcdi.com>.
+      All rights reserved. 
+      Use of this source code is governed by the [ident] license that can be found in the LICENSE file.
```
Hoverbear, line 28: This line is obvious (literally all source code is guarded by its license) and doesn't need to be included.

.licensure.yml
```diff
@@ -1,14 +1,57 @@
...
-  - ".*\\.sh"
-  - load-order.txt
+licenses:
+  - files: any
+    ident: MIT
+    authors:
+      - name: ZomboDB, LLC and Technology Concepts & Design, Inc.
+        email: support@tcdi.com
+    template: |
+      Portions Copyright 2019-2021 ZomboDB, LLC.
```
Hoverbear, line 25: Copyright statements are typically one line.

.licensure.yml
```diff
@@ -1,14 +1,57 @@
...
+licenses:
+  - files: any
+    ident: MIT
+    authors:
+      - name: ZomboDB, LLC and Technology Concepts & Design, Inc.
+        email: support@tcdi.com
+    template: |
+      Portions Copyright 2019-2021 ZomboDB, LLC. 
+      Portions Copyright 2021-[year] Technology Concepts & Design, Inc. <support@tcdi.com>.
+      All rights reserved.
```
Hoverbear, line 27: I found previous artifacts of this statement in WebRTC's repo: https://webrtc.github.io/webrtc-org/license/software/

I did note that this may cause some confusion to some folks: https://github.com/PowerShell/PowerShell/issues/2129

.licensure.yml
```diff
@@ -1,14 +1,57 @@
...
   - ".*\\.yml"
-  - ".*\\.sh"
-  - load-order.txt
+licenses:
+  - files: any
+    ident: MIT
+    authors:
+      - name: ZomboDB, LLC and Technology Concepts & Design, Inc.
+        email: support@tcdi.com
+    template: |
```
Hoverbear, line 24: Noting that the MIT license requires no such header for either the original or any derivative works: https://choosealicense.com/licenses/mit/

So this could be removed entirely.

.licensure.yml
```diff
@@ -1,14 +1,57 @@
...
   - ".*\\.md"
+  - ".*\\.nix"
   - ".*\\.yml"
-  - ".*\\.sh"
-  - load-order.txt
+licenses:
+  - files: any
+    ident: MIT
+    authors:
+      - name: ZomboDB, LLC and Technology Concepts & Design, Inc.
```
Hoverbear, line 22: According to the header below showing "Portions Copyright 2019-2021 ZomboDB, LLC." then later "Portions Copyright 2021-[year] Technology Concepts & Design, Inc. <support@tcdi.com>.", it seems incorrect to list `ZomboDB, LLC` as a current copyright holder? Should `ZomboDB, LLC` have their copyright header portion changed to include `2022`?

eeeebbbbrrrr: This is the only one that needs to be fixed, and I'll do that right now.  This `licensure` tool is quite terrible and I hadn't had any caffeine this morning when I pushed this.

PR #512, opened 2022-03-28
Add pg_operator.h

einarmo:
Really just the `FormData_pg_operator` struct and a few associated methods, which is useful when looking up operators from queries.

I'm noticing that regenerating bindings has a tendency to make a few OS and even user related changes. Not sure if you care about that.


Hoverbear:
Thanks for this @einarmo !

And yes, it can't be helped, this is why we don't ask contributors to update the bindings anymore and instead it's part of the release step.

It seems you've generated those bindings again though anyways, but it's not a problem since they'll be regenerated.

Hoverbear (approved):

PR #513, opened 2022-03-28
Pass RUSTFLAGS to rustc on schema generate

Hoverbear:
On some operating systems, such as Alpine, `rustc` requires special arguments to be able to generate a `cdylib`. (Eg: https://github.com/zombodb/zombodb/blob/f5848c652786e18dc2d1029f2a0747b4b1da0e3b/docker-build-system/zombodb-build-alpine-3.12/Dockerfile#L28)

This patch changes `cargo-pgx schema` to use any existing `RUSTFLAGS` in the environment.

**Please note it does not read `.cargo/config`.**


eeeebbbbrrrr:
LGTM.  Merged

PR #526, opened 2022-04-02
#[pg_guard] cleanup

eeeebbbbrrrr:
This centralizes all the code `#[pg_guard]` otherwise generates into a common function, which takes a closure that is a wrapper around the desired internal Postgres function call.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,3 +24,6 @@ members = [
     "pgx-examples/strings",
     "pgx-examples/triggers",
 ]
+
+[profile.dev.build-override]
+opt-level = 3
--- a/pgx-pg-sys/build.rs
+++ b/pgx-pg-sys/build.rs
@@ -14,7 +14,7 @@ use eyre::{eyre, WrapErr};
 use pgx_utils::pg_config::{PgConfig, PgConfigSelector, Pgx};
 use pgx_utils::prefix_path;
 use pgx_utils::rewriter::PgGuardRewriter;
-use quote::quote;
+use quote::{quote, ToTokens};
 use rayon::prelude::*;
 use std::{
     collections::{HashMap, HashSet},
@@ -135,7 +135,7 @@ fn main() -> color_eyre::Result<()> {
                 .wrap_err_with(|| format!("bindgen failed for pg{}", major_version))?;
 
             let oids = extract_oids(&bindgen_output);
-            let rewritten_items = rewrite_items(bindgen_output)
+            let rewritten_items = rewrite_items(&bindgen_output)
                 .wrap_err_with(|| format!("failed to rewrite items for pg{}", major_version))?;
 
             let dest_dirs = if std::env::var("PGX_PG_SYS_GENERATE_BINDINGS_FOR_RELEASE")
@@ -192,10 +192,8 @@ fn write_rs_file(
     file: &PathBuf,
     header: proc_macro2::TokenStream,
 ) -> eyre::Result<()> {
-    let contents = quote! {
-        #header
-        #code
-    };
+    let mut contents = header;
+    contents.extend(code);
 
     std::fs::write(&file, contents.to_string())?;
     rust_fmt(&file)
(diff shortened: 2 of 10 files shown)


Cargo.toml
```diff
@@ -24,3 +24,6 @@ members = [
     "pgx-examples/strings",
     "pgx-examples/triggers",
 ]
+
+[profile.dev.build-override]
```
Hoverbear, line 28: It seems this is dead code or accidentally left in?

pgx-pg-sys/build.rs
```diff
@@ -317,33 +312,24 @@ fn impl_pg_node(items: &Vec<syn::Item>) -> eyre::Result<Vec<syn::Item>> {
...
                         write!(f, "{}", crate::node_to_string_for_display(self.as_node_ptr() as *mut crate::Node))
                     }
                 }
-            })?,
-        ));
+            });
     }
 
-    // sort the pgnode impls by their struct name so that we have a consistent ordering in our bindings output
-    pgnode_impls.sort_by(|(a_name, _), (b_name, _)| a_name.cmp(b_name));
```
Hoverbear, line 343: Do we want to maintain this sorting behavior to help reduce diffs on binding generation?

Cargo.toml
eeeebbbbrrrr: Sorry, I had a pending review of my own that commented on this.  See this comment:  https://github.com/tcdi/pgx/pull/526#discussion_r841260088

pgx-pg-sys/build.rs
eeeebbbbrrrr: I thought about it, but decided it doesn't matter.  The diffs for the bindings files are typically large anyways when we do change them.  I was also trying to just eliminate some overhead during build.rs.

You can convince me otherwise

pgx-pg-sys/src/submodules/setjmp.rs
```diff
@@ -0,0 +1,37 @@
+#[inline(always)]
+pub unsafe fn postgres_function_guard<T, F: FnOnce() -> T>(f: F) -> T {
```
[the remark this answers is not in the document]
Hoverbear: I guess `pg_guard` works but also doesn't hint to me what it's guarding?

pgx-pg-sys/build.rs
Hoverbear: I'd prefer to keep the generated diffs each release commit as small as we can, and if there are changes, I prefer them to be meaningful. This is useful for tracking possible regressions. 

... How about we only sort the bindings we write for docs.rs?

Cargo.toml
```diff
@@ -24,3 +24,6 @@ members = [
...
+opt-level = 3
```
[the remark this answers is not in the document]
Hoverbear: TIL about https://doc.rust-lang.org/cargo/reference/profiles.html#build-dependencies, very neat.

pgx-pg-sys/build.rs
eeeebbbbrrrr: > .. How about we only sort the bindings we write for docs.rs?

oh, great idea!

Cargo.toml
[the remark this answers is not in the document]
eeeebbbbrrrr: yeah, there.  So this means `build.rs` **and** its dependencies will always be compiled in release mode.  Which makes `build.rs` go about 2x faster, even if `build.rs` itself has to be recompiled because of a different `pgXX` feature flag -- its dependencies are still in release mode so they don't have to be recompiled.

pgx-pg-sys/src/submodules/setjmp.rs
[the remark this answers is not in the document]
eeeebbbbrrrr: Maybe we make this `pub(crate)` for now.  Since we wholesale apply this anyways, users won't need to use it.

Down the road if we make `#[pg_guard]` optional, then we can expose this to the world.

eeeebbbbrrrr:
> Does this improve build time at all? If so I'd love to have a sample I could snippet for the release notes.

Yes, on my machine a rebuild of `pgx-pg-sys` where `build.rs` runs has gone from ~36s to ~10s.  When build.rs doesn't run, it's gone from ~8s to ~2.5s.

It's quite significant.

pgx-pg-sys/src/submodules/setjmp.rs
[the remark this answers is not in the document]
eeeebbbbrrrr: `pub(crate) unsafe fn pg_guard_ffi_boundary(...)` is the new spelling.

Hoverbear:
Way to go! I'm happy to see this merge, you seem to have a couple pending todos according to comments though!

pgx-pg-sys/build.rs
eeeebbbbrrrr: Committed in https://github.com/tcdi/pgx/pull/526/commits/8021953305de4a86bc2a6af8e8f7a57b44c42e1b

PR #528, opened 2022-04-04
Adding get_typeid and renaming apis to use get_

einhverfr:
Improving introspection

### DIFF
--- a/pgx/src/spi.rs
+++ b/pgx/src/spi.rs
@@ -624,7 +624,9 @@ impl SpiHeapTupleData {
         }
     }
     pub fn get_nattrs(&self) -> i32 {
-        (*self.tupdesc).natts
+        unsafe {
+            (*self.tupdesc).natts
+        }
     }
     pub fn fname(&self, ordinal: i32) -> String {
          unsafe {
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr:
Interestingly enough, I'm working on a new `PgHeapTuple` type that allows getting **and** setting(!) attributes by name and ordinal.  Along with having runtime type checks for those actions (to ensure the Rust type is compatible with the Postgres type in the TupleDesc).  Trying to make things much safer...

Once I've got this all sussed out I was going to retool our SPI stuff to use it.  

I didn't consider adding these functions to the new `PgHeapTuple`, but I definitely will.

That said, I think I'm going to close this PR rather than merge it.  I appreciate your efforts for sure, but I don't want to introduce more to the API that'll likely be breaking in a (very) future pgx release.

eeeebbbbrrrr:
As an aside, I'll have a PR up in the next few days and I'll make sure to ping you to see if we can collab on that in any way.

einhverfr:
As long as we get the ordinal range, field names, and values so can properly introspect that will work for me.

I have a project where I am trying to have an extension (Oracle migration, can you guess) which takes an arbitrary query in, executes it, and writes a dbase file out which means I have to introspect the SPI table type for relevant type info.

Issue #499, opened 2022-03-23
RFC: Support User Defined types


Hoverbear:
`pgx` should be able to handle user defined types (such as those defined via `CREATE TYPE...`) in function argument and return value position.

Since user defined types can be created, removed, or altered at runtime, the `pgx` extension itself cannot possibly know the structure of the type at build time, thus we cannot possibly create some equivalent Rust type and define a mapping between them.

Instead, `pgx` needs to determine that some type is a user defined type, and use some `UserDefinedType` type which provides some `serde_json::Value`-like interface with runtime safety. We can determine the runtime layout of a type via inspecting the result of a query like `\dT+ example.type_name` in psql, there are also methods via SPI through the `pg_catalog.pg_type` table.

So if the user defined something like:

```rust
#[pgx::pg_extern]
fn user_function(
    arg: pgx::UserDefinedType
) -> pgx::UserDefinedType {
    unimplemented!();
}
```

However this doesn't communicate enough information to the SQL generator to create a valid function signature (unless we fall back to something like `any`, which is undesirable).

In order to augment this information we could use something like a `type!() macro that operates similar to the `name!()` macro. We could also consider having some trait `pgx::UserDefinedType` which users could do like:

```rust
mod pgx {
    pub trait UserDefinedType {
        const SQL_TYPE: &'static str;
    }
}

struct SomeUserType;

impl pgx::UserDefinedType for SomeUserType {
    const SQL_TYPE: &'static str = "MySqlType";
}

#[pg_extern]
fn user_function(
    arg: SomeUserType
) -> SomeUserType {
    unimplemented!();
}
```

Hoverbear:
If we had https://github.com/rust-lang/rust/issues/95174 we could do like this and then capture it with the SQL generator macros:

```rust
#![feature(adt_const_params)]

mod pgx {
    pub trait UserDefined<const SQL_TYPE: &'static str> {}
    pub struct UserDefinedType<const T: &'static str>;
}

struct SomeUserType;

impl pgx::UserDefined<"BEAN"> for SomeUserType { }


fn user_function(
    arg: SomeUserType
) -> SomeUserType {
    unimplemented!();
}

fn other_user_function(
    arg: pgx::UserDefinedType<"BEET">
) -> SomeUserType {
    unimplemented!();
}
```

Hoverbear:
From a "Correct Rust" level, this is what we ~should~ be doing, lacking `adt_const_params`:

```rust
[...]

PR #532, opened 2022-04-06
Improved "HeapTuple" support

### LINKED ISSUE #502: RFC: Improved `HeapTuple` support
Right now we have basic support for `HeapTuple`, but we'd like to expand on it.

Currently these types are quite basic:

https://github.com/tcdi/pgx/blob/a96c1b75380cfda210029bc4671b342c4e904fe7/pgx/src/spi.rs#L69-L79

We'd like to be able to look up attributes via an ordinal index, or via attribute name. 

This is partially to eventually support triggers (which `OLD` and `NEW` are heaptuples).

We'd like it for there to be a way to (without causing unnecessary copies) view the data inside these heaptuples, as well as bindings to their internal postgres counterparts which could mutate and/or create these heaptuples.

It may be that we need to distinct types, but we'd like to explore having one common type.

eeeebbbbrrrr:
This PR provides a new `PgHeapTuple` type that is capable of being constructed in various ways, along with reading **and** writing the individual attribute values in the underlying `HeapTuple`.

Fixes #502.

## Breaking changes

*  `FromDatum::from_datum`'s type signature has changed. It no longer needs an `oid` parameter.

## Other changes

* A `HeapTuple` type has been introduced which provides nice wrappers around `heap_getattr` and friends.
* Added `FromDatum::try_from_datum()` which returns a `Result<T, TryFromDatumError>` instead of an `Option` allowing users to get context of why a converstion failed.


eeeebbbbrrrr:
Once this gets merged, I can move on to doing work on issue #499 (which I'm mentally calling "composite type" support) and then also retool our `Spi` support to use the new `PgHeapTuple`.  I don't think we have an issue for that...

cargo-pgx/src/command/schema.rs
```diff
@@ -502,6 +502,14 @@ fn create_stub(
 
     let mut so_rustc_invocation = Command::new("rustc");
     so_rustc_invocation.stderr(Stdio::inherit());
+
+    if let Some(rustc_flags_str) = std::env::var("RUSTFLAGS").ok() {
```
Hoverbear, line 506: (I note for @workingjubilee this is unrelated to this PR itself, it appears to be from 0.4.2 release.)

cargo-pgx/src/command/test.rs
```diff
@@ -35,9 +35,6 @@ pub(crate) struct Test {
     /// Path to Cargo.toml
     #[clap(long, parse(from_os_str))]
     manifest_path: Option<PathBuf>,
-    /// Test all packages in the workspace
```
Hoverbear, line 38: (I note for @workingjubilee this is unrelated to this PR itself, it appears to be from 0.4.2 release.)

pgx-examples/composite_types/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+    composite
+        .set_by_name("name", "Someone Else")
+        .expect("no such attribute");
+
+    let composite = composite.into_composite_datum().unwrap();
+    AnyElement::from_datum(composite, false)
+}
+
+#[pg_extern]
+fn hello_composite_types() -> &'static str {
```
Hoverbear, line 53: Probably can be removed as it's very unrelated?

pgx/src/datum/array.rs
```diff
@@ -456,4 +436,9 @@ where
     fn type_oid() -> u32 {
         unsafe { pg_sys::get_array_type(T::type_oid()) }
     }
+
+    #[inline]
+    fn is_compatible_with(other: pg_sys::Oid) -> bool {
```
Hoverbear, line 441: I notice this pattern in several implementations, I wonder if we could provide a default implementation in the trait?

pgx/src/datum/into.rs
```diff
@@ -32,6 +33,60 @@ pub trait IntoDatum {
...
+
+    /// Is a Datum of this type compatible with another Postgres type?
+    ///
+    /// An example of this are the Postgres `text` and `varchar` types, which are both
+    /// technically compatible from a Rust type perspective.  They're both represented in Rust as
+    /// `String` (or `&str`), but the underlying Postgres types are different.
+    ///
+    /// If implementing this yourself, you likely want to follow a pattern like this:
+    ///
+    /// ```text
```
Hoverbear, line 45:
```suggestion
    /// ```rust,no_run
```

pgx-examples/composite_types/src/lib.rs
eeeebbbbrrrr: I'm gonna yank this example from this PR.  This "composite type" stuff ain't quite finished yet and really should be part of the work I do for #499.

pgx/src/datum/anyelement.rs
```diff
@@ -12,35 +12,26 @@ use crate::{pg_sys, FromDatum, IntoDatum};
...
     pub fn datum(&self) -> pg_sys::Datum {
         self.datum
     }
 
-    pub fn oid(&self) -> pg_sys::Oid {
-        self.typoid
-    }
-
     #[inline]
     pub fn into<T: FromDatum>(&self) -> Option<T> {
```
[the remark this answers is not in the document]
eeeebbbbrrrr: No?  It's safe to convert a Rust thing we own into a Datum.

pgx/src/datum/into.rs
eeeebbbbrrrr: No, because it's not exactly valid Rust code and I think the docs system will still try to ensure it parses.  I'll take a look at making it actually valid rust.

pgx/src/heap_tuple/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+#[derive(Debug, Copy, Clone)]
+pub enum PgHeapTupleCreationError {
+    NullTuple,
+    IncorrectAttributeCount,
+}
+
+pub type PgHeapTupleCreationResult<'a, AllocatedBy> =
+    Result<PgHeapTuple<'a, AllocatedBy>, PgHeapTupleCreationError>;
+
+/// A [PgHepTuple] is a lightweight wrapper around Postgres' [pg_sys::HeapTuple] object and a [PgTupleDesc].
```
Hoverbear, line 18:
```suggestion
/// A [PgHeapTuple] is a lightweight wrapper around Postgres' [pg_sys::HeapTuple] object and a [PgTupleDesc].
```

pgx/src/heap_tuple/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+///
+/// In order to access the attributes within a [pg_sys::HeapTuple], the [PgTupleDesc] is required
+/// to describe its structure.
+///
+/// [PgHeapTuple]s can be created from existing (Postgres-provided) [pg_sys:HeapTuple] pointers, from
+/// [pg_sys::TriggerData] pointers, from a composite datum, or created from scratch using raw Datums.
+///
+/// A [PgHeapTuple] can either be considered to be allocated by Postgres or by the Rust runtime. If
+/// allocated by Postgres, it is not mutable until [::into_owned()] is called.
```
Hoverbear, line 27: `[::into_owned()]` this isn't a valid link.

pgx/src/heap_tuple/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+pub struct PgHeapTuple<'a, AllocatedBy: WhoAllocated<pg_sys::HeapTupleData>> {
+    tuple: PgBox<pg_sys::HeapTupleData, AllocatedBy>,
+    tupdesc: PgTupleDesc<'a>,
+}
+
+impl<'a> PgHeapTuple<'a, AllocatedByPostgres> {
+    /// Creates a new [PgHeapTuple] from a [PgTupleDesc] and a [pg_sys::HeapTuple] pointer.  The
+    /// returned [PgHeapTuple] will be considered by have been allocated by Postgres and is not mutable
+    /// until [::into_owned()] is called.
```
Hoverbear, line 36:
```rust
warning: unresolved link to `::into_owned`
  --> pgx/src/heap_tuple/mod.rs:52:10
   |
52 |     /// [::into_owned()] is called.  
   |          ^^^^^^^^^^^^^^ no item named `` in scope
   ```

File: pgx/src/heap_tuple/mod.rs [hunk 10]
```diff
@@ -0,0 +1,337 @@
...
+    pub unsafe fn from_heap_tuple(tupdesc: PgTupleDesc<'a>, heap_tuple: pg_sys::HeapTuple) -> Self {
+        Self {
+            tuple: PgBox::from_pg(heap_tuple),
+            tupdesc,
+        }
+    }
+
+    /// Creates a new [PgHeapTuple] from one of the two (`Current` or `New`) trigger tuples.  The returned
+    /// [PgHeapTuple] will be considered by have been allocated by Postgres and is not mutable until
+    /// [::into_owned()] is called.
```
Hoverbear, line 52: This is a dead link :(

```rust

warning: unresolved link to `::into_owned`
  --> pgx/src/heap_tuple/mod.rs:52:10
   |
52 |     /// [::into_owned()] is called.  
   |          ^^^^^^^^^^^^^^ no item named `` in scope
```

pgx/src/heap_tuple/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+
+pub type PgHeapTupleCreationResult<'a, AllocatedBy> =
+    Result<PgHeapTuple<'a, AllocatedBy>, PgHeapTupleCreationError>;
+
+/// A [PgHepTuple] is a lightweight wrapper around Postgres' [pg_sys::HeapTuple] object and a [PgTupleDesc].
+///
+/// In order to access the attributes within a [pg_sys::HeapTuple], the [PgTupleDesc] is required
+/// to describe its structure.
+///
+/// [PgHeapTuple]s can be created from existing (Postgres-provided) [pg_sys:HeapTuple] pointers, from
```
Hoverbear, line 23:
```suggestion
/// [PgHeapTuple]s can be created from existing (Postgres-provided) [pg_sys::HeapTuple] pointers, from
```

pgx/src/trigger_support.rs
```diff
@@ -11,6 +11,21 @@ Use of this source code is governed by the MIT license that can be found in the
 
 use crate::{is_a, pg_sys};
 
+/// Indicates which trigger tuple to convert into a [PgHeapTuple].
```
Hoverbear, line 14: This is a dead link:

```rust
warning: unresolved link to `PgHeapTuple`
  --> pgx/src/trigger_support.rs:14:54
   |
14 | /// Indicates which trigger tuple to convert into a [PgHeapTuple].
   |                                                      ^^^^^^^^^^^ no item named `PgHeapTuple` in scope
   |
   = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`
```

pgx/src/datum/into.rs
workingjubilee: If all the type and imports are correct, sure it is. This builds just fine ([Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cf6526e9cd236a2d51a8ab1fbd420287)):
```rust
pub mod pg_sys {
    #[derive(PartialEq)]
    pub struct Oid();
    pub const TEXTOID: Oid = Oid();
}

struct Something;
impl Something {
    fn type_oid() -> pg_sys::Oid {
        pg_sys::Oid()
    }

    fn is_compatible_with(other: pg_sys::Oid) -> bool {
        // first, if our type is the other type, then we're compatible
        Self::type_oid() == other

     // and here's the other type we're compatible with
     || other == pg_sys::TEXTOID
    }
}
```

pgx/src/heap_tuple/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+            Some((attnum, _)) => self.set_by_index(attnum, value),
+        }
+    }
+
+    /// Given the index for an attribute in this [PgHeapTuple], change its value.
+    ///
+    /// Attribute numbers start at 1, not 0.
+    ///
+    /// ## Errors
+    /// - return [TryFromDatumError::NoSuchAttributeIndex] if the attribute does not exist
```
Hoverbear, line 172: This is a dead link: 

```rust

warning: unresolved link to `TryFromDatumError::NoSuchAttributeIndex`
   --> pgx/src/heap_tuple/mod.rs:172:19
    |
172 |     /// - return [TryFromDatumError::NoSuchAttributeIndex] if the attribute does not exist
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the enum `TryFromDatumError` has no variant or associated item named `NoSuchAttributeIndex`

```

pgx/src/heap_tuple/mod.rs
```diff
@@ -0,0 +1,337 @@
+//! Provides a safe interface to Postgres `HeapTuple` objects.
```
Hoverbear, line 1: This can be made into `pgx/src/heap_tuple.rs` perhaps?

pgx/src/heap_tuple/mod.rs
Hoverbear, line 172:
```suggestion
    /// - return [pgx::datum::::from::TryFromDatumError::NoSuchAttributeIndex] if the attribute does not exist
```

This is a dead link: 

```rust

warning: unresolved link to `TryFromDatumError::NoSuchAttributeIndex`
   --> pgx/src/heap_tuple/mod.rs:172:19
    |
172 |     /// - return [TryFromDatumError::NoSuchAttributeIndex] if the attribute does not exist
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the enum `TryFromDatumError` has no variant or associated item named `NoSuchAttributeIndex`

```

pgx/src/heap_tuple/mod.rs
eeeebbbbrrrr: ehh... yah, I suppose.  I think it's going to be self-contained now

pgx/src/datum/into.rs
eeeebbbbrrrr: Yah, I got something hacked that's now `rust,no_run`.

pgx/src/heap_tuple/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+    /// [PgHeapTuple] will be considered by have been allocated by Postgres and is not mutable until
+    /// [::into_owned()] is called.  
+    ///
+    /// ## Errors
+    /// - [PgHeapTupleCreationError::NullTuple] if the specified `TriggerTuple` is `null`
+    pub fn from_trigger_data(
+        trigger_data: &'a pg_sys::TriggerData,
+        which_tuple: TriggerTuple,
+    ) -> PgHeapTupleCreationResult<'a, AllocatedByPostgres> {
+        unsafe {
```
Hoverbear: Is there a reason this isn't an unsafe function, or the unsafe scope isn't smaller?

eeeebbbbrrrr: I'll make it unsafe.

pgx/src/heap_tuple/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+                pg_sys::heap_form_tuple(tupdesc.as_ptr(), datums.as_mut_ptr(), nulls.as_mut_ptr());
+
+            Ok(Self {
+                tuple: PgBox::<pg_sys::HeapTupleData, AllocatedByRust>::from_rust(formed_tuple),
+                tupdesc,
+            })
+        }
+    }
+
+    /// Creates a new [PgHeapTuple] from an opaque Datum that is should be a "composite" type.
```
Hoverbear:
```suggestion
    /// Creates a new [PgHeapTuple] from an opaque Datum that should be a "composite" type.
```

pgx/src/datum/anyelement.rs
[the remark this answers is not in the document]
Hoverbear: Ok!

pgx/src/datum/from.rs
```diff
@@ -50,29 +70,48 @@ pub trait FromDatum {
...
-        typoid: pg_sys::Oid,
     ) -> Option<Self>
     where
         Self: Sized,
     {
-        memory_context.switch_to(|_| FromDatum::from_datum(datum, is_null, typoid))
+        memory_context.switch_to(|_| FromDatum::from_datum(datum, is_null))
+    }
+
+    /// `try_from_datum` is a convenience wrapper around `FromDatum::from_datum` that returns a
```
Hoverbear, line 85: Is there any reason to use `from_datum` now, given that `try_from_datum` exists? Perhaps the documentation should suggest `try_from_datum` as the primary API, not `from_datum`.

eeeebbbbrrrr: In most cases, `from_datum` is preferred because we know the types are compatible -- most especially for the args to `#[pg_extern]` functions.  Additionally, there's less overhead in `from_datum` as we don't have to go digging for the Datum's type from $somewhere.

This was actually a thing @JLockerman once complained about and is how we ended up with that, now removed, `const NEEDS_TYPID: bool = false` thing in `FromDatum`.

Hoverbear: Hmm ok cool!

eeeebbbbrrrr: Places where we don't know the types are when dealing with composite/row types, random HeapTuples, and technically with Spi.

So once this is merged, I'll plumb the usage of `try_from_datum` into Spi and also wrap up the composite type support.

pgx/src/heap_tuple.rs
```diff
@@ -0,0 +1,340 @@
+//! Provides a safe interface to Postgres `HeapTuple` objects.
+use crate::{
+    heap_getattr_raw, pg_sys, AllocatedByPostgres, AllocatedByRust, FromDatum, FromDatumResult,
+    IntoDatum, PgBox, PgTupleDesc, TriggerTuple, TryFromDatumError, WhoAllocated,
+};
+use std::num::NonZeroUsize;
+
+/// Describes errors that can occur when trying to create a new [PgHeapTuple].
+#[derive(Debug, Copy, Clone)]
+pub enum PgHeapTupleCreationError {
```
Hoverbear, line 10: Anything returned in a `Result<T, E>` as the `E` type should always `impl Error`.  I suggest using `thiserror` which interacts well with `eyre` and `anyhow` to `#[derive(Error)]` this and similar types: https://github.com/dtolnay/thiserror

pgx/src/datum/from.rs
```diff
@@ -10,36 +10,56 @@ Use of this source code is governed by the MIT license that can be found in the
...
-    PgMemoryContexts,
+    pg_sys, text_to_rust_str_unchecked, varlena_to_byte_slice, AllocatedByPostgres, IntoDatum,
+    PgBox, PgMemoryContexts,
 };
 use std::ffi::CStr;
+use std::num::NonZeroUsize;
 
-/// Convert a `(pg_sys::Datum, is_null:bool, type_oid:pg_sys::Oid)` tuple into a Rust type
+/// If converting a Datum to a Rust type fails, this is the set of possible reasons why.
+#[derive(Debug)]
```
Hoverbear, line 20: Anything returned in a `Result<T, E>` as the `E` type should always `impl Error`.  I suggest using `thiserror` which interacts well with `eyre` and `anyhow` to `#[derive(Error)]` this and similar types: https://github.com/dtolnay/thiserror

pgx/src/heap_tuple.rs
eeeebbbbrrrr: I have no desire to import more dependencies for simple error handling. 

That said I’ll go rtfm on impl Error

pgx/src/heap_tuple.rs
```diff
@@ -0,0 +1,340 @@
...
+    ///
+    /// ## Errors
+    /// - return [TryFromDatumError::NoSuchAttributeNumber] if the attribute does not exist
+    /// - return [TryFromDatumError::IncompatibleTypes] if the Rust type of the `value` is not
+    /// compatible with the attribute's Postgres type
+    pub fn set_by_index<T: IntoDatum>(
+        &mut self,
+        attno: NonZeroUsize,
+        value: T,
+    ) -> FromDatumResult<()> {
```
Hoverbear, line 182: It strikes me that this we could have some of this function extracted into a `replace_at_index(..) -> Result<Self, _>` which returns the new heaptuple, and call it here as well as exposing that.

pgx/src/heap_tuple.rs
```diff
@@ -0,0 +1,340 @@
...
+                ),
+            );
+            let old_tuple = std::mem::replace(&mut self.tuple, new_tuple);
+            drop(old_tuple);
+            Ok(Some(()))
+        }
+    }
+}
+
+impl<'a, AllocatedBy: WhoAllocated<pg_sys::HeapTupleData>> PgHeapTuple<'a, AllocatedBy> {
```
Hoverbear, line 220: Noting for @workingjubilee this is a similar pattern to how `PgBox` works.

pgx/src/heap_tuple.rs
```diff
@@ -0,0 +1,340 @@
...
+    pub fn into_composite_datum(self) -> Option<pg_sys::Datum> {
+        unsafe {
+            Some(pg_sys::heap_copy_tuple_as_datum(
+                self.tuple.as_ptr(),
+                self.tupdesc.as_ptr(),
+            ))
+        }
+    }
+
+    /// How many attributes does this [PgHeapTuple] have?
```
Hoverbear, line 239: Generally doc comments that are questions are discouraged in favor of the more bland style. Copying `Vec`'s:

```rust
/// Returns the number of attributes in the [PgHeapTuple], also referred to as its ‘length’.
```

pgx/src/heap_tuple.rs
eeeebbbbrrrr: I don't follow?

pgx/src/trigger_support.rs
```diff
@@ -11,6 +11,21 @@ Use of this source code is governed by the MIT license that can be found in the
 
 use crate::{is_a, pg_sys};
 
+/// Indicates which trigger tuple to convert into a [crate::PgHeapTuple].
+#[derive(Debug, Copy, Clone)]
+pub enum TriggerTuple {
+    /// The row for which the trigger was fired. This is the row being inserted, updated, or deleted.
+    /// If this trigger was fired for an INSERT or DELETE then this is what you should return from the
```
Hoverbear, line 18: "then this is what you" -- So I return the enum from the function?

pgx/src/heap_tuple.rs
eeeebbbbrrrr: Done.

pgx/src/datum/from.rs
eeeebbbbrrrr: Done.  Again, just `impl Error` -- no importing of more dependencies.

pgx/src/trigger_support.rs
eeeebbbbrrrr: These were copy-pasta'd from Postgres' docs.  To me it's clear that you'd return the underlying tuple provided to you when you got the one specified from the enum variant.

pgx/src/heap_tuple.rs
eeeebbbbrrrr: That's really poorly named.  Instead of `WhoAllocated` it should be `WhoNeedsToFreeThisThing`.

The reason is, despite if it's `AllocatedByRust` or `AllocatedByPostgres`, it's a postgres `palloc()`'d value.

We can address this at a later time.  Not relevant to this PR.

pgx/src/heap_tuple.rs
Hoverbear: Given that this function always must create a copy, it might desirable to support a pattern like:

```rust
let cloned = my_tuple.replace_at_index(1, "thing");
```

Instead of

```rust
let cloned = my_tuple.clone();
cloned.set_at_index(1, "thing");
```

This would allow the user to avoid an extra clone happening. (and this function could call that one, so it's not added complexity on our side)

pgx/src/heap_tuple.rs
workingjubilee: WhoDrops

pgx/src/heap_tuple.rs
eeeebbbbrrrr: PgHeapTuple doesn’t implement Clone so I still don’t follow…

pgx/src/heap_tuple.rs
eeeebbbbrrrr: Ya know, @workingjubilee, why don't you put up a PR for that (just off of current `develop` branch) and otherwise give `PgBox` an audit that only someone with your skills have.

It's looking like our next release is going to have some pretty big API breakage anyways, so now would be a good time to rejigger naming things...

eeeebbbbrrrr:
oh, I wanted to /cc @[einhverfr](https://github.com/einhverfr) https://github.com/tcdi/pgx/pull/528.

This doesn't touch any of the Spi code yet, but this will become the foundation for retooling our Spi API.

workingjubilee:
Reviewing this to my satisfaction (and getting a start on auditing PgBox's "drop logic") required doing a deep-dive into how PGX and indeed Postgres handles transmitting data types, and auditing that code took my yesterday. However, at my current level of understanding, this looks good to me.

eeeebbbbrrrr:
> Reviewing this to my satisfaction (and getting a start on auditing PgBox's "drop logic") required doing a deep-dive into how PGX and indeed Postgres handles transmitting data types, and auditing that code took my yesterday. However, at my current level of understanding, this looks good to me.

Cool.  Thanks a bunch.  Go ahead and smash that "approve" button.

I'll probably want to hold off on merging this until you're done poking around with `PgBox`.

workingjubilee:
Hmm. I would probably feel best if this went ahead first, precisely because my audit started touching on potentially conflicting areas and it would give me freedom to rampage over this (or more likely it would reduce the diff, really).

pgx/src/datum/array.rs
```diff
@@ -414,11 +389,16 @@ where
...
+        Self::type_oid() == other || other == unsafe { pg_sys::get_array_type(T::type_oid()) }
+    }
 }
 
 impl<'a, T> IntoDatum for &'a [T]
 where
-    T: IntoDatum + Copy,
+    T: IntoDatum + Copy + 'a,
```
workingjubilee, line 401: I wonder why this now needs a lifetime annotation here?

eeeebbbbrrrr:
> Hmm. I would probably feel best if this went ahead first, precisely because my audit started touching on potentially conflicting areas and it would give me freedom to rampage over this (or more likely it would reduce the diff, really).

Okay. Let’s ask @Hoverbear to merge it in the morning.

pgx/src/datum/into.rs
```diff
@@ -32,6 +33,73 @@ pub trait IntoDatum {
...
+            || my_type == std::any::TypeId::of::<bool>()
+            || my_type == std::any::TypeId::of::<()>()
+            || my_type == std::any::TypeId::of::<crate::Time>()
+            || my_type == std::any::TypeId::of::<crate::TimeWithTimeZone>()
+            || my_type == std::any::TypeId::of::<crate::Timestamp>()
+            || my_type == std::any::TypeId::of::<crate::TimestampWithTimeZone>()
+            || my_type == std::any::TypeId::of::<crate::Date>()
+            || my_type == std::any::TypeId::of::<PgOid>()
+            || my_type == std::any::TypeId::of::<pg_sys::Datum>()
+            || my_type == std::any::TypeId::of::<Option<pg_sys::Datum>>()
```
workingjubilee, line 101: Wait. Sorry.

I re-skimmed this and this line is incorrect. Part of my audit PR makes it correct, but this line should go until then.

```suggestion
```

Alternatively, this is also true, but I suspect was not what you meant:
```suggestion
            || my_type == std::any::TypeId::of::<Option<&pg_sys::Datum>>()
```

workingjubilee (commented):
Ack, sorry. I re-rechecked this before hitting "Approve" and had almost missed a line.

Hoverbear:
After @workingjubilee 's feedback I'd love to see this merge, and can't wait to see what they follow up with!

pgx/src/datum/into.rs
eeeebbbbrrrr: I don't understand what you're asking me to do on this one, or why what I had is wrong or why making it a reference to a `pg_sys::Datum` would instead be correct?  What am I missing?

pgx/src/datum/array.rs
eeeebbbbrrrr:
```
error[E0309]: the parameter type `T` may not live long enough
   --> pgx/src/datum/array.rs:442:9
    |
399 | impl<'a, T> IntoDatum for &'a [T]
    |          - help: consider adding an explicit lifetime bound...: `T: 'a`
...
442 |         Self::type_oid() == other || other == unsafe { pg_sys::get_array_type(T::type_oid()) }
    |         ^^^^ ...so that the reference type `&'a [T]` does not outlive the data it points at
```

pgx/src/datum/into.rs
workingjubilee: So my understanding is that currently, Datum has a repr of a usize, and only things that fit within a usize are "pass by value". An Option requires has a tag bit to distinguish between Some and None, which means that on a 64-bit target, `Option<usize>` requires 65 bits. There is a trick called the "null pointer optimization", but that only applies to pointers, not numbers. Any integer value is valid, some pointers are not. `&usize` only has to occupy as much space as a pointer does, so even though `&usize` and `usize` are both 64 bits, `Option<&usize>` still only requires 64 bits.

workingjubilee: Ahhh I discussed this and turns out I was just confused.

workingjubilee (approved):
My question turned out to mostly be a point of confusion: the function doesn't necessarily use the _Postgres_ model for `is_pass_by_value` but IntoDatum's model, which asks this question re: `Option<Datum>`.

PR #539, opened 2022-04-13
Get TimestampWithTimeZone from OffsetDateTime

einarmo:
Doing this currently means splitting an OffsetDateTime into Date, Time and UtcOffset, which is a little silly when it just gets converted back into an OffsetDateTime.


eeeebbbbrrrr (approved):
LGTM!  Thanks!

PR #541, opened 2022-04-14
Updates Github Action runners for PGX

BradyBonnette:
This update includes a few things that help speed up Github Action runs and also give slightly better output for the end-user.

* A proper balance of throughput between number of parallel jobs and execution time per job, while also making it clearer about what job is ran and what has failed
* Uses Github Actions cache + sccache within each runner to speed up subsequent runs when only codebase code has changed
* Includes a MacOS build to ensure that any changes to the codebase will still build on a Mac (macos-11)


eeeebbbbrrrr:
Just a few comments from our discussion this morning:

- Trim down the `cargo pgx init` jobs to **only** do `pg14`.
- Shorten job titles to "pgx-tests & examples", "cargo pgx init", and "MacOS build & test"

eeeebbbbrrrr (approved):
Please see my comments on the PR itself.  After those are addressed I believe this is good to merge!

PR #546, opened 2022-04-21
Optionally emit versioned .so

JamesGuthrie:
`pgx` now experimentally supports the option to produce a versioned
shared library. This option is activated by removing the
`module_pathname` parameter in the extension's `.control` file.

When activated, it does the following:

- appends current package version to shared library name
- points generated function SQL to the versioned shared library (instead
  of `MODULE_PATHNAME`)
- replaces `@MODULE_PATHNAME@` placeholder with the path to the versioned
  shared library

Background:

There is an implicit requirement that C extensions maintain ABI
compatibility between versions. Emitting a versioned .so allows for ABI
breaks between versions.

This functionality is experimental and that care must be taken in the
following scenarios:

- when using shared memory
- when using query planner hooks
- when producing SQL schema migrations

### DIFF
--- a/cargo-pgx/README.md
+++ b/cargo-pgx/README.md
@@ -686,3 +686,94 @@ OPTIONS:
     -V, --version
             Print version information
 ```
+
+## EXPERIMENTAL: Versioned shared-object support
+
+`pgx` experimentally supports the option to produce a versioned shared library. This allows multiple versions of the
+extension to be installed side-by-side, and can enable the deprecation (and removal) of functions between extension
+versions. There are some caveats which must be observed when using this functionality. For this reason it is currently
+experimental.
+
+### Activation
+
+Versioned shared-object support is enabled by removing the `module_pathname` configuration value in the extension's
+`.control` file.
+
+### Concepts
+
+Postgres has the implicit requirement that C extensions maintain ABI compatibility between versions. The idea behind
+this feature is to allow interoperability between two versions of an extension when the new version is not ABI
+compatible with the old version.
+
+The mechanism of operation is to version the name of the shared library file, and to hard-code function definitions to
+point to the versioned shared library file. Without versioned shared-object support, the SQL definition of a C function
+would look as follows:
+
+```SQL
+CREATE OR REPLACE FUNCTION "hello_extension"() RETURNS text /* &str */
+STRICT
+LANGUAGE c /* Rust */
+AS 'MODULE_PATHNAME', 'hello_extension_wrapper';
+```
+
+`MODULE_PATHNAME` is replaced by Postgres with the configured value in the `.control` file. For pgx-based extensions,
+this is  usually set to `$libdir/<extension-name>`.
+
+When using versioned shared-object support, the same SQL would look as follows:
+
+```SQL
+CREATE OR REPLACE FUNCTION "hello_extension"() RETURNS text /* &str */
+STRICT
+LANGUAGE c /* Rust */
+AS '$libdir/extension-0.0.0', 'hello_extension_wrapper';
+```
+
+Note that the versioned shared library is hard-coded in the function definition. This corresponds to the
+`extension-0.0.0.so` file which `pgx` generates.
+
+It is important to note that the emitted SQL is version-dependent. This means that all previously-defined C functions
+must be redefined to point to the current versioned-so in the version upgrade script. As an example, when updating the
+extension version to 0.1.0, the shared object will be named `<extension-name>-0.1.0.so`, and `cargo pgx schema` will
+produce the following SQL for the above function:
+
+```SQL
+CREATE OR REPLACE FUNCTION "hello_extension"() RETURNS text /* &str */
+STRICT
+LANGUAGE c /* Rust */
+AS '$libdir/extension-0.1.0', 'hello_extension_wrapper';
+```
+
+This SQL must be used in the upgrade script from `0.0.0` to `0.1.0` in order to point the `hello_extension` function to
+the new shared object. `pgx` _does not_ do any magic to determine in which version a function was introduced or modified
+and only place it in the corresponding versioned so file. By extension, you can always expect that the shared library
+will contain _all_ functions which are still defined in the extension's source code.
+
+This feature is not designed to assist in the backwards compatibility of data types.
+
+### `@MODULE_PATHNAME@` Templating
+
+In case you are already providing custom SQL definitions for Rust functions, you can use the `@MODULE_PATHNAME@`
+template in your custom SQL. This value will be replaced with the path to the actual shared object. 
+
+The following example illustrates how this works:
+
+```rust
+#[pg_extern(sql = r#"
+    CREATE OR REPLACE FUNCTION tests."overridden_sql_with_fn_name"() RETURNS void
+    STRICT
+    LANGUAGE c /* Rust */
+    AS '@MODULE_PATHNAME@', '@FUNCTION_NAME@';
+"#)]
+fn overridden_sql_with_fn_name() -> bool {
+    true
+}
+```
+
+### Caveats
+
+There are some scenarios which are entirely incompatible with this feature, because they rely on some global state in
+Postgres, so loading two versions of the shared library will cause trouble.
+
+These scenarios are:
+- when using shared memory
+- when using query planner hooks
--- a/cargo-pgx/src/command/install.rs
+++ b/cargo-pgx/src/command/install.rs
@@ -14,8 +14,8 @@ use crate::{
 use cargo_toml::Manifest;
 use eyre::{eyre, WrapErr};
 use owo_colors::OwoColorize;
-use pgx_utils::get_target_dir;
 use pgx_utils::pg_config::PgConfig;
+use pgx_utils::{get_target_dir, versioned_so_name};
 use std::{
     io::BufReader,
     path::{Path, PathBuf},
@@ -114,6 +114,8 @@ pub(crate) fn install_extension(
         ));
     }
 
+    let versioned_so = get_property(&package_manifest_path, "module_pathname")?.is_none();
+
     let build_command_output = build_extension(
         user_manifest_path.as_ref(),
         user_package,
@@ -152,7 +154,13 @@ pub(crate) fn install_extension(
     {
         let mut dest = base_directory.clone();
         dest.push(&pkgdir);
-        dest.push(format!("{}.so", extname));
+        let so_name = if versioned_so {
+            let extver = get_version(&package_manifest_path)?;
+            versioned_so_name(&extname, &extver)
+        } else {
+            extname.clone()
+        };
+        dest.push(format!("{}.so", so_name));
 
         if cfg!(target_os = "macos") {
             // Remove the existing .so if present. This is a workaround for an
@@ -301,7 +309,7 @@ fn get_target_sql_file(
     let mut dest = base_directory.clone();
     dest.push(extdir);
 
-    let (_, extname) = crate::command::get::find_control_file(&manifest_path)?;
+    let (_, extname) = find_control_file(&manifest_path)?;
     let version = get_version(&manifest_path)?;
     dest.push(format!("{}--{}.sql", extname, version));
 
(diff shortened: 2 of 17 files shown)


cargo-pgx/src/command/install.rs
```diff
@@ -146,13 +152,18 @@ pub(crate) fn install_extension(
...
+            versioned_so,
         )?;
     }
 
     {
         let mut dest = base_directory.clone();
         dest.push(&pkgdir);
-        dest.push(format!("{}.so", extname));
+        if versioned_so {
+            dest.push(format!("{}-{}.so", extname, extver));
```
eeeebbbbrrrr, line 163: this incantation also exists over in `pgx_sql.rs`.  How can we centralize them?

cargo-pgx/src/command/install.rs
```diff
@@ -459,7 +476,17 @@ pub(crate) fn format_display_path(path: impl AsRef<Path>) -> eyre::Result<String
...
+fn filter_contents(
+    manifest_path: impl AsRef<Path>,
+    mut input: String,
+    versioned_so: bool,
+) -> eyre::Result<String> {
+    if versioned_so {
+        input = input.replace(
+            "module_pathname =",
+            "# commented out by pgx because versioned .so requested\n# module_pathname =",
+        )
```
eeeebbbbrrrr, line 488: I'm having a hard time visualizing what this is replacing, and where.  Can you clarify?

cargo-pgx/src/command/run.rs
```diff
@@ -46,6 +46,9 @@ pub(crate) struct Run {
     /// Use an existing `pgcli` on the $PATH.
     #[clap(env = "PGX_PGCLI", long)]
     pgcli: bool,
+    /// EXPERIMENTAL: Emit a versioned shared object file (and schema)
+    #[clap(long)]
+    versioned_so: bool,
```
eeeebbbbrrrr, line 51: Do we need this for `cargo pgx run`?  I'm leaning towards no, but could be convinced otherwise...

eeeebbbbrrrr: If we do do this for `run`, I suppose we'd want to do it for `test` too?

eeeebbbbrrrr (changes requested):
I think overall this is pretty close.

Before merging, I'd like to see the `cargo-pgx/README.md` updated to reflect this new argument and detail what it does any why it might be necessary.

cargo-pgx/src/command/run.rs
JamesGuthrie: It's not necessary for `run`, but it sure made testing this change a lot easier for me, as I could just `cargo pgx run` instead of having to `install` and then startup and connect to PG. I can add it for test as well. I don't really see the utility in that, but it would make the CLI API surface a little more consistent. What would you prefer?

cargo-pgx/src/command/install.rs
JamesGuthrie: In the `.control` file, one can set the `module_pathname` (path to shared object). With the default pgx extension template, the value of `module_pathname` is set to `$libdir/<extension-name>`. When using the `--versioned-so` feature, this default value for `module_pathname` will be incorrect.

Instead if leaving it in there to break stuff, this change comments it out of the `.control` file during the install phase.

Another option would be to check if a `module_pathname` entry is present in the `.control` file when `--versioned-so` is used, and explicitly throw an error. Actually I think that this approach may be more explicit in what it does.

JamesGuthrie: Another other option would be to make `module_pathname` templated in the control file, for instance setting it to `@MODULE_PATHNAME@`. This would bring it in line with how `default_version` works.

cargo-pgx/src/command/install.rs
JamesGuthrie: I don't think that there is any value in trying to do that. There are similarities, yes, but by the time you would have refactored everything to pull out the commonality, the common piece would be this:

```
pub fn build_versioned_so_name(extname: String, extver: String) -> String {
    format!("{}-{}", extname, extver)
}
```

cargo-pgx/src/command/run.rs
JamesGuthrie: In the latest iteration I've implemented this for `test`.

JamesGuthrie:
> I think overall this is pretty close.
> 
> Before merging, I'd like to see the `cargo-pgx/README.md` updated to reflect this new argument and detail what it does any why it might be necessary.

I've updated the docs. I'm having difficulty seeing where somebody who is unfamiliar with these concepts will have difficulty, so if there's anything that you find unclear please let me know and I'll take another pass.

cargo-pgx/src/command/install.rs
JamesGuthrie: I see. I've fixed this in the most recent iteration.

cargo-pgx/src/command/install.rs
eeeebbbbrrrr: Based on Postgres' docs for `module_pathname`, it really doesn't mean anything unless `MODULE_PATHNAME` exists in any of the extension sql scripts.  And with this PR, I guess we wouldn't.

Here's an alternative thought...

What if instead of having the `--versioned-so` CLI arg, we instead look for `module_pathname` in the .control file.  If it's there, we do the default behavior, if it's not, then we do the versioned-so behavior?  

I'm still not 100% convinced that if you want versioned .so's that you'd want to turn them off from time to time or from cargo pgx command to command.

Thoughts?

JamesGuthrie: > And with this PR, I guess we wouldn't.

Yes and no. By user error, it's possible that `MODULE_PATHNAME` would still be set in an SQL override.

> What if instead of having the --versioned-so CLI arg, we instead look for module_pathname in the .control file. If it's there, we do the default behavior, if it's not, then we do the versioned-so behavior?

That's an elegant way to "hide" the optionality. I kinda like it. I can't think of a scenario which it causes problems with, but maybe that's my lack of creativity. One downside it that it's quite implicit.

> I'm still not 100% convinced that if you want a versioned .so that there'd also be times when you wouldn't. That doesn't jive in my little brain.

I think that you're right on this.

JamesGuthrie: > You wanna go ahead and make these changes? Sorry for the churn, but we really need to get this right so it's not surprising to users.

Yup, I've changed it to work as you described. Please take another look.

pgx-examples/versioned_so/versioned_so.control
```diff
@@ -0,0 +1,4 @@
+comment = 'versioned_so:  Created by pgx'
+default_version = '@CARGO_VERSION@'
+relocatable = false
+superuser = false
```
eeeebbbbrrrr, line 4: Just for clarity, add a commented out `module_pathname` property here, with some explanation as to what it means to **not** have defined.

eeeebbbbrrrr (changes requested):

pgx-examples/versioned_so/versioned_so.control
eeeebbbbrrrr: Also, please add a run line for the example in the appropriate place around here:  https://github.com/tcdi/pgx/blob/c4ce4b9a5d74d589345fe95f3175a9e656dc6b3f/.github/workflows/tests.yml#L135

eeeebbbbrrrr: Also, also, please add the example to the top-level workspace here:  https://github.com/tcdi/pgx/blob/master/Cargo.toml

eeeebbbbrrrr:
@JamesGuthrie this is super nice now.  Great work and thanks for sticking through the process.  When you address those three minor comments above this will be ready to merge.  

Oh, please retarget the PR to the `develop` branch as well.  We're weird and don't merge directly into master, unless you're me and are prepared to be chastised by @Hoverbear for doing it.  ;)

pgx-examples/versioned_so/versioned_so.control
JamesGuthrie: Done. And added the example to the GitHub workflow.

JamesGuthrie:
> @JamesGuthrie this is super nice now. Great work and thanks for sticking through the process. When you address those three minor comments above this will be ready to merge.
> 
> Oh, please retarget the PR to the `develop` branch as well. We're weird and don't merge directly into master, unless you're me and are prepared to be chastised by @Hoverbear for doing it. ;)

No problem, thanks for the quick review! I've fixed the three things above and retargeted against develop.

eeeebbbbrrrr (approved):

eeeebbbbrrrr:
Approved.  Lets see what CI says and then I'll merge it.

And as promised, I'm happy to put out a v0.4.4 this week so y'all can get back to your real work.

cargo-pgx/README.md
```diff
@@ -686,3 +686,77 @@ OPTIONS:
...
+
+```SQL
+CREATE OR REPLACE FUNCTION "hello_extension"() RETURNS text /* &str */
+STRICT
+LANGUAGE c /* Rust */
+AS '$libdir/extension-0.0.0', 'hello_extension_wrapper';
+```
+
+It is important to note that the emitted SQL is version-dependent. This means that all previously-defined C functions
+must be redefined to point to the current versioned-so in the version upgrade script.
```
Hoverbear, line 732: So the 0.0.1 schema would contain: 

```sql

```SQL
CREATE OR REPLACE FUNCTION "hello_extension"() RETURNS text /* &str */
STRICT
LANGUAGE c /* Rust */
AS '$libdir/extension-0.0.1', 'hello_extension_wrapper';
```

Still?

eeeebbbbrrrr:
hmm, the new `versioned_so` test failed:  https://github.com/tcdi/pgx/runs/6161094219?check_suite_focus=true

Error message sucks.  Did it work for you locally?

cargo-pgx/README.md
```diff
@@ -686,3 +686,77 @@ OPTIONS:
...
+this is  usually set to `$libdir/<extension-name>`.
+
+When using versioned shared-object support, the same SQL would look as follows:
+
+```SQL
+CREATE OR REPLACE FUNCTION "hello_extension"() RETURNS text /* &str */
+STRICT
+LANGUAGE c /* Rust */
+AS '$libdir/extension-0.0.0', 'hello_extension_wrapper';
+```
```
Hoverbear, line 729: Does versioned shared-object support also make the output `.so` have a different name?

eeeebbbbrrrr: > Does versioned shared-object support also make the output .so have a different name?

Yup

pgx-examples/versioned_so/versioned_so.control
```diff
@@ -0,0 +1,6 @@
+comment = 'versioned_so:  Created by pgx'
+default_version = '@CARGO_VERSION@'
+# commenting-out module_pathname results in this extension being built/run/tested in "versioned shared-object mode"
```
Hoverbear, line 3: I love how you left this comment to explain it for readers. :)

Hoverbear (commented):
Some minor readme expansions requested, but otherwise I'm happy with how this turned out!

cargo-pgx/README.md
JamesGuthrie: I've mentioned it here in the readme.

cargo-pgx/README.md
JamesGuthrie: I've added and explicit example.

JamesGuthrie:
Okay, managed to get the build to work locally. Let's hope that CI accepts it.

Hoverbear (approved):

Hoverbear:
This is great James. :)

eeeebbbbrrrr:
Thanks a bunch, @JamesGuthrie!  Merged.  We'll get it out in a new release in the next few days.

pgx-examples/versioned_so/.cargo/config
```diff
@@ -0,0 +1,3 @@
+[build]
+# Postgres symbols won't be available until runtime
+rustflags = ["-C", "link-args=-Wl,-undefined,dynamic_lookup"]
```
Hoverbear, line 3: We didn't catch this during review, but this should have been removed as it causes some conflicts with the workspace override. @eeeebbbbrrrr is fixing this up before release.

eeeebbbbrrrr: yah.  totally our bad.

In a workspace crate environment like pgx is, you **only** need the `.cargo/config` at the root workspace crate.  If it's in any "child" crates, the compiler goes stupid and you spend all day trying to figure it out and then you're thankful @Hoverbear is on your team!

PR #552, opened 2022-05-03
Fix namespace.rs

eeeebbbbrrrr:
Fix a problem with leaking a `std::ffi::CString` along with Postgres' expectation that the argument to `pg_sys::makeString()` is a `palloc()`'d string.

This was reported by @JamesGuthrie via discord.


eeeebbbbrrrr:
WTF?

Issue #503, opened 2022-03-23
RFC: Trigger Support


Hoverbear:
We'd like PGX (and, as a consequence, PL/Rust) to support triggers.

Part of that support is #502 . 

Triggers could be defined by

```rust
#[pg_trigger(
  before(insert, update),
  each(row),
)]
fn trigger_example(ctx: Trigger) -> pgx::Trigger {
  let old = &ctx.OLD;
  let new = &ctx.NEW;
  let relname = &ctx.TG_RELNAME;

  new
}
```

The syntax remains undecided for now, but we should ensure we could at least support the options available to `CREATE TRIGGER`: https://www.postgresql.org/docs/current/sql-createtrigger.html

timClicks:
What is a `Trigger` context object? Is this something defined by `pgx`?

eeeebbbbrrrr:
Looks pretty good!

Hoverbear:
This is done in the `develop-v0.5.0` branch.

johnrballard:
## Tasks

- [ ] #502 
- [x] #558
- [ ] #604
- [ ] An immutable “OLD” HeapTuple instance,
- [ ] A mutable “NEW” HeapTuple instance,
- [ ] Access to trigger constant string arguments,
- [ ] Access to all PostgreSQL state variables as outlined here: https://www.postgresql.org/docs/current/plpgsql-trigger.html
- [ ] Ability to return “NULL”, “OLD”, or “NEW”

PR #558, opened 2022-05-09
Trigger support improvements

Hoverbear:
Improves Trigger support roughly in line with #503 :

```rust
use pgx::{pg_trigger, pg_sys, PgHeapTuple, WhoAllocated, PgHeapTupleError, PgTrigger};

#[derive(thiserror::Error, Debug)]
enum CustomTriggerError {
    #[error("No current HeapTuple")]
    NoCurrentHeapTuple,
    #[error("pgx::PgHeapTupleError: {0}")]
    PgHeapTuple(PgHeapTupleError),
}

#[pg_trigger]
fn example_custom_error(trigger: &PgTrigger) -> Result<
    PgHeapTuple<'_, impl WhoAllocated<pg_sys::HeapTupleData>>,
    CustomTriggerError,
> {
    unsafe { trigger.current() }.ok_or(CustomTriggerError::NoCurrentHeapTuple)
}
```


Hoverbear:
I'd like to add some more tests...

eeeebbbbrrrr:
> `fn trigger_example<'a>(trigger: &'a pgx::PgTrigger<'a>) -> Result<Option<PgHeapTuple<'a, impl WhoAllocated<pgx::pg_sys::HeapTupleData>>>, TriggerError> {`

That's special.  Wonder if we could make a type alias like:

```rust
type TriggerResult = PgHeapTuple<'a, impl WhoAllocated<pgx::pg_sys::HeapTupleData>
```

Hoverbear:
You can't use `impl Trait` in type aliases sadly.

pgx/src/trigger_support/pg_trigger.rs
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    /// A reference to the underlaying fcinfo
+    pub fn fcinfo(&self) -> &pg_sys::FunctionCallInfo {
+        self.fcinfo.borrow()
+    }
+
+    /// The new HeapTuple
+    // Derived from `pgx_pg_sys::TriggerData.tg_newtuple` and `pgx_pg_sys::TriggerData.tg_newslot.tts_tupleDescriptor`
+    pub unsafe fn new(&self) -> Option<PgHeapTuple<'_, AllocatedByPostgres>> {
```
eeeebbbbrrrr, line 61: I really feel like these functions don't need to be `unsafe`.

The `::from_fcinfo()` function definitely does, but at that point the developer is saying "hey compiler, I decided what I gave you is correct".  So for us to continue to not trust that fcinfo object seems wrong to me.

Think about things like `String::from_utf8_unchecked()`.  Yes, that function is unsafe, but `String` itself is safe and otherwise trusts the backing Vec.

Hoverbear: Do you also think `PgHeapTuple::from_trigger_data` should be safe?

Hoverbear: I think if we check `trigger_data.tg_trigger` during `from_fcinfo` we can make `name()`, `old_transition_table_name()`, and `new_transition_table_name()` safe.

eeeebbbbrrrr: No, that should absolutely be unsafe because it can't prove the pointer is correct.  But once the user calls it as:

```rust
let pg_trigger = unsafe { PgTrigger::from_trigger_data(...) };
```

I think we're clear to assume whatever was given to us is valid -- after all, the programmer told us so b/c the used `unsafe {}`.

Hoverbear: `PgRelation::open` is `unsafe` itself, I can't really see us making that safe unless we eagerly open it during `from_fcinfo`.

eeeebbbbrrrr: Sorry, how is `PgRelation` involved here?

Hoverbear: `table_name` and `table_schema` are got via `PgRelation::open` on `fcinfo.trigger_data.relation_data.rd_id` which is unsafe.

eeeebbbbrrrr: I'm confused.  I'm saying that `PgTrigger::from_trigger_data(...)` **should** be `unsafe`.  As such, the programmer is telling us the provided pointer is properly defined.  How does that impact that function's usage of `PgReation::open()`?

Hoverbear: Well, to start the `from_trigger_data` function you're referring to has never existed.. 😅 

I sat down and wrote some safety rules on the `from_fcinfo` function and I'm now making several more functions safe because the user has... well, agreed to my terms.

pgx/src/trigger_support/pg_trigger.rs
```diff
@@ -0,0 +1,292 @@
...
+    /// # Panics
+    /// 
+    /// If the relation was recently deleted, this function will panic.
+    /// 
+    /// # Safety
+    /// 
+    /// The caller should already have at least AccessShareLock on the relation ID, else there are nasty race conditions.
+    /// 
+    /// As such, this function is unsafe as we cannot guarantee that this requirement is true.
+    pub unsafe fn relation(&self) -> Result<crate::PgRelation, PgTriggerError> {
```
Hoverbear, line 188: What's your take on the safety of this one, @eeeebbbbrrrr ?

eeeebbbbrrrr (approved):

PR #560, opened 2022-05-10
Don't use NullableDatum on pg10,11

Hoverbear:
I experienced a build error building pgx-tests for pg10 without this.


workingjubilee:
:pray: Thank you.

PR #562, opened 2022-05-13
Remove dubious thread assertion

workingjubilee:
This assertion is misleading: if the Rust runtime becomes nested,
it creates a "new" thread, without actually spawning a second thread.
This `debug_assert!` then yields erroneous assertions.

### DIFF
--- a/pgx-pg-sys/src/submodules/guard.rs
+++ b/pgx-pg-sys/src/submodules/guard.rs
@@ -76,17 +76,6 @@ fn take_panic_location() -> PanicLocation {
 thread_local! { pub(crate) static IS_MAIN_THREAD: once_cell::sync::OnceCell<()> = once_cell::sync::OnceCell::new() }
 
 pub fn register_pg_guard_panic_handler() {
-    // first, lets ensure we're not calling ourselves twice
-    #[cfg(debug_assertions)]
-    {
-        if IS_MAIN_THREAD.with(|v| v.get().is_some()) {
-            panic!("IS_MAIN_THREAD has already been set")
-        }
-
-        // it's expected that this function will only ever be called by `pg_module_magic!()` by the main thread
-        IS_MAIN_THREAD.with(|v| v.set(()).expect("failed to set main thread sentinel"));
-    }
-
     std::panic::set_hook(Box::new(|info| {
         #[cfg(debug_assertions)]
         {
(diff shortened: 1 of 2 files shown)


Hoverbear (commented):
Is there a way for us to guard at all?

workingjubilee:
These are only debug asserts to begin with, so they only work if they are part of testing, but: not without completely revising the way we approach this, and probably not in a way that we would reasonably consider "trustworthy".

That is, we **could** simply check if we share the identity of a Rust main thread. But this seems easily forged, so it would be more security theater than security wall.

pgx-pg-sys/src/submodules/guard.rs
```diff
@@ -76,17 +76,6 @@ fn take_panic_location() -> PanicLocation {
 thread_local! { pub(crate) static IS_MAIN_THREAD: once_cell::sync::OnceCell<()> = once_cell::sync::OnceCell::new() }
```
eeeebbbbrrrr, line 76: Seems we should remove this too.

eeeebbbbrrrr (changes requested):

eeeebbbbrrrr:
> These are only debug asserts to begin with, so they only work if they are part of testing, but: not without completely revising the way we approach this, and probably not in a way that we would reasonably consider "trustworthy".
> 
> That is, we **could** simply check if we share the identity of a Rust main thread. But this seems easily forged, so it would be more security theater than security wall.

Yeah, this was an ill-considered thing from the beginning.  The intention was just to let the programmer know they used a Postgres-internal function in a background thread.  Ya know, just to be "helpful".

As we work towards the "postgrestd" stuff, I'm imagining that `std::thread` won't even be available, so...

pgx-pg-sys/src/submodules/guard.rs
workingjubilee: I was slightly more hesitant to remove all of that machinery because it was alive for "not just in debug", so I couldn't immediately tell if it was important or not that it worked this way for the non-debug cases. But if you are confident that it is okay to remove, I will happily do so!

workingjubilee:
I imagine we might offer the facilities for naming the current thread, but probably not spawning one unless we can somehow make sure things go through a properly reentrant interface.

PR #575, opened 2022-05-25
Require rust-version "1.58".

epgts:
It would be nice to support older versions, but currently pgx makes use
of captured identifers in format strings:
https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html#captured-identifiers-in-format-strings

At least this way users get a comprehensible error message:

    error: package `pgx-pg-sys v0.4.5` cannot be built
    because it requires rustc 1.58 or newer, while the
    currently active rustc version is 1.57.0


Hoverbear:
We seem to have done this at the same time! https://github.com/tcdi/pgx/commit/0da3b3f5f53751c4ce5f4aeaac3a24ef34036958 Thanks for your contribution. :)

PR #577, opened 2022-05-26
Assert BackgroundWorker funcs are called from a registered bgworker

### LINKED ISSUE #576: BackgroundWorker get_name/get_extra are unsound, segfaulting when called outside BackgroundWorker
As per title, several BackgroundWorker associated functions are unsound and cause segfault when called outside a background worker, specifically:

- get_name
- get_extra
- connect_worker_to_spi

EdMcBane:
Fixes #576

### DIFF
--- a/pgx/src/bgworkers.rs
+++ b/pgx/src/bgworkers.rs
@@ -73,6 +73,7 @@ impl BackgroundWorker {
         const LEN: usize = 96;
 
         unsafe {
+            assert!(!pg_sys::MyBgworkerEntry.is_null(), "BackgroundWorker associated functions can only be called from a registered background worker");
             CStr::from_ptr(std::mem::transmute::<&[c_char; LEN], *const c_char>(
                 &(*pg_sys::MyBgworkerEntry).bgw_name,
             ))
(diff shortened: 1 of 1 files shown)


Hoverbear (approved):
This looks good to me, but I'll ask @eeeebbbbrrrr to take a peek since I'm not that familiar with the BGWorkers.

pgx/src/bgworkers.rs
```diff
@@ -73,6 +73,7 @@ impl BackgroundWorker {
         const LEN: usize = 96;
 
         unsafe {
+            assert!(!pg_sys::MyBgworkerEntry.is_null(), "BackgroundWorker associated functions can only be called from a registered background worker");
```
Hoverbear, line 76: I think these should be `debug_assert`s to avoid a perf impact in release?

EdMcBane: I agree that it is unlikely that a mistake such as invoking background-worker methods gets into a release build, as it should manifest pretty quickly in development.

On the other hand, I believe only `BackgroundWorker::transaction()` has the potential impact performance in any meaningful way, as others are unlikely to be called in any kind of tight(ish) loop.

Hoverbear (approved):

pgx/src/bgworkers.rs
workingjubilee: Dereferencing the null pointer is unsound. If it is at all in question whether it is null, it should be asserted on or the function should return a sum type. These functions return bare types, so they **must** assert.

workingjubilee (approved):

PR #579, opened 2022-05-28
Add support for Spi cursors

EdMcBane:
This extends the `SpiClient` interface to expose methods to open a cursor based on a query, returning a `SpiCursor` that can be used to fetch results incrementally. Opened cursors can be closed early, or left open and turned into their (portal) name, to be later retrieved in a different Spi session in the same transaction.

### DIFF
--- a/pgx/src/spi.rs
+++ b/pgx/src/spi.rs
@@ -398,8 +398,9 @@ impl<'a> SpiClient<'a> {
         }
     }
 
-    /// sets up a cursor that will execute the specified query.
-    /// Rows may be then fetched using [`SpiCursor::fetch`]
+    /// Set up a cursor that will execute the specified query
+    ///
+    /// Rows may be then fetched using [`SpiCursor::fetch`].
     ///
     /// See [`SpiCursor`] docs for usage details.
     pub fn open_cursor(
@@ -448,8 +449,10 @@ impl<'a> SpiClient<'a> {
         SpiCursor { ptr }
     }
 
-    /// find a cursor in transaction by name.
-    /// Name should usually be retrieved via [`SpiCursor::into_name`]
+    /// Find a cursor in transaction by name
+    ///
+    /// A cursor for a query can be opened using [`SpiClient::open_cursor`].
+    /// Cursor name can be retrieved via [`SpiCursor::into_name`].
     ///
     /// See [`SpiCursor`] docs for usage details.
     pub fn find_cursor(
@@ -468,6 +471,8 @@ impl<'a> SpiClient<'a> {
     }
 }
 
+/// An SPI Cursor from a query
+///
 /// Represents a Postgres cursor (internally, a portal), allowing to retrieve result rows a few
 /// at a time. Moreover, a cursor can be left open within a transaction, and accessed in
 /// multiple independent Spi sessions within the transaction.
@@ -477,8 +482,9 @@ impl<'a> SpiClient<'a> {
 /// retrieved by name (in the same transaction) via [`SpiClient::find_cursor()`].
 ///
 /// # Important notes about memory usage
-/// Result sets returned by [`SpiCursor::fetch()`] will not be freed until the current
-/// Spi session is complete; this is a Pgx limitation that might get lifted in the future.
+/// Result sets ([`SpiTupleTable`]s) returned by [`SpiCursor::fetch()`] will not be freed until
+/// the current Spi session is complete;
+/// this is a Pgx limitation that might get lifted in the future.
 ///
 /// In the meantime, if you're using cursors to limit memory usage, make sure to use
 /// multiple separate Spi sessions, retrieving the cursor by name.
@@ -493,7 +499,7 @@ impl<'a> SpiClient<'a> {
 ///     assert_eq!(Some(2u32), cursor.fetch(2).get_one());
 ///     assert_eq!(Some(3u32), cursor.fetch(3).get_one());
 ///     Ok(None::<()>)
-///     // <--- all three result sets get freed only at this point
+///     // <--- all three SpiTupleTable get freed by Spi::connect at this point
 /// });
 /// ```
 ///
@@ -503,14 +509,16 @@ impl<'a> SpiClient<'a> {
 /// let cursor_name = Spi::connect(|mut client| {
 ///     let mut cursor = client.open_cursor("SELECT * FROM generate_series(1, 5)", None);
 ///     assert_eq!(Some(1u32), cursor.fetch(1).get_one());
-///     Ok(Some(cursor.into_name()))
-///     // <--- first result set gets freed at this point
+///     Ok(Some(cursor.into_name())) // <-- cursor gets dropped here
+///     // <--- first SpiTupleTable gets freed by Spi::connect at this point
 /// }).unwrap();
 /// Spi::connect(|mut client| {
 ///     let mut cursor = client.find_cursor(cursor_name);
 ///     assert_eq!(Some(2u32), cursor.fetch(1).get_one());
+///     drop(cursor); // <-- cursor gets dropped here
+///     // ... more code ...
 ///     Ok(None::<()>)
-///     // <--- second result set gets freed at this point
+///     // <--- second SpiTupleTable gets freed by Spi::connect at this point
 /// });
 /// ```
 pub struct SpiCursor {
@@ -518,7 +526,9 @@ pub struct SpiCursor {
 }
 
 impl SpiCursor {
-    /// fetch some rows from a cursor
+    /// Fetch up to `count` rows from the cursor, moving forward
+    ///
+    /// If `fetch` runs off the end of the available rows, an empty [`SpiTupleTable`] is returned.
     pub fn fetch(
         &mut self,
         count: i64
@@ -546,7 +556,9 @@ impl SpiCursor {
         }
     }
 
-    /// consumes the cursor and returns its name.
+    /// Consume the cursor, returning its name
+    ///
+    /// The actual Postgres cursor is kept alive for the duration of the transaction.
     /// This allows to fetch it in a later SPI session within the same transaction
     /// using [`SpiClient::find_cursor()`]
     pub fn into_name(self) -> String {
@@ -554,8 +566,9 @@ impl SpiCursor {
             .to_str().expect("non-utf8 cursor name").to_string()
     }
 
-    /// closes the cursor and releases its resources.
-    /// All cursors are closed automatically at the end of a transaction anyway.
+    /// Close the cursor, releasing its resources
+    ///
+    /// All cursors are closed automatically at the end of a transaction.
     /// Avoid closing the cursor if planning to retrieve it in a later Spi session within the same
     /// transaction using [`SpiClient::find_cursor()`].
     pub fn close(mut self) {
--- a/pgx/src/spi.rs
+++ b/pgx/src/spi.rs
@@ -517,7 +517,7 @@ impl<'a> SpiClient<'a> {
 /// ```
 pub struct SpiCursor<'client> {
     ptr: pg_sys::Portal,
-    _phantom: PhantomData<&'client SpiClient>
+    _phantom: PhantomData<&'client SpiClient>,
 }
 
 impl SpiCursor<'_> {
--- a/pgx/src/spi.rs
+++ b/pgx/src/spi.rs
@@ -16,6 +16,7 @@ use std::fmt::Debug;
 use std::marker::PhantomData;
 use std::mem;
 use std::ops::{Index, IndexMut};
+use std::ptr::NonNull;
 
 /// These match the Postgres `#define`d constants prefixed `SPI_OK_*` that you can find in `pg_sys`.
 #[derive(Debug, PartialEq)]
@@ -433,7 +434,7 @@ impl<'a> SpiClient<'a> {
             }
         }
 
-        let ptr = unsafe {
+        let ptr = NonNull::new(unsafe {
             pg_sys::SPI_cursor_open_with_args(
                 std::ptr::null_mut(), // let postgres assign a name
                 src.as_ptr(),
@@ -444,7 +445,8 @@ impl<'a> SpiClient<'a> {
                 false,
                 0,
             )
-        };
+        })
+        .expect("Portal ptr was null");
         SpiCursor { ptr, _phantom: PhantomData }
     }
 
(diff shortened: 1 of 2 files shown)


EdMcBane:
I've tried to keep the API consistent with the existing Spi API, although I have a couple of doubts, specifically:

1. I find signed row counts not very intuitive, although I understand they are consistent with the underlying C API (e.g. see `SpiClient::select` `limit` parameter); that might be addressed in the future Spi API overhaul mentioned on Discord, I believe?
2. There is no way to free the memory associated to returned `SpiTupleTableData`; this is more of an issue with cursors as there are use cases where several fetch operations are performed in the same Spi session.

pgx/src/spi.rs
```diff
@@ -378,8 +378,125 @@ impl SpiClient {
...
+        };
+        if ptr.is_null() {
+            error!("cursor named \"{}\" not found", name);
+        }
+        SpiCursor { ptr }
+    }
+}
+
+
+pub struct SpiCursor {
```
Hoverbear, line 448: Since this lacks a constructor, could you include some doc comment pointing out how to create it?

pgx/src/spi.rs
```diff
@@ -378,8 +378,125 @@ impl SpiClient {
...
+        unsafe { std::ffi::CStr::from_ptr((*self.ptr).name) }
+            .to_str().expect("non-utf8 cursor name").to_string()
+    }
+
+    /// closes the cursor and releases its resources.
+    /// Not mandatory, as all open cursors are closed automatically at the end of a transaction.
+    pub fn close(mut self) {
+        let ptr = std::mem::replace(&mut self.ptr, std::ptr::null_mut());
+        unsafe {
+            pg_sys::SPI_cursor_close(ptr)
```
Hoverbear, line 494: I wonder if this should be part of `Drop`...

EdMcBane: Leaving cursors open to be retrieved by name later in transaction is a core use case, so always closing on `Drop` would not work.

I experimented with having `SpiCursor::into_name()` invalidating the inner `ptr`, and closing conditionally in `Drop` in case `into_inner` hadn't been called, so that if you did not retrieve the name, it would be closed, since you would not have a handle for `find_cursor` later. 
Unfortunately then you would need to call `into_name` _every_time_ you retrieved it via `find_cursor`, or else it would be closed on Drop.

Of course this could be handled by having a `close_on_drop` flag, and setting it to false for cursors retrieved by name, but it looked like too much complexity for little gain.

For now I've tried to make the doc clearer on `close`, but I'm open to better ideas.

pgx/src/spi.rs
EdMcBane: Sure thing.

pgx/src/spi.rs
```diff
@@ -378,8 +378,176 @@ impl SpiClient {
...
+///     Ok(None::<()>)
+///     // <--- second result set gets freed at this point
+/// });
+/// ```
+pub struct SpiCursor {
+    ptr: pg_sys::Portal,
 }
 
+impl SpiCursor {
+    /// fetch some rows from a cursor
```
Hoverbear, line 502:
```suggestion
    /// Fetch `count` rows from the cursor
```

I'd like to also specify what the behavior is if the count is more than the number of rows available.

pgx/src/spi.rs
```diff
@@ -378,8 +378,176 @@ impl SpiClient {
...
+            tupdesc: if unsafe { pg_sys::SPI_tuptable }.is_null() {
+                None
+            } else {
+                Some(unsafe { (*pg_sys::SPI_tuptable).tupdesc })
+            },
+            current: -1,
+        }
+    }
+
+    /// consumes the cursor and returns its name.
```
Hoverbear, line 530:
```suggestion
    /// Consumes the cursor, returning its name
```

pgx/src/spi.rs
```diff
@@ -378,8 +378,176 @@ impl SpiClient {
...
+    /// This allows to fetch it in a later SPI session within the same transaction
+    /// using [`SpiClient::find_cursor()`]
+    pub fn into_name(self) -> String {
+        unsafe { std::ffi::CStr::from_ptr((*self.ptr).name) }
+            .to_str().expect("non-utf8 cursor name").to_string()
+    }
+
+    /// closes the cursor and releases its resources.
```
Hoverbear, line 538:
```suggestion
    /// Closes the cursor, releasing its resources
    ///
```

*Note:* Add newline for `cargo doc`.

pgx/src/spi.rs
```diff
@@ -378,8 +378,176 @@ impl SpiClient {
...
+    /// All cursors are closed automatically at the end of a transaction anyway.
```
Hoverbear, line 539:
```suggestion
    /// All cursors are closed automatically at the end of a transaction.
```

pgx/src/spi.rs
```diff
@@ -378,8 +378,176 @@ impl SpiClient {
...
+            pg_sys::SPI_cursor_find(name.as_pg_cstr())
+        };
+        if ptr.is_null() {
+            error!("cursor named \"{}\" not found", name);
+        }
+        SpiCursor { ptr }
+    }
+}
+
+/// Represents a Postgres cursor (internally, a portal), allowing to retrieve result rows a few
```
Hoverbear, line 452: `cargo doc` will give significant prominent in module level docs to the first line which should be a short sentence describing the role of the type. It should not have a period at the end.

```suggestion
/// An SPI cursor from a query
///
/// Represents a Postgres cursor (internally, a portal), allowing to retrieve result rows a few
```

pgx/src/spi.rs
```diff
@@ -378,8 +378,176 @@ impl SpiClient {
...
+/// at a time. Moreover, a cursor can be left open within a transaction, and accessed in
+/// multiple independent Spi sessions within the transaction.
+///
+/// A cursor can be created via [`SpiClient::open_cursor()`] from a query; a cursor left open (i.e.
+/// not [`Self::close()`]d) can be turned into a cursor name via [`Self::into_name()`] and later
+/// retrieved by name (in the same transaction) via [`SpiClient::find_cursor()`].
+///
+/// # Important notes about memory usage
+/// Result sets returned by [`SpiCursor::fetch()`] will not be freed until the current
+/// Spi session is complete; this is a Pgx limitation that might get lifted in the future.
```
Hoverbear, line 462: @zombodb I wonder if we could resolve this?

pgx/src/spi.rs
```diff
@@ -378,8 +378,176 @@ impl SpiClient {
...
+///     let mut cursor = client.open_cursor("SELECT * FROM generate_series(1, 5)", None);
+///     assert_eq!(Some(1u32), cursor.fetch(1).get_one());
+///     Ok(Some(cursor.into_name()))
+///     // <--- first result set gets freed at this point
+/// }).unwrap();
+/// Spi::connect(|mut client| {
+///     let mut cursor = client.find_cursor(cursor_name);
+///     assert_eq!(Some(2u32), cursor.fetch(1).get_one());
+///     Ok(None::<()>)
+///     // <--- second result set gets freed at this point
```
Hoverbear, line 494: This one is right!
```suggestion
///     // <--- second cursor gets freed at this point
```

Hoverbear (commented):
Asked for a few changes in the docs, this is looking great, @EdMcBane .

pgx/src/spi.rs
EdMcBane: I'm afraid I was way too terse, I could have clarified it better, let me try.

The issue I'm referring to is that zero-copy `FromDatum` implementations, such as `&str`, borrow the data from the `pg_sys::SPITupleTable` (wrapped in the `pgx::spi::SpiTupleTable`) returned by `select` and `fetch`, but this is not reflected by the type system (i.e. the returned `&str` has an inferred lifetime not bound by the `SpiTupleTable` it borrows from). 

As such, the borrow checker has no way to enforce proper lifetimes; this has been handled in the existing `pgx::spi` API by keeping all returned `SpiTupleTable`s around for the entire Spi session (by not exposing Postgres' `spi_freetuptable`) and having `Spi::connect` rely on `spi_finish` to release them all at the end, after copying any returned `FromDatum` in the upper memory context.

This works reasonably well for selects, where the user is already retrieving all the data at once.
Cursor users, on the other side, usually do so to avoid allocating memory for the entire result set all at once, so keeping `SpiTupleTable`s around might become a problem.

Thinking about it, I came up with a couple of options:

- recommending users to create separate Spi sessions if they care about freeing memory (current solution);
- having `fetch` take a closure as param, making it work more like `Spi::connect`, copying any returned data in the upper memory context and freeing the result set right away;
- having `fetch` return a different type, similar to `pgx::spi::PgTupleTable`, that yields borrowed data with a limited lifetime, forcing the user to copy it explicitly before dropping the source; this would have significant impact in term of dev experience I feel.

To start, I went with the simplest one, but I'm really interested in exploring it further based on your feedback.

EdMcBane: Makes sense 👍

pgx/src/spi.rs
EdMcBane: 👍 Done

pgx/src/spi.rs
EdMcBane: Done 👍

pgx/src/spi.rs
EdMcBane: Ditto, done 👍

EdMcBane:
I've integrated your changes, with some tweaks as I found other things that could be cleared up. Let me know what you think. Thanks for the feedback, @Hoverbear .

eeeebbbbrrrr:
This has bitrotted a little bit.  @EdMcBane, are you interested in bringing it up-to-date?  We can get it merged soon...

EdMcBane:
Sure thing, @eeeebbbbrrrr,  I'll give it a go in the next few days.

EdMcBane:
Should be good to go.

eeeebbbbrrrr:
I haven't checked out this branch to play with it, but can a cursor escape an Spi::connect() block?  Like, does this compile:

```rust
         let cursor = Spi::connect(|mut client| {
             let cursor = client.open_cursor("SELECT 1", None);
             return Ok(Some(cursor));
         });
```

If so, that seems pretty bad.

Also, I wonder if this PR needs to be rebased again -- we've had some other API changes around SPI and I dunno if this has them or not.

I would like to get this merged tho.  @EdMcBane, you've been working at this for awhile now and I feel bad we haven't given it much attention.  :(

EdMcBane:
> I haven't checked out this branch to play with it, but can a cursor escape an Spi::connect() block? Like, does this compile:
> 
> ```rust
>          let cursor = Spi::connect(|mut client| {
>              let cursor = client.open_cursor("SELECT 1", None);
>              return Ok(Some(cursor));
>          });
> ```

Good catch, it used to compile, now it does not, I've made sure to phantom-borrow client to make it so.

> Also, I wonder if this PR needs to be rebased again? We've had some other API changes around SPI and I dunno if this has them or not.

Just did it; the above fix needed to be adapted, but the rest rebased cleanly.

> I would like to get this merged tho. @EdMcBane, you've been working at this for awhile now and I feel bad we haven't given it much attention. :(

Nothing to worry about :-)
Cheers

EdMcBane:
`SpiCursor` can implement `Drop` instead of `close()`, as long as we use a sentinel value (null_ptr, or Option::None) to skip calling `SPI_cursor_close` when dropping it within `SpiCursor::into_name`.
I kinda think it would be best.

eeeebbbbrrrr:
> `SpiCursor` can implement `Drop` instead of `close()`, as long as we use a sentinel value (null_ptr, or Option::None) to skip calling `SPI_cursor_close` when dropping it within `SpiCursor::into_name`.
> I kinda think it would be best.

Makes sense.  I missed the `into_name()` function.  Is that even necessary as part of the API?  It looks a lot like leaking the cursor (waiting for Postgres to clean it up later).

Is it meant to be used for "WITH HOLD" cursors that can survive transactions?

EdMcBane:
`into_name` is meant to allow leaving a cursor open **within the same transaction** across multiple Spi sessions.
I was not aware of "WITH HOLD" cursors, interesting.

eeeebbbbrrrr:
> `into_name` is meant to allow leaving a cursor open **within the same transaction** across multiple Spi sessions.

Isn't this just `std::mem::forget()`?  I think if you make `SpiCursor` `#[repr(transparent)]` then that wouldn't ever leak Rust-allocated memory, and Postgres would clean up at the end of the transaction, and the `find_cursor()` function will still work.

> I was not aware of "WITH HOLD" cursors, interesting.

Yeah, that just popped into my head right now.  Incredibly useful feature -- I've used it a lot over the years.  I don't think we need to tackle it with this PR tho.  File it away for the future.

EdMcBane:
> Isn't this just `std::mem::forget()`? I think if you make `SpiCursor` `#[repr(transparent)]` then that wouldn't ever leak Rust-allocated memory, and Postgres would clean up at the end of the transaction, and the `find_cursor()` function will still work.

I'm wondering wether `#[repr(transparent)]` is actually necessary, as `into_name` takes self by value, thus moving it on the stack.
I'll replace close() with drop() and get back to you.

eeeebbbbrrrr:
> I'm wondering wether #[repr(transparent)] is actually necessary, as into_name takes self by value, thus moving it on the stack.

True.  Maybe it'll just be good to declare our intentions.  Also, I guess now the pointer can be `NonNull<pg_sys::PortalData>` now?

eeeebbbbrrrr:
This has turned out really nice.  Great work and thanks for sticking with it and thanks for your patience. 

Would you mind adding some // SAFETY comments to the few unsafe blocks?

Then I’ll merge.

eeeebbbbrrrr (approved):

EdMcBane:
Glad to hear it, will add // SAFETY comments later today.

I've noticed that `SpiTupleTable` doesn't ever early free memory via `SPI_freetuptable`, relying on the memory context being cleaned at Spi session completion. This makes cursors use a little bit awkward, as explained in my code:

```rust
/// # Important notes about memory usage
/// Result sets ([`SpiTupleTable`]s) returned by [`SpiCursor::fetch()`] will not be freed until
/// the current Spi session is complete;
/// this is a Pgx limitation that might get lifted in the future.
///
/// In the meantime, if you're using cursors to limit memory usage, make sure to use
/// multiple separate Spi sessions, retrieving the cursor by name.
```

Solving this is a bit involved, as it requires a `Drop` implementation on `SpiTupleTable` and to create a lifetime dependency between `SpiHeapTupleData`/`SpiHeapTupleDataEntry` and their parent `SpiTupleTable`.
Should we address this now or should I open a new issue for it?

EdMcBane:
> Solving this is a bit involved, as it requires a `Drop` implementation on `SpiTupleTable` and to create a lifetime dependency between `SpiHeapTupleData`/`SpiHeapTupleDataEntry` and their parent `SpiTupleTable`. Should we address this now or should I open a new issue for it?

After taking a stab at it, it is even more involved than I thought. `SpiTupleTable` _is_ an `Iterator` and holds status for the current line; an Iterator cannot borrow from itself, as such the `Iterator` implementation and all mutable state would need to go into some `SpiTupleTableIterator` that borrows from `SpiTupleTable`, but this:
1. is a breaking API change
2. conflicts with all data access methods on `SpiTupleTable` such as `one()`, `get_datum()`, `get_heap_tuple` which rely on the `SpiTupleTable` holding iteration status.

I'm thinking this would be worthwhile, but definitely farther reaching and in a separate issue / PR.

PR #581, opened 2022-05-31
Reunite develop and develop-v0.5.0 

workingjubilee:


Hoverbear:
Noting we should probably *not* squash this one.

.github/workflows/tests.yml
```diff
@@ -9,6 +9,7 @@ on:
     branches:
       - master
       - develop
+      - develop-v0.5.0
```
Hoverbear, line 12: We will need to yank this

PR #554, opened 2022-05-03
Do not use CREATE OR REPLACE FUNCTION

Hoverbear:
Fixes up #506

PR #584, opened 2022-06-01
Adding a flag that forces `CREATE OR REPLACE FUNCTION` statements

sumerman:
Adding a flag that forces `CREATE OR REPLACE FUNCTION` statements  instead of plain `CREATE FUNCTION` statements pgx now produces by default.

The new default (#554) is potentially less destructive to pre-existing objects and more secure. Unfortunately, it complicates the upgrade path when using versioned shared objects.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -181,6 +181,8 @@ jobs:
       run: cargo test --package triggers --features "pg$PG_VER" --no-default-features
 
     - name: Run versioned_so example tests
+      env: 
+        PGX_FORCE_CREATE_OR_REPLACE: true
       run: cargo test --package versioned_so --features "pg$PG_VER" --no-default-features
 
     # Attempt to make the cache payload slightly smaller.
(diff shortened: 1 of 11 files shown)


sumerman:
> LGTM. Much appreciated

Thank you! Could you approve the workflow to run?

eeeebbbbrrrr:
> Thank you! Could you approve the workflow to run?

Done.

workingjubilee:
I have opened #683 which serves as an alternative solution to this problem.

sumerman:
> I have opened #683 which serves as an alternative solution to this problem.

Thank you for addressing this! #683 is ever so slightly more work for us, but we'll be fine 😄

workingjubilee:
I don't want to create more work for you, truly! We just are inevitably going to get "so, how about something that lets me have one function CREATEd and one function REPLACEd?" and then we'll have wanted #683 all along.

sumerman:
> I don't want to create more work for you, truly! We just are inevitably going to get "so, how about something that lets me have one function CREATEd and one function REPLACEd?" and then we'll have wanted #683 all along.

No worries! As I’ve said — we’ll be fine 😅 I completely agree that #683 is more flexible and covers more use cases.

PR #588, opened 2022-06-10
Add mapping for fdw_handler

Hoverbear:
This should fix https://github.com/tcdi/pgx/issues/587, also adds a regression test for the future.


workingjubilee (approved):
This looks fine! I don't think we need to worry too much about making it functional if the desired function is, in fact, making sure the signature validates.

PR #591, opened 2022-06-17
Update Nix flake template

steve-chavez:
Closes https://github.com/tcdi/pgx/issues/590.


Hoverbear:
Ah ya the templated flake lock? I guess that makes sense.

Thanks for this contribution!

Hoverbear (approved):

PR #592, opened 2022-06-17
Composite Type Support

Hoverbear:
Addresses https://github.com/tcdi/pgx/issues/499, https://github.com/tcdi/pgx/issues/203, https://github.com/tcdi/pgx/issues/326. Helps along https://github.com/tcdi/pgx/issues/451.

Adds "Composite Type" support to PGX.

# What's changed

* A new `pgx::composite_type!("name")` macro now exists. It can be used in most PGX related macros where one would use an argument. It expands to create a `PgHeapTuple`, but the PGX related macros are able to derive additional metadata from this macro for SQL generation.
* `PgHeapTuple` now implements `FromDatum`.
* `PgHeapTuple` acquired a `new_composite_type` constructor which allows users to create a heap tuple shaped for a given composite type string. Eg `gHeapTuple::new_composite_type("dog")`.
* `PgHeapTuple` now has a `into_trigger_datum` function which will result in the correct `Datum` for returning from a `#[pg_trigger]` function. Since this was done as part of macro expanded code, no user changes should be required.
* The `IntoDatum` implementation of `PgHeapTuple` changed to better work with composite types. Specialized trigger related functions exist and should work transparently for users.
* `PgTupleDesc` now has a `for_composite_type` function which allows users to construct one for a given composite type by name.

# Breaking Changes

* You can no longer do `Option<default!(i32, 1)>` in your `#[pg_externs]`. Change it to `default!(Option<i32>, 1)`.
* The ordering of pgx related function argument macros is now fixed. Build errors should correctly guide users into the correct ordering with location context.
* Various types inside `pgx::utils::sql_entity_graph` were updated to include fields for composite types. `MaybeNamedVariadicTypeList` was removed, a new `UsedType` was added, and various internals around macro-expansion-time type resolution were tinkered with.
  > *As a reminder, while these structures are public, they are not recommended for users unless they are ok with them breaking nearly every `0.x` version.*
* If you are not using `pgx::pg_module_magic!();` (or `pgx::pg_sql_graph_magic!();`) in your extension, they now expand to expose a `pub extern "C" fn __pgx_sql_mappings() -> ::pgx::utils::sql_entity_graph::RustToSqlMapping` instead of their prior. This was done to both reduce the number of exported symbols extensions you, as well as support composite types.


Hoverbear:
This is probably ready for a preliminary look over now, but it's by no means 100% ready

Hoverbear:
@eeeebbbbrrrr I'd appreciate you giving this a lookie-loo if you can. :)

pgx-examples/composite_type/README.md
```diff
@@ -0,0 +1,5 @@
+An example of how to create an Aggregate with `pgx`.
+
+Demonstrates how to create a `IntegerAvgState` aggregate.
+
+This example also demonstrates the use of `PgVarlena<T>` and how to use `#[pgvarlena_inoutfuncs]` with `#[derive(PostgresType)]`.
```
Hoverbear, line 5: Whoops...

Hoverbear (commented):
I should probably do an aggregate test...

pgx-utils/src/lib.rs
```diff
@@ -517,7 +517,8 @@ pub fn anonymonize_lifetimes(value: &mut syn::Type) {
                             match arg {
                                 // rename lifetimes to the anonymous lifetime
```
eeeebbbbrrrr, line 518: comment no longer matches the code below.  Why are we now using `'static`?

pgx/src/datum/array.rs
```diff
@@ -331,23 +331,6 @@ impl<T: FromDatum> FromDatum for Vec<T> {
     }
 }
 
-impl<T: FromDatum> FromDatum for Vec<Option<T>> {
```
eeeebbbbrrrr, line 334: Why did this go away?  This is for supporting postgres `ARRAY[]`s that might contain `NULL` elements

pgx/src/datum/from.rs
```diff
@@ -408,3 +393,34 @@ impl<T> FromDatum for PgBox<T, AllocatedByPostgres> {
...
+{
+    #[inline]
+    unsafe fn from_datum(datum: pg_sys::Datum, is_null: bool) -> Option<Self> {
+        match is_null {
+            true => None,
+            false => Some(T::from_datum(datum, is_null)),
+        }
+    }
+
+    unsafe fn from_datum_in_memory_context(
```
eeeebbbbrrrr, line 409: I don't see a need to implement this -- the default implementation should suffice.  Also, this is basically what the default implementation is anyways.

Hoverbear: iirc we added this when trying to debug an ownership issue. Let me yank it! Good catch.

pgx/src/datum/array.rs
Hoverbear: If I re-add this and remove that conflicting implementation I get a lot of type failures, eg

```rust
error[E0277]: the trait bound `std::option::Option<heap_tuple::PgHeapTuple<'static, pgbox::AllocatedByRust>>: datum::PostgresType` is not satisfied
   --> pgx/src/lib.rs:263:26
    |
263 |             TypeId::of::<Array<Option<PgHeapTuple<'static, AllocatedByRust>>>>(),
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `datum::PostgresType` is not implemented for `std::option::Option<heap_tuple::PgHeapTuple<'static, pgbox::AllocatedByRust>>`
    |
note: required because of the requirements on the impl of `from::FromDatum` for `std::option::Option<heap_tuple::PgHeapTuple<'static, pgbox::AllocatedByRust>>`
   --> pgx/src/datum/varlena.rs:344:14
```

pgx-utils/src/lib.rs
Hoverbear: Ah, thanks!

We need anything using a lifetime to use the same lifetime so that the `TypeId`s match (`&'a Foo` and `&'b Foo` have different `TypeId`s)

pgx/src/datum/from.rs
Hoverbear: Works fine with it removed!

Hoverbear:
@zombodb I added many changes around aggregate support, did you want to take another review pass or shall I merge once I'm comfortable with it? I need to do a self review still.

pgx/src/tupdesc.rs
```diff
@@ -96,17 +97,17 @@ impl<'a> PgTupleDesc<'a> {
     ///
     /// This method is unsafe as we cannot validate that the provided `pg_sys::TupleDesc` is valid
     /// or requires reference counting.
-    pub unsafe fn from_pg_copy<'b>(ptr: pg_sys::TupleDesc) -> PgTupleDesc<'b> {
+    pub unsafe fn from_rust_copy<'b>(ptr: pg_sys::TupleDesc) -> PgTupleDesc<'b> {
```
Hoverbear, line 100: @zombodb can I specifically get your eyes on this bit here? We discussed renaming these at some point and I want confirmation I did the right thing.

pgx/src/tupdesc.rs
```diff
@@ -129,9 +130,9 @@ impl<'a> PgTupleDesc<'a> {
     ///
     /// This method is unsafe as we cannot validate that the provided `pg_sys::TupleDesc` is valid
     /// or is actually a copy that requires a `pfree()` on Drop.
-    pub unsafe fn from_pg_is_copy<'b>(ptr: pg_sys::TupleDesc) -> PgTupleDesc<'b> {
+    pub unsafe fn from_rust<'b>(ptr: pg_sys::TupleDesc) -> PgTupleDesc<'b> {
```
Hoverbear, line 133: @zombodb can I specifically get your eyes on this bit here? We discussed renaming these at some point and I want confirmation I did the right thing.

PR #600, opened 2022-06-22
Silence some nightly name collisions

Hoverbear:
Disables some warnings about nightly name collisions.


workingjubilee (approved):
If the pointer API we're using gets its `std` version stabilized quickly we may get a rude awakening and some complaints, but that's fine. We'll just fix it then.

PR #604, opened 2022-06-28
Triggers return result options

Hoverbear:
> **This is based off #592, it's not useful to review it prior to that merging.**

Enables triggers to return NULLs.


eeeebbbbrrrr:
this has bitrotted quite a bit but the gist (triggers being able to return NULL) was taken care of in https://github.com/tcdi/pgx/pull/1055.  This does have a "composite_type" example included, but looks like some form of that got committed back in #592.

closing

PR #606, opened 2022-07-04
Documentation touchups

Hoverbear:
Just a smattering of little documentation updates.


pgx/src/fcinfo.rs
```diff
@@ -7,10 +7,10 @@ All rights reserved.
 Use of this source code is governed by the MIT license that can be found in the LICENSE file.
 */
 
-//! Helper macros and functions for creating Postgres UDFs.
+//! Helper functions for creating Postgres UDFs
 //!
-//! Other than the exported macros, typically these functions are not necessary to call directly
-//! as they're used behind the scenes by the code generated by the `#[pg_extern]` macro.
+//! Other typically these functions are not necessary to call directly as they're used behind
```
workingjubilee, line 12: Orphaned "Other"

```suggestion
//! Typically these functions are not necessary to call directly as they're used behind
```

pgx/src/list.rs
```diff
@@ -7,10 +7,10 @@ All rights reserved.
 Use of this source code is governed by the MIT license that can be found in the LICENSE file.
 */
 
-//! A safe wrapper around Postgres' internal `List` structure.
+//! A safe wrapper around Postgres' internal [`List`][crate::pg_sys::List] structure.
 //!
-//! It functions similarly to a Rust `Vec`, including Iterator support, but provides separate
-//! understandings of Lists of Oids, Integers, and Pointers.
+//! It functions similarly to a Rust [`Vec`][std::vec::Vec], including Iterator support, but provides separate
```
workingjubilee, line 12:
```suggestion
//! It functions similarly to a Rust [`Vec`][std::vec::Vec], including iterator support, but provides separate
```

Iterator capitalized mid-sentence suggests the trait, otherwise we mean iterators generally, and Vec only impls `IntoIterator`.

pgx/src/datum/time_with_timezone.rs
```diff
@@ -12,6 +12,7 @@ use crate::{pg_sys, FromDatum, IntoDatum, PgBox};
 use std::ops::{Deref, DerefMut};
 use time::format_description::FormatItem;
 
+/// A `time with timezone` type from PostgreSQL
```
workingjubilee, line 15: TIL that's an actual type name with Postgres and not the version with underscores.

pgx/src/bgworkers.rs
```diff
@@ -9,7 +9,7 @@ Use of this source code is governed by the MIT license that can be found in the
 
 //! Safely create Postgres Background Workers, including with full SPI support
 //!
-//! See: [https://www.postgresql.org/docs/12/bgworker.html](https://www.postgresql.org/docs/12/bgworker.html)
+//! See: [https://www.postgresql.org/docs/14/bgworker.html](https://www.postgresql.org/docs/14/bgworker.html)
```
workingjubilee, line 12: Let's make this the last time we have to change the URL, instead:
```suggestion
//! See: [https://www.postgresql.org/docs/current/bgworker.html](https://www.postgresql.org/docs/current/bgworker.html)
```

pgx/src/datum/array.rs
```diff
@@ -13,6 +13,39 @@ use std::marker::PhantomData;
...
+use pgx::*;
+
+#[pg_extern]
+fn with_vec(elems: Array<String>) {
+    // Elements converted one by one
+    for maybe_elem in elems {
+        let elem = maybe_elem.unwrap();
+        todo!()
+    }
+}
```
workingjubilee, line 46: Lovely! It might be useful to note explicitly the parallel with an iterator and `Iterator::next` or `FnMut` closures which warp their captured contents, here, as "lazy" is somewhat of a "term of art".

workingjubilee (changes requested):
Thank you for annotating these.

Using `git merge develop -X theirs` resolves almost all the merge conflicts without losing aaalmost anything if you want to be lazy about making this work against `develop` (only change to reapply is in pgx/src/datum/from.rs).

workingjubilee (approved):

workingjubilee:
Thank you. :sparkling_heart:

PR #607, opened 2022-07-12
Refactor version defaulting

workingjubilee:
This boots the code for selecting versions from pgx-utils to cargo-pgx so as to make it easier to use pgx-utils without having to think or be aware of networks existing, making pgx-utils closer to a "pure" macro crate.


Hoverbear (approved):
Moving this makes sense to me and since next version is 0.5.0 it's the perfect time to break ABI if someone is depending on it.

PR #614, opened 2022-08-01
Added utils/catcache.h to includes in pg_sys

icewind:
Added catcache.h file to includes to generate mappings for `CatCList` structure which is returned by `SearchSysCacheList`. For more information please refer to https://github.com/tcdi/pgx/issues/608


Hoverbear:
Is there a reason you deleted this? Besides needing a rebase off develop, it looked fine!

icewind:
Yes. During development I found that not only catcache is required but also catalog/pg_auth_members. So you can cast the result tuple to a proper struct. I will check it once more and will do a proper PR.

PR #615, opened 2022-08-02
New sql type mapping

Hoverbear:
This revamps the SQL type mapping system to be based around the trait `SqlTranslatable` instead of the current `TypeId` based system.

Releasing this has breaking changes and requires a 0.5.0

# What's changed

* `pgx` is now able to map Rust types to SQL based on the existence of an `SqlTranslatable` trait (and its results). This trait is implemented for all types `pgx` supports, and implementations are generated in `#[derive(PostgresType)]` and `#[derive(PostgresEnum)]` macros for users.
* Foundation added to handle const generics for tasks such as supporting `NUMERIC(1, 10)`.
* Foundation added for generic types for tasks such as supporting users returning `Result<T,E>` once codegen is updated.
* The `pgx-utils/src/rewriter.rs` itemfn code was mostly migrated into the `sql_entity_graph` code and integrated.
* Source-only SQL mapping still exists and happens, however @workingjubilee and I discussed that there was a way to get this working with a newtype pattern. This was left for the future.
* SQL generation is, in general, quite a bit more strict about how it works, and users should see more, earlier build time errors when they try to do things that won't work.

# Breaking changes

* We no longer maintain a big hashmap of "Rust type -> SQL type" for builtin types like `i32` and `char` which is passed around during the SQL generation process. The `pgx_graph_magic!()` (and thus `pgx_module_magic!()`) macros have had their ABIs changed. Old version of `cargo-pgx` won't work anymore.
* It is no longer possible to return `impl Iterator<Item = ...>` from `#[pg_extern]`s. Instead, we have `SetOfIterator<'a, T>` and `TableIterator<'a, T>` which generate `SETOF T` and `TABLE T` in SQL generation respectively.
  - Users can transform set returning functions from:
      ```rust
      fn fluffy() -> impl Iterator<Item = i32> {
         vec![1,2,3].into_iter()
      }
      ```
    Into:
      ```rust
      fn fluffy() -> SetOfIterator<'static, i32> {
         SetOfIterator::new(vec![1,2,3].into_iter())
      }
      ```
  - Users can transform table returning functions from:
      ```rust
      fn fluffy() -> impl Iterator<Item = (named!(name, String), named!(i32),)> {
         vec![1,2,3].into_iter()
      }
      ```
    Into:
      ```rust
      fn fluffy() -> TableIterator<'static, (named!(name, String), named!(i32),)> {
         TableIterator::new(vec![1,2,3].into_iter())
      }
      ```
  - If your function accepts some borrows with lifetimes, it may need to use those lifetimes instead of `'static`.


workingjubilee:
I tested this with <REDACTED> and it only required modest changes to make everything work out.

pgx-examples/bad_ideas/src/lib.rs
```diff
@@ -35,9 +35,12 @@ fn exec(
...
 
-    (
-        output.status.code(),
-        String::from_utf8(output.stdout).expect("stdout is not valid utf8"),
+    TableIterator::new(
+        [(
+            output.status.code(),
+            String::from_utf8(output.stdout).expect("stdout is not valid utf8"),
+        )]
+        .into_iter(),
```
workingjubilee, line 43: This is common enough that it really ought to be addressed by its own method, e.g. `TableIterator::once`.

pgx-utils/src/lib.rs
```diff
@@ -404,15 +404,83 @@ pub fn staticize_lifetimes(value: &mut syn::Type) {
...
+            for elem in &mut type_tuple.elems {
+                staticize_lifetimes(elem);
+            }
+        }
+
+        _ => {}
+    }
+}
+
+pub fn anonymonize_lifetimes_in_type_path(value: syn::TypePath) -> syn::TypePath {
```
workingjubilee, line 424: "anonymize"

pgx/src/datum/array.rs
```diff
@@ -10,12 +10,13 @@ Use of this source code is governed by the MIT license that can be found in the
 use crate::{array::RawArray, layout::*, pg_sys, FromDatum, IntoDatum, PgMemoryContexts};
 use bitvec::slice::BitSlice;
 use core::ptr::NonNull;
+use pgx_utils::sql_entity_graph::metadata::{
+    ArgumentError, ReturnVariant, ReturnVariantError, SqlTranslatable, SqlVariant,
+};
```
workingjubilee, line 15: At first I was like "eh it's imported everywhere, sure: because it needs to be implemented everywhere".

Then I started wondering: Is this an inappropriate leakage/coupling of implementation details?

Then I remembered that this is no more peculiar than a Rust struct requiring Debug in order to be `format!("{:?}", debug_formatted)`.

pgx/src/datum/array.rs
```diff
@@ -285,6 +277,91 @@ impl<'a, T: FromDatum> Array<'a, T> {
...
+    #[deprecated(
+        since = "0.5.0",
+        note = "this function cannot be safe and is not generically sound\n\
+        even `unsafe fn as_slice(&self) -> &[T]` is not sound for all `&[T]`\n\
+        if you are sure your usage is sound, consider RawArray"
+    )]
+    #[allow(deprecated)]
+    pub fn as_slice(&self) -> &[T] {
+        self.0.as_slice()
+    }
```
workingjubilee, line 334: Having to introduce this only to instantly deprecate it displeases me but at least one test depends on it and VariadicArray needs to reimpl Array's full API in order for this to work, so...

pgx/src/datum/array.rs
```diff
@@ -285,6 +277,91 @@ impl<'a, T: FromDatum> Array<'a, T> {
...
+        Self(Array::<T> {
+            _ptr: NonNull::new(std::ptr::null_mut()),
+            raw: None,
+            elem_layout: None,
+            nelems,
+            elem_slice: slice::from_raw_parts(elements, nelems).into(),
+            null_slice: slice::from_raw_parts(nulls, nelems).into(),
+            _marker: PhantomData,
+        })
+    }
```
workingjubilee, line 315:
```suggestion
```

Maybe we can remove this without breaking anything?

pgx/src/datum/array.rs
```diff
@@ -505,3 +610,89 @@ where
...
+            })) => Ok(ReturnVariant::Plain(SqlVariant::SourceOnly {
+                requires_array_brackets: true,
+            })),
+            Ok(ReturnVariant::Plain(SqlVariant::Skip)) => Err(ReturnVariantError::SkipInArray),
+            Ok(ReturnVariant::SetOf(_)) => Err(ReturnVariantError::SetOfInArray),
+            Ok(ReturnVariant::Table(_)) => Err(ReturnVariantError::TableInArray),
+            err @ Err(_) => err,
+        }
+    }
+}
```
workingjubilee, line 653: This is... moderately complex.

pgx/src/datum/array.rs
```diff
@@ -505,3 +610,89 @@ where
...
+            Ok(ReturnVariant::Plain(SqlVariant::Skip)) => Err(ReturnVariantError::SkipInArray),
+            Ok(ReturnVariant::SetOf(_)) => Err(ReturnVariantError::SetOfInArray),
+            Ok(ReturnVariant::Table(_)) => Err(ReturnVariantError::TableInArray),
+            err @ Err(_) => err,
+        }
+    }
+
+    fn variadic() -> bool {
+        true
+    }
```
workingjubilee, line 697: Why a function and not an associated const?

pgx-tests/src/tests/heap_tuple.rs
```diff
@@ -466,85 +467,94 @@ mod sql_generator_tests {
...
-        Item = (
+    fn iterable_named_table_optional_array_optional_elems() -> TableIterator<
+        'static,
+        (
             name!(dog, Option<Vec<Option<::pgx::composite_type!("Dog")>>>),
             name!(cat, Option<Vec<Option<::pgx::composite_type!("Cat")>>>),
         ),
     > {
-        Vec::new().into_iter()
+        TableIterator::new(Vec::new().into_iter())
```
workingjubilee, line 525: This makes me wonder if the winning move is I: IntoIterator?

pgx/src/iterators.rs
```diff
@@ -0,0 +1,129 @@
+use pgx_utils::sql_entity_graph::metadata::{
+    ArgumentError, ReturnVariant, ReturnVariantError, SqlTranslatable, SqlVariant,
+};
+
+use crate::{pg_sys, IntoDatum};
+
+pub struct SetOfIterator<'a, T>
+where
+    T: std::panic::UnwindSafe + std::panic::RefUnwindSafe,
```
workingjubilee, line 9: We don't need these paths everywhere.

```suggestion
    T: UnwindSafe + RefUnwindSafe,
```

pgx/src/lib.rs
```diff
@@ -41,6 +40,7 @@ pub mod hooks;
 pub mod htup;
 pub mod inoutfuncs;
 pub mod itemptr;
+pub mod iterators;
```
workingjubilee, line 43:
```suggestion
pub mod iter;
```

pgx/src/lib.rs
```diff
@@ -81,6 +81,7 @@ pub use hooks::*;
 pub use htup::*;
 pub use inoutfuncs::*;
 pub use itemptr::*;
+pub use iterators::*;
```
workingjubilee, line 84:
```suggestion
```
Also see remarks in #678 about this.

pgx/src/rel.rs
```diff
@@ -352,3 +355,14 @@ impl Drop for PgRelation {
         }
     }
 }
+
+impl SqlTranslatable for PgRelation {
+    fn argument_sql() -> Result<SqlVariant, ArgumentError> {
+        Ok(SqlVariant::Mapped(String::from("regclass")))
+    }
+    fn return_sql() -> Result<ReturnVariant, ReturnVariantError> {
+        Ok(ReturnVariant::Plain(SqlVariant::Mapped(String::from(
```
workingjubilee, line 364: This boilerplate could be greatly reduced by a trait or function.

workingjubilee (changes requested):
This didn't have the time to receive a lot of polish, but there's some ergonomics concerns in this that should be addressed just before merging it. Overall I think it's a lot better than the current way of handling this, though, and it's amazing you got it working at all, honestly!

pgx/src/iterators.rs
```diff
@@ -0,0 +1,129 @@
+use pgx_utils::sql_entity_graph::metadata::{
+    ArgumentError, ReturnVariant, ReturnVariantError, SqlTranslatable, SqlVariant,
+};
+
+use crate::{pg_sys, IntoDatum};
+
+pub struct SetOfIterator<'a, T>
```
workingjubilee, line 7: `'a` binds the lifetime of the iterator, but the iterator is in a Box, so when is the lifetime not actually the lifetime of `T`?

Hoverbear: I believe this was for `'static` in certain cases.

pgx-tests/src/tests/heap_tuple.rs
Hoverbear: I recall this created some curious errors in some existing code. 🤔 It is something I initially tried.

pgx-examples/bad_ideas/src/lib.rs
Hoverbear: Agree, ya

pgx/src/iterators.rs
workingjubilee: I wonder if GATs can make this go away, in effect.

workingjubilee (approved):

pgx/src/datum/array.rs
Hoverbear: I think it could be! 🤔

Hoverbear:
I recall this still presents some challenges in ZDB, I want to make sure it works there before we merge this. :)

workingjubilee:
Since we can't seal it due to it having to live in both pgx and the pgx "compiler", I suspect `SqlTranslatable` should be an `unsafe trait`. It could potentially trick the C or Rust interfaces of PGX and Postgres to perform what is tantamount to `mem::transmute`, which is only safe Sometimes.

workingjubilee:
@Hoverbear Alright. I now have this building both ZomboDB and `<REDACTED>`, and the latter's tests even all pass. Less sure about ZomboDB, some do, some probably weren't set up correctly in my env, but also that has a slower release cadence and so we can probably defer the release if we need to fix bugs.

workingjubilee:
This PR will now also fix #615.

workingjubilee:
Apparently this now works for ZomboDB on build environments that aren't my local? I have no idea what's going wrong on my local. I can fix it later if necessary.

workingjubilee:
Okay, so there are a few lingering test errors at https://github.com/zombodb/zombodb/pull/763 but

> [3:02 PM] Hoverbear: I want to look at the timestamp thing
[3:04 PM]  Hoverbear: Hmmm yes it looks fine
[3:04 PM]  Hoverbear: I think it's just a serde thing
[3:04 PM]  Hoverbear: Okay I feel comfortable-ish with the pgx new-type-mapping pr merging
[3:05 PM]  Hoverbear: If you wanna, go ahead ❤️

These errors were likely introduced by https://github.com/tcdi/pgx/pull/643 which otherwise we expect to have significant performance/correctness improvements from, so we are not reverting that (but we might be adjusting the serialization slightly). Otherwise we're good, it looks like.

Hoverbear:
Yay!

Hoverbear:
We noticed this seems to have missed some dead code at https://github.com/tcdi/pgx/blob/cafefaaf8810143059f7e731e6cd464a38499c6a/pgx/src/lib.rs#L156

workingjubilee:
Opened https://github.com/tcdi/pgx/pull/697

PR #617, opened 2022-08-11
Adding -infinity/infinity to Date/Timestamp(Tz)

mhov:
Added infinity/-infinity support for Date, Timestamp, TimestampWithTimeZone

without this `date'infinity'` or `'infinity'::timestamp` would cause a i64 overflow.

+/- infinity for Date/Timestamps are defined in PG as the i64 Min/Max values and have special significance
```
#define DT_NOBEGIN		PG_INT64_MIN
#define DT_NOEND		PG_INT64_MAX
```

### DIFF
--- a/pgx/src/datum/time_stamp_with_timezone.rs
+++ b/pgx/src/datum/time_stamp_with_timezone.rs
@@ -14,7 +14,10 @@ use std::{
     convert::TryFrom,
     ops::{Deref, DerefMut},
 };
+use time::macros::time;
 use time::{format_description::FormatItem, UtcOffset};
+
+pub(crate) const MAX_TIME: time::Time = time!(23:59:59.999_999);
 
 #[derive(Debug, Copy, Clone)]
 pub struct TimestampWithTimeZone(time::OffsetDateTime);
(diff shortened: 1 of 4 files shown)


pgx/src/datum/time_stamp_with_timezone.rs
```diff
@@ -36,45 +37,54 @@ impl FromDatum for TimestampWithTimeZone {
...
-            let date = time::Date::from_calendar_date(
-                tm.tm_year,
-                time::Month::try_from(tm.tm_mon as u8)
-                    .expect("Got month outside of range in TimestampWithTimeZone::from_datum"),
-                tm.tm_mday as u8,
-            )
-            .expect("failed to create date from TimestampWithTimeZone");
+            let datum_val = datum.value() as i64;
+            let (date, time, tz) = match datum_val {
+                i64::MIN => (time::Date::MIN, time::Time::MIDNIGHT, 0i32),
```
mhov, line 42: The only questionable decision i made was just setting the date portion to MIN/MAX and not the time. is it worth making the infinity timestamp have a time of 23:59:59.999?

workingjubilee: Yes. If we use the MIN/MAX values that's one thing but it should be an unambiguous comparison for the multi-component case, so that when negative infinity is mapped, all other values are greater-or-equal, and when positive infinity is mapped, all other values are lesser-or-equal.

mhov: I am now reminded that postgres can represent dates way higher than `time::Date::MAX` `9999-12-31 23:59:59.999999` ex. pg dates go from 4713 BC to 5874897 AD and timestamp from 4713BC to 294276 AD.  So we could easily break the current mapping with just `date'10001-01-01'` I've heard `chrono` has a wider range, but is it possible to map to a rust enum?

```
enum PgDate {
    NegInfinity,
    OutOfRange(i64), // epoch days do what you will
    Date(time::Date),
    Infinity
}
```

workingjubilee:
As mentioned in chat, this PR may want to account for these issues:
- https://github.com/tcdi/pgx/issues/618
- https://github.com/tcdi/pgx/issues/504

However, addressing neither is required!
They just are potential inputs into how things evolve in the future.

PR #622, opened 2022-08-12
Opaque Date type with +/- infinity support

mhov:
I switched the Date type with one that's an alias for the `i32` datum.  `try_get_date(&self)` will get you a `time:Date` if it's within the `time` crate's bounds, or an `Err(pg_epoch_days: i32)` if out of range.  `-infinity` and `infinity` are handled on their own.

I moved the serialization `#[test]` to `#[pg_test]` so that serialization can leverage PG's built in serialization.  This is especially important for values outside of `time:Date` bounds which we couldn't format/serialize otherwise.

### DIFF
--- a/pgx/src/datum/date.rs
+++ b/pgx/src/datum/date.rs
@@ -10,6 +10,9 @@ Use of this source code is governed by the MIT license that can be found in the 
 use crate::{pg_sys, FromDatum, IntoDatum};
 use std::ffi::CStr;
 
+pub const POSTGRES_EPOCH_JDATE: i32 = pg_sys::POSTGRES_EPOCH_JDATE as i32;
+pub const UNIX_EPOCH_JDATE: i32 = pg_sys::UNIX_EPOCH_JDATE as i32;
+
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
 #[repr(transparent)]
 pub struct Date(i32);
@@ -59,7 +62,7 @@ impl Date {
 
     #[inline]
     pub fn from_date(date: time::Date) -> Date {
-        Self::from_pg_epoch_days(date.to_julian_day() - pg_sys::POSTGRES_EPOCH_JDATE as i32)
+        Self::from_pg_epoch_days(date.to_julian_day() - POSTGRES_EPOCH_JDATE)
     }
 
     #[inline]
@@ -78,8 +81,23 @@ impl Date {
     }
 
     #[inline]
-    pub fn to_julian_day(&self) -> i32 {
+    pub fn to_julian_days(&self) -> i32 {
+        self.0 + POSTGRES_EPOCH_JDATE
+    }
+
+    #[inline]
+    pub fn to_pg_epoch_days(&self) -> i32 {
         self.0
+    }
+
+    #[inline]
+    fn to_unix_epoch_days(&self) -> i32 {
+        self.0 + POSTGRES_EPOCH_JDATE - UNIX_EPOCH_JDATE
+    }
+
+    #[inline]
+    pub fn to_posix_time(&self) -> libc::time_t {
+        libc::time_t::from(self.to_unix_epoch_days()) * libc::time_t::from(pg_sys::SECS_PER_DAY)
     }
 
     pub fn try_get_date(&self) -> Result<time::Date, i32> {
@@ -87,8 +105,7 @@ impl Date {
         const INNER_RANGE_END: i32 = time::Date::MAX.to_julian_day();
         match self.0 {
             INNER_RANGE_BEGIN..=INNER_RANGE_END => {
-                time::Date::from_julian_day(self.0 + pg_sys::POSTGRES_EPOCH_JDATE as i32)
-                    .or_else(|_e| Err(self.0))
+                time::Date::from_julian_day(self.0 + POSTGRES_EPOCH_JDATE).or_else(|_e| Err(self.0))
             }
             v => Err(v),
         }
@@ -118,7 +135,7 @@ impl serde::Serialize for Date {
                 _ => {
                     let mut pg_tm: pg_sys::pg_tm = Default::default();
                     pg_sys::j2date(
-                        &self.0 + pg_sys::POSTGRES_EPOCH_JDATE as i32,
+                        &self.0 + POSTGRES_EPOCH_JDATE,
                         &mut pg_tm.tm_year,
                         &mut pg_tm.tm_mon,
                         &mut pg_tm.tm_mday,
--- a/pgx/src/datum/date.rs
+++ b/pgx/src/datum/date.rs
@@ -10,7 +10,8 @@ Use of this source code is governed by the MIT license that can be found in the 
 use crate::{pg_sys, FromDatum, IntoDatum};
 use std::ffi::CStr;
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+#[repr(transparent)]
 pub struct Date(i32);
 
 impl TryFrom<pg_sys::Datum> for Date {
@@ -43,8 +44,8 @@ impl FromDatum for Date {
 }
 
 impl Date {
-    pub const NEG_INFINITY: i32 = i32::MIN;
-    pub const INFINITY: i32 = i32::MAX;
+    pub const NEG_INFINITY: Self = Date(i32::MIN);
+    pub const INFINITY: Self = Date(i32::MAX);
 
     #[inline]
     pub fn from_date(date: time::Date) -> Date {
@@ -58,12 +59,12 @@ impl Date {
 
     #[inline]
     pub fn is_infinity(&self) -> bool {
-        self.0 == Self::INFINITY
+        self == &Self::INFINITY
     }
 
     #[inline]
     pub fn is_neg_infinity(&self) -> bool {
-        self.0 == Self::NEG_INFINITY
+        self == &Self::NEG_INFINITY
     }
 
     #[inline]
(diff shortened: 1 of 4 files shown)


pgx/src/datum/date.rs
```diff
@@ -67,13 +92,29 @@ impl serde::Serialize for Date {
...
-        )
+        let cstr;
+        unsafe {
+            let buf = [0u8; pg_sys::MAXDATELEN as usize + 1].as_mut_ptr() as *mut i8;
+            match &self.0 {
+                &Self::NEG_INFINITY | &Self::INFINITY => {
+                    pg_sys::EncodeSpecialDate(self.0, buf);
+                }
+                _ => {
+                    let mut pg_tm: pg_sys::pg_tm = Default::default();
```
mhov, line 103: I copied the logic from PG's `Datum date_out(PG_FUNCTION_ARGS)` for serialization.  I'm not sure if my approach to buffer pointers are best practice.  PG always uses 128byte string buffers, followed by `pstrdup(buf)` to trim down to the final length cstring.

I'm not sure if a cleaner approach like this would have too much overhead vs what i did
```
let cstr = direct_function_call::<&std::ffi::CStr>(
                pg_sys::date_out,
                vec![Some(datum_val)],
            )
            .expect("failed to convert date to a cstring");
 ```

**Review by workingjubilee — CHANGES_REQUESTED** [review]
This looks good! I mostly have a few remarks about the question of the `unsafe` code / PGX <-> Postgres interop.

File: pgx/src/datum/date.rs [hunk 2]
```diff
@@ -8,54 +8,79 @@ Use of this source code is governed by the MIT license that can be found in the
 */
 
 use crate::{pg_sys, FromDatum, IntoDatum};
-use std::ops::{Deref, DerefMut};
-use time::format_description::FormatItem;
+use std::ffi::CStr;
 
 #[derive(Debug)]
```
workingjubilee, line 13: Oh, also I think we can at least provide equality tests.
```suggestion
#[derive(Debug, PartialEq)]
```

pgx/src/datum/date.rs
```diff
@@ -8,54 +8,79 @@ Use of this source code is governed by the MIT license that can be found in the
...
-pub struct Date(time::Date);
+pub struct Date(i32);
```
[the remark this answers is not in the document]
mhov: done

pgx/src/datum/date.rs
```diff
@@ -67,13 +92,29 @@ impl serde::Serialize for Date {
...
+                    pg_sys::j2date(
+                        &self.0 + pg_sys::POSTGRES_EPOCH_JDATE as i32,
+                        &mut pg_tm.tm_year,
+                        &mut pg_tm.tm_mon,
+                        &mut pg_tm.tm_mday,
+                    );
+                    pg_sys::EncodeDateOnly(&mut pg_tm, pg_sys::USE_XSD_DATES as i32, buf)
+                }
+            }
+            cstr = CStr::from_ptr(pg_sys::pstrdup(buf));
```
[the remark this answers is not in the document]
mhov: before I used pstrdup the cstr values was padded with a ton of \0 null terminators and still 128bytes in length.  I noticed the PG code used pstrdup to trim the length of a buffer.  `libc` apparently fixed that, thank you!

pgx/src/datum/date.rs
mhov: I went with `#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]` because they should be comparable, sortable, etc.  I'm having doubts about `Ord` though is that only or string based data?

workingjubilee: `Ord` is for data which exposes a **total** order. Thus there is no `Ord` implementation for `f32`, as `NAN != NAN` and `-0.0 == 0.0`, where both break the total order.

As long as `#[derive(PartialOrd)]` is correct, then implementing Ord is fine.

pgx/src/datum/date.rs
mhov: Without knowing entirely how Serialization is used in pgx (other than the `json!(..)` tests), I'm assuming this wouldn't be a code hotspot/botleneck so if we can trade `direct_function_call`'s simplicity for an allocation and an an ffi call, maybe? `¯\_(ツ)_/¯`

workingjubilee: That would also be fine yes.

workingjubilee: The main catch is that serialization is used for anything with JSON, as you may have noticed, so emitting json (including into memory) that must serialize dates will hit this. Which may or may not matter. Frankly I have no idea.

mhov: ok we'll leave it with the current stack-allocated approach.  can always fall back or even make a convenience c-shim later.  There's a REALLY convenient PG function `datum_to_json(..)` which would handle ANY type we give it, but it's only in PG 12 and later, so  11 & 10 would need a c-shim to call non-exported functions with similar functionality.  i'll resolve this comment unless there's any further concern

pgx/src/datum/date.rs
mhov: I needed `Eq` to make this match work
``` 
match self {
    &Self::NEG_INFINITY | &Self::INFINITY => {
          pg_sys::EncodeSpecialDate(self.0, buf);
    ...
}
```
If `Ord` doesn't hurt i'll leave it in there.

pgx/src/datum/date.rs
```diff
@@ -67,13 +92,29 @@ impl serde::Serialize for Date {
...
+        }
+
+        serializer
+            .serialize_str(cstr.to_str().unwrap())
```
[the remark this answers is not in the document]
workingjubilee: So I rechecked and specifically, Postgres handles string outputs here via `#define`s and indexing into a digit table, in a way that only supports the Postgres formatting styles, none of which have a locale variance on that count, and we only support one date format. Postgres does in fact only ever write ASCII here as a result.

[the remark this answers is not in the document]
mhov: I'm not sure I understand the implication or the desired solution for this comment.
`pg_sys::USE_XSD_DATES` locks it to an ISO date format right? Are you saying people's locale's might mix up UTF-16 with the UTF-8 returned here?  If this approach doesn't feel write we can start going down the  `datum_to_json(..)` PG functions (with shims for PG 10/11) that essentially is what `to_json(...)` would return in PSQL.  Sorry if i'm misunderstanding your comment.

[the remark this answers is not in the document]
workingjubilee: No actual code changes required, I just want the comment to be precise.

pgx/src/datum/date.rs
```diff
@@ -8,54 +8,80 @@ Use of this source code is governed by the MIT license that can be found in the
...
-}
 
-impl Date {
-    pub fn new(date: time::Date) -> Self {
-        Date(date)
+    #[inline]
+    pub fn is_infinity(&self) -> bool {
+        self == &Self::INFINITY
     }
-}
```
workingjubilee, line 46: We should break Deref and DerefMut, but we shouldn't break `Date::new` immediately. `pub fn new` should be brought back, but with
```rust
#[deprecated(since = "0.5.0",
    note = "the repr of pgx::Date is no longer time::Date \
    and this fn will be removed in a future version")]
```
applied to it.

workingjubilee (approved):
I think this looks good!

My only remarks are, now that the dust has settled on everything else and I can see the conversions we are going to provide: When the conversion is for a Rust type and not a C typedef, we can use the conversion traits that Rust normally provides via the prelude to define the conversion.

There are reasons to do otherwise, like type inference on the conversion and syntactic niceties, but I don't think they apply here because it's a fairly direct conversion. Rather, the pressure is on to provide it ourselves, since only two other crates can implement these conversions (and in practice only one).

workingjubilee:
I think by returning the deprecated `fn new` to existence, this changeset has minimized the blast radius, and... I really, really hope no one was actually implicitly using methods via `time::Date`. If they did... well, it should only be a quick `time::Date::from(date)`/`date.into()` away. :see_no_evil:

Hopefully they will forgive us when we say it's ultimately for performance reasons.

pgx/src/datum/date.rs
```diff
@@ -8,54 +8,107 @@ Use of this source code is governed by the MIT license that can be found in the
...
-        &mut self.0
+    pub fn try_get_date(&self) -> Result<time::Date, i32> {
+        const INNER_RANGE_BEGIN: i32 = time::Date::MIN.to_julian_day();
+        const INNER_RANGE_END: i32 = time::Date::MAX.to_julian_day();
+        match self.0 {
+            INNER_RANGE_BEGIN..=INNER_RANGE_END => {
+                time::Date::from_julian_day(self.0 + POSTGRES_EPOCH_JDATE).or_else(|_e| Err(self.0))
+            }
+            v => Err(v),
+        }
```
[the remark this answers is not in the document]
mhov: Good insight! I believe my last commit addresses this.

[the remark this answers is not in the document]
mhov: @workingjubilee I wasn't clear on whether we should keep `pub fn Date::from_date(time::Date)` or `pub fn try_get_date(&self)` (calling their new From/TryFrom implementations) to keep it clear to the user what the intended usage is?  When using into/from conversions the language service/auto complete doesn't really show me the intent to create a `pgx::Date` from `time::Date`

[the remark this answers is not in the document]
workingjubilee: Yeah, this is one of the Rust Experience Things you accumulate.
You just kinda Know that
```rust
fn from_time_date_to_pgx_date(input: time::Date) -> pgx::Date {
    input.into()
}
```
does that conversion. So I think leaving them out is fine. We definitely want the `{Try,}From` implementations, and we can always readd convenience methods later.

Hoverbear:
I think it's reasonable. :)

PR #632, opened 2022-08-16
including utils/rangetypes.h in pg_sys

mhov:
This opens up some PG Range utility functions to `pg_sys`.  It will help support proper a `pgx::Range<T>` type in a later PR, but lets users work directly with Range datums in the meantime.


workingjubilee:
Are there any macros that we should eagerly cshim here?

mhov:
> Are there any macros that we should eagerly cshim here?
Anything worth using is already exposed by `range_serialize(...)` and `range_deserialize(..)`

Hoverbear (approved):

workingjubilee:
Righto then!

Issue #631, opened 2022-08-16
Risky soundness holding both pointers and references


workingjubilee:
For a given `struct SqlWidget`, we often are wrapping an underlying PostgresWidget type, and then we have both a safe reference, `&PostgresWidget` and a raw pointer, `*mut PostgresWidget`.

This is extremely dangerous in terms of soundness, and adds several pointer sizes to our structs besides. Instead, we should be retaining references with the appropriate properties and then working back from them, using `unsafe` to peel back and use the raw pointer when we need it.

workingjubilee:
An example of why comingling these is bad. According to a hypothetical Rust optimizer, if Widget does not express internal mutability, then this is probably a valid optimization:
```rust
let ptr: *mut Widget = get_widget_ptr();
let len = eval_len(raw);
let sliced: &[Widget] = unsafe { slice::from_raw_parts(ptr, len) };
let i = pick_an_index(sliced); // gives a random valid index
let read = sliced[i];
let written = pick_a_widget(sliced); // gives a random valid value
assert_ne!(read, written); // We guarantee that the new value isn't the same.
let iptr = unsafe { ptr.add(i) };
unsafe { iptr.write(written); }
let new_read = sliced[i]; // We repeat the read at the same index.
// This is where opts kick in: read and new_read are actually the "same read".
assert_eq!(read, new_read); // CAN PASS; these may be equal
assert_ne!(new_read, written); // CAN PASS; these may be non-equal
// It's not just the reference, the entire pointer write may be dead code.
let ptr_read = unsafe { iptr.read() };
assert_ne!(ptr_read, written); // CAN PASS; these may be non-equal
// But the ptr_read there might see something different, which is if anything worse:
assert_ne!(ptr_read, new_read);
```

...and this is _definitely_ valid if `let sliced: &mut [Widget]`.
It is best not to tempt fate here.

workingjubilee:
https://github.com/tcdi/pgx/commit/1d80c9557549edd9a8774cfb7d4fd7d62bfc4333 has implications for further solutions to this problem, and also for handling varlena objects in general:

https://github.com/tcdi/pgx/blob/1d80c9557549edd9a8774cfb7d4fd7d62bfc4333/pgx/src/array.rs#L35-L39

[...]

Issue #627, opened 2022-08-13
Incorrect size assumptions in `Array::as_slice()` mean it must be destroyed


workingjubilee:
Arrays currently can be converted into `&[T]` using `as_slice`, but in actuality their repr is boxed in a `usize`. Thus for types which are smaller than the machine's pointer width, this results in exposing various zero-initialized instances of the types in question, with tragic results when this is interacted with further.

This can be fixed by ceasing to use `deconstruct_array` to generate them, which resizes types like `i32` into `Datum`s.

To fix:
- [x] land tcdi/pgx#636
- [x] followup PR introducing null bitmap handling into Array, to allow not using `deconstruct_array`
- [x] remove it

workingjubilee:
In the process of trying to creating a diff that goes from #662 to solving this, I have discovered that we can create slices with Rust types that exceed the size of their Postgres layout, and exceed 8 bytes, the maximum Datum size, at 16 bytes. These include the recently introduced HeapTuple types, but there are almost certainly more.

This means there is currently **no safe way** to offer this function. In practice, even an `unsafe` version that mimicked this interface would not be correct, and would instead be a giant footgun. So I will instead use the recently introduced low-level interface tools to offer increasingly safe ways to do random access into the "flat array" representation and reduce the overhead of `Array<'_, T>`.

Hopefully in offering these tools it can make up for removing `Array::as_slice` entirely.

Hoverbear:
Nice detective work, @workingjubilee :)

workingjubilee:
Because this function is common in Rust code, it is usually sound, and there are probably some users of this function "in the wild", I have opened https://github.com/tcdi/pgx/pull/672 to begin a deprecation process rather than simply removing it outright.

workingjubilee:
It will now also panic if any of the values are null, as it is quite probable that if a valid is null it is not something that should be witnessed by Rust code.

I will work on an alternative function at some future point which describes a `NullableSlice<'_, T>` that can be used to interact with data structures in Postgres with associated null mappings.

workingjubilee:
I think I was a bit panicky with "no safe way", but I think it's still a massive footgun. I think it might be safe for most cases at the moment, and if the bounds were tightened, I might be confident in its soundness. It's still going away, I just need something I can feel good about using to rewrite the existing `as_slice` tests with first, since they've unfortunately caught quite a few "improvements" to Array I attempted experimentally.

Followups in this vein:
[...]

PR #634, opened 2022-08-17
Fix unsound interaction between `Array::over` and `Drop`

### LINKED ISSUE #633: `Array::over` with Array's drop impl is unsound
Upholding the contract of `Array::over` by passing valid pointers and lengths is not enough, these tests can break:

```rust
use pgx::*;

#[pg_extern]
fn over_implicit_drop() -> Vec<i64> {
    // Create an array of exactly Datum-sized numbers.
    let mut vec: Vec<i64> = vec![1, 2, 3, 4, 5];
    let mut nulls = vec![false, false, true, false, false];
    // Verify we uphold the length contract.
    assert_eq!(vec.len(), nulls.len());
    let len = vec.len();
    // Create an Array...
    let _arr = unsafe { Array::<'_, i64>::over(vec.as_mut_ptr().cast(), nulls.as_mut_ptr(), len) };
    // core::mem::forget(_arr); // Uncomment me to make the tests pass.
    vec
    // Implicit drop of _arr
}

#[cfg(any(test, feature = "pg_test"))]
#[pgx::pg_schema]
mod tests {
    use pgx::*;
    #[pg_test]
    fn test_array_over_direct() {
        let vals = crate::tests::array_tests::over_implicit_drop();
        assert_eq!(vals, &[1, 2, 3, 4, 5]);
    }

    #[pg_test]
    fn test_array_over_spi() {
        let vals: Vec<i64> =
            Spi::get_one("SELECT over_implicit_drop();").expect("over machine broke");
        assert_eq!(vals, &[1, 2, 3, 4, 5]);
    }
}
```

This is due to `Drop::drop` for Array calling `pg_sys::pfree`, assuming having pointers means owning them, which is why `mem::forget` can "solve" the problem.

workingjubilee:
This closes #633.
It also partially addresses #631 (i.e. for this type).
It introduces a new problem (leaking arrays) that I hope to fix soon,
as part of trying to fix, in the future, #627.

### DIFF
--- a/pgx-tests/src/tests/array_tests.rs
+++ b/pgx-tests/src/tests/array_tests.rs
@@ -99,6 +99,21 @@ fn return_zero_length_vec() -> Vec<i32> {
     Vec::new()
 }
 
+#[pg_extern]
+fn over_implicit_drop() -> Vec<i64> {
+    // Create an array of exactly Datum-sized numbers.
+    let mut vec: Vec<i64> = vec![1, 2, 3, 4, 5];
+    let mut nulls = vec![false, false, true, false, false];
+    // Verify we uphold the length contract.
+    assert_eq!(vec.len(), nulls.len());
+    let len = vec.len();
+    // Create an Array...
+    let _arr = unsafe { Array::<'_, i64>::over(vec.as_mut_ptr().cast(), nulls.as_mut_ptr(), len) };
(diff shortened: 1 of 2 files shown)


workingjubilee:
cc @mhov you may be interested in knowing about the other problems with the array type.

PR #636, opened 2022-08-17
Low-level interop for PG arrays

mhov:
These PG function shims will allow for future work to improve the `pgx::Array<T>` type, as well as manual access to ArrayType data in the meantime.

### DIFF
--- a/pgx-pg-sys/cshim/pgx-cshim.c
+++ b/pgx-pg-sys/cshim/pgx-cshim.c
@@ -117,9 +117,9 @@ char *pgx_ARR_DATA_PTR(ArrayType *arr) {
     return ARR_DATA_PTR(arr);
 }
 
-PGDLLEXPORT int pgx_ARR_NELEMS(ArrayType *arr);
-int pgx_ARR_NELEMS(ArrayType *arr) {
-    return ArrayGetNItems(ARR_NDIM(arr), ARR_DIMS(arr));
+PGDLLEXPORT int pgx_ARRNELEMS(ArrayType *arr);
+int ARRNELEMS(ArrayType *arr) {
+    return ArrayGetNItems(arr->ndim, ARR_DIMS(arr));
 }
 
 PGDLLEXPORT bits8 *pgx_ARR_NULLBITMAP(ArrayType *arr);
@@ -135,4 +135,9 @@ int pgx_ARR_NDIM(ArrayType *arr) {
 PGDLLEXPORT bool pgx_ARR_HASNULL(ArrayType *arr);
 bool pgx_ARR_HASNULL(ArrayType *arr) {
     return ARR_HASNULL(arr);
+}
+
+PGDLLEXPORT int *pgx_ARR_HASNULL(ArrayType *arr);
+int *pgx_ARR_DIMS(ArrayType *arr){
+    return ARR_DIMS(arr);
 }
--- a/pgx-pg-sys/cshim/pgx-cshim.c
+++ b/pgx-pg-sys/cshim/pgx-cshim.c
@@ -117,8 +117,8 @@ char *pgx_ARR_DATA_PTR(ArrayType *arr) {
     return ARR_DATA_PTR(arr);
 }
 
-PGDLLEXPORT int pgx_ARRNELEMS(ArrayType *arr);
-int ARRNELEMS(ArrayType *arr) {
+PGDLLEXPORT int pgx_ARR_NELEMS(ArrayType *arr);
+int pgx_ARR_NELEMS(ArrayType *arr) {
     return ArrayGetNItems(arr->ndim, ARR_DIMS(arr));
 }
 
@@ -137,7 +137,7 @@ bool pgx_ARR_HASNULL(ArrayType *arr) {
     return ARR_HASNULL(arr);
 }
 
-PGDLLEXPORT int *pgx_ARR_HASNULL(ArrayType *arr);
+PGDLLEXPORT int *pgx_ARR_DIMS(ArrayType *arr);
 int *pgx_ARR_DIMS(ArrayType *arr){
     return ARR_DIMS(arr);
 }
--- a/pgx-pg-sys/cshim/pgx-cshim.c
+++ b/pgx-pg-sys/cshim/pgx-cshim.c
@@ -25,6 +25,7 @@ Use of this source code is governed by the MIT license that can be found in the 
 #include "parser/parsetree.h"
 #include "utils/memutils.h"
 #include "utils/builtins.h"
+#include "utils/array.h"
 
 
 PGDLLEXPORT MemoryContext pgx_GetMemoryContextChunk(void *ptr);
@@ -109,4 +110,29 @@ Oid pgx_HeapTupleHeaderGetOid(HeapTupleHeader htup_header) {
 PGDLLEXPORT char *pgx_GETSTRUCT(HeapTuple tuple);
 char *pgx_GETSTRUCT(HeapTuple tuple) {
     return GETSTRUCT(tuple);
+}
+
+PGDLLEXPORT char *pgx_ARR_DATA_PTR(ArrayType *arr);
+char *pgx_ARR_DATA_PTR(ArrayType *arr) {
+    return ARR_DATA_PTR(arr);
+}
+
+PGDLLEXPORT int pgx_ARR_NELEMS(ArrayType *arr);
+int pgx_ARR_NELEMS(ArrayType *arr) {
+    return ArrayGetNItems(ARR_NDIM(arr), ARR_DIMS(arr));
+}
+
+PGDLLEXPORT bits8 *pgx_ARR_NULLBITMAP(ArrayType *arr);
+bits8 *pgx_ARR_NULLBITMAP(ArrayType *arr) {
+    return ARR_NULLBITMAP(arr);
+}
+
+PGDLLEXPORT int pgx_ARR_NDIM(ArrayType *arr);
+int pgx_ARR_NDIM(ArrayType *arr) {
+    return ARR_NDIM(arr);
+}
+
+PGDLLEXPORT bool pgx_ARR_HASNULL(ArrayType *arr);
+bool pgx_ARR_HASNULL(ArrayType *arr) {
+    return ARR_HASNULL(arr);
 }
--- a/pgx-pg-sys/src/lib.rs
+++ b/pgx-pg-sys/src/lib.rs
@@ -240,6 +240,11 @@ mod all_versions {
         pub fn pgx_list_nth_oid(list: *mut super::List, nth: i32) -> super::Oid;
         pub fn pgx_list_nth_cell(list: *mut super::List, nth: i32) -> *mut super::ListCell;
         pub fn pgx_GETSTRUCT(tuple: pg_sys::HeapTuple) -> *mut std::os::raw::c_char;
+        pub fn pgx_ARR_DATA_PTR(arrayType: *mut pg_sys::ArrayType) -> *mut u8;
+        pub fn pgx_ARR_NELEMS(arrayType: *mut pg_sys::ArrayType) -> i32;
+        pub fn pgx_ARR_NDIM(arrayType: *mut pg_sys::ArrayType) -> i32;
+        pub fn pgx_ARR_NULLBITMAP(arrayType: *mut pg_sys::ArrayType) -> *mut pg_sys::bits8;
+        pub fn pgx_ARR_HASNULL(arrayType: *mut pg_sys::ArrayType) -> bool;
     }
 
     #[inline]
@@ -444,6 +449,42 @@ mod all_versions {
             >,
             context: *mut ::std::os::raw::c_void,
         ) -> bool;
+    }
+
+    #[inline]
+    pub fn get_arr_data_ptr<T>(arr: *mut super::ArrayType) -> *mut T {
+        unsafe { pgx_ARR_DATA_PTR(arr) as *mut T }
+    }
+
+    #[inline]
+    pub fn get_arr_nelems(arr: *mut super::ArrayType) -> i32 {
+        unsafe { pgx_ARR_NELEMS(arr) }
+    }
+
+    #[inline]
+    pub fn get_arr_ndim(arr: *mut super::ArrayType) -> i32 {
+        unsafe { pgx_ARR_NDIM(arr) }
+    }
+
+    #[inline]
+    pub fn get_arr_nullbitmap<'a>(arr: *mut super::ArrayType) -> &'a [pg_sys::bits8] {
+        unsafe {
+            let len = (pgx_ARR_NELEMS(arr) + 7) / 8;
+            std::slice::from_raw_parts(pgx_ARR_NULLBITMAP(arr), len as usize)
+        }
+    }
+
+    #[inline]
+    pub fn get_arr_nullbitmap_mut<'a>(arr: *mut super::ArrayType) -> &'a mut [u8] {
+        unsafe {
+            let len = (pgx_ARR_NELEMS(arr) + 7) / 8;
+            std::slice::from_raw_parts_mut(pgx_ARR_NULLBITMAP(arr), len as usize)
+        }
+    }
+
+    #[inline]
+    pub fn get_arr_hasnull(arr: *mut super::ArrayType) -> bool {
+        unsafe { pgx_ARR_HASNULL(arr) }
     }
 }
 
--- a/pgx-pg-sys/src/lib.rs
+++ b/pgx-pg-sys/src/lib.rs
@@ -240,11 +240,6 @@ mod all_versions {
         pub fn pgx_list_nth_oid(list: *mut super::List, nth: i32) -> super::Oid;
         pub fn pgx_list_nth_cell(list: *mut super::List, nth: i32) -> *mut super::ListCell;
         pub fn pgx_GETSTRUCT(tuple: pg_sys::HeapTuple) -> *mut std::os::raw::c_char;
-        pub fn pgx_ARR_DATA_PTR(arrayType: *mut pg_sys::ArrayType) -> *mut u8;
-        pub fn pgx_ARR_NELEMS(arrayType: *mut pg_sys::ArrayType) -> i32;
-        pub fn pgx_ARR_NDIM(arrayType: *mut pg_sys::ArrayType) -> i32;
-        pub fn pgx_ARR_NULLBITMAP(arrayType: *mut pg_sys::ArrayType) -> *mut pg_sys::bits8;
-        pub fn pgx_ARR_HASNULL(arrayType: *mut pg_sys::ArrayType) -> bool;
     }
 
     #[inline]
@@ -449,42 +444,6 @@ mod all_versions {
             >,
             context: *mut ::std::os::raw::c_void,
         ) -> bool;
-    }
-
-    #[inline]
-    pub fn get_arr_data_ptr<T>(arr: *mut super::ArrayType) -> *mut T {
-        unsafe { pgx_ARR_DATA_PTR(arr) as *mut T }
-    }
-
-    #[inline]
-    pub fn get_arr_nelems(arr: *mut super::ArrayType) -> i32 {
-        unsafe { pgx_ARR_NELEMS(arr) }
-    }
-
-    #[inline]
-    pub fn get_arr_ndim(arr: *mut super::ArrayType) -> i32 {
-        unsafe { pgx_ARR_NDIM(arr) }
-    }
-
-    #[inline]
-    pub fn get_arr_nullbitmap<'a>(arr: *mut super::ArrayType) -> &'a [pg_sys::bits8] {
-        unsafe {
-            let len = (pgx_ARR_NELEMS(arr) + 7) / 8;
-            std::slice::from_raw_parts(pgx_ARR_NULLBITMAP(arr), len as usize)
-        }
-    }
-
-    #[inline]
-    pub fn get_arr_nullbitmap_mut<'a>(arr: *mut super::ArrayType) -> &'a mut [u8] {
-        unsafe {
-            let len = (pgx_ARR_NELEMS(arr) + 7) / 8;
-            std::slice::from_raw_parts_mut(pgx_ARR_NULLBITMAP(arr), len as usize)
-        }
-    }
-
-    #[inline]
-    pub fn get_arr_hasnull(arr: *mut super::ArrayType) -> bool {
-        unsafe { pgx_ARR_HASNULL(arr) }
     }
 }
 
--- a/pgx-tests/src/tests/array_tests.rs
+++ b/pgx-tests/src/tests/array_tests.rs
@@ -108,18 +108,10 @@ fn get_arr_nelems(arr: Array<i32>) -> libc::c_int {
 
 #[pg_extern]
 fn get_arr_data_ptr_nth_elem(arr: Array<i32>, elem: i32) -> Option<i32> {
-    let len = arr.len();
-    let arr_type = arr.into_array_type();
-
     unsafe {
-        let ptr = array::get_arr_data_ptr::<i32>(arr_type as *mut ArrayType);
-        let elem = elem as usize;
-        let data = std::slice::from_raw_parts(ptr, len);
-        if elem < len {
-            Some(data[elem])
-        } else {
-            None
-        }
+        let raw = RawArray::from_array(arr).unwrap().data_slice::<i32>();
+        let slice = &(*raw.as_ptr());
+        slice.get(elem as usize).copied()
     }
 }
 
--- a/pgx-tests/src/tests/array_tests.rs
+++ b/pgx-tests/src/tests/array_tests.rs
@@ -102,9 +102,8 @@ fn return_zero_length_vec() -> Vec<i32> {
 }
 
 #[pg_extern]
-fn get_arr_nelems(arr: Array<i32>) -> i32 {
-    let arr_type = arr.into_array_type();
-    array::get_arr_nelems(arr_type as *mut ArrayType)
+fn get_arr_nelems(arr: Array<i32>) -> libc::c_int {
+    unsafe { RawArray::from_array(arr).unwrap().len() }
 }
 
 #[pg_extern]
--- a/pgx-tests/src/tests/array_tests.rs
+++ b/pgx-tests/src/tests/array_tests.rs
@@ -7,7 +7,7 @@ All rights reserved.
 Use of this source code is governed by the MIT license that can be found in the LICENSE file.
 */
 
-use pgx::array::ArrayPtr;
+use pgx::array::{ArrayPtr, RawArray};
 use pgx::*;
 use serde_json::*;
 
@@ -114,7 +114,7 @@ unsafe fn get_arr_data_ptr_nth_elem(arr: Array<i32>, elem: i32) -> Option<i32> {
     // SAFETY: this is Known to be an Array from ArrayType,
     // and the index has to be a valid one.
     unsafe {
-        let raw = ArrayPtr::from_array(arr).unwrap().data::<i32>();
+        let raw = RawArray::from_valid(ArrayPtr::from_array(arr).unwrap()).data::<i32>();
         let slice = &(*raw.as_ptr());
         slice.get(elem as usize).copied()
     }
@@ -122,7 +122,7 @@ unsafe fn get_arr_data_ptr_nth_elem(arr: Array<i32>, elem: i32) -> Option<i32> {
 
 #[pg_extern]
 fn display_get_arr_nullbitmap(arr: Array<i32>) -> String {
-    let raw = unsafe { ArrayPtr::from_array(arr) }.unwrap();
+    let raw = unsafe { RawArray::from_valid(ArrayPtr::from_array(arr).unwrap()) };
 
     if let Some(slice) = raw.nulls() {
         // SAFETY: If the test has gotten this far, the ptr is good for 0+ bytes,
--- a/pgx-tests/src/tests/array_tests.rs
+++ b/pgx-tests/src/tests/array_tests.rs
@@ -125,7 +125,8 @@ fn display_get_arr_nullbitmap(arr: Array<i32>) -> String {
     let raw = unsafe { RawArray::from_array(arr) }.unwrap();
 
     if let Some(slice) = raw.nulls() {
-        // SAFETY: If the test has gotten this far, the ptr is good for 0+ bytes.
+        // SAFETY: If the test has gotten this far, the ptr is good for 0+ bytes,
+        // so reborrow NonNull<[u8]> as &[u8] for the hot second we're looking at it.
         let slice = unsafe { &*slice.as_ptr() };
         // might panic if the array is len 0
         format!("{:#010b}", slice[0])
--- a/pgx-tests/src/tests/array_tests.rs
+++ b/pgx-tests/src/tests/array_tests.rs
@@ -7,6 +7,7 @@ All rights reserved.
 Use of this source code is governed by the MIT license that can be found in the LICENSE file.
 */
 
+use pgx::array::RawArray;
 use pgx::{pg_sys::ArrayType, *};
 use serde_json::*;
 
@@ -135,9 +136,8 @@ fn display_get_arr_nullbitmap(arr: Array<i32>) -> String {
 }
 
 #[pg_extern]
-fn get_arr_ndim(arr: Array<i32>) -> i32 {
-    let arr_type = arr.into_array_type();
-    array::get_arr_ndim(arr_type as *mut ArrayType)
+fn get_arr_ndim(arr: Array<i32>) -> libc::c_int {
+    unsafe { RawArray::from_array(arr).unwrap().ndims() }
 }
 
 #[pg_extern]
--- a/pgx-tests/src/tests/array_tests.rs
+++ b/pgx-tests/src/tests/array_tests.rs
@@ -138,7 +138,7 @@ fn display_get_arr_nullbitmap(arr: Array<i32>) -> String {
 #[pg_extern]
 fn get_arr_ndim(arr: Array<i32>) -> libc::c_int {
     // SAFETY: This is a valid ArrayType and it's just a field access.
-    unsafe { ArrayPtr::from_array(arr) .unwrap().ndims() }
+    unsafe { ArrayPtr::from_array(arr).unwrap().ndims() }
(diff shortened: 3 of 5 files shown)


pgx-pg-sys/cshim/pgx-cshim.c
```diff
@@ -110,3 +111,28 @@ PGDLLEXPORT char *pgx_GETSTRUCT(HeapTuple tuple);
...
+
+PGDLLEXPORT bits8 *pgx_ARR_NULLBITMAP(ArrayType *arr);
+bits8 *pgx_ARR_NULLBITMAP(ArrayType *arr) {
+    return ARR_NULLBITMAP(arr);
+}
+
+PGDLLEXPORT int pgx_ARR_NDIM(ArrayType *arr);
+int pgx_ARR_NDIM(ArrayType *arr) {
+    return ARR_NDIM(arr);
+}
```
workingjubilee, line 133: Hmm. Shouldn't we also directly expose `ARR_DIMS`, as it might be desired by Rust code?

I thought `ARR_NDIM(ArrayType*)` was just a direct field read,
so if I remember my C, `ARR_NELEMS` could be written:
```c
int pgx_ARR_NELEMS(ArrayType *arr) {
    return ArrayGetNItems(arr->ndim, ARR_DIMS(arr));
}
```

pgx-pg-sys/src/lib.rs
```diff
@@ -445,6 +450,42 @@ mod all_versions {
             context: *mut ::std::os::raw::c_void,
         ) -> bool;
     }
+
+    #[inline]
+    pub fn get_arr_data_ptr<T>(arr: *mut super::ArrayType) -> *mut T {
+        unsafe { pgx_ARR_DATA_PTR(arr) as *mut T }
+    }
```
workingjubilee, line 457: This is not a safe function. It is unsound to call this if `*mut ArrayType` is the null pointer or an invalid pointer, as it involves **reading a field** of the pointee to determine what to do next. A function must **not** be safe if it is memory-unsafe to call it with **any** valid input for that type, **especially** including the null pointer.

This function may be desirable but it **must** be `unsafe`.

pgx-pg-sys/src/lib.rs
```diff
@@ -445,6 +450,42 @@ mod all_versions {
...
+        unsafe {
+            let len = (pgx_ARR_NELEMS(arr) + 7) / 8;
+            std::slice::from_raw_parts_mut(pgx_ARR_NULLBITMAP(arr), len as usize)
+        }
+    }
+
+    #[inline]
+    pub fn get_arr_hasnull(arr: *mut super::ArrayType) -> bool {
+        unsafe { pgx_ARR_HASNULL(arr) }
+    }
```
workingjubilee, line 488: I believe none of these are actually safe due to the prior reason: a field read on a possibly-invalid, possibly-null pointer happens.

workingjubilee (changes requested):
Do the helper functions have to live in `pgx-pg-sys`? Because they should be purely additive to the sys crate, I would somewhat prefer they were exposed through a `mod` in the main `pgx` crate, so as to make it easier to thread them through the other code in `pgx` without violating coherence.

I also would like it if all new `unsafe` code that lands in PGX to document its public `/// # Safety` requirements as precisely as possible, and for every invocation of `unsafe` to come with `// SAFETY:` annotations,  and I would prefer it to come with `#[deny(unsafe_op_in_unsafe_fn)]` on the containing module so as to require **both of those** even inside an `unsafe fn`.

It is the last one, the extreme safety requirements, that is completely impractical to implement throughout `pgx-pg-sys`. Note that the comments I desire can be **very** terse, they do not have to be elaborate, they just have to make clear all the little details like "this does a field read".

workingjubilee:
The burden of assuring memory safety must end the moment code leaves an `unsafe` context. This means that a safe function may not impose any requirements whatsoever on its callers beyond that which it imposes on the purely programmatic level. It is permissible to induce logic errors for illogical inputs, but trying to read values out of the null page (or, really, any field offset from the null pointer) is definitely not one of them.

e.g. It is fine if an impl of `Ord` is incorrect for some comparisons, because Ord is safe code. Thus when your `sort` fn relies on an impl of `Ord` to be correct in order to return a correct sort, it is fine if the resulting sort is wrong.

It is not fine if you then rely on the impl of `Ord` being correct for memory safety purposes. This is an actual, in-the-wild concern in real code that is not Rust, as glibc's `qsort` in the past has induced segfaults on incorrect comparison functions. Rust does not tolerate that kind of shenanigans, so glibc's `qsort` would be considered `unsafe`. The direct implication is that safe code is allowed to "shit the bed", from the perspective of `unsafe` code, and the `unsafe` code must either pass the burden on to a caller, or clean it up. Thus, for a function which provides a safe wrapper for `unsafe` code, **the buck stops there**. It must guarantee all safely-obtained inputs in Rust (and it is "safe" code to write `std::ptr::null_mut()`) to the `unsafe` block inside do not violate memory safety.

...Directly, mind. Unfortunately running `DROP DATABASE IF EXISTS prod` is in the category of things that are considered "logic errors", not "memory safety violations", despite having a marvelous potential to erase data you may have wanted to keep around.

pgx/src/array.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /// A raw slice of the dimensions.
+    /// Oxidized form of ARR_DIMS(ArrayType*)
+    ///
+    /// # Safety
+    ///
+    /// Be aware that if you get clever and use this pointer beyond owning RawArray, it's wrong!
+    /// Raw pointer validity is **asserted on dereference, not construction**,
+    /// and this slice is no longer valid if you do not also hold RawArray.
```
workingjubilee, line 103: This isn't quite right. It's fine to use this if you also "consumed" RawArray to produce the slice.

pgx-pg-sys/src/lib.rs
workingjubilee: I have introduced safe wrappers for everything except the null bitmap functions, which will require another wave of commits.

workingjubilee (dismissed):
Needs a second pass, as already mentioned.

pgx/src/array.rs
```diff
@@ -0,0 +1,166 @@
...
+    pub fn pgx_ARR_DIMS(arrayType: *mut ArrayType) -> *mut libc::c_int;
+}
+
+#[inline]
+pub fn get_arr_nullbitmap<'a>(arr: *mut ArrayType) -> &'a [bits8] {
+    unsafe {
+        let len = (pgx_ARR_NELEMS(arr) + 7) / 8;
+        std::slice::from_raw_parts(pgx_ARR_NULLBITMAP(arr), len as usize)
+    }
+}
```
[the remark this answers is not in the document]
workingjubilee: This was verrry incorrect to use, as it went from a possibly null, possibly invalid raw pointer, to a slice.

pgx-tests/src/tests/array_tests.rs
```diff
@@ -99,6 +100,52 @@ fn return_zero_length_vec() -> Vec<i32> {
...
+}
+
+/// # Safety
+/// Don't call with an actually-null index, or you might see an invalid value.
+#[pg_extern]
+#[deny(unsafe_op_in_unsafe_fn)]
+unsafe fn get_arr_data_ptr_nth_elem(arr: Array<i32>, elem: i32) -> Option<i32> {
+    // SAFETY: this is Known to be an Array from ArrayType,
+    // and the index has to be a valid one.
+    unsafe {
```
BradyBonnette, line 116: What is the reason for marking the fn as `unsafe` but also having the body of the function also wrapped in an `unsafe` block? The `get_arr_nelems` function above does not have it. Is it because this fn returns an `Option<>`?

pgx-tests/src/tests/array_tests.rs
```diff
@@ -99,6 +100,52 @@ fn return_zero_length_vec() -> Vec<i32> {
...
+    }
+}
+
+#[pg_extern]
+fn display_get_arr_nullbitmap(arr: Array<i32>) -> String {
+    let raw = unsafe { RawArray::from_array(arr) }.unwrap();
+
+    if let Some(slice) = raw.nulls() {
+        // SAFETY: If the test has gotten this far, the ptr is good for 0+ bytes.
+        let slice = unsafe { &*slice.as_ptr() };
```
BradyBonnette, line 129: `&*slice.as_ptr()`  👀

pgx-tests/src/tests/array_tests.rs
```diff
@@ -255,6 +302,51 @@ mod tests {
...
+    }
+
+    #[pg_test]
+    fn test_display_get_arr_nullbitmap() {
+        let bitmap_str = Spi::get_one::<String>(
+            "SELECT display_get_arr_nullbitmap(ARRAY[1,NULL,3,NULL,5]::int[])",
+        )
+        .expect("failed to get SPI result");
+
+        assert_eq!(bitmap_str, "0b00010101");
```
BradyBonnette, line 328: Is this to show which indexes have nulls and which indexes have non-null (apologies for the crude ascii diagram that may not render correctly)?

```
0b00010101
    [5n3n1]
```

pgx/src/array.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// Must only be used on a "valid" (Postgres-constructed) ArrayType
+    fn pgx_ARR_NELEMS(arrayType: *mut ArrayType) -> i32;
+    /// # Safety
+    /// Does a field access, but doesn't deref out of bounds of ArrayType
+    fn pgx_ARR_NULLBITMAP(arrayType: *mut ArrayType) -> *mut bits8;
+}
+
+/// Handle describing a bare, "untyped" pointer to an array,
+/// offering safe accessors to the various fields of one.
+#[repr(transparent)]
```
BradyBonnette, line 23: 👀 Had to look up `repr(transparent)`

pgx/src/array.rs
```diff
@@ -0,0 +1,215 @@
...
+    pub fn len(&self) -> libc::c_int {
+        // SAFETY: Validity asserted on construction, and...
+        // ...well, hopefully Postgres knows what it's doing.
+        unsafe {
+            pgx_ARR_NELEMS(self.at.as_ptr())
+            // FIXME: While this was indeed a function that returns a c_int,
+            // using a usize is more idiomatic in Rust, to say the least.
+            // In addition, the actual sizes are under various restrictions,
+            // so we probably can further constrain the value, honestly.
+            // However, PGX has trouble with returning usizes
```
BradyBonnette, line 133: `// However, PGX has trouble with returning usizes`

I saw this comment earlier. Is this something that'll need to be addressed at some point? Can it even be addressed?

pgx/src/array.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// Gets the offset to the ArrayType's data.
+    /// Note that this should not be "taken literally".
+    fn data_offset(&self) -> i32 {
+        // SAFETY: Validity asserted on construction.
+        unsafe { (*self.at.as_ptr()).dataoffset }
+        // This field is an "int32" in Postgres
+    }
+
+    /// Equivalent to ARR_HASNULL(ArrayType*)
+    /// Note this means that it only asserts that there MIGHT be a null
```
BradyBonnette, line 152: `MIGHT` 🤔

pgx/src/array.rs
```diff
@@ -0,0 +1,215 @@
...
+    ///
+    /// This is not inherently typesafe!
+    /// Thus you must know the implied type of the underlying ArrayType when calling this.
+    /// In addition, the raw slice is not guaranteed to be legible at any given index,
+    /// e.g. it may be an "SQL null" if so indicated in the null bitmap.
+    /// But even if the index is not marked as null, the value may be equal to nullptr,
+    /// thus leaving it correct to read the value but incorrect to then dereference.
+    ///
+    /// That is why this returns `NonNull<[T]>`: if it returned `&mut [T]`,
+    /// then for many possible types, that would actually be UB, as it would assert
```
BradyBonnette, line 192: What's `UB` here? "Upper bound"?

pgx/src/array.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// that each particular index was a valid `T`.
+    pub unsafe fn data<T>(&mut self) -> NonNull<[T]> {
+        let len = self.len() as usize;
```
BradyBonnette, line 195: Does using `usize` here have any implications, since earlier it was mentioned that PGX has issues returning `usize`?

pgx-tests/src/tests/array_tests.rs
workingjubilee: It's because I think we should assume that if a Postgres varlena array has "nullable indices", that we should treat the resulting slice as [`&[MaybeUninit<T>]`](https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html) so it is incorrect to read the value of an index that was nulled.

The latter is because I want to move towards `#[deny(unsafe_op_in_unsafe_fn)]` across the repository.

pgx/src/array.rs
workingjubilee: Yes, it should be addressed.
I have no idea how best to do so currently.

pgx/src/array.rs
workingjubilee: undefined behavior

pgx/src/array.rs
workingjubilee: Basically Rust indices by `usize`, but Postgres indices with `c_int`, so we can usually safely upgrade a Postgres index into a Rust index because either the machine is 32-bit or 64-bit and we're not really running on 16-bit machines.

pgx-tests/src/tests/array_tests.rs
workingjubilee: Correct. @mhov wrote that test, but it basically prints the bitwise repr of the null bitmap that Postgres uses internally.

pgx-tests/src/tests/array_tests.rs
workingjubilee: This pattern is called a "reborrow".
It brings us from `NonNull<[T]>` to `&[T]`.

pgx-tests/src/tests/array_tests.rs
workingjubilee: I believe it's actually the reverse though:
```
00001010
   5n3n1
```

pgx/src/array.rs
workingjubilee: PERHAPS UNNECESSARY CAPITALIZATION, or?

pgx/src/array.rs
workingjubilee: Hm. I'm not sure if it actually should be, tbh. Might want to look up the `len` and cache it on construction.

pgx/src/array.rs
BradyBonnette: Ah ok

pgx/src/array.rs
BradyBonnette: No, not at all, it's a good callout. It was more like "Im calling something that might have null" instead of "I am calling something I know for sure will be null or not".

pgx/src/array.rs
BradyBonnette: I don't know why I keep forgetting that, lol

pgx/src/array.rs
BradyBonnette: Ohhhhhhhh ok, I see. Thanks!

pgx-tests/src/tests/array_tests.rs
BradyBonnette: Ah, more reading material. Thanks 😄

pgx/src/array.rs
```diff
@@ -18,8 +18,19 @@ extern "C" {
...
+///
+/// # On sizes and subscripts
+///
+/// Postgres uses C's `int` (`c_int` in Rust) for sizes, and Rust uses `usize`.
+/// Thus various functions of RawArray return `c_int` values, but you must convert to usize.
+/// On 32-bit or 64-bit machines with 32-bit `c_int`s, you can losslessly upgrade to `as usize`,
+/// except with negative indices, which Postgres asserts against creating.
+/// PGX currently only intentionally supports 64-bit machines,
+/// and while support for ILP32 or I64LP128 C data models may become possible,
+/// PGX will **not** support 16-bit machines in any practical case, even though Rust does.
```
BradyBonnette, line 33: Thanks for the expanded explanation 😃

pgx/src/array.rs
workingjubilee: > Hm. I'm not sure if it actually should be, tbh. Might want to look up the `len` and cache it on construction.

Ah, I believe I will learn what we should do when I construct the followups to this that wrap RawArray further to allow more "fully safe" handling of it. It may require changing some design specs given here.

pgx/src/array.rs
```diff
@@ -0,0 +1,235 @@
...
+
+        // SAFETY: Welcome to the infernal bowels of FFI.
+        // Because the initial ptr was NonNull, we can assume this is also NonNull.
+        // As validity of the initial ptr was asserted on construction of RawArray,
+        // this can assume the data ptr is also valid, allowing making the slice.
+        // This code doesn't assert validity per se, but in practice,
+        // the caller will probably immediately turn this into a borrowed slice,
+        // opening up the methods that are available on borrowed slices.
+        //
+        // Most importantly, the caller has asserted this is in fact a valid [T],
```
workingjubilee, line 225: You don't specifically document that it's asserting `[T]`, you're only asking for `T`, and there's some nuances between those.

pgx/src/array.rs
```diff
@@ -0,0 +1,235 @@
...
+        // postgres/src/include/utils/array.h
+        // #define ARR_HASNULL
+        self.data_offset() != 0
+    }
+
+    /// Oxidized form of ARR_NULLBITMAP(ArrayType*)
+    pub fn nulls(&self) -> Option<NonNull<[u8]>> {
+        // for expected behavior, see:
+        // postgres/src/include/utils/array.h
+        // #define ARR_NULLBITMAP
```
workingjubilee, line 175: Same question: maybe this should path should be public?

pgx/src/array.rs
```diff
@@ -0,0 +1,235 @@
...
+    ///
+    /// This requires &mut to collect the slice, but be aware: this is a raw pointer!
+    /// If you find ways to store it, you are probably violating ownership.
+    /// Raw pointer validity is **asserted on dereference, not construction**,
+    /// so remember, this slice is only guaranteed to be valid almost immediately
+    /// after obtaining it, or if you continue to hold the RawArray.
+    pub fn dims(&mut self) -> NonNull<[libc::c_int]> {
+        // for expected behavior, see:
+        // postgres/src/include/utils/array.h
+        // #define ARR_DIMS
```
workingjubilee, line 112: Should this path be visible in the public documentation? I suppose they can click through to see the source, but the public docs will support links.

pgx/src/array.rs
```diff
@@ -0,0 +1,235 @@
...
+#[deny(unsafe_op_in_unsafe_fn)]
+impl RawArray {
+    // General implementation notes:
+    // RawArray is not Copy or Clone, making it harder to misuse versus *mut ArrayType.
+    // But this also offers safe accessors to the fields, like &ArrayType,
+    // so it requires validity assertions in order to be constructed.
+    // The main reason it uses NonNull and denies Clone, however, is access soundness:
+    // It is not sound to go from &mut Type to *mut T if *mut T is not a field of Type.
+    // This creates an obvious complication for handing out pointers into varlenas.
+    // Thus also why this does not use lifetime-bounded borrows.
```
workingjubilee, line 49: Talking about the intent of this type might be useful to publicly discuss in its documentation.

pgx/src/array.rs
```diff
@@ -0,0 +1,235 @@
...
+    /// Must only be used on a "valid" (Postgres-constructed) ArrayType
+    fn pgx_ARR_NELEMS(arrayType: *mut ArrayType) -> i32;
+    /// # Safety
+    /// Does a field access, but doesn't deref out of bounds of ArrayType
+    fn pgx_ARR_NULLBITMAP(arrayType: *mut ArrayType) -> *mut bits8;
+}
+
+/// Handle describing a bare, "untyped" pointer to a Postgres varlena array,
+/// offering safe accessors to the various fields of the ArrayType,
+/// and access to the various slices that are implicitly present in the varlena.
```
workingjubilee, line 23: You may want a more terse initial comment, I think, for the "snippet form" in the docs.

pgx/src/array.rs
```diff
@@ -0,0 +1,235 @@
...
+    /// * The pointer must be properly aligned.
+    /// * It must be "dereferenceable" in the sense defined in [the std documentation].
+    /// * The pointer must point to an initialized instance of `ArrayType`.
+    /// * This can be considered a unique, "owning pointer",
+    ///   so it won't be aliased while RawArray is held,
+    ///   and it points to data in the Postgres ArrayType format.
+    ///
+    /// It should be noted as RawArray is not inherently lifetime-bound, it can be racy and unsafe!
+    ///
+    /// [the std documentation]: core::ptr#safety
```
workingjubilee, line 65: Style elsewhere in the repo has been to use `/*` (block), `/*!` (inner block doc), and `/**` (block doc) for multiline comments, consider applying that here?

workingjubilee (changes requested):
This needs improvement in the documentation before it lands, even though there is going to be a followup PR soon.

pgx-tests/src/tests/array_tests.rs
workingjubilee: You should open an issue if you want to do this across the entire repo, @workingjubilee, so other people working with you can help with it.a

workingjubilee: Opened:
- https://github.com/tcdi/pgx/issues/653

pgx-tests/src/tests/array_tests.rs
workingjubilee: This tendency in the formatting detail needs to be documented.

workingjubilee: What were you saying, @workingjubilee? It is what Brady just said, as witnessed here in this very test. The nullbitmap uses a 0 to mark "is null", unlike all the rest of Postgres and `NullableDatum`, which uses `true` bools to mark "is null", because Postgres has very good ideas.

pgx/src/array.rs
```diff
@@ -0,0 +1,235 @@
...
+    pub unsafe fn from_raw(at: NonNull<ArrayType>) -> RawArray {
+        RawArray { at }
+    }
+
+    /// # Safety
+    ///
+    /// Array must have been constructed from an ArrayType pointer.
+    pub unsafe fn from_array<T: FromDatum>(arr: Array<T>) -> Option<RawArray> {
+        let array_type = arr.into_array_type() as *mut _;
```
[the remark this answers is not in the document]
workingjubilee: I was going to change it to not do that NULL-check in a followup PR. Should I fix that now?

pgx/src/array.rs
```diff
@@ -0,0 +1,235 @@
...
+    /// Oxidized form of ARR_DIMS(ArrayType*)
+    ///
+    /// # Safety
+    ///
+    /// This requires &mut to collect the slice, but be aware: this is a raw pointer!
+    /// If you find ways to store it, you are probably violating ownership.
+    /// Raw pointer validity is **asserted on dereference, not construction**,
+    /// so remember, this slice is only guaranteed to be valid almost immediately
+    /// after obtaining it, or if you continue to hold the RawArray.
+    pub fn dims(&mut self) -> NonNull<[libc::c_int]> {
```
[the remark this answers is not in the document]
workingjubilee: Mmh. Isn't there a case people can make an incorrect `&[T]` slice if we don't offer this ourselves?

pgx/src/array.rs
workingjubilee: Yeah, actually, directly linking is a good idea. And that way we could also link to the doxygen.

pgx/src/array.rs
```diff
@@ -0,0 +1,235 @@
...
+            let len = self.ndims() as usize;
+            NonNull::new_unchecked(slice_from_raw_parts_mut(
+                pgx_ARR_DIMS(self.at.as_ptr()),
+                len,
+            ))
+        }
+    }
+
+    /// The flattened length of the array.
+    pub fn len(&self) -> libc::c_int {
```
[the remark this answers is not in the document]
workingjubilee: Well, it's the number of elements in the entire array, covering both those that are invalid (i.e. marked null in the nullbitmap) and valid, without regard to any given dimension.

I guess it's more traditional to call that "size" or something? Postgres uses `nitems`, and elsewhere PGX uses `nelems`.

pgx/src/array.rs
```diff
@@ -0,0 +1,235 @@
...
+        // #define ARR_HASNULL
+        self.data_offset() != 0
+    }
+
+    /// Oxidized form of ARR_NULLBITMAP(ArrayType*)
+    pub fn nulls(&self) -> Option<NonNull<[u8]>> {
+        // for expected behavior, see:
+        // postgres/src/include/utils/array.h
+        // #define ARR_NULLBITMAP
+        let len = (self.len() as usize + 7) >> 3; // Obtains 0 if len was 0.
```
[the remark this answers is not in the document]
workingjubilee: It's just going to be optimized to this anyways, and sometimes `%` and `/` are missed opts by LLVM.

pgx/src/array.rs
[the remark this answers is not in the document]
workingjubilee: Oh nah that's fine, just leave this unresolved until the followup PR, it's nonblocking.

pgx/src/array.rs
```diff
@@ -0,0 +1,235 @@
...
+            ))
+        }
+    }
+
+    /// The flattened length of the array.
+    pub fn len(&self) -> libc::c_int {
+        // SAFETY: Validity asserted on construction, and...
+        // ...well, hopefully Postgres knows what it's doing.
+        unsafe {
+            pgx_ARR_NELEMS(self.at.as_ptr())
```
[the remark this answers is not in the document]
workingjubilee: Okay, I get what you are saying, I think. We should really not deref the slices we are exposing to users, and only do the absolute minimum possible here. That also weakens the requirements on this type. Then the next type can build on top of this with a safer abstraction, so the nullbitmap ptr and the data ptr both should point to `Option<NonNull<u8>>` and `Option<NonNull<T>` respectively. The user of this type can cache the len and wrap this properly and return slices. I'm just worried about people misusing these pointers.

pgx/src/array.rs
```diff
@@ -0,0 +1,235 @@
...
+    /// Note this means that it only asserts that there MIGHT be a null
+    pub fn nullable(&self) -> bool {
+        // for expected behavior, see:
+        // postgres/src/include/utils/array.h
+        // #define ARR_HASNULL
+        self.data_offset() != 0
+    }
+
+    /// Oxidized form of ARR_NULLBITMAP(ArrayType*)
+    pub fn nulls(&self) -> Option<NonNull<[u8]>> {
```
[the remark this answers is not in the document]
workingjubilee: I like the pithy one-word names, tho', and `nullbitmap()` is not really idiomatic for Rust, it would be `null_bitmap`, I think, which feels much worse, and it would be the `nulls_bitmap` anyways, since it describes the indices of `nulls`, and then you see how this got back to just `nulls`, yes?

pgx/src/array.rs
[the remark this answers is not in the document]
workingjubilee: Yeah, this should remain as-is, and actually `ndims` is probably the fn that should go, because they can call `dims().len()`

pgx/src/array.rs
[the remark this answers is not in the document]
workingjubilee: Ugh, the results of this are too painful to use. Having two types for the same low-level accessors feels a bit silly???

pgx/src/array.rs
```diff
@@ -232,20 +235,68 @@ impl ArrayPtr {
...
 A pointer into a Postgres varlena.
 */
-pub struct RawArray<T> {
-    _ptr: ArrayPtr,
-    _len: usize,
-    ty: PhantomData<T>,
+pub struct RawArray {
+    ptr: ArrayPtr,
+    len: usize,
+}
```
workingjubilee, line 248: This was a cute idea for a few seconds but then you dropped the PhantomData type. These are definitely not different things.

pgx/src/array.rs
```diff
@@ -232,20 +235,68 @@ impl ArrayPtr {
...
+        // the validity was asserted on construction. However, unlike the other cases,
+        // it isn't correct to trust it. Instead, this gets null-checked.
+        // This is because, while the initial pointer is NonNull,
+        // ARR_NULLBITMAP can return a nullptr!
+        Some(unsafe { NonNull::new_unchecked(slice_from_raw_parts_mut(self.ptr.nulls()?.as_ptr(), len)) } )
+    }
+
+    /// # Safety
+    /// Asserts the data is mutable, properly aligned, all that jazz.
+    pub unsafe fn data<T>(&mut self) -> NonNull<[T]> {
```
workingjubilee, line 290: I don't think this needs to be unsafe if it loads the requirements into construction.

workingjubilee (changes requested):

pgx/src/array.rs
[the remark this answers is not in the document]
workingjubilee: Really? Sad.

pgx/src/array.rs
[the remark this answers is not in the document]
workingjubilee: Fair enough.

pgx/src/array.rs
workingjubilee: Well, also, invalid pointers aren't unsafe to construct, yadda yadda.

pgx/src/array.rs
workingjubilee: Doxygen is too unstable, will point to master and lines might change. Instead used gitweb links.

pgx/src/array.rs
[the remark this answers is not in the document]
BradyBonnette: In my mind, if the intention is for it to be used as a bitmap by the caller (as opposed to a "regular" map), then I don't think there's anything inherently wrong with having the world "bitmap" in there somewhere. It at least makes the intention clear, and a few extra characters probably won't hurt anything.

I don't think I can necessarily speak to how idiomatic either/all names would be, but having a clear intention is nice.

Though on second look, the one thing that did sorta confuse me (as an early-stage Rust developer, mind) is that the intention is "this is a bitmap of things which indicate that something is Null" but the Option type is `NonNull`. I get that `NonNull` is a pointer type (I think?), and more seasoned Rust developers would understand that in a heartbeat. It just sorta threw me off for a quick second.

pgx/src/array.rs
```diff
@@ -0,0 +1,235 @@
...
+    /// Equivalent to ARR_HASNULL(ArrayType*)
+    /// Note this means that it only asserts that there MIGHT be a null
+    pub fn nullable(&self) -> bool {
+        // for expected behavior, see:
+        // postgres/src/include/utils/array.h
+        // #define ARR_HASNULL
+        self.data_offset() != 0
+    }
+
+    /// Oxidized form of ARR_NULLBITMAP(ArrayType*)
```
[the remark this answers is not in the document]
BradyBonnette: How much detail would be required to make it "make sense" to a newcomer without also making it overbearing to read? Would a consumer of this need to fully understand what Postgres does with it?

pgx/src/array.rs
```diff
@@ -0,0 +1,235 @@
...
+        unsafe { (*self.at.as_ptr()).dataoffset }
+        // This field is an "int32" in Postgres
+    }
+
+    /// Equivalent to ARR_HASNULL(ArrayType*)
+    /// Note this means that it only asserts that there MIGHT be a null
+    pub fn nullable(&self) -> bool {
+        // for expected behavior, see:
+        // postgres/src/include/utils/array.h
+        // #define ARR_HASNULL
```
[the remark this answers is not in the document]
BradyBonnette: I don't think it'd hurt, and having it reference Postgres' source is nice to know. Not as in an actual hyperlink that could change over time, but just a general "look here" as you have I think is sufficient. Actually there was a point in time while reviewing this stuff I looked at the Postgres source mirror on Github just to reference what you had written.

pgx/src/array.rs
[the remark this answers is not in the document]
BradyBonnette: I kinda wonder if your answer to your own question should also be in the comments?  😃

pgx/src/array.rs
BradyBonnette: From a pgx-maintainers perspective, how much of a maintenance nightmare is it linking to the doxygen pages? How often would that change?

workingjubilee: Potentially every time someone lands a commit into postgresql/master.

pgx/src/array.rs
```diff
@@ -0,0 +1,288 @@
...
+    /// Will be in 0..=[pg_sys::MAXDIM].
+    fn ndim(&self) -> libc::c_int {
+        // SAFETY: Validity asserted on construction.
+        unsafe {
+            (*self.ptr.as_ptr()).ndim
+            // FIXME: While this is a c_int, the max ndim is normally 6
+            // While the value can be set higher, it is... unlikely
+            // that it is going to actually challenge even 16-bit pointer widths.
+            // It would be preferable to return a usize instead,
+            // however, PGX has trouble with that, unfortunately.
```
workingjubilee, line 116: Stray big multiline comment with `//`

workingjubilee (approved):
Wow, that was... quite a few commits, and only 11% were rustfmt commits! Nicely done!

pgx/src/array.rs
```diff
@@ -0,0 +1,286 @@
...
+    Write to them in order.
+    */
+    pub unsafe fn dims_mut(&mut self) -> (&mut libc::c_int, NonNull<libc::c_int>, &mut usize) {
+        // SAFETY: Validity asserted on construction, origin ptr is non-null.
+        let dims_ptr = unsafe { NonNull::new_unchecked(pgx_ARR_DIMS(self.ptr.as_ptr())) };
+        let len = &mut self.len;
+
+        // SAFETY: Validity asserted on construction. Just reborrowing a subfield.
+        (unsafe { &mut self.ptr.as_mut().ndim }, dims_ptr, len)
+    }
```
workingjubilee, line 165: You ain't gonna need it, tho'.

workingjubilee: Okaaaay it was just fun to write.

PR #637, opened 2022-08-17
Hash functions correctly depend on _eq functions

### LINKED ISSUE #611: Missing DAG edge between "PostgresEq" and "PostgresHash"
When a struct derives both `PostgresEq` and `PostgresHash`, the generated SQL will include:
```sql
CREATE OPERATOR = (
  -- ...
```
and:
```sql
CREATE OPERATOR FAMILY ...
CREATE OPERATOR CLASS ...
    OPERATOR 3 =,
```
However, the order of these two items is not defined in the DAG, and if the `OPERATOR FAMILY` happens to come first, PostgreSQL will not be able to load the extension, with an error message like `ERROR:  operator does not exist: foo = foo`. This causes the built extension to not work about 50% of the time. See [discussion in Discord](https://discord.com/channels/561648697805504526/749453981494804590/1001507547745894420) from this morning.

willmurnane:
When a function is marked with `#[derive(PostgresHash, ...)]` SQL code like the following is generated:
```
CREATE OPERATOR FAMILY Foo_hash_ops USING hash;
CREATE OPERATOR CLASS Foo_hash_ops DEFAULT FOR TYPE Foo USING hash FAMILY Foo_hash_ops AS
	OPERATOR    1   =  (Foo, Foo),
	FUNCTION    1   foo_hash(Foo);
```
However, in the pgx schema DAG, there is no dependency currently emitted between the `OPERATOR CLASS` and the `foo_eq` function (or the `OPERATOR =` associated with it). This means about half the time, the hash stuff is generated before the definition of equals, which means the extension will fail to install.

This merge request adds a dependency to the DAG, causing the equals function to always be emitted before the hash function. Closes #611 .

### DIFF
--- a/pgx-utils/src/sql_entity_graph/pgx_sql.rs
+++ b/pgx-utils/src/sql_entity_graph/pgx_sql.rs
@@ -210,6 +210,7 @@ impl PgxSql {
         connect_externs(
             &mut graph,
             &mapped_externs,
+            &mapped_hashes,
             &mapped_schemas,
             &mapped_types,
             &mapped_enums,
@@ -961,6 +962,7 @@ fn initialize_externs(
 fn connect_externs(
     graph: &mut StableGraph<SqlGraphEntity, SqlGraphRelationship>,
     externs: &HashMap<PgExternEntity, NodeIndex>,
+    hashes: &HashMap<PostgresHashEntity, NodeIndex>,
     schemas: &HashMap<SchemaEntity, NodeIndex>,
     types: &HashMap<PostgresTypeEntity, NodeIndex>,
     enums: &HashMap<PostgresEnumEntity, NodeIndex>,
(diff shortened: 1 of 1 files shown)


pgx-utils/src/sql_entity_graph/pgx_sql.rs
```diff
@@ -1017,6 +1020,21 @@ fn connect_externs(
             );
         }
 
+        // The hash function must be defined after the {typename}_eq function.
+        for (hash_item, &hash_index) in hashes {
+            tracing::debug!(
```
Hoverbear, line 1025: Could we prune out this message? I think it's extraneous.

Hoverbear (commented):
Rest looks good!

pgx-utils/src/sql_entity_graph/pgx_sql.rs
willmurnane: Good catch, done!

workingjubilee:
This looks lovely! Thank you!

PR #638, opened 2022-08-18
Example aggregate allows for NULLs

willmurnane:
The example aggregation as provided crashes if a NULL value is aggregated (to see what I mean, add the example test without making other changes). This PR fixes this behavior.


workingjubilee (approved):
Thank you!

PR #640, opened 2022-08-18
Eliminate duplicate function modifiers.

willmurnane:
Some conditions cause the `STRICT` function modifier to be added to the `extern_attrs` list multiple times, which causes misbehavior like #639.

Closes #639.

PR #641, opened 2022-08-18
Input and output functions should be strict

willmurnane:
Per [discussion in Discord](https://discord.com/channels/561648697805504526/749454482450153594/1009841644805300276) this morning, the input and output functions (used to convert from {json, strings, varlena} to custom types) should either be strict or have `Option<T>` as parameters. Otherwise, statements like `SELECT NULL::customtype` will crash the server.

### DIFF
--- a/pgx-macros/src/lib.rs
+++ b/pgx-macros/src/lib.rs
@@ -772,13 +772,13 @@ fn impl_postgres_type(ast: DeriveInput) -> proc_macro2::TokenStream {
             impl #generics JsonInOutFuncs #inout_generics for #name #generics {}
 
             #[doc(hidden)]
-            #[pg_extern(immutable,parallel_safe)]
+            #[pg_extern(immutable,parallel_safe,strict)]
             pub fn #funcname_in #generics(input: &#lifetime pgx::cstr_core::CStr) -> #name #generics {
                 #name::input(input)
             }
 
             #[doc(hidden)]
-            #[pg_extern(immutable,parallel_safe)]
+            #[pg_extern(immutable,parallel_safe,strict)]
             pub fn #funcname_out #generics(input: #name #generics) -> &#lifetime pgx::cstr_core::CStr {
                 let mut buffer = StringInfo::new();
                 input.output(&mut buffer);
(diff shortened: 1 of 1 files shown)


willmurnane:
FYI: This will probably not build until #640 is merged.

pgx-macros/src/lib.rs
```diff
@@ -772,13 +772,13 @@ fn impl_postgres_type(ast: DeriveInput) -> proc_macro2::TokenStream {
             impl #generics JsonInOutFuncs #inout_generics for #name #generics {}
 
             #[doc(hidden)]
-            #[pg_extern(immutable,parallel_safe)]
+            #[pg_extern(immutable,parallel_safe,strict)]
             pub fn #funcname_in #generics(input: &#lifetime pgx::cstr_core::CStr) -> #name #generics {
                 #name::input(input)
             }
```
workingjubilee, line 778: Are we sure these are not going to be Options?

willmurnane: I think there are a couple of ways this could work:
* Functions are strict, return a concrete value, and `panic!` if they can't return a value
* Functions are non-strict, return a concrete value, and `panic!` if they can't return a value
* Functions are non-strict, return an `Option`, and `error!` or `warn!` or whatever if they decide they want to, and just return `None` sometimes.
* Functions are non-strict, and return a `Result` (or even `Result<Option>`) which is somehow transmuted into a `panic!` in the `Err` case.

In the first case, `NULL` always translates to `NULL` as a custom type. I think this is reasonable, but maybe there's some domain where `NULL` should translate to something else: an empty string, 0, banana flavor. If we think we need to support that case, then this option won't work.

In the second, everyone has to handle `Option`s, but hey it's not too bad. This doesn't expose a way of explicitly returning `NULL` from non-null input text.

The third is probably my favorite of the non-strict options. Explicit handling of input and output NULLs, panic or error or whatever if you want to.

The fourth is maybe more rust-friendly, if this function is going to be called from non-pgx code. Something like this (and similar for `PgVarlenaInOutFuncs` and `JsonInOutFuncs`):
```rust
pub trait InOutFuncs<E> {
    fn input(input: &Option<crate::cstr_core::CStr>) -> Result<Option<Self>, E>
    where
        Self: Sized;
    fn output(&self, buffer: &mut StringInfo);
}
```
This would need a small shim to unwrap the result, but shouldn't be too bad.

Hoverbear:
This is a good catch. I'm a bit puzzled why we didn't do this before...

PR #643, opened 2022-08-19
Opaque Timestamp/Tz types

mhov:
Similar to `pgx::Date` I've made `pgx::Timestamp` and `pgx::TimestampWithTimeZone` simply represented by their `i64` microsecond datum.  Also `+/-infinity` is supported.  Various `TryFrom` implementations will provide conversions to/from `time::PrimitiveDateTime` and `time::OffsetDateTime`.  Note that conversion will lose any embedded timezones as that is not part of the storage datum, but applied by PG at display time.

I moved the serialization #[test] to #[pg_test] so that serialization can leverage PG's built in serialization. This is especially important for boundary case and +/- infinity.

### DIFF
--- a/pgx/src/datum/time_stamp_with_timezone.rs
+++ b/pgx/src/datum/time_stamp_with_timezone.rs
@@ -19,7 +19,7 @@ const PG_EPOCH_DATETIME: time::PrimitiveDateTime = date!(2000 - 01 - 01).midnigh
 
 // taken from /include/datatype/timestamp.h
 const MIN_TIMESTAMP_USEC: i64 = -211_813_488_000_000_000;
-const END_TIMESTAMP_USEC: i64 = 9_223_371_331_200_000_000;
+const END_TIMESTAMP_USEC: i64 = 9_223_371_331_200_000_000 - 1; // dec by 1 to accommodate exclusive range match pattern
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
 #[repr(transparent)]
@@ -71,10 +71,7 @@ impl TryFrom<pg_sys::TimestampTz> for TimestampWithTimeZone {
         match usec {
             i64::MIN => Ok(Self::NEG_INFINITY),
             i64::MAX => Ok(Self::INFINITY),
-            // https://github.com/rust-lang/rust/issues/37854 for funny exlusive range guard
-            MIN_TIMESTAMP_USEC..=END_TIMESTAMP_USEC if usec < END_TIMESTAMP_USEC => {
-                Ok(TimestampWithTimeZone(usec))
-            }
+            MIN_TIMESTAMP_USEC..=END_TIMESTAMP_USEC => Ok(TimestampWithTimeZone(usec)),
             _ => Err(TimestampConversionError::OutOfRangeMicroseconds),
         }
     }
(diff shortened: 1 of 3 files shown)


pgx/src/datum/time_stamp_with_timezone.rs
```diff
@@ -8,143 +8,169 @@ Use of this source code is governed by the MIT license that can be found in the
...
+impl TryFrom<pg_sys::TimestampTz> for TimestampWithTimeZone {
+    type Error = TimestampConversionError;
+
+    fn try_from(value: pg_sys::TimestampTz) -> Result<Self, Self::Error> {
+        let usec = value as i64;
+        match usec {
+            i64::MIN => Ok(Self::NEG_INFINITY),
+            i64::MAX => Ok(Self::INFINITY),
+            // https://github.com/rust-lang/rust/issues/37854 for funny exlusive range guard
+            MIN_TIMESTAMP_USEC..=END_TIMESTAMP_USEC if usec < END_TIMESTAMP_USEC => {
```
workingjubilee, line 75: This is... not idiomatic. Please decrease the end constant by 1 and omit the guard.

pgx/src/datum/time_stamp_with_timezone.rs
```diff
@@ -8,143 +8,169 @@ Use of this source code is governed by the MIT license that can be found in the
...
+                Ok(TimestampWithTimeZone(usec))
+            }
+            _ => Err(TimestampConversionError::OutOfRangeMicroseconds),
```
workingjubilee, line 78: Then check to see if it's an exhaustive match by testing removing the fallthrough case.

pgx/src/datum/time_stamp_with_timezone.rs
```diff
@@ -8,143 +8,169 @@ Use of this source code is governed by the MIT license that can be found in the
...
+                    .expect("Error converting timestamp with time zone datum"),
+            )
+        }
     }
 }
 
+#[derive(thiserror::Error, Debug, Clone, Copy)]
+pub enum TimestampConversionError {
+    #[error("timestamp value is negative infinity and shouldn't map to time::PrimitiveDateTime")]
+    TimestampNegativeInfinity,
```
workingjubilee, line 165: Rust code tends to prefer terse names as an offset to other tendencies in the language that lead to verbosity. More like C, less like Java. In particular, this repeats "Timestamp", and thus is expressing the "stuttering" problem. I would expect this variant to be named `NegInfinity`, and so on for the other variants.

pgx/src/datum/time_stamp_with_timezone.rs
```diff
@@ -8,143 +8,169 @@ Use of this source code is governed by the MIT license that can be found in the
...
+    #[error("timestamp value is negative infinity and shouldn't map to time::PrimitiveDateTime")]
+    TimestampInfinity,
+    #[error("time::PrimitiveDateTime was unable to convert this timestamp")]
+    TimeCrateConversionError,
```
workingjubilee, line 169: This will almost always be written `TimestampConversionError::TimeCrateConversionError`, so we do not need to see a full spelling of `ConversionError` twice.

workingjubilee (changes requested):
My apologies, for I have some comments about names, and so it is time for us to get out our buckets of paint.

pgx/src/datum/time_stamp_with_timezone.rs
mhov: it did not accept removing the `_ =>` case, it errored as non-exhaustive.

pgx/src/datum/time_stamp_with_timezone.rs
mhov: done

workingjubilee (approved):
This looks good to me. I will take another look in the morning when my head is clearer and I am better at making decisions just to make sure, before merging this. See https://github.com/tcdi/pgx/pull/664 if you want to know what the followup PR is going to be, for reference.

pgx/src/datum/time_stamp_with_timezone.rs
```diff
@@ -8,143 +8,164 @@ Use of this source code is governed by the MIT license that can be found in the
...
+pub enum TimestampConversionError {
+    #[error("timestamp value is negative infinity and shouldn't map to time::PrimitiveDateTime")]
+    NegInfinity,
+    #[error("timestamp value is negative infinity and shouldn't map to time::PrimitiveDateTime")]
+    Infinity,
+    #[error("time::PrimitiveDateTime was unable to convert this timestamp")]
+    TimeCrate,
+    #[error("usec outside of PG's defined Timestamp range")]
+    OutOfRangeMicroseconds,
+}
```
workingjubilee, line 167: With this more proper error type, I feel like this should be generalized for the other temporal types, but that expands the scope of work, so it is clearly a followup PR.

PR #644, opened 2022-08-20
Calling "cargo pgx install" from build script always enables --release

TimeToogo:
The `PROFILE` env var is set by cargo within build scripts.
I believe clap is setting the `release` field to `true` when `PROFILE` is set regardless of whether it is set to `debug` or `release` as I assume it is coercing any non-empty string to `true`.
Since cargo respects the `PROFILE` env var it should just inherit without explicit argument passing.

### DIFF
--- a/cargo-pgx/src/command/install.rs
+++ b/cargo-pgx/src/command/install.rs
@@ -33,7 +33,7 @@ pub(crate) struct Install {
     #[clap(long, parse(from_os_str))]
     manifest_path: Option<PathBuf>,
     /// Compile for release mode (default is debug)
-    #[clap(env = "PROFILE", long, short)]
+    #[clap(long, short)]
     release: bool,
     /// Build in test mode (for `cargo pgx test`)
     #[clap(long)]


workingjubilee:
@Hoverbear Does this seem correct to you? I have low confidence this change is correct or incorrect since either way things seem nonintuitive here and you last touched this.

TimeToogo:
Would be good to get this merged in for the next release if at all possible.
Simple reproduction is to run the following command on any pgx extension:
```bash
PROFILE=debug cargo pgx install 
```
And observe it will be compiled with the `--release` flag.

Hoverbear:
I'm a bit confused here, you're running `cargo pgx install` inside a `build.rs`?

Hoverbear:
@TimeToogo `cargo pgx install` is intended to use release by default, you'd need to do `cargo pgx install --debug` for debug.

TimeToogo:
> I'm a bit confused here, you're running `cargo pgx install` inside a `build.rs`? You can do `Command.new("cargo-pgx").args(["pgx", "install"]).env_remove("PROFILE")` if you really want.

@Hoverbear Yes, I have a workspace in which some crates (integration tests mainly) are dependent on having the extension installed on the local posters.

TimeToogo:
> @TimeToogo `cargo pgx install` is intended to use release by default, you'd need to do `cargo pgx install --debug` for debug.

@Hoverbear does this work? From memory I tried this but there is no debug flag defined in the struct used for capturing args.

TimeToogo:
@Hoverbear am I just wrong in thinking that the `env = "PROFILE"` on a `release: bool` makes no sense? Would welcome an explanation of how that is intended to work 😀

Hoverbear:
I think it's safe to remove this and I am not entirely sure it is intended. I think we could merge this. :)

Hoverbear (approved):

workingjubilee:
Now I'm more confused. :sweat_smile:

Hoverbear:
`cargo pgx install` defaults to release, but should be able to be toggled down to debug with `cargo pgx install --debug`. The user is reporting because their own crate's `build.rs` calls `cargo pgx install` the `PROFILE` environment which clap automagically sets from is forcing it into release mode even if they use `--debug`.

I am unclear if `cargo pgx install` from a `build.rs` is a good idea, but I do think this is something we can merge to correct what may be a confusing behavior.

workingjubilee:
Mmmh. I think I understand that, I'm just concerned this may cause users running the command in the "normal" way to stop having installations enter `--release` by default because of some unforeseen interaction.

eeeebbbbrrrr:
`cargo pgx install` is intended to use `--debug` by **default**.  Only `cargo pgx package` defaults to `--release`.

The typical usage of `cargo pgx install` is for general development where you want to run your extension against a user-installed Postgres instance, rather than the one pgx manages.  In general, it's the same as `cargo pgx run`, but installs the extension locally and doesn't immediately run psql.  

Having `cargo pgx install` default to `--release` will significantly disrupt that workflow.

@thomcc has added proper support for cargo profiles with a new `--profile` argument, in #758.

I'm not sure of the effects of the changeset here (and I'm currently too lazy to go read clap's docs).  @thomcc, do you think this PR is necessary now that #758 exists?

thomcc:
I did notice the PROFILE argument and thought it was a bit odd but didn't want to disrupt too much in my patch, I think its reasonable to remove if we don't have a need for it.

thomcc:
This is actually somewhat incomplete tho, there are a few other places with `env = "PROFILE"`. I'll add this (and the other places) to my PR in #758.

thomcc:
Yep (and `schema` also defaults to debug).

thomcc:
Closing in favor of #758.

eeeebbbbrrrr:
> Yep (and `schema` also defaults to debug).

Oh sure, `schema`.  

On its own (as in `cargo pgx schema`), yeah it should default to debug.  But when that code is run as part of `test, run, install, package`, it needs to use whatever they use to avoid recompilations.

PR #645, opened 2022-08-22
Moves unnecessary deps out of critical build path

BradyBonnette:
These are changes that coincide with https://github.com/tcdi/pgx/issues/597

Mostly these changes are spent trying to pull out as much stuff out of `pgx-utils` where applicable, but there could be more to it as well.


BradyBonnette:
Also note that this branch is experimental for now, and the major reason for creating a pull request at this moment is to get CI to build it and see what results from the build. Depending on how that goes and the feedback given, this could very well end up in `develop`.

workingjubilee:
So I noticed you inlined single-consumer things into their consumers. `pgx-pg-sys-stub` is such a single-consumer entity. Would it be better to lock that down?

pgx-paths/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+        serde_json::from_slice(&output.stdout).wrap_err("Invalid `cargo metada` response")?;
+    let target_dir = json.get("target_directory");
+    match target_dir {
+        Some(JsonValue::String(target_dir)) => Ok(target_dir.into()),
+        v => Err(eyre!(
+            "could not read target dir from `cargo metadata` got: {:?}",
+            v,
+        )),
+    }
+}
```
workingjubilee, line 40: This function has a single consumer, in `cargo pgx`.

pgx-paths/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+pub fn prefix_path<P: Into<PathBuf>>(dir: P) -> String {
+    let mut path = std::env::split_paths(&std::env::var_os("PATH").expect("failed to get $PATH"))
+        .collect::<Vec<_>>();
+
+    path.insert(0, dir.into());
+    std::env::join_paths(path)
+        .expect("failed to join paths")
+        .into_string()
+        .expect("failed to construct path")
+}
```
workingjubilee, line 51: All consumers of this function also need pgx_pg_config, so maybe this should be part of that crate?

pgx-paths/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+
+pub fn versioned_so_name(extension_name: &str, extension_version: &str) -> String {
+    format!("{}-{}", extension_name, extension_version)
+}
```
workingjubilee, line 55: As a pure String manipulation, it's actually fine to leave this behind.

pgx-utils/src/lib.rs
```diff
@@ -7,29 +7,15 @@ All rights reserved.
...
+use crate::sql_entity_graph::PositioningRef;
 use proc_macro2::{TokenStream, TokenTree};
 use quote::{format_ident, quote, ToTokens, TokenStreamExt};
-use serde_json::value::Value as JsonValue;
-use std::{
-    collections::HashSet,
-    path::PathBuf,
-    process::{Command, Stdio},
-    str::FromStr,
-};
```
workingjubilee, line 21: I know it doesn't look like it but this is probably one of the most important improvements here.

workingjubilee (commented):
Thank you for taking a look at this!

I have some thoughts about a few more moves in this game of shufflepuck, but stripping pgx-utils down to almost the bare essentials is very nice to see.

pgx-paths/src/lib.rs
BradyBonnette: I am also seeing it in `pgx-tests`. Maybe I am overlooking something:

image

pgx-paths/src/lib.rs
BradyBonnette: Ah yes, definitely. Doing that now.

pgx-paths/src/lib.rs
workingjubilee: Oh.
...Is my find not working?

pgx-paths/src/lib.rs
BradyBonnette: I am not sure I am clear on "leave this behind"?

Now that you mention it, I wonder if the one single line in this method should just be "inlined" into the only two spots that need it, and getting it out of this `pgx-paths` crate?

Ideally I'd like to get rid of this `pgx-paths` crate because it's dangerously close to a "common" or a "utils" crate, which is most of what this pull request was trying to eliminate :)

pgx-paths/src/lib.rs
BradyBonnette: This is what I am finding with some `rg`:

```
$ rg "get_target_dir"
cargo-pgx/src/command/install.rs
17:use pgx_paths::{get_target_dir, versioned_so_name};
463:        .strip_prefix(get_target_dir()?.parent().unwrap())

cargo-pgx/src/command/package.rs
16:use pgx_paths::get_target_dir;
123:    let mut target_dir = get_target_dir()?;

cargo-pgx/src/command/schema.rs
21:use pgx_paths::get_target_dir;
180:    let mut target_dir_with_profile = get_target_dir()?;

cargo-pgx/src/command/test.rs
12:use pgx_paths::get_target_dir;
125:    let target_dir = get_target_dir()?;

pgx-paths/src/lib.rs
14:pub fn get_target_dir() -> eyre::Result<PathBuf> {

pgx-tests/src/framework.rs
16:use pgx_paths::get_target_dir;
261:        .env("CARGO_TARGET_DIR", get_target_dir()?);
528:    let mut target_dir = get_target_dir()?;

File: pgx-utils/src/lib.rs (hunk 4 above)
Comment (BradyBonnette) [reply pgx-utils/src/lib.rs:21]: Yeah, based on what you said earlier this is huge.

File: pgx-paths/src/lib.rs (hunk 3 above)
Comment (workingjubilee) [reply pgx-paths/src/lib.rs:55]: Yes, I agree, inlining seems best.

File: pgx-paths/src/lib.rs (hunk 1 above)
Comment (workingjubilee) [reply pgx-paths/src/lib.rs:40]: Yeah. I think maybe the same thought about "this could also live in pgx-pg-config" still applies.

File: pgx-pg-config/src/lib.rs [hunk 5]
```diff
@@ -24,6 +24,13 @@ use url::Url;
 pub static BASE_POSTGRES_PORT_NO: u16 = 28800;
 pub static BASE_POSTGRES_TESTING_PORT_NO: u16 = 32200;
 
+// These methods were originally in `pgx-utils`, but in an effort to consolidate
+// dependencies, the decision was made to package them into wherever made the
+// most sense. In this case, it made the most sense to put them into this
+// pgx-pg-config crate. That doesnt mean they can't be moved at a later date.
+mod path_methods;
+pub use path_methods::{get_target_dir, prefix_path};
+
```
BradyBonnette, line 33: @workingjubilee 

This combined with the new `path_methods.rs` file is what we discussed in Discord. Does this seem right to you?

workingjubilee: Yes!

cargo-pgx/src/command/install.rs
```diff
@@ -156,7 +155,10 @@ pub(crate) fn install_extension(
         dest.push(&pkgdir);
         let so_name = if versioned_so {
             let extver = get_version(&package_manifest_path)?;
-            versioned_so_name(&extname, &extver)
+            // Note: This originally was a call into versioned_so_name, but in an effort to eliminate crate dependencies,
+            // and because it was a single format! line, it was decided that inlining the single format! line was best in
+            // order to reduce a dependency.
+            format!("{}-{}", &extname, &extver)
```
workingjubilee, line 161: Perhaps simply

```suggestion
            // note: versioned so-name format must agree with pgx-utils
            format!("{}-{}", &extname, &extver)
```

pgx-utils/src/sql_entity_graph/pgx_sql.rs
```diff
@@ -548,7 +547,10 @@ impl PgxSql {
         return if self.versioned_so {
             let extname = &self.extension_name;
             let extver = &self.control.default_version;
-            format!("$libdir/{}", versioned_so_name(extname, extver))
+            // Note: This originally was a call into versioned_so_name, but in an effort to eliminate crate dependencies,
+            // and because it was a single format! line, it was decided that inlining the single format! line was best in
+            // order to reduce a dependency.
+            format!("$libdir/{}-{}", extname, extver)
```
workingjubilee, line 553: Perhaps simply

```suggestion
            // Note: versioned so-name format must agree with cargo pgx
            format!("$libdir/{}-{}", extname, extver)
```

workingjubilee (commented):
The pgx-pg-sys-stub crate could probably still be moved in as a submodule in cargo-pgx.

pgx-utils/src/sql_entity_graph/pgx_sql.rs
BradyBonnette: 👍

BradyBonnette:
@workingjubilee I added a couple new commits based on earlier discussions. Let me know what you think!

workingjubilee (approved):
Excellent!

BradyBonnette:
@workingjubilee One last closing thought before this gets merged in.

We had talked about possibly renaming `pgx-utils` into something more meaningful, such as your suggested `pgx-sqlgen` since the only thing that's left in `pgx-utils` now is SQL generation stuff.

Do you feel that we should still do that? I bring it up since renaming it would (more than likely) break some things for any other project that relies on PGX, but since the next release is going to be breaking anyways it might be the best opportunity to do it. If it's something we should consider, then the best course of action might be to create a separate issue just for the rename, and maybe do it right before release so we're not stomping on all of the other incoming changes from the other contributors. The things I refactored out of `pgx-utils` in this PR were fairly benign.

workingjubilee:
Functionally, `pgx-utils` is really not meant to be used by external consumers to `pgx`, despite the fact we must publish it in order to make `pgx` work. However, I agree that we should leave things as-is at the current moment and leave a moment to paint a bikeshed.

BradyBonnette:
Ok, that's fair. I went ahead and made a new issue so that maybe we won't forget and perhaps visit in the future: https://github.com/tcdi/pgx/issues/655

PR #646, opened 2022-08-23
Input functions should take Option

willmurnane:
This PR contrasts with [PR 641](https://github.com/tcdi/pgx/pull/641). The primary motivation here is that the various input functions take optional arguments, and produce optional results. This is implemented by changing the interface in `pgx/src/inoutfuncs.rs`, then the macro in `pgx-macros/src/lib.rs`, then teaching the SQL generator that this is acceptable in `pgx/src/datum/mod.rs`. Finally, I changed all the implementations of those functions that I could find, so that `cargo test --workspace` passes with these changes.

### DIFF
--- a/pgx-macros/src/lib.rs
+++ b/pgx-macros/src/lib.rs
@@ -773,8 +773,13 @@ fn impl_postgres_type(ast: DeriveInput) -> proc_macro2::TokenStream {
 
             #[doc(hidden)]
             #[pg_extern(immutable,parallel_safe)]
-            pub fn #funcname_in #generics(input: &#lifetime pgx::cstr_core::CStr) -> #name #generics {
-                #name::input(input)
+            pub fn #funcname_in #generics(input: Option<&#lifetime pgx::cstr_core::CStr>) -> Option<#name #generics> {
+                input.map_or_else(|| {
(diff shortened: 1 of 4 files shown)


workingjubilee:
I think this needs at least one test to verify that the NULL -> NULL and NULL -> error message cases both work correctly. I am hoping the former is already effectively tested for (haven't checked yet) so it would be the latter, mostly. The latter just to make sure that the Rust &str we return from these functions is correctly used by Postgres during erroring.

willmurnane:
I added a pair of tests, and fixed up the interface in a minor way (using a constant rather than a function, so that implementors can override it).

workingjubilee:
Well, an implementation can override a default impl of a function too, no? I think it's mostly about ergonomics and if e.g. we would want to offer dynamic messages (I can't really think of why or how we would, however, since it's just a simple in/out).

workingjubilee (commented):
This fixes some erroneous behavior and is fully compatible with #615. If any final retouch edits disrupt it I will resolve them myself, though I am still going to land #615 first.

workingjubilee (approved):
Looks like this passes tests!

PR #650, opened 2022-08-25
Skip initdb during cargo pgx init if `$USER == "root"`

anth0nyleung:
## Notes 
In some use case, we would set up rust toolchain and pgx as root user, we do not necessarily need pgx to spin up running postgres instance. However during `cargo pgx init`, as pgx is trying to `initdb` for postgres, it will fail to do so since `initdb` cannot be executed as root user.

## Changes
If current user is root user, skip doing `initdb`.

## Testing

`cargo test --features pg14 --no-default-features` passed

```
> ls -l ~/.pgx   
total 0

> cargo pgx init --pg14 /usr/local/pgsql/bin/pg_config 
   Validating /usr/local/pgsql/bin/pg_config
 Initializing data directory at /Users/antholeu/.pgx/data-14

> ls -l ~/.pgx                                        
total 8
drwx------  24 antholeu  staff   768B Aug 25 16:36 data-14
-rw-r--r--   1 antholeu  staff    48B Aug 25 16:36 config.toml
```

```
> ls -l ~/.pgx                                             
total 0

> sudo cargo pgx init --pg14 /usr/local/pgsql/bin/pg_config
   Validating /usr/local/pgsql/bin/pg_config
   Skipping initdb as current user is root user

> ls -l ~/.pgx                                             
total 8
-rw-r--r--  1 root  staff    48B Aug 25 16:37 config.toml
```

### DIFF
--- a/cargo-pgx/src/command/init.rs
+++ b/cargo-pgx/src/command/init.rs
@@ -60,6 +60,9 @@ pub(crate) struct Init {
     /// If installed locally, the path to PG14's `pgconfig` tool, or `download` to have pgx download/compile/install it
     #[clap(env = "PG14_PG_CONFIG", long)]
     pg14: Option<String>,
+    /// Skip performing initdb for postgres
+    #[clap(long)]
+    no_init_db: bool,
     #[clap(from_global, parse(from_occurrences))]
     verbose: usize,
 }
@@ -87,7 +90,7 @@ impl CommandExecute for Init {
 
         if versions.is_empty() {
             // no arguments specified, so we'll just install our defaults
-            init_pgx(&pgx_default(SUPPORTED_MAJOR_VERSIONS)?)
+            init_pgx(&pgx_default(SUPPORTED_MAJOR_VERSIONS)?, self.no_init_db)
         } else {
             // user specified arguments, so we'll only install those versions of Postgres
             let mut default_pgx = None;
@@ -110,13 +113,13 @@ impl CommandExecute for Init {
                 pgx.push(config);
             }
 
-            init_pgx(&pgx)
+            init_pgx(&pgx, self.no_init_db)
         }
     }
 }
 
 #[tracing::instrument(skip_all, fields(pgx_home = %Pgx::home()?.display()))]
-pub(crate) fn init_pgx(pgx: &Pgx) -> eyre::Result<()> {
+pub(crate) fn init_pgx(pgx: &Pgx, no_init_db: bool) -> eyre::Result<()> {
     let dir = Pgx::home()?;
 
     let output_configs = Arc::new(Mutex::new(Vec::new()));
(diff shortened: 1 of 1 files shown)


cargo-pgx/src/command/init.rs
```diff
@@ -166,10 +169,18 @@ pub(crate) fn init_pgx(pgx: &Pgx) -> eyre::Result<()> {
...
+                "{} initdb as --no-init-db is specified {}",
+                "   Skipping ".bold().green(),
+                no_init_db.bold().green(),
+            );
+        } else {
+            let datadir = pg_config.data_dir()?;
+            let bindir = pg_config.bin_dir()?;
+            if !datadir.exists() {
+                initdb(&bindir, &datadir)?;
+            }
```
workingjubilee, line 183: Currently we avoid running `initdb` if `pg_config.data_dir()` exists.

But presumably, in the case that the user has specified a given path, the database should in fact _always_ not be initialized by us, as the user wants a specific database to be used and likely has it configured the way they like it, or will want to do so themselves. It may be that we will not be able to execute successive commands using `cargo pgx` if `initdb` has not been run, but we will run into that issue if this flag is passed as well, so we must handle that case anyways. After all, `cargo pgx` cannot simply assume that the system state has remain unchanged between runtime sessions. This is especially true if `cargo pgx` is pointed at a specific PostgreSQL database and so is not "in control" of it.

Is this hypothesis incorrect?

workingjubilee (commented):
This looks good per se! I just have some design-level questions so I can better understand why to add this, and not other solutions.

cargo-pgx/src/command/init.rs
workingjubilee: This hypothesis **is** incorrect! PGX expects a specific path for the `data_dir` to exist, it's not actually controlled by `pg_config`, see: https://github.com/tcdi/pgx/blob/master/pgx-utils/src/pg_config.rs#L199-L202

cargo-pgx/src/command/init.rs
```diff
@@ -60,6 +60,9 @@ pub(crate) struct Init {
     /// If installed locally, the path to PG14's `pgconfig` tool, or `download` to have pgx download/compile/install it
     #[clap(env = "PG14_PG_CONFIG", long)]
     pg14: Option<String>,
+    /// Skip performing initdb for postgres
+    #[clap(long)]
+    no_init_db: bool,
```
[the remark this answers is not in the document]
anth0nyleung: yes. i think the first approach is better. Avoids adding new flag and simplifies the changes

cargo-pgx/src/command/init.rs
anth0nyleung: `datadir` is currently hardcoded to `data-<major-version>`, user is not allow to specify this.

workingjubilee:
Thank you!

PR #656, opened 2022-08-26
Refactor to use NonNull in datum::Array

workingjubilee:
This removes a small but, IMO, pointless `panic!` in our exposed call surface. It is not unsafe to return the null pointer. That's why `ptr::null()` is safe to call. `panic!` should be used to avoid unsafe or invalid conditions, not just undesirable ones.

This also begins the deprecation process for `Array::over`.

### DIFF
--- a/pgx/src/datum/array.rs
+++ b/pgx/src/datum/array.rs
@@ -16,7 +16,7 @@ use std::{mem, ptr, slice};
 pub type VariadicArray<'a, T> = Array<'a, T>;
 
 pub struct Array<'a, T: FromDatum> {
-    _ptr: *mut pg_sys::varlena,
+    _ptr: Option<NonNull<pg_sys::varlena>>,
     array_type: Option<NonNull<pg_sys::ArrayType>>,
     nelems: usize,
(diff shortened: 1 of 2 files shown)


Hoverbear (approved):

pgx/src/datum/array.rs
```diff
@@ -75,8 +91,8 @@ impl<'a, T: FromDatum> Array<'a, T> {
     /// - `nulls` is non-null
     /// - both `elements` and `nulls` point to a slice of equal-or-greater length than `nelems`
     unsafe fn from_pg(
-        _ptr: *mut pg_sys::varlena,
-        array_type: *mut pg_sys::ArrayType,
+        _ptr: Option<NonNull<pg_sys::varlena>>,
+        array_type: Option<NonNull<pg_sys::ArrayType>>,
         elements: *mut pg_sys::Datum,
         nulls: *mut bool,
```
workingjubilee, line 97: These are being kept as-is for the moment because the actual fix for them is using the new `RawArray` abstraction plus `bitvec` to handle things, removing usage of `deconstruct_array`, and that's a more complex followup.

PR #658, opened 2022-08-27
[docs] pgx init subcommand requires lowercase download

flynneva:
- Instead of `downLoad`
- only updates the README.md for the `pgx init` sub command


workingjubilee (approved):
Thank you!

Issue #659, opened 2022-08-28
Double up null-checks in `FromDatum`


workingjubilee:
It is not actually safe to trust the input to `FromDatum` to be correctly marked `is_null` if it is a pointer type, per the findings in
- https://github.com/tcdi/pgx/issues/648

So I am going to make a big tracking item again for doubling these null checks tomorrow.

Issue #648, opened 2022-08-23
Converting `NULL` to rust-defined type causes server segfault


willmurnane:
First, the test case:
```
git checkout develop
cargo install --path cargo-pgx/
cd pgx-examples/custom_types
cargo pgx run
custom_types=# select null::animals;
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
!?> 
```
After attaching a debugger, I observed that [this function](https://github.com/tcdi/pgx/blob/develop/pgx/src/datum/from.rs#L289-L301) is being called with its `Datum` argument null, but `is_null` set to `false`. This causes a null pointer dereference when the CStr object is initialized, and an attempt is made to calculate the length. See [discord discussion](https://discord.com/channels/561648697805504526/749454482450153594/1011689402814177310) for more context.

I think this may be due to a misunderstanding somewhere of what the semantics of `NullableDatum.isnull` are. There's [at least one place in the PostgreSQL source code](https://github.com/postgres/postgres/blob/master/src/backend/executor/execExpr.c#L3640-L3642) that sets value `NULL` but isnull `false`. Thus, it may be a logic error specific to the `impl FromDatum` for `CStr` types.

One possible fix for this is to do the null check ourselves, in [get_nullable_datum](https://github.com/tcdi/pgx/blob/master/pgx/src/fcinfo.rs#L169-L177):
```diff
--- a/pgx/src/fcinfo.rs
+++ b/pgx/src/fcinfo.rs
@@ -158,7 +158,9 @@ mod pg_12_13_14 {
         unsafe {
             let nargs = fcinfo.nargs;
             let len = std::mem::size_of::<pg_sys::NullableDatum>() * nargs as usize;
-            fcinfo.args.as_slice(len)[num].clone()
+            let mut result = fcinfo.args.as_slice(len)[num].clone();
+            result.isnull |= result.value.is_null();
+            result
         }
     }
```
Another is to check in the associated `FromDatum` implementation for pointer-y types, like this:
```diff
@@ -292,7 +292,7 @@ impl<'a> FromDatum for &'a crate::cstr_core::CStr {
        datum: pg_sys::Datum,
        is_null: bool,
    ) -> Option<&'a crate::cstr_core::CStr> {
-        if is_null {
+        if is_null || datum.is_null() {
            None
        } else {
            Some(crate::cstr_core::CStr::from_ptr(datum.ptr_cast()))
```
[...]

PR #661, opened 2022-08-29
Check for null Datum objects, even if isnull is false.

willmurnane:
In at least some cases, PostgreSQL can provide a `NullableDatum` instance with a null pointer and its `isnull` value set to `false`. This commit checks for that misbehavior whenever we're preparing to dereference a pointer, so that downstream code just gets `None` instead of a segfault. See more discussion at #648.


willmurnane:
Closes #659, presumably.

workingjubilee (approved):
Oh, that's fewer callsites than I was expecting, for some reason. Thank you!

PR #663, opened 2022-08-30
Range<T> Type mapping

mhov:
A second attempt at mapping the built-in PG range types to PGX. 

As per recent type mapping improvements, the mapping type will be a shallow/opaque wrapper to the datum `Range<T>`, with the option perform a non-trivial deserialization to a more accessible form `RangeData<T>` only when needed. 

Users have the option to work with the direct `T` lower/upper values or the the lower/upper `pg_sys::RangeBound` structs which carry additional information about bound inclusivity/infinity.

Attempts to map back and forth with `std::ops::Range<T>` were made but proved to be leaky abstractions:
- no mapping for `empty` pg ranges
- having to dyn map to the various `trait std::ops::RangeBounds<T>` implementations depending on lower_inc/upper_inc, lower_inf/upper_inf states
- Some `std::ops::Range<T>` states that would be parsable but illegal as a PG range ex `(1..0)`

If a user wants to create their own `std::ops::Range<T>` representations, nothing is stopping them.

Note this does not cover PG14 multirange types, which might build upon @workingjubilee 's Array<T> improvements.


mhov:
I had to add `pg_sys::{subtype}RANGEOID` shims for PG10 as bindgen can't get them in the form they're in `pg_type.h` looks like this 
```
/* range types */
DATA(insert OID = 3904 ( int4range		PGNSP PGUID  -1 f r R f t \054 0 0 3905 range_in range_out range_recv range_send - - range_typanalyze i x f 0 -1 0 0 _null_ _null_ _null_ ));
DESCR("range of integers");
#define INT4RANGEOID		3904
DATA(insert OID = 3905 ( _int4range		PGNSP PGUID  -1 f b A f t \054 0 3904 0 array_in array_out array_recv array_send - - array_typanalyze i x f 0 -1 0 0 _null_ _null_ _null_ ));
DATA(insert OID = 3906 ( numrange		PGNSP PGUID  -1 f r R f t \054 0 0 3907 range_in range_out range_recv range_send - - range_typanalyze i x f 0 -1 0 0 _null_ _null_ _null_ ));
DESCR("range of numerics");
DATA(insert OID = 3907 ( _numrange		PGNSP PGUID  -1 f b A f t \054 0 3906 0 array_in array_out array_recv array_send - - array_typanalyze i x f 0 -1 0 0 _null_ _null_ _null_ ));
DATA(insert OID = 3908 ( tsrange		PGNSP PGUID  -1 f r R f t \054 0 0 3909 range_in range_out range_recv range_send - - range_typanalyze d x f 0 -1 0 0 _null_ _null_ _null_ ));
DESCR("range of timestamps without time zone");
DATA(insert OID = 3909 ( _tsrange		PGNSP PGUID  -1 f b A f t \054 0 3908 0 array_in array_out array_recv array_send - - array_typanalyze d x f 0 -1 0 0 _null_ _null_ _null_ ));
DATA(insert OID = 3910 ( tstzrange		PGNSP PGUID  -1 f r R f t \054 0 0 3911 range_in range_out range_recv range_send - - range_typanalyze d x f 0 -1 0 0 _null_ _null_ _null_ ));
DESCR("range of timestamps with time zone");
DATA(insert OID = 3911 ( _tstzrange		PGNSP PGUID  -1 f b A f t \054 0 3910 0 array_in array_out array_recv array_send - - array_typanalyze d x f 0 -1 0 0 _null_ _null_ _null_ ));
DATA(insert OID = 3912 ( daterange		PGNSP PGUID  -1 f r R f t \054 0 0 3913 range_in range_out range_recv range_send - - range_typanalyze i x f 0 -1 0 0 _null_ _null_ _null_ ));
DESCR("range of dates");
DATA(insert OID = 3913 ( _daterange		PGNSP PGUID  -1 f b A f t \054 0 3912 0 array_in array_out array_recv array_send - - array_typanalyze i x f 0 -1 0 0 _null_ _null_ _null_ ));
DATA(insert OID = 3926 ( int8range		PGNSP PGUID  -1 f r R f t \054 0 0 3927 range_in range_out range_recv range_send - - range_typanalyze d x f 0 -1 0 0 _null_ _null_ _null_ ));
DESCR("range of bigints");
DATA(insert OID = 3927 ( _int8range		PGNSP PGUID  -1 f b A f t \054 0 3926 0 array_in array_out array_recv array_send - - array_typanalyze d x f 0 -1 0 0 _null_ _null_ _null_ ));
```
only INT4RANGEOID is parseable 

is there a better way than how i did it in the last commit? when do you plan to drop PG10 support?

workingjubilee:
November 11, 2022, probably.

You should probably instead just use `#[cfg(not(feature = "pg10")]`.

mhov:
Ok cool at which scope should I have this? on the range.rs and any range related tests?

workingjubilee:
On top of the "entire thing", yes. You can put it on top of the `mod range;` to make the module disappear.

pgx/src/datum/mod.rs
```diff
@@ -44,6 +45,7 @@ pub use item_pointer_data::*;
 pub use json::*;
 pub use numeric::*;
 use once_cell::sync::Lazy;
+pub use range::*;
```
workingjubilee, line 48: I have begun to want to replace these kinds of reexports with a "prelude" module. You may omit this.

pgx/src/datum/range.rs
```diff
@@ -0,0 +1,332 @@
...
+                range_data.is_empty,
+            );
+
+            Range::<T>::from_pg(range_type as *mut pg_sys::varlena, range_type)
+        }
+    }
+}
+
+/// This trait allows a struct to be a valid subtype for a RangeType
+pub trait RangeSubType {
```
workingjubilee, line 282: Using "subtype" is potentially confounding. In Rust, this trait can be used in a generic bound. Rust does not have "subtypes" as usually considered, except in the form of the lifetime system (lifetime subtyping allowing different lifetimes to satisfy the same constraint, as long as each "`'a` outlives `'b`" and so on constraint is met).

If you mean that it satisfies a Postgres subtype relationship, then that is another thing, in which case this trait should probably be sealed.

If you mean that they all implement the same behavior, then they should just implement the same behavior via the trait.

mhov:
excluded `Range<T>` support from `pg10` with `#[cfg(not(feature = "pg10")]`

workingjubilee (dismissed):
This PR will need to be updated to use the new type mapping strategy after tcdi/pgx#615 is merged, so 0.5.1 is the more practical time for this to be released.

workingjubilee:
#615 was landed.

mhov:
Will update asap. Thanks

mhov:
Updated for #615

pgx/src/datum/range.rs
```diff
@@ -0,0 +1,391 @@
...
+    /// - datum represents a RangeType datum
+    fn try_from(datum: pg_sys::Datum) -> Result<Self, Self::Error> {
+        if datum.is_null() {
+            Err(RangeConversionError::NullDatum)
+        } else {
+            unsafe {
+                let ptr = datum.ptr_cast();
+                let range_type =
+                    pg_sys::pg_detoast_datum(datum.ptr_cast()) as *mut pg_sys::RangeType;
+                let _ = range_type.as_ref().expect("RangeType * was NULL");
```
thomcc, line 63: Is there a reason not to return an Err here (and expand the RangeConversionError to include this)?

pgx/src/datum/range.rs
```diff
@@ -0,0 +1,332 @@
...
+//! Utility functions for working with `pg_sys::RangeType` structs
+use crate::{
+    pg_sys, void_mut_ptr, Date, FromDatum, IntoDatum, Numeric, Timestamp, TimestampWithTimeZone,
+};
+use pgx_pg_sys::{Oid, RangeBound};
+use std::marker::PhantomData;
+
+/// Represents Datum to serialized RangeType PG struct
+pub struct Range<T: FromDatum + IntoDatum + RangeSubType> {
+    ptr: *mut pg_sys::varlena,
```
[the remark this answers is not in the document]
workingjubilee: We only retain the varlena ptr in a case where we think we might need to interact with it somehow, and in that case the only time we do, is when we think we might need to pfree some detoasted data. It should not be preserved if we should not.

pgx/src/datum/range.rs
workingjubilee: If it is possible for the trait implementation to cause unsound results (e.g. if a wrong answer from `range_type_oid()` could, when used with other functions, cause unsoundness, or if implementing this for the _wrong_ type can cause unsoundness) then the bound cannot be fully "trusted" unless this is marked as a "trusted" trait (i.e. is an `unsafe Trait`). Note this does not mean the trait's functions must be `unsafe`: any `fn` of an `unsafe Trait` may be itself safe or `unsafe` to call, the `unsafe` marker on the trait instead says the entire trait has a compiler-unverified invariant and thus must be implemented correctly by the programmer.

pgx/src/datum/range.rs
workingjubilee: ...Could the returned pointer be unaligned?

pgx/src/datum/range.rs
```diff
@@ -0,0 +1,371 @@
...
+impl<T> Range<T>
+where
+    T: FromDatum + IntoDatum + RangeSubType,
+{
+    /// ## Safety
+    /// function requires that
+    /// - datum is null OR
+    /// - datum represents a RangeType datum
+    #[inline]
+    unsafe fn from_pg(ptr: *mut pg_sys::varlena, range_type: *mut pg_sys::RangeType) -> Self {
```
workingjubilee, line 36: This doesn't accept a Datum? Do you mean the pointers? Could you handle the invariants earlier and make this function safe?

pgx/src/datum/range.rs
```diff
@@ -0,0 +1,371 @@
...
+where
+    T: FromDatum + IntoDatum + RangeSubType,
+{
+    type Error = RangeConversionError;
+
+    /// ## Safety
+    /// function requires that
+    /// - datum is null OR
+    /// - datum represents a RangeType datum
+    fn try_from(datum: pg_sys::Datum) -> Result<Self, Self::Error> {
```
workingjubilee, line 51: This function cannot be safe if it has safety requirements like this, which points at not using TryFrom in this case.

pgx/src/datum/range.rs
```diff
@@ -0,0 +1,371 @@
...
+impl<T> Drop for Range<T>
+where
+    T: FromDatum + IntoDatum + RangeSubType,
+{
+    fn drop(&mut self) {
+        if !self.range_type.is_null() && self.range_type as *mut pg_sys::varlena != self.ptr {
+            unsafe {
+                pg_sys::pfree(self.range_type as void_mut_ptr);
+            }
+        }
```
workingjubilee, line 111: It would be a good idea to document why we need this unsafe block and why those preconditions have to be upheld.

pgx/src/datum/range.rs
```diff
@@ -0,0 +1,371 @@
...
+
+            let mut lower_bound = range_data.lower;
+            let mut upper_bound = range_data.upper;
+
+            let range_type = pg_sys::make_range(
+                typecache,
+                &mut lower_bound,
+                &mut upper_bound,
+                range_data.is_empty,
+            );
```
workingjubilee, line 257: This does a bunch of invocations into Postgres, it'd be nice to at least get a general idea of what the postconditions and preconditions are.

pgx/src/datum/range.rs
```diff
@@ -0,0 +1,371 @@
...
+
+            pg_sys::range_deserialize(
+                typecache,
+                range.range_type,
+                &mut lower_bound,
+                &mut upper_bound,
+                &mut is_empty,
+            );
+        }
```
workingjubilee, line 234: unsafe block, preconditions, postconditions, etc., you can keep it brief but it'd be nice.

pgx/src/datum/mod.rs
```diff
@@ -21,6 +21,8 @@ mod into;
 mod item_pointer_data;
 mod json;
 mod numeric;
+#[cfg(not(feature = "pg10"))]
+mod range;
```
workingjubilee, line 25:
```suggestion
#[deny(unsafe_op_in_unsafe_fn)]
mod range;
```

workingjubilee (changes requested):
I think we can narrow down the unsafe code required if we identify the invariants more clearly, and then contain it into a smaller box.

pgx/src/datum/range.rs
workingjubilee: Alternately, using actual error returns.

pgx/src/datum/range.rs
mhov: I've made RangeSubType an unsafe trait

pgx/src/datum/range.rs
mhov: I removed this vestigial check.

pgx/src/datum/range.rs
mhov: blindly copied (at the time)[ 0.4.5 array.rs](https://github.com/tcdi/pgx/blob/1a89778046547c52c714fe6885069eab5638f965/pgx/src/datum/array.rs#L74) as an example of varlena handling.  Will rework with a datum argument

pgx/src/datum/range.rs
mhov: removed the safety comments, I was guided towards `try_from` for general type conversion a while back, is that no longer the case for most PG types?

pgx/src/datum/range.rs
mhov: Another example of me blindly copying [0.4.5 array.rs logic](https://github.com/tcdi/pgx/blob/1a89778046547c52c714fe6885069eab5638f965/pgx/src/datum/array.rs#L266
) when working with a varlena/detoasted type. 
I'm still too green at Rust to tell you what's right or wrong here.  added safety messaging.

pgx/src/datum/mod.rs
mhov: should i strip all the `pg10` conditionals?

workingjubilee: That would be safe, yes.

pgx/src/datum/range.rs
workingjubilee: It's the case for "by value" types. It's a much more complicated story for those which are potentially null pointers.

pgx/src/datum/range.rs
workingjubilee: Using the pointer arguments is fine, really, I just wanted to make sure everything was correct.

mhov: i had moved it around to accept a datum and getting the ptr/range_type from that.

pgx/src/datum/range.rs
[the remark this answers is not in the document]
mhov: *RangeType is varlena/detoastable, so we should keep it right?

pgx/src/datum/range.rs
```diff
@@ -78,8 +78,9 @@ where
             None
         } else {
             let ptr = datum.cast_mut_ptr();
+            // Datum should be non-null and point to PG RangeType
             let range_type =
-                pg_sys::pg_detoast_datum(datum.cast_mut_ptr()) as *mut pg_sys::RangeType;
+                unsafe { pg_sys::pg_detoast_datum(datum.cast_mut_ptr()) as *mut pg_sys::RangeType };
```
workingjubilee, line 83: AHA! I knew there was going to be one of these hiding. Thank you for taking care of this.

pgx/src/datum/range.rs
```diff
@@ -67,6 +61,10 @@ impl<T> FromDatum for Range<T>
 where
     T: FromDatum + IntoDatum + RangeSubType,
 {
+    /// ## Safety
+    /// function requires that
+    /// - is_null is true
```
workingjubilee, line 66: Do you mean
```suggestion
    /// - is_null is correct
```
If this is a pointer based type, I don't know that we can ask for that invariant to be upheld because there are lots of places that these functions are called generically without knowledge of the underlying type's pointerness or valueness. I think we need to check the nullptrness ourselves.

workingjubilee (changes requested):
This looks much better, thank you! I have one remaining concern and I need to take another closer look a bit later to make sure this is okay for merging, but it looks like we are about there.

pgx/src/datum/range.rs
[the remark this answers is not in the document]
workingjubilee: I suppose yes.

workingjubilee (approved):
I think I can see some improvements that can be made but I think those can all be deferred for future work. This looks great as-is! Thank you so much, and sorry for taking so much time to review it all!

PR #665, opened 2022-08-31
Increases verbosity of pgx-tests in various places

BradyBonnette:
It was brought up in issue https://github.com/tcdi/pgx/issues/324 that any errors in pgx-tests that resulted from a failed `CREATE EXTENSION` call were not propagated properly back to the user. While some of it could be displayed by running the test with the `--nocapture` flag, it's less than ideal since normal Rust tests would at least output something more meaningful instead of simply saying "failed" as it does in this case.

Upon inspection it turns out that there were other places where more verbose output should be warranted. These changes aims to make other errors more visible as well and include other small tweaks along the way.

Also note that these changes can also be referenced as part of https://github.com/tcdi/pgx/issues/612

### DIFF
--- a/cargo-pgx/src/command/install.rs
+++ b/cargo-pgx/src/command/install.rs
@@ -127,8 +127,7 @@ pub(crate) fn install_extension(
     let build_command_messages =
         build_command_stream.collect::<Result<Vec<_>, std::io::Error>>()?;
 
-    println!();
-    println!("installing extension");
+    println!("{} extension", "  Installing".bold().green(),);
     let pkgdir = make_relative(pg_config.pkglibdir()?);
     let extdir = make_relative(pg_config.extension_dir()?);
     let shlibpath = find_library_file(&manifest, &build_command_messages)?;
@@ -287,8 +286,14 @@ pub(crate) fn build_extension(
     let command = command.stderr(Stdio::inherit());
     let command_str = format!("{:?}", command);
     println!(
-        "building extension with features `{}`\n{}",
-        features_arg, command_str
+        "{} extension with features {}",
+        "    Building".bold().green(),
+        features_arg.cyan()
+    );
+    println!(
+        "{} command {}",
+        "     Running".bold().green(),
+        command_str.cyan()
     );
     let cargo_output = command
         .output()
--- a/pgx-pg-config/src/lib.rs
+++ b/pgx-pg-config/src/lib.rs
@@ -455,9 +455,7 @@ pub fn createdb(
     let command_str = format!("{:?}", command);
 
     let child = command.spawn().wrap_err_with(|| {
-        format!(
-            "Failed to spawn process for creating database using command: '{command_str}': "
-        )
+        format!("Failed to spawn process for creating database using command: '{command_str}': ")
     })?;
 
     let output = child.wait_with_output().wrap_err_with(|| {
--- a/pgx-pg-config/src/lib.rs
+++ b/pgx-pg-config/src/lib.rs
@@ -456,15 +456,13 @@ pub fn createdb(
 
     let child = command.spawn().wrap_err_with(|| {
         format!(
-            "Failed to spawn process for creating database using command: '{}': ",
-            command_str
+            "Failed to spawn process for creating database using command: '{command_str}': "
         )
     })?;
 
     let output = child.wait_with_output().wrap_err_with(|| {
         format!(
-            "Failed waiting for spawned process attempting to create database using command: '{}': ",
-            command_str
+            "failed waiting for spawned process to create database using command: '{command_str}': "
         )
     })?;
 
--- a/pgx-pg-config/src/lib.rs
+++ b/pgx-pg-config/src/lib.rs
@@ -451,9 +451,22 @@ pub fn createdb(
         .arg(dbname)
         .stdout(Stdio::piped())
         .stderr(Stdio::piped());
+
     let command_str = format!("{:?}", command);
 
-    let output = command.output()?;
+    let child = command.spawn().wrap_err_with(|| {
+        format!(
+            "Failed to spawn process for creating database using command: '{}': ",
+            command_str
+        )
+    })?;
+
+    let output = child.wait_with_output().wrap_err_with(|| {
+        format!(
+            "Failed waiting for spawned process attempting to create database using command: '{}': ",
+            command_str
+        )
+    })?;
 
     if !output.status.success() {
         return Err(eyre!(
--- a/pgx-tests/src/framework.rs
+++ b/pgx-tests/src/framework.rs
@@ -76,28 +76,32 @@ where
     match result {
         Ok(result) => Ok(result),
         Err(e) => {
-            let e = e.as_db_error().unwrap();
-            Err(eyre!(
-                "Postgres query failed:
-Query: {},
-Query Params: {:?},
-Error Message: {}
-Code: {:?}
-Severity: {}
-Detail: {}
-Hint: {}
-Schema: {}
-Table: {}",
-                query.unwrap(),
-                query_params,
-                e.message(),
-                e.code(),
-                e.severity(),
-                e.detail().unwrap_or("None"),
-                e.hint().unwrap_or("None"),
-                e.schema().unwrap_or("None"),
-                e.table().unwrap_or("None"),
-            ))
+            let dberror = e.as_db_error().unwrap();
+            let query = query.unwrap();
+            let query_params = query_params.unwrap_or(&[]);
+            let query_message = dberror.message();
+
+            let mut message = format!(
+                "postgres query failed:\nquery: {query}\nquery params: {:?}\nmessage: {query_message}",
+                query_params
+            );
+
+            if let Ok(var) = std::env::var("RUST_BACKTRACE") {
+                if var.eq("1") {
+                    let code = dberror.code().code();
+                    let severity = dberror.severity();
+                    let detail = dberror.detail().unwrap_or("None");
+                    let hint = dberror.hint().unwrap_or("None");
+                    let schema = dberror.hint().unwrap_or("None");
+                    let table = dberror.table().unwrap_or("None");
+                    let more_info = format!(
+                        "\ncode: {code}\nseverity: {severity}\ndetail: {detail}\nhint: {hint}\nschema: {schema}\ntable: {table}"
+                    );
+                    message.push_str(more_info.as_str());
+                }
+            }
+
+            Err(eyre!(message))
         }
     }
 }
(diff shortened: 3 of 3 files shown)


cargo-pgx/src/command/install.rs
```diff
@@ -127,8 +127,7 @@ pub(crate) fn install_extension(
     let build_command_messages =
         build_command_stream.collect::<Result<Vec<_>, std::io::Error>>()?;
 
-    println!();
-    println!("installing extension");
+    println!("{} extension", "  Installing".bold().green(),);
```
BradyBonnette, line 130: This is only to make things look a little nicer.

cargo-pgx/src/command/install.rs
```diff
@@ -287,8 +286,14 @@ pub(crate) fn build_extension(
...
-        "building extension with features `{}`\n{}",
-        features_arg, command_str
+        "{} extension with features {}",
+        "    Building".bold().green(),
+        features_arg.cyan()
+    );
+    println!(
+        "{} command {}",
+        "     Running".bold().green(),
+        command_str.cyan()
```
BradyBonnette, line 296: This is only to make things look a little nicer.

pgx-pg-config/src/lib.rs
```diff
@@ -451,9 +451,22 @@ pub fn createdb(
...
+            command_str
+        )
+    })?;
+
+    let output = child.wait_with_output().wrap_err_with(|| {
+        format!(
+            "Failed waiting for spawned process attempting to create database using command: '{}': ",
+            command_str
+        )
+    })?;
```
BradyBonnette, line 469: This is only to make things look a little nicer.

pgx-tests/src/framework.rs
```diff
@@ -162,9 +204,12 @@ fn initialize_test_framework(
...
     let mut state = TEST_MUTEX.lock().unwrap_or_else(|_| {
-        // if we can't get the lock, that means it was poisoned,
-        // so we just abruptly exit, which cuts down on test failure spam
-        std::process::exit(1);
+        // This used to immediately throw an std::process::exit(1), but it
+        // would consume both stdout and stderr, resulting in error messages
+        // not being displayed unless you were running tests with --nocapture.
+        panic!(
+            "Could not obtain test mutex. Please check for errors in pervious tests to find the root cause."
+        );
```
BradyBonnette, line 212: I do believe this was the reason none of the errors were showing up at all. `std::process::exit(1)` would completely consume any errors that wanted to be displayed in the output. I am not 100% sure `panic!` is the right way to go, but in my local testing the failure "spam" wasn't all that significant. I felt like it was better to have more information than not enough.

pgx-tests/src/framework.rs
```diff
@@ -56,14 +55,61 @@ where
...
+                "Postgres query failed:
+Query: {},
+Query Params: {:?},
+Error Message: {}
+Code: {:?}
+Severity: {}
+Detail: {}
+Hint: {}
+Schema: {}
+Table: {}",
```
BradyBonnette, line 90: I feel like there might be a better way to do this?

pgx-tests/src/framework.rs
```diff
@@ -256,7 +327,7 @@ fn install_extension() -> eyre::Result<()> {
         .arg("--pg-config")
         .arg(pg_config.path().ok_or(eyre!("No pg_config found"))?)
         .stdout(Stdio::inherit())
-        .stderr(Stdio::inherit())
+        .stderr(Stdio::piped())
```
BradyBonnette, line 330: Setting this to `piped` means that we can capture it in the child process so that we can display it upon an error condition

BradyBonnette:
Here is an example of output before any changes applied. This is a result of an invalid Postgres query by running `cargo test -p aggregate --lib`:

image

Here is the output with the changes implemented, throwing the same error:

image

It should be noted that I used the same exact command -- I did not append `--nocapture` at the end.

Because of the way the test framework is currently set up, there may be some instances where a single test failure means you get a huge infodump of other abruptly halted tests. Typically this is because the tests are behind a shared mutex that will shut down any running postgres instances and fail them outright -- all due to one failure. In those situations, the other abruptly-halted tests will display something along the lines of:

image

Though, in the above screenshot, the reason for the failure is because I specifically injected a failure in the `CREATE EXTENSION` query, which would technically fail on every single test. It feels sorta "spam-y" but with the way the framework is built, there was no other option.

Here is another example of a framework setup error that happens when a postgres client fails to set up properly:

image

pgx-tests/src/framework.rs
workingjubilee: Well, it's somewhat common to use lowercase for framing text like this in Rust error messages so as not to distract from the more important (and possibly semantically capitalized) error messaging text. And we're still definitely adding a lot of vertical height to query failure, and a lot more for things that may be "None".

Query, yes. Error message, yes.

For the others... can it be made more compact so it doesn't add newlines when those are used? We're on an error path so we can afford to spend extra compute on formatting.

I also think that instead of `{}` for each case you should probably be assigning keys to these, so sev = e.severity(), and then using it up there somewhere.

pgx-pg-config/src/lib.rs
```diff
@@ -451,9 +451,22 @@ pub fn createdb(
...
+        format!(
+            "Failed to spawn process for creating database using command: '{}': ",
+            command_str
+        )
+    })?;
+
+    let output = child.wait_with_output().wrap_err_with(|| {
+        format!(
+            "Failed waiting for spawned process attempting to create database using command: '{}': ",
+            command_str
```
workingjubilee, line 467:
```suggestion
        format!(
            "failed waiting for spawned process to create database using command: '{command_str}': "
```

pgx-pg-config/src/lib.rs
```diff
@@ -451,9 +451,22 @@ pub fn createdb(
...
         .stdout(Stdio::piped())
         .stderr(Stdio::piped());
+
     let command_str = format!("{:?}", command);
 
-    let output = command.output()?;
+    let child = command.spawn().wrap_err_with(|| {
+        format!(
+            "Failed to spawn process for creating database using command: '{}': ",
+            command_str
```
workingjubilee, line 460: These format strings can have problems if you rely on them indexing like that, so I generally recommend avoiding that. Also tends to eat source-level vertical space.
```suggestion
            "Failed to spawn process for creating database using command: '{command_str}': "
```

pgx-tests/src/framework.rs
workingjubilee: This implies something about test history without stating the cause outright. How about
```suggestion
        // This used to immediately throw an std::process::exit(1), but it
        // would consume both stdout and stderr, resulting in error messages
        // not being displayed unless you were running tests with --nocapture.
        panic!("Could not obtain test mutex. A previous test may have hard-aborted while holding it.");
```

pgx-tests/src/framework.rs
```diff
@@ -379,8 +492,13 @@ fn start_pg(loglines: LogLines) -> eyre::Result<String> {
 
     // add a shutdown hook so we can terminate it when the test framework exits
     add_shutdown_hook(move || unsafe {
-        let message_string =
-            std::ffi::CString::new("Stopping Postgres\n\n".bold().blue().to_string()).unwrap();
+        let message_string = std::ffi::CString::new(
+            format!("Stopping Postgres PID {}\n\n", pgpid)
```
workingjubilee, line 496: should have more symmetry between the two pid markers in a test's output over a test run's lifetime, and probably don't need double newlines as the next entry will probably have its own framing. or maybe I'm wrong on that last bit.
```suggestion
            format!("stopping postgres (pid={pgpid})\n")
```

cargo-pgx/src/command/install.rs
workingjubilee: I love it!

workingjubilee (changes requested):
This looks wonderful! I think the only major Thing is clawing back a bit of the vertical height and making sure we aren't distracting from the important parts. I agree that a little bit of "spam" isn't so bad but we should generally prefer to fight against giving up height in the terminal as excess noise will increase the tendency to tune out the test output.

Likewise, we should generally make anything that the user will see again and again look more subdued, while gearing in the highlighting on the things that will change every time.

If something needs to be more verbose, we can offer toggles for that. We could also piggyback off the setting for `RUST_BACKTRACE` instead of asking for a unique configuration just for PGX, cranking everything to "maximum volume" when they use `"full"`, and up slightly when they use `1`, *IF* we think that's appropriate (probably "yes" if it's implicitly "a backtrace", more dubious if it's not definitely a trace).

pgx-tests/src/framework.rs
workingjubilee: Is that really the thing that was swallowing all those messages? **Thank you**.

cargo-pgx/src/command/install.rs
BradyBonnette: ❤️

pgx-tests/src/framework.rs
BradyBonnette: Looks good! We should use that.

pgx-tests/src/framework.rs
BradyBonnette: I don't believe this particular thing was swallowing all of the test messages, but it was definitely swallowing any error messages as a result of the test framework installing the extension via `cargo pgx install ...`. The main reason for no errors showing up at all was the `std::process::exit(1)` call that I eliminated earlier.

This change was a result of me going through as many paths as possible, causing them to break, then seeing the result on the screen as the test runs. In this particular case, if `cargo pgx install ...` failed for any reason, it would fail silently, the test would attempt to continue then fail for "some other reason" that was not really visible to the user.

pgx-tests/src/framework.rs
BradyBonnette: Looks good!

BradyBonnette: Regarding the newline: They were there before I started doing this work, so while we have a chance to improve it we should. This could probably also be colored too if we wanted to get Super Fancy.

pgx-pg-config/src/lib.rs
BradyBonnette: Ah right, I keep forgetting you can do that with `format!()`.

BradyBonnette:
@workingjubilee 

Here are some updated screenshots of errors that have been abbreviated as per your suggestions.

Without `RUST_BACKTRACE=1`:

image

Same error with `RUST_BACKTRACE=1` (remaining stuff cut off from the screenshot because this 4k monitor doesnt have _that_ much real estate):

image

workingjubilee (approved):
I like this! I did mention a slight formatting adjustment for sliiightly prettier and more searchable errors on Discord. By and large though, this looks nice!

BradyBonnette:
Yup, working on those now and getting feedback from you :)

BradyBonnette:
@workingjubilee Quick update after we discussed some things.

Example of a simple error without `RUST_BACKTRACE=1`:

image

Example with a backtrace:

image

PR #666, opened 2022-09-01
Rewrite Spi{Ok,Error} without num-traits

workingjubilee:
This thins our dependency tree with fairly light lifting.
It also makes it much harder to make mistakes,
as we no longer rely on the odd split of needing negative values
in order to do the status code check properly.

### DIFF
--- a/pgx/src/spi.rs
+++ b/pgx/src/spi.rs
@@ -18,6 +18,7 @@ use std::ops::{Index, IndexMut};
 /// These match the Postgres `#define`d constants prefixed `SPI_OK_*` that you can find in `pg_sys`.
 #[derive(Debug, PartialEq)]
 #[repr(i32)]
+#[non_exhaustive]
 pub enum SpiOk {
     Connect = 1,
     Finish = 2,
@@ -36,7 +37,7 @@ pub enum SpiOk {
     RelRegister = 15,
     RelUnregister = 16,
     TdRegister = 17,
-    #[cfg(feature = "pg15")]
+    /// Added in Postgres 15
     Merge = 18,
 }
 
--- a/pgx/src/spi.rs
+++ b/pgx/src/spi.rs
@@ -36,6 +36,8 @@ pub enum SpiOk {
     RelRegister = 15,
     RelUnregister = 16,
     TdRegister = 17,
+    #[cfg(feature = "pg15")]
+    Merge = 18,
 }
 
 /// These match the Postgres `#define`d constants prefixed `SPI_ERROR_*` that you can find in `pg_sys`.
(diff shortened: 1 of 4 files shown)


pgx/src/spi.rs
```diff
@@ -37,25 +37,61 @@ pub enum SpiOk {
...
+#[repr(i32)]
 pub enum SpiError {
-    // NB:  These are #define'd as negative, but we redefine them as positive so that
-    // #[derive(Primitive)] will work.  We just need to negate result codes from the
-    // various SPI_xxx functions when looking for errors
-    Connect = 1,
-    Copy = 2,
-    Opunknown = 3,
-    Unconnected = 4,
+    Null = 0,
```
workingjubilee, line 43: This one is mildly odd as there are functions that will return a nullptr to signal errors happened, and the result code gets set to 0 at the start of most functions that set the result code. So it's not a formal "variant", but if we see it, it suggests something has gone wrong.

pgx/src/spi.rs
```diff
@@ -37,25 +37,61 @@ pub enum SpiOk {
...
-    RelNotFound = 13,
+    Cursor = -5, /* not used anymore */
+    Argument = -6,
+    Param = -7,
+    Transaction = -8,
+    NoAttribute = -9,
+    NoOutFunc = -10,
+    TypUnknown = -11,
+    RelDuplicate = -12,
+    RelNotFound = -13,
```
workingjubilee, line 57: This might seem like a huge break but it's only an actual functional break if they specifically cast to an integer and then match on it. As-is it only breaks on the ones I gave CamelCase because I don't think there's any actual reason not to. I guess we could keep them as-is...?

pgx/src/spi.rs
BradyBonnette: Is it advisable to even use a nullptr to signal "something bad happened"? Or should it consistently be `Ok/Err` wherever it's applicable?

pgx/src/spi.rs
BradyBonnette: Well, considering 0.5.0 is going to be breaking anyways, my opinion might be that doing whatever feels more natural (and also reducing dependencies) would be a really good thing. 

What are the integer values used for? Is it only for matching error results? Are the primitives needed for some kind of postgres functionality?

workingjubilee: Assigning integer values enforces that these values are, and allows you to cast `result_enum as i32`.

pgx/src/spi.rs
workingjubilee: I would say "no" but this is part of what Postgres does. However it's not actually clear it ever arises in a codepath we have hit yet, so thinking about it, I don't know if we should.

cc @willmurnane

pgx/src/spi.rs
```diff
@@ -36,6 +36,8 @@ pub enum SpiOk {
     RelRegister = 15,
     RelUnregister = 16,
     TdRegister = 17,
+    #[cfg(feature = "pg15")]
+    Merge = 18,
```
workingjubilee, line 40: This is the wrong approach, as it requires users to `cfg` around it too, instead of letting them using one codebase for different versions of Postgres.

workingjubilee: This is worst in migration mode.

PR #667, opened 2022-09-01
Note reasons for all deps in pgx core and sort

workingjubilee:


pgx/Cargo.toml
```diff
@@ -33,26 +33,33 @@ no-default-features = true
...
-serde_cbor = "0.11.2"
-serde_json = "1.0.81"
-time = { version = "0.3.9", features = ["formatting", "parsing", "alloc", "macros"] } # TODO(0.6.0): add `optional = true`
-atomic-traits = "0.3.0"
-heapless = "0.7.13"
-uuid = { version = "1.0.0", features = [ "v4" ] }
-once_cell = "1.10.0"
+
+# used to internally impl things
+once_cell = "1.10.0" # polyfill until std::lazy::OnceCell stabilizes
```
workingjubilee, line 44: lol as if that'll ever happen

pgx/Cargo.toml
```diff
@@ -33,26 +33,33 @@ no-default-features = true
...
+
+# error handling and logging
 eyre = "0.6.8"
+thiserror = "1.0"
 tracing = "0.1.34"
 tracing-error = "0.2.0"
-quote = "1.0.18"
-thiserror = "1.0"
+
+# exposed in public API
```
workingjubilee, line 54: fwiw I considered something "exposed" if it's "encapsulated but basically we've documented it definitely uses this, so we're stuck using it or completely reimplementing it or something similar", which is where seahash is at, but most of these aren't even that, they're just Right There.

johnrballard (approved):
This is great. Super helpful.

LGTM.

BradyBonnette:
👀 

Is this something we should consider doing from this point forward?

johnrballard:
I think so, maybe we put a little comment in the dependcy file that gives some direction to that end?

> 👀
> 
> Is this something we should consider doing from this point forward?

Yes I think so.  Perhaps we note that in the file?

pgx/Cargo.toml
```diff
@@ -33,26 +33,33 @@ no-default-features = true
...
 
 [dependencies]
-bitflags = "1.3.2"
-bitvec = "1.0" # processing array nullbitmaps
-cstr_core = "0.2.5"
 enum-primitive-derive = "0.2.2"
 num-traits = "0.2.15"
-seahash = "4.1.0"
-libc = "0.2.126"
+
```
johnrballard, line 38: Add a note on line 36 to make sure that developers give a reason for the addition.

pgx/Cargo.toml
```diff
@@ -33,26 +33,33 @@ no-default-features = true
...
+thiserror = "1.0"
 tracing = "0.1.34"
 tracing-error = "0.2.0"
-quote = "1.0.18"
-thiserror = "1.0"
+
+# exposed in public API
+atomic-traits = "0.3.0" # PgAtomic and shmem init
+bitflags = "1.3.2" # BackgroundWorker
+bitvec = "1.0" # processing array nullbitmaps
```
workingjubilee, line 57: despite the similar names, these actually do fairly different things.

pgx/Cargo.toml
workingjubilee: I think we should add a note to CONTRIBUTING.md instead since there's multiple Cargo.tomls in the repo and this PR only sorts/annotates one. I think the "peer pressure" of seeing the small notes will assist in enforcing this norm in-location.

johnrballard (approved):
LGTM !

Super great!

PR #668, opened 2022-09-01
Add a test which calls a user-defined function with NULL argument

willmurnane:
I found this incorrect behavior, but I haven't figured out how to fix it.

### DIFF
--- a/pgx-tests/src/tests/postgres_type_tests.rs
+++ b/pgx-tests/src/tests/postgres_type_tests.rs
@@ -61,6 +61,15 @@ impl InOutFuncs for CustomTextFormatSerializedType {
     }
 }
 
+#[pg_extern(immutable)]
+fn fn_takes_option(input: Option<CustomTextFormatSerializedType>) -> String {
+    input.map_or("nothing".to_string(), |c| {
+        let mut b = StringInfo::new();
+        c.output(&mut b);
+        b.to_string()
+    })
+}
+
 #[derive(Serialize, Deserialize, PostgresType)]
 pub struct JsonType {
     a: f32,
@@ -86,6 +95,20 @@ mod tests {
         assert_eq!(result.a, 1.0);
         assert_eq!(result.b, 2.0);
         assert_eq!(result.c, 3);
+    }
+
+    #[pg_test]
+    fn test_call_with_value() {
+        let result = Spi::get_one::<String>(
+            "SELECT fn_takes_option('1.0,2.0,3'::CustomTextFormatSerializedType);",
+        );
+        assert_eq!("1,2,3", result.unwrap());
+    }
+
+    #[pg_test]
+    fn test_call_with_null() {
+        let result = Spi::get_one::<String>("SELECT fn_takes_option(NULL);");
+        assert_eq!(Some(String::from("nothing")), result);
     }
 
     #[pg_test]


pgx-utils/src/lib.rs
```diff
@@ -129,8 +129,10 @@ pub enum FunctionArgs {
 
 #[derive(Debug)]
 pub enum CategorizedType {
-    Iterator(Vec<String>),
-    OptionalIterator(Vec<String>),
+    Iterator {
+        types: Vec<String>,
+        is_optional: bool,
+    },
```
workingjubilee, line 135: I'm not sure I understand why we're now adding a bool into the same variant, as opposed to using a different variant, as previously done?

pgx-utils/src/rewriter.rs
```diff
@@ -718,21 +694,18 @@ impl FunctionSignatureRewriter {
...
                             quote_spanned! {ident.span()=>
-                                let #name = pgx::pg_getarg::<#option_type>(#fcinfo_ident, #i);
+                                let #name : #type_ = pgx::pg_getarg::<#option_type>(#fcinfo_ident, #i);
                             }
                         } else if type_matches(&type_, "pg_sys :: FunctionCallInfo")
                             || type_matches(&type_, "pgx :: pg_sys :: FunctionCallInfo")
                         {
                             quote_spanned! {ident.span()=>
-                                let #name = #fcinfo_ident;
+                                let #name: pgx::pg_sys::FunctionCallInfo  = #fcinfo_ident;
```
workingjubilee, line 708: rustfmt often leaves formatting in macro contexts untouched. Here, `#name:` is used, elsewhere, `#name :` is used, this must be consistent, and preferably consistent with the default type annotation formatting that rustfmt emits.

pgx-utils/src/rewriter.rs
```diff
@@ -566,8 +537,18 @@ impl PgGuardRewriter {
                         if suffix_arg_name && ident.ident.to_string() != "fcinfo" {
                             let ident = Ident::new(&format!("{}_", ident.ident), ident.span());
                             arg_list.extend(quote! { #ident, });
+                            if type_matches(ty.ty.as_ref(), "()")
+                                || extract_type_from_option(ty.ty.as_ref()).is_some()
+                                || type_matches(&ty.ty, "pg_sys :: FunctionCallInfo")
+                                || type_matches(&ty.ty, "pgx :: pg_sys :: FunctionCallInfo")
```
workingjubilee, line 543: Why the spaces in `pgx :: pg_sys :: FunctionCallInfo`?

pgx-utils/src/rewriter.rs
```diff
@@ -482,7 +449,7 @@ impl PgGuardRewriter {
             func.sig.ident.span(),
         );
 
-        let arg_list = PgGuardRewriter::build_arg_list(&sig, false);
+        let (arg_list, _matchable) = PgGuardRewriter::build_arg_list(&sig, false);
```
workingjubilee, line 452: This function had its return signature altered but the latter half of the tuple is used in only one location? Is there another way to do this?

workingjubilee (commented):
Apologies for the delay in review. I find the content of this segment of the code hard to follow in general as I am not as experienced in it, and as I start working on attempting to update and land #615, it's likely to induce conflicts with this set of changes. So my initial remarks are just from a relatively cursory pass... in general this has proven to be an extremely touchy segment of the code, in the past, especially around SRFs, so any changes touching it could stand to add some comments and focus, a lot, on legibility.

pgx-utils/src/lib.rs
willmurnane: The different variant was translated into "call the same code, but with a boolean flag" anyways. I don't mind reverting this change, but it collapsed the code in `item_fn_with_rewrite` quite a bit. Combined with factoring out the `if types.len() == 1` check, that collapsed all 4 cases down to one code path.

pgx-utils/src/rewriter.rs
willmurnane: I think that's due to how [`format!("{}", quote!`](https://github.com/tcdi/pgx/blob/f2c54dc38e6b28cb7dcf43ef9f84df36e7108278/pgx-utils/src/rewriter.rs#L750) works. I was just matching the style seen elsewhere.

pgx-utils/src/rewriter.rs
willmurnane: I'm not sure it's entirely correct that I'm ignoring it in this case. I'm not totally clear on how `PgGuardRewriter .item_fn` is called and for what reasons. It looks like it's usually called from [here](https://github.com/tcdi/pgx/blob/develop/pgx-macros/src/lib.rs#L592-L598), where `rewrite_args` is always supplied as `false`, but is also called from [here](https://github.com/tcdi/pgx/blob/develop/pgx-macros/src/lib.rs#L46), where it's also hard-coded to a value of `false`. So based on I think it might be okay to entirely remove `item_fn_with_rewrite`, the argument `rewrite_args`, and so on. I didn't want to do that surgery without understanding more, though.

pgx-utils/src/rewriter.rs
willmurnane: Good catch. I'll update.

willmurnane:
I agree with the overall challenging nature of this code, especially given the behavior of Rust macros ("[macro expansion to text is a lossy process](https://github.com/dtolnay/cargo-expand#disclaimer)"). I could try a couple of things to make this easier to review:
* break this PR into smaller, more digestible chunks that do exactly one thing
* add more comments: what is happening when this code is called, and why does (this) need to happen in response
* rebase onto #615
* dive into some of the major surgery I mentioned in review comments above.
I would be happy to do any or all of these. Let me know what would be helpful.

pgx-utils/src/rewriter.rs
willmurnane: Oops - I misread. In the first case, it's got a value of `true` hard-coded, not `false`. Not sure how I managed that one...

pgx-utils/src/lib.rs
workingjubilee: I would actually expect it to be such an... "overly ritualistic" form? It does suggest this code needs to be refactored, but there's something in the back of my mind that suggests this particular refactoring will reach a local maxima faster and thus lower than other ones which try to address the code factoring from a "one-level up" perspective.

workingjubilee:
I think this is fine as one PR, but comments comments comments.

I don't think you want to rebase onto it yet. Soon maybe. If you want to give it a quick experimental pass on your local to see if there are any conflicts you think are particularly noteworthy, sure, but that PR itself needs to be updated.

workingjubilee (changes requested):
I want to land this before 0.5.0, truly I do, but this PR too is also dashed upon the rocks of #615.

workingjubilee:
#615 has landed.

willmurnane:
#615 fixed the problem that this PR was addressing. I've winnowed it down to just add the test that previously failed.

workingjubilee (approved):
Which also makes it much easier to review. Nice!

PR #672, opened 2022-09-06
Deprecate `Array<'a, T>::as_slice`

workingjubilee:
This marks the formal beginning of the deprecation process for the `Array<'_, T>::as_slice` function. As currently implemented, it cannot be safe without inducing at minimum these panics, which means it must be `unsafe`, which would already be a breaking change to introduce. And this does not actually make it fully correct, either, but we can expect code "in the wild" probably exercises this function as it is fairly idiomatic to use in Rust, so removing it abruptly without offering these directions would make the upgrade path for 0.5.0 even more annoying.

I expect we can introduce a new variant of this function in the future which offers a very similar interface and is either safe or `unsafe` (and arguably RawArray offers the `unsafe` version already), but I expect that to not happen before 0.6.1 or 0.7.0, as it involves some fairly sweeping structural changes in PGX.

### DIFF
--- a/pgx-examples/arrays/src/lib.rs
+++ b/pgx-examples/arrays/src/lib.rs
@@ -14,21 +14,18 @@ pg_module_magic!();
 
 #[pg_extern]
 fn sq_euclid_pgx(a: Array<f32>, b: Array<f32>) -> f32 {
-    a.as_slice()
-        .iter()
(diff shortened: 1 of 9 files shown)


workingjubilee:
I had to add several tests to feel confident merging this but I feel it is ready now.

PR #676, opened 2022-09-09
Propagate TryFromIntError in `TryFrom<Datum> for Date`

workingjubilee:
The error we went with before in this impl doesn't make sense to me here, on further consideration.

### DIFF
--- a/pgx/src/datum/date.rs
+++ b/pgx/src/datum/date.rs
@@ -8,6 +8,7 @@ Use of this source code is governed by the MIT license that can be found in the 
 */
 
 use crate::{pg_sys, FromDatum, IntoDatum};
+use core::num::TryFromIntError;
 use std::ffi::CStr;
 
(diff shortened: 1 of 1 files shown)


workingjubilee:
...Ah, it appears that the top bytes of the Datum may be "whatever". Very well then!

pgx/src/datum/date.rs
```diff
@@ -41,7 +42,11 @@ impl FromDatum for Date {
         if is_null {
             None
         } else {
-            Some(datum.try_into().expect("Error converting date datum"))
+            if cfg!(feature = "pg10") {
+                Some(Date(datum.value() as i32))
```
Hoverbear, line 46: Oh deer.

Hoverbear (approved):

pgx/src/datum/date.rs
workingjubilee: :upside_down_face: PostgreSQL has a... sometimes interesting way of handling Datums, yes.

PR #679, opened 2022-09-09
Updates versions script because of #645

BradyBonnette:
We want to be sure any new crates need to also be part of the release process. https://github.com/tcdi/pgx/pull/645 adds a new crate that needs to be considered for release.


workingjubilee (approved):
Thank you!

PR #680, opened 2022-09-10
Recommend `cargo install --locked`

workingjubilee:
Otherwise Cargo will try to select "whatever seems compatible" in terms of crates when they install, which can in rare cases result in them trying to pull in versions which the Rust toolchain we intend to build with cannot actually compile. This was used to solve a dependency/version problem a while ago in tcdi/plrust#73 and it seems like a better recommendation for our users in general.


BradyBonnette:
Interesting, I would have thought default behavior of cargo would be to use exactly what is in the `Cargo.lock` file at all times if it's present.

workingjubilee:
Sadly not the case!

PR #682, opened 2022-09-12
Add a function to run arbitrary sql with arguments.

montanalow:
This helps to ensure args are properly encoded, rather than relying on mechanisms like format!() that may be less correct or secure.

### DIFF
--- a/pgx/src/spi.rs
+++ b/pgx/src/spi.rs
@@ -171,6 +171,12 @@ impl Spi {
             client.update(query, None, None);
         })
     }
+    
+    fn run_with_args(query: &str, args: Vec<(PgOid, Option<pg_sys::Datum>)>) {
+        Spi::execute(|mut client| {
+            client.update(query, None, Some(args));
+        })
+    }
 
     /// explain a query, returning its result in json form
     pub fn explain(query: &str) -> Json {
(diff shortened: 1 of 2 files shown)


workingjubilee:
Hello! I think this is a fine addition to have but I would greatly enjoy a few tests for it to verify a couple trivial use-cases, simply so that it is in the testing infra and I can reference it if we build further on this feature (I think it is incredibly plausible that may happen).

pgx/src/spi.rs
```diff
@@ -171,6 +171,12 @@ impl Spi {
             client.update(query, None, None);
         })
     }
+    
+    fn run_with_args(query: &str, args: Vec<(PgOid, Option<pg_sys::Datum>)>) {
```
workingjubilee, line 175: You will want to make this function `pub` and also add a comment.
```suggestion
    pub fn run_with_args(query: &str, args: Vec<(PgOid, Option<pg_sys::Datum>)>) {
```

montanalow:
I added a couple tests, but since this function doesn't return anything... It's sorta basic. Is there a Postgres level assert I should be using in the SQL instead, or is this what you were looking for?

workingjubilee:
Please do not submit an unrelated change as part of this PR.

workingjubilee:
Maybe this PR should also have `explain_with_args` and you should test the return of that instead.

montanalow:
> Please do not submit an unrelated change as part of this PR.

That wasn't intentional, I didn't realize I'd open this PR tracking against the master branch. I'll need to close this PR and open a new one on a dedicated branch. I'll include `explain_with_args` as well.

PR #685, opened 2022-09-13
Remove unused code

willmurnane:
There are two use cases which call `item_fn`:
* One when handling `#[pg_guard]`
* One when handling `#[pg_extern]`, which (almost always) eventually emits `#[pg_guard]` and leads to the first case
This lead to an unnecessary amount of confusion between the two cases. This PR splits the two cases into two functions, and does some small cleanup around the edges afterwards.


workingjubilee (changes requested):
This will conflict with #615, which may still yet receive final-pass revisions. It cannot land before then, and after it does, I will be pushing towards `0.5.0`.

willmurnane:
This now winds up just eliminating a big swath of unused code :)

pgx-utils/src/rewriter.rs
```diff
@@ -505,18 +72,14 @@ impl PgGuardRewriter {
...
-        let sql_graph_entity_submission = entity_submission.cloned().into_iter();
-
         quote_spanned! {func.span()=>
             #prolog
             #[doc(hidden)]
             #vis #sig {
                 #[allow(non_snake_case)]
                 #func
-                #func_call
+                pg_sys::guard::guard( || #func_name(#arg_list) )
```
willmurnane, line 81: This is effectively the code from line 491, but we're always doing the `no_guard = false` case so I inlined it.

workingjubilee:
what, seriously? half this code is just unused...?

willmurnane:
Nope, 3/4s. There used to be two ways that `item_fn` was called, and it in turn called `item_fn_with_rewrite` or `item_fn_without_rewrite` as appropriate. One of those was eliminated in #615, and the other greatly simplified, so that only `item_fn_without_rewrite` remained. So we can eliminate all of the `item_fn_with_rewrite` code and all of its dependencies.

workingjubilee (approved):
amazing.

PR #686, opened 2022-09-13
Eliminate tuple udf

willmurnane:
There's a lot of song and dance about how functions which return a single tuple are turned into something that Postgres can call. This eliminates that: when a function which returns a tuple is discovered, the `func_call` is rewritten into an iterator on the spot, and then it's treated as if it were a function which returns an iterator (and thus passes through the `impl_table_srf` function). This removes:
* the `impl_tuple_udf` function
* the tuple return type of `item_fn`
* the edge case of "this might not need a `pg_guard`"


workingjubilee (changes requested):
This will conflict with #615, which may still yet receive final-pass revisions. It cannot land before then, and after #615 does, I will be pushing towards `0.5.0`.

willmurnane:
Closing in favor of #685. Now that #615 has landed, *all* of these functions are unused and can be removed.

PR #687, opened 2022-09-14
Fix undefined behaviour when loading control file

Smittyvb:
The type of `__pgx_marker` is `extern "C" fn() -> eyre::Result<ControlFile>`, but it was incorrectly called as `extern "C" fn() -> SqlGraphEntity` in `cargo-pgx`, which is undefined behavior since it's UB to assume that a valid value for an `eyre::Result` is also a valid value for a `SqlGraphEntity`. On my M1 macOS device the undefined behavior results in an unconditional `Failed to get control file information` panic. This PR fixes that.

The only reason this ever worked was because on some platforms `eyre::Result<ControlFile>` and `SqlGraphEntity` both can have a `ControlFile` after a 1 byte long prefix. And it happened to be the case that `Result::Ok` and `SqlGraphEntity::ExtensionRoot` both correpond to a prefix of `0`.

I've also wrote [a similar fix for the same issue 0.4](https://github.com/Smittyvb/pgx/commit/25b6a31298dfcdc641fcfd3583f089b274fdcb26), I can make a PR against `master` with that commit if you want to backport this to the 0.4.x release series.

### DIFF
--- a/cargo-pgx/src/command/schema.rs
+++ b/cargo-pgx/src/command/schema.rs
@@ -367,7 +367,13 @@ pub(crate) fn generate_schema(
     let mut entities = Vec::default();
     let sql_mapping;
 
+    #[rustfmt::skip] // explict extern "Rust" is more clear here
     unsafe {
+        // SAFETY: Calls foreign functions with the correct type signatures.
+        // Assumes that repr(Rust) enums are represented the same in this crate as in the external
+        // binary, which is the case in practice when the same compiler is used to compile the
+        // external crate.
+
         POSTMASTER_LIBRARY
             .get_or_try_init(|| {
                 libloading::os::unix::Library::open(
@@ -384,14 +390,14 @@ pub(crate) fn generate_schema(
             .wrap_err_with(|| format!("Couldn't libload {}", lib_so.display()))?;
 
         let sql_mappings_symbol: libloading::os::unix::Symbol<
-            unsafe extern "C" fn() -> ::pgx_utils::sql_entity_graph::RustToSqlMapping,
+            unsafe extern "Rust" fn() -> ::pgx_utils::sql_entity_graph::RustToSqlMapping,
         > = lib
             .get("__pgx_sql_mappings".as_bytes())
             .expect(&format!("Couldn't call __pgx_sql_mappings"));
         sql_mapping = sql_mappings_symbol();
 
         let symbol: libloading::os::unix::Symbol<
-            unsafe extern "C" fn() -> eyre::Result<ControlFile>,
+            unsafe extern "Rust" fn() -> eyre::Result<ControlFile>,
         > = lib
             .get("__pgx_marker".as_bytes())
             .expect(&format!("Couldn't call __pgx_marker"));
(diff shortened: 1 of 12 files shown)


workingjubilee:
There are... many fixes in 0.5.0 that will not appear in 0.4. This will be another one of them.

cargo-pgx/src/command/schema.rs
```diff
@@ -390,10 +390,14 @@ pub(crate) fn generate_schema(
             .expect(&format!("Couldn't call __pgx_sql_mappings"));
         sql_mapping = sql_mappings_symbol();
 
-        let symbol: libloading::os::unix::Symbol<unsafe extern "C" fn() -> SqlGraphEntity> = lib
+        let symbol: libloading::os::unix::Symbol<
+            unsafe extern "C" fn() -> eyre::Result<ControlFile>,
+        > = lib
```
workingjubilee, line 395: It seems unlikely a Result returning function like this is actually FFI-safe, `extern "C" fn` or no. Rust enums are usually not FFI-safe except in some qualified cases. This probably works but it is probably still UB, unfortunately. Maybe this should be using `extern "Rust" fn`?

Smittyvb: That does look like UB to me. That issue also affects the other two external function calls in this file (to `__pgx_sql_mappings` and `__pgx_internals*`), which all return `repr(Rust)` `enum`s.

I think even a `extern "Rust" fn` would still technically be UB since the compiler doesn't guarantee that the representation of an item is the same across compilation sessions (although in practice it usually is, at least when the target and compiler version are the same). Serializing data to a C array as it goes across the FFI boundary would remove that UB (although doing so would probably result in a minor hit to performance).

Smittyvb:
I was able to determine that the cause of this issue is a rustc change sometime between beta 1.64 and nightly 2022-09-11. I don't think having an aarch64 CPU actually caused the issue, but instead the LLVM 14 -> 15 bump might be responsible.

cargo-pgx/src/command/schema.rs
workingjubilee: I agree the guarantees are very thin here, but a `repr(Rust)` enum should have a consistent result, in practice, with the same _compiler_ (there are various reasons why this _has_ to be true), and is less likely to be a bear (on performance or otherwise).

Smittyvb: I updated the PR to make the relevant functions `extern "Rust"`. `extern "Rust"` is implied if nothing is specified but I wrote it out explicitly on the relevant functions for clarity.

I also documented that the `cargo-pgx` compiler should match the compiler used to compile crates that use it.

Smittyvb: It turns out rustfmt doesn't like explicit `extern "Rust"`, so I've made them implict. (rustfmt only changed the ones that weren't in a macro but I did all of them for consistency)

workingjubilee: I think `#[rustfmt::skip]` is better here.

Smittyvb: Updated to use `#[rustfmt::skip]`.

workingjubilee (approved):
Looks like everything is good to go! Thank you!

PR #689, opened 2022-09-14
Add a function to run arbitrary sql with arguments.

montanalow:
This helps to ensure args are properly encoded, rather than relying on mechanisms like format!() that may be less correct or secure.

Follow up to #682 with proper git branching.


workingjubilee:
```
    left: `Array([Object({"Plan": Object({"Async Capable": Bool(false), "Node Type": String("Result"), "Parallel Aware": Bool(false), "Plan Rows": Number(1), "Plan Width": Number(1), "Startup Cost": Number(0.0), "Total Cost": Number(0.01)})})])`,
    right: `Array([Object({"Plan": Object({"Node Type": String("Result"), "Parallel Aware": Bool(false), "Plan Rows": Number(1), "Plan Width": Number(1), "Startup Cost": Number(0.0), "Total Cost": Number(0.01)})})])`
```
The fields of Explain can vary based on PG version. You may want to match on the fields you know will be present in all versions. I will also accept `#[cfg(feature = "pg__")]`

workingjubilee:
Alright. This looks good. But, while I do not mean to keep this long, this isn't a bugfix or a critical break, and I need time to think on whether there needs to be any differences w/r/t this API before I can merge it, seeing as how you will be immediately depending on it (probably). So it looks like this will be 0.5.1

workingjubilee:
We can merge this but @eeeebbbbrrrr mentioned not feeling great about the general pattern our interfaces have taken here, so I want to doublecheck with him before doing so.

pgx-tests/src/tests/spi_tests.rs
```diff
@@ -124,6 +124,47 @@ mod tests {
...
+    #[pg_test]
+    fn test_spi_run_with_args() {
+        let i = 1 as i32;
+        let j = 2 as i64;
+
+        Spi::run_with_args(
+            "SELECT $1 + $2 = 3",
+            Some(vec![
+                (PgBuiltInOids::INT4OID.oid(), i.into_datum()),
+                (PgBuiltInOids::INT8OID.oid(), j.into_datum()),
```
workingjubilee, line 141:
```suggestion
                (PgBuiltInOids::INT4OID.oid(), Some(i.into()),
                (PgBuiltInOids::INT8OID.oid(), Some(j.into())),
```

pgx-tests/src/tests/spi_tests.rs
```diff
@@ -124,6 +124,47 @@ mod tests {
...
+    #[pg_test]
+    fn test_spi_explain_with_args() {
+        let i = 1 as i32;
+        let j = 2 as i64;
+
+        let result = Spi::explain_with_args(
+            "SELECT $1 + $2 = 3",
+            Some(vec![
+                (PgBuiltInOids::INT4OID.oid(), i.into_datum()),
+                (PgBuiltInOids::INT8OID.oid(), j.into_datum()),
```
workingjubilee, line 161:
```suggestion
                (PgBuiltInOids::INT4OID.oid(), Some(i.into()),
                (PgBuiltInOids::INT8OID.oid(), Some(j.into())),
```

workingjubilee (commented):
Whoops, sorry. Imports shifted around.

eeeebbbbrrrr (approved):
LGTM!

Thanks!

pgx-tests/src/tests/spi_tests.rs
```diff
@@ -124,6 +124,47 @@ mod tests {
...
+
+    #[pg_test]
+    fn test_spi_explain_with_args() {
+        let i = 1 as i32;
+        let j = 2 as i64;
+
+        let result = Spi::explain_with_args(
+            "SELECT $1 + $2 = 3",
+            Some(vec![
+                (PgBuiltInOids::INT4OID.oid(), Some(i.into()),
```
workingjubilee, line 160:
```suggestion
                (PgBuiltInOids::INT4OID.oid(), Some(i.into())),
```

pgx-tests/src/tests/spi_tests.rs
```diff
@@ -124,6 +124,47 @@ mod tests {
...
+
+    #[pg_test]
+    fn test_spi_run_with_args() {
+        let i = 1 as i32;
+        let j = 2 as i64;
+
+        Spi::run_with_args(
+            "SELECT $1 + $2 = 3",
+            Some(vec![
+                (PgBuiltInOids::INT4OID.oid(), Some(i.into()),
```
workingjubilee, line 140:
```suggestion
                (PgBuiltInOids::INT4OID.oid(), Some(i.into())),
```

workingjubilee (commented):
Amazing, I can't believe I missed that.

workingjubilee:
I will rustfmt this \_\_\_\_\_\_\_\_\_\_\_\_\_ myself.

PR #692, opened 2022-09-18
Specify logging config options when starting postgres from test framework

TimeToogo:
In postgres installations which have logging options which redirect output to files (`logging_collector = on`), the test framework will hang indefinitely waiting for `database system is ready to accept connections` line.

### DIFF
--- a/pgx-tests/src/framework.rs
+++ b/pgx-tests/src/framework.rs
@@ -498,6 +498,10 @@ fn start_pg(loglines: LogLines) -> eyre::Result<String> {
                 .expect("unable to determine test port")
                 .to_string(),
         )
+        .arg("-c")
+        .arg("log_destination=stderr")
+        .arg("-c")
+        .arg("logging_collector=off")
         .stdout(Stdio::inherit())
         .stderr(Stdio::piped());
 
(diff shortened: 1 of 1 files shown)


pgx-tests/src/framework.rs
```diff
@@ -498,6 +498,10 @@ fn start_pg(loglines: LogLines) -> eyre::Result<String> {
                 .expect("unable to determine test port")
                 .to_string(),
         )
+        .arg("-c")
+        .arg("log_destination=stderr")
+        .arg("-c")
+        .arg("logging_collector=off")
```
workingjubilee, line 504: These all have the lifetime `'static`, so they can be a single array of strings, and they are all set for the same reason, so they should be grouped as a single set so they can all have the same comment attached to them:
```suggestion
        // Redirecting logs to files can hang the test framework, override it
        .args([
            "-c",
            "log_destination=stderr",
            "-c",
            "logging_collector=off",
        ])
```

workingjubilee: I feel like this should be an actual single line, but that requires setting a rustfmt.toml

workingjubilee (approved):
Thank you! This makes sense.

PR #694, opened 2022-09-19
Block FATAL or PANIC in plrust via elog or ereport

workingjubilee:
Maybe this should be a different `cfg`, since really this is a plrust security level restriction, and postgrestd itself (for example) would be fine doing this itself internally.

Not sure.

### DIFF
--- a/pgx/src/log.rs
+++ b/pgx/src/log.rs
@@ -456,6 +456,19 @@ pub fn elog(level: PgLogLevel, message: &str) {
     use std::ffi::CString;
     use std::os::raw::c_char;
 
+    #[cfg(feature = "postgrestd")]
+    let level = {
+        use PgLogLevel::*;
+
+        // PL/Rust isn't permitted to abort the entire database.
+        match level {
+            level @ (DEBUG1 | DEBUG2 | DEBUG3 | DEBUG4 | DEBUG5 | LOG | LOG_SERVER_ONLY | INFO
+            | NOTICE | WARNING | ERROR) => level,
+            FATAL => ERROR,
+            PANIC => ERROR,
+        }
+    };
+
     unsafe {
         extern "C" {
             fn pgx_elog(level: i32, message: *const c_char);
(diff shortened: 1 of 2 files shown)


pgx/src/log.rs
```diff
@@ -456,6 +456,19 @@ pub fn elog(level: PgLogLevel, message: &str) {
     use std::ffi::CString;
     use std::os::raw::c_char;
 
+    #[cfg(feature = "postgrestd")]
```
johnrballard, line 459: I would think that the default behavior, trusted or not, for any extension would be don't kill the database abort the transaction.

> Jubliee: "some extensions manage multi-transaction state. These basically are the ones that do a lot more heavy-lifting, like ZomboDB or TimescaleDB or Citus or etc. and thus they have the "right" to terminate the database, in a sense, because they need to be able to pull the plug for the same reasons Postgres does (if everything is well and truly screwed, the only thing to do is to terminate the program and hope the user sorts it out)

johnrballard: So maybe its more of a "postgres safe" option rather than a "postgrestd" or maybe "postgrestd" means "postgres safe" and we define what that means ??

JohnHVancouver: I think you do want something other than `postgrestd` as the cfg option since I imagine there are use-cases for pgx extensions to do some of what would be considered concerning in PL/Rust.

For instance, I could see potential utility in wanting to run something like: `pgx_utils::createdb` from an extension but it wouldn't make sense to do so in PL/Rust. 

---
https://docs.rs/pgx-utils/0.5.0-beta.0/pgx_utils/fn.createdb.html

workingjubilee: That command shells out to `pg_config`, it's already blocked.

workingjubilee: I guess I think the main question is if we want to assume postgrestd should be safe-by-default in this way or if we want pgx extensions in general to be buildable with postgrestd. And we have at least some intention to generalize postgrestd a little beyond PL/Rust so the answer is yes. I will add a `"plrust"` feature for now.

PR #698, opened 2022-09-21
Use `unguarded_switch_to` with `postgrestd` as intended

workingjubilee:
This was missed during merging #615, it is necessary to get PL/Rust back to work, though the PL/Rust repo also needs some updates.


workingjubilee:
Error seems to be spurious, something to do with the network.

workingjubilee:
...continuous timing out. Well.

workingjubilee:
\_\_\_\_ it, we ball.

PR #699, opened 2022-09-21
`use pgx::prelude::*;`

workingjubilee:
As an alternative to exporting all identifiers from PGX, this PR introduces a module for importing all _commonly-used_, _distinctive_ identifiers from PGX. It also makes it much less likely that a random name introduced by a user will shadow any of PGX's own names by fully qualifying the names used in various macro expansions.


workingjubilee:
Assuming PL/Rust works with `use pgx::prelude::*;`, is this more like what you had in mind? @JohnHVancouver

Hoverbear (commented):
This means we can clean up a lot of the `pgx::*` stuff ya? Love it.

JohnHVancouver:
@workingjubilee  does that mean the user-code can no longer access all of pgx? Or just makes it easier/cleaner to detect

workingjubilee:
Currently the latter, but it makes the former humanly possible without completely breaking... uh, everything.

workingjubilee:
I mentioned that a lot of PGX's logic is deeply entangled, and this PR involves sorting... most of that out. Before this PR, PGX has what amounts to an _existential dependency_ on the glob-export that you do not enjoy. #615 made the situation significantly better, but it still remains a problem, even after this amount of refactoring.

workingjubilee:
This now makes all of `pgx-tests` work without the `use pgx::*;` glob. I am going to address `pgx-examples` in a later PR. Though it removes `heap_tuple` and `iter` from the public exports, these were added during the current 0.4.5 ~ 0.5.0 interstitial period, so hopefully no one is already depending on them.

workingjubilee:
Well... except PL/Rust.

PR #700, opened 2022-09-21
Adding Clone to Timestamp and Date types 

jsaied99:
Adding Clone to the traits of pgx::Timestamp.


workingjubilee (changes requested):
Thank you for catching this! Yeah, any simple values like this one should be at least Clone. Probably Copy, honestly. Can you make sure all of the `Date` and `Time*` types implement Clone?

workingjubilee (approved):
Excellent!

PR #702, opened 2022-09-22
Paint bikeshed for `Datum::cast_mut_ptr`

workingjubilee:
Before 0.5.0 goes gold, paint a bikeshed:
- With [`pointer::cast_{const, mut}` stabilizing in 1.65.0](https://github.com/rust-lang/rust/issues/92675),`Datum::ptr_cast` seems off. Instead, follow the ordering implied by `cast_mut`, `as_mut`, and `as_mut_ptr`: cast, mutability, then `_ptr` or rarely `_ref` for pointer-kind. This gives us `Datum::cast_mut_ptr::<T>` instead.
- `Datum::to_void` was used too rarely to warrant a unique fn, and `cast_mut_ptr` is more correct in 2 of those cases. Drop it.

Also, add some further verbiage about reference safety.

### DIFF
--- a/pgx-pg-sys/src/submodules/datum.rs
+++ b/pgx-pg-sys/src/submodules/datum.rs
@@ -33,8 +33,17 @@ struct DatumBlob {
 ///
 /// Only create Datums from non-pointers when you know you want to pass a value, as
 /// it is erroneous for `unsafe` code to dereference the address of "only a value" as a pointer.
-/// It is still a "safe" operation to create such pointers: validity is asserted by dereferencing.
-/// For all intents and purposes, however, Postgres counts as `unsafe` code that may be relying
+/// It is still a "safe" operation to create such pointers: validity is asserted by dereferencing,
+/// **or by creating a safe reference such as &T or &mut T**. Also be aware that the validity
+/// of Datum's Copy is premised on the same implicit issues with pointers being Copy:
+/// while any `&T` is live, other `*mut T` must not be used to write to that `&T`,
+/// and `&mut T` implies no other `*mut T` even exists outside an `&mut T`'s borrowing ancestry.
+/// It is thus of dubious soundness for Rust code to receive `*mut T`, create another `*mut T`,
+/// cast the first to `&mut T`, and then later try to use the second `*mut T` to write.
+/// It _is_ sound for Postgres itself to pass a copied pointer as a Datum to Rust code, then later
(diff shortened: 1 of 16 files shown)


pgx-pg-sys/src/submodules/datum.rs
```diff
@@ -48,18 +57,14 @@ impl Datum {
...
-
     /// True if the datum is equal to the null pointer.
     pub fn is_null(self) -> bool {
         self.0.is_null()
     }
 
     /// Assume the datum is a pointer and cast it to point to T.
-    pub fn ptr_cast<T>(self) -> *mut T {
+    /// It is recommended to explicitly use `datum.cast_mut_ptr::<T>()`.
+    pub fn cast_mut_ptr<T>(self) -> *mut T {
```
Hoverbear, line 67: I wonder if it would be useful to work like `Arc::clone()` and require the user to do `T::cast_mut_ptr(thing)`

Hoverbear:
Is there a way for users to get the old functionality of `Datum::to_void` in the new way? If so, can we make sure to document it in the changelog?

workingjubilee:
> Is there a way for users to get the old functionality of Datum::to_void in the new way? If so, can we make sure to document it in the changelog?

Yes. `datum.cast_mut_ptr::<core::ffi::c_void>()`

pgx-pg-sys/src/submodules/datum.rs
workingjubilee: I think it's too high an impedance in this case, as normal pointer functions don't require it, and there's a very specific reason Arc requires that for certain functions:

> ## [Deref behavior](https://doc.rust-lang.org/std/sync/struct.Arc.html#deref-behavior)
> Arc<T> automatically dereferences to T (via the [Deref](https://doc.rust-lang.org/std/ops/trait.Deref.html) trait), so you can call T’s methods on a value of type Arc<T>. To avoid name clashes with T’s methods, the methods of Arc<T> itself are associated functions, called using [fully qualified syntax](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#fully-qualified-syntax-for-disambiguation-calling-methods-with-the-same-name):

Note this is not actually required for `Arc::clone`:
> Arc<T>’s implementations of traits like Clone may also be called using fully qualified syntax. Some people prefer to use fully qualified syntax, while others prefer using method-call syntax.

PR #706, opened 2022-09-23
re`impl Drop for pgx::datum::Array<'a, T>`

workingjubilee:
While discussing the 0.5.0 release with @Hoverbear, the current situation regarding leaking memory stuck out as a sore spot. We discussed ways to make the `Drop` impl sound again, and I found this implementation after a bit of thought. It should actually be correct this time.

This effectively reverses #633 but is more correct than the prior state.

### DIFF
--- a/pgx/src/datum/array.rs
+++ b/pgx/src/datum/array.rs
@@ -18,9 +18,11 @@ use std::marker::PhantomData;
 use std::{mem, ptr, slice};
 
 pub struct Array<'a, T: FromDatum> {
-    _ptr: Option<NonNull<pg_sys::varlena>>,
+    ptr: Option<NonNull<pg_sys::varlena>>,
     raw: Option<RawArray>,
     nelems: usize,
+    // Remove this field if/when we figure out how to stop using pg_sys::deconstruct_array
+    datum_palloc: Option<NonNull<pg_sys::Datum>>,
     elem_slice: &'a [pg_sys::Datum],
     null_slice: NullKind<'a>,
     elem_layout: Option<Layout>,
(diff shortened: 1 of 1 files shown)


Hoverbear (approved):
I kinda wish I understood this better but it does read logically to me and not appear to have obvious issues!

pgx/src/datum/array.rs
```diff
@@ -69,6 +71,33 @@ impl<'a, T: FromDatum + serde::Serialize> serde::Serialize for Array<'a, T> {
...
+            // No conflict possible with ditching this.
+            unsafe { pg_sys::pfree(data.as_ptr().cast()) };
+
+            // Now we check for detoasting clones
+            let raw = raw.take().map(|r| r.into_ptr());
+            let ptr = ptr.take();
+            match (ptr, raw) {
+                // SAFETY: if pgx detoasted a clone of this varlena, pfree the clone
+                (Some(p), Some(r)) if r.cast() != p => unsafe { pg_sys::pfree(r.as_ptr().cast()) },
+                _ => (),
```
workingjubilee, line 95: Postgres also **may** allocate when you call `pg_detoast_datum`: the canonical way to check if it did is checking pointer equivalence. 🫠 So this `.take()`s both the varlena pointer from earlier and the RawArray from earlier, and checks that they are both `Some`, mapping RawArray to `*mut ArrayType`, casting from there to `*mut pg_sys::varlena` to compare, then casting again to `*mut c_void` to free it.

pgx/src/datum/array.rs
```diff
@@ -168,10 +195,23 @@ impl<'a, T: FromDatum> Array<'a, T> {
...
+        let pallocd_null_slice = unsafe { slice::from_raw_parts(nulls, nelems) };
+        for i in 0..nelems {
+            assert_eq!(null_slice.get(i).unwrap(), pallocd_null_slice[i]);
+        }
+
+        // Throw away the slice we made.
+        mem::drop(pallocd_null_slice);
+        // SAFETY: We made it, we can break it. Or Postgres can, at least.
+        unsafe { pg_sys::pfree(nulls.cast()) };
+
```
workingjubilee, line 209: This code continues to call `deconstruct_array` to maintain compatibility for current schemes for now, but it immediately frees the pallocated slice of `bool` as unnecessary after checking that our `NullKind`-wrapped `BitSlice` "understands" the null mapping in the same way.

pgx/src/datum/array.rs
```diff
@@ -69,6 +71,33 @@ impl<'a, T: FromDatum + serde::Serialize> serde::Serialize for Array<'a, T> {
...
+            raw,
+            datum_palloc: Some(data),
+            elem_slice,
+            ..
+        } = self
+        {
+            // It's just a slice, dropping it doesn't "do" anything, but out of an abundance of caution:
+            mem::drop(elem_slice);
+            // No conflict possible with ditching this.
+            unsafe { pg_sys::pfree(data.as_ptr().cast()) };
```
workingjubilee, line 87: While dropping the Array, check if the `datum_palloc` field is Some, and then if so, drop the `elem_slice` (it's a no-op, but it means there's no other pointers) and `pfree` the pointer to the allocation of Datums.

workingjubilee (commented):
Ah, so, the issue is that when `pg_sys::deconstruct_array` is called, Postgres builds an alternate "expanded" representation. This was selected by PGX because it is more... "user-friendly". However, in doing so, it allocates: once for the `[Datum]`s, once for the `[bool]`s.

Hoverbear:
Understand better now, thanks, :) I didn't get much chance to dig into arrays yet.

PR #707, opened 2022-09-25
Allow denying unsafe ops in PGX-using crates

workingjubilee:
The most basic element that attempts to reinforce PGX's safety guarantees are various bits of "guard" code combined with code that handles the PostgreSQL function call interface for programmers, rather than allowing them to do it themselves. Rust RFC 2585 allows using `unsafe { }` blocks inside `unsafe fn`, allowing adding tightly, narrowly-scoped information on safety, even inside an `unsafe fn`. However, this can't be done at the crate level if the crate is expanded by macros to include "sloppy" code: the lint fires after the macro expansion. These two facts are rather at odds: PGX should make it easier to write code for PostgreSQL, including `unsafe` code or heavily linted code, not harder.

To fix this, use `unsafe { }` throughout macro-expanded code with a combination of `#[allow(unused_unsafe)]` and `#[warn(unsafe_op_in_unsafe_fn)]` on various items to prevent these lints from improperly firing. Some experimentation was done with a "user" crate (plrust) to help validate this approach.

### DIFF
--- a/pgx-utils/src/sql_entity_graph/pg_extern/mod.rs
+++ b/pgx-utils/src/sql_entity_graph/pg_extern/mod.rs
@@ -488,7 +488,6 @@ impl PgExtern {
                         let mut funcctx: ::pgx::PgBox<pg_sys::FuncCallContext>;
                         let mut iterator_holder: ::pgx::PgBox<IteratorHolder<#retval_ty_resolved>>;
 
-                        // SAFETY: I know what I'm doing. Scout's honor!
                         unsafe {
                             if ::pgx::srf_is_first_call(#fcinfo_ident) {
                                 funcctx = pgx::srf_first_call_init(#fcinfo_ident);
@@ -590,7 +589,6 @@ impl PgExtern {
                         let mut funcctx: pgx::PgBox<pg_sys::FuncCallContext>;
                         let mut iterator_holder: pgx::PgBox<IteratorHolder<#retval_tys_tuple>>;
 
-                        // SAFETY: I know what I'm doing. Scout's honor!
                         unsafe {
                             if ::pgx::srf_is_first_call(#fcinfo_ident) {
                                 funcctx = ::pgx::srf_first_call_init(#fcinfo_ident);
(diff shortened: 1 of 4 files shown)


pgx-utils/src/sql_entity_graph/pg_extern/mod.rs
```diff
@@ -484,29 +488,33 @@ impl PgExtern {
                         let mut funcctx: ::pgx::PgBox<pg_sys::FuncCallContext>;
                         let mut iterator_holder: ::pgx::PgBox<IteratorHolder<#retval_ty_resolved>>;
 
-                        if ::pgx::srf_is_first_call(#fcinfo_ident) {
-                            funcctx = pgx::srf_first_call_init(#fcinfo_ident);
-                            funcctx.user_fctx = pgx::PgMemoryContexts::For(funcctx.multi_call_memory_ctx).palloc_struct::<IteratorHolder<#retval_ty_resolved>>() as *mut ::core::ffi::c_void;
+                        // SAFETY: I know what I'm doing. Scout's honor!
```
workingjubilee, line 491:
```suggestion
```
I should at some point either give a real reason or I should add equally-flippant `// SAFETY:` remarks to other parts of the PGX expansions so that all of the functions we make pass relevant clippy lints, but I am deciding I should joke around slightly less at the moment and not add `// SAFETY` remarks if I am not going to explain them (so we deliberately don't pass the relevant clippy lints).

pgx-utils/src/sql_entity_graph/pg_extern/mod.rs
```diff
@@ -578,31 +590,35 @@ impl PgExtern {
                         let mut funcctx: pgx::PgBox<pg_sys::FuncCallContext>;
                         let mut iterator_holder: pgx::PgBox<IteratorHolder<#retval_tys_tuple>>;
 
-                        if ::pgx::srf_is_first_call(#fcinfo_ident) {
-                            funcctx = ::pgx::srf_first_call_init(#fcinfo_ident);
-                            funcctx.user_fctx = pgx::PgMemoryContexts::For(funcctx.multi_call_memory_ctx).palloc_struct::<IteratorHolder<#retval_tys_tuple>>() as *mut ::core::ffi::c_void;
-                            funcctx.tuple_desc = pgx::PgMemoryContexts::For(funcctx.multi_call_memory_ctx).switch_to(|_| {
-                                let mut tupdesc: *mut pgx::pg_sys::TupleDescData = std::ptr::null_mut();
+                        // SAFETY: I know what I'm doing. Scout's honor!
```
workingjubilee, line 593:
```suggestion
```

as before.

workingjubilee:
:cowboy_hat_face:

PR #708, opened 2022-09-25
Allow returning values from a transaction in a bgworker

yrashk:
In some scenarios, it'd be beneficial to be able to return a result from a successfully executed transactio.

To allow this, `BackgroundWorker::transaction` has been extended to return a return value from the provided function. It's backwards compatible, as previously these functions were required to return the unit, and it `transaction` returned unit as well.

### DIFF
--- a/pgx/src/bgworkers.rs
+++ b/pgx/src/bgworkers.rs
@@ -182,21 +182,22 @@ impl BackgroundWorker {
     }
 
     /// Once connected to SPI via `connect_worker_to_spi()`, begin a transaction to
-    /// use the `pgx::Spi` interface.
-    pub fn transaction<F: FnOnce() + std::panic::UnwindSafe + std::panic::RefUnwindSafe>(
+    /// use the `pgx::Spi` interface. Returns the return value of the `F` function.
+    pub fn transaction<F: FnOnce() -> R + std::panic::UnwindSafe + std::panic::RefUnwindSafe, R>(
         transaction_body: F,
-    ) {
+    ) -> R {
         unsafe {
             assert!(!pg_sys::MyBgworkerEntry.is_null(), "BackgroundWorker associated functions can only be called from a registered background worker");
             pg_sys::SetCurrentStatementStartTimestamp();
             pg_sys::StartTransactionCommand();
             pg_sys::PushActiveSnapshot(pg_sys::GetTransactionSnapshot());
         }
-        pg_sys::guard(|| transaction_body());
+        let result = pg_sys::guard(|| transaction_body());
         unsafe {
             pg_sys::PopActiveSnapshot();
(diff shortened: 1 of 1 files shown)


pgx/src/bgworkers.rs
```diff
@@ -182,21 +182,22 @@ impl BackgroundWorker {
     }
 
     /// Once connected to SPI via `connect_worker_to_spi()`, begin a transaction to
-    /// use the `pgx::Spi` interface.
-    pub fn transaction<F: FnOnce() + std::panic::UnwindSafe + std::panic::RefUnwindSafe>(
+    /// use the `pgx::Spi` interface. Returns the return value of the `F` function.
+    pub fn transaction<F: FnOnce() -> R + std::panic::UnwindSafe + std::panic::RefUnwindSafe, R>(
```
workingjubilee, line 186: This actually is not always backward-compatible as back-compat would rely on the second type parameter being inferred or defaulted.

workingjubilee (changes requested):
We are very thin on tests for `BackgroundWorker::transaction` (I think we only have one example, and by default the examples are only checked for the case that they build at all, not necessarily full-blown tests?). And this claims backcompat. So I would like to see some tests added that
- demonstrate returning a value works (a simple assert is fine)
- demonstrate this is fully compatible, as in builds and actually executes and works, with preexisting code (if you can find an existing example of BackgroundWorker usage in the wild, that would be cool, if not, a fictional example is fine, but also maybe that's a good reason not to care much).

yrashk:
> We are very thin on tests for `BackgroundWorker::transaction` (I think we only have one example, and by default the examples are only checked for the case that they build at all, not necessarily full-blown tests?). And this claims backcompat. So I would like to see some tests added that
> 
> * demonstrate returning a value works (a simple assert is fine)
> * demonstrate this is fully compatible, as in builds and actually executes and works, with preexisting code (if you can find an existing example of BackgroundWorker usage in the wild, that would be cool, if not, a fictional example is fine, but also maybe that's a good reason not to care much).

The best way to add these tests would be to wait until #710 is merged so that there's bgworker test suite.

yrashk:
To add the tests, I've rebased this PR off #710 (don't merge it before #710!).

yrashk:
@workingjubilee resolved confilicts here

yrashk:
Anything that can be done here to get it finalized?

workingjubilee (approved):
Hm. Out of an abundance of caution I intend to merge this with the next necessarily-breaking PR (#830) so no one complains if the type inference actually does fail, as that means it will show up in pgx 0.6.0 instead. But this looks good.

BradyBonnette:
@workingjubilee #830 is merged in

PR #710, opened 2022-09-26
Implement dynamic background worker registration

yrashk:


workingjubilee (changes requested):
Oh cool. This looks interesting.
A test confirming this works would be fantastic!

yrashk:
@workingjubilee I've added the first draft of the test; please let me know what you think.

yrashk:
@workingjubilee I've updated this PR to be based on the latest `develop` and removed shmem startup hook as it doesn't make sense in the dynamic registration scenario.

I've considered changing the API for the startup hook to make it impossible to use dynamic registration if the hook is provided, but for now, I decided to keep it as is unless you want me to. Something like:

```rust
pub struct BackgroundWorkerBuilder<const Dynamic: bool = true> { .... }

pub fn enable_shmem_access(mut self: Self) -> Self { ... }
pub fn shmem_startup_hook(mut self: Self, unsafe extern "C" fn()) -> BackgroundWorkerBuilder<false> { ... }

impl BackgroundWorkerBuilder<true> {
  pub fn load_dynamic(self: Self) -> BackgroundWorkerHandle { ... }
}
```

pgx/src/bgworkers.rs
```diff
@@ -210,6 +212,97 @@ unsafe extern "C" fn worker_spi_sigterm(_signal_args: i32) {
...
+        match status {
+            BackgroundWorkerStatus::Started => Ok(pid),
+            _ => Err(status),
+        }
+    }
+
+    /// Block until the background worker exits, or postmaster dies. When the background worker exits, the return value is unit,
+    /// if postmaster dies it will return error with `BackgroundWorkerStatus::PostmasterDied` status
+    ///
+    /// Requires `BackgroundWorkerBuilder.bgw_notify_pid` to be set to `pg_sys::MyProcPid`
```
eeeebbbbrrrr, line 283: What happens if it is not?

pgx/src/bgworkers.rs
```diff
@@ -210,6 +212,97 @@ unsafe extern "C" fn worker_spi_sigterm(_signal_args: i32) {
...
+    pub fn wait_for_shutdown(self) -> Result<(), BackgroundWorkerStatus> {
+        TerminatingBackgroundWorkerHandle(self.0).wait_for_shutdown()
+    }
+}
+
+pub struct TerminatingBackgroundWorkerHandle(*mut pg_sys::BackgroundWorkerHandle);
```
eeeebbbbrrrr, line 289: A sentence or two as doccomments would be nice for this struct.

eeeebbbbrrrr (changes requested):

pgx/src/bgworkers.rs
yrashk: If it is not set, this will block forever.

yrashk:
@workingjubilee I've updated the documentation as requested.

pgx/src/bgworkers.rs
eeeebbbbrrrr: Is that ideal?  Should it instead return an `Err`?

blocking forever due to some kind of programmer mistake seems to violate POLA...

yrashk: @eeeebbbbrrrr, let me see what I can do. I agree it's not perfect.

yrashk:
@eeeebbbbrrrr I've replaced the blocking on `wait_*` with an error.

pgx/src/bgworkers.rs
```diff
@@ -210,10 +212,125 @@ unsafe extern "C" fn worker_spi_sigterm(_signal_args: i32) {
...
+
+/// Dynamic background worker status
+#[derive(Debug, Clone, Copy)]
+pub enum BackgroundWorkerStatus {
+    Started,
+    NotYetStarted,
+    Stopped,
+    PostmasterDied,
+    Untracked,
+    Unknown,
```
eeeebbbbrrrr, line 232: `::Unknown` seems to be dead code.

eeeebbbbrrrr (changes requested):

pgx/src/bgworkers.rs
yrashk: I can replace it with `BackgroundWorkerStatus` made non-exhaustive (just in case something new appears). Or I can just remove it. The only question is, what do we do if (for some reason?) BgwHandleStatus will have more variants in the future. I understand the possibility of this is slim.

eeeebbbbrrrr: We own this code, so fixing pgx is fine.  If it pushes compilation errors down to pgx users, that's a fine breaking change to ensure they're handling all the proper cases.

A blanket "Unknown", on the other hand, is kinda mean.  Especially since we do know, at least as of how Postgres works today.

pgx/src/bgworkers.rs
```diff
@@ -210,10 +212,125 @@ unsafe extern "C" fn worker_spi_sigterm(_signal_args: i32) {
...
+pub type Pid = pg_sys::pid_t;
+
+/// Dynamic background worker status
+#[derive(Debug, Clone, Copy)]
+pub enum BackgroundWorkerStatus {
+    Started,
+    NotYetStarted,
+    Stopped,
+    PostmasterDied,
+    Untracked,
```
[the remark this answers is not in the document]
yrashk: @eeeebbbbrrrr I've updated the docs, let me know if this works?

pgx/src/bgworkers.rs
yrashk: Agree. I pushed an update removing it.

pgx/src/bgworkers.rs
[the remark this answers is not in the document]
eeeebbbbrrrr: Yeah. Even better.

eeeebbbbrrrr:
I think this looks good now. Let’s have @workingjubilee confirm next week before we merge. 

Thanks for the work, @yrashk!

yrashk:
Anything else I can do here to make this ready to be merged?

thomcc (commented):
This looks fine to me, but there might be something I'm missing, so it's probably better if @workingjubilee takes a look

pgx/src/bgworkers.rs
```diff
@@ -210,10 +212,130 @@ unsafe extern "C" fn worker_spi_sigterm(_signal_args: i32) {
...
+    fn from(s: pg_sys::BgwHandleStatus) -> Self {
+        match s {
+            pg_sys::BgwHandleStatus_BGWH_STARTED => BackgroundWorkerStatus::Started,
+            pg_sys::BgwHandleStatus_BGWH_NOT_YET_STARTED => BackgroundWorkerStatus::NotYetStarted,
+            pg_sys::BgwHandleStatus_BGWH_STOPPED => BackgroundWorkerStatus::Stopped,
+            pg_sys::BgwHandleStatus_BGWH_POSTMASTER_DIED => BackgroundWorkerStatus::PostmasterDied,
+            _ => unreachable!(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
workingjubilee: I should note this is not a blocking concern, or even necessarily a design I would like more, it's just an implicit decision that is being made here that I felt I should note, and ideally we would document why we made the decision one way or another. I made it slightly differently when writing `impl TryFrom<c_int> for Spi{Ok,Error}`, but that case may be sufficiently different from this one to justify different decisions (likely most relative to probable evolution).

https://github.com/tcdi/pgx/blob/b92c60974c203e07e05cf1320581084be20a8e8a/pgx/src/spi.rs

pgx/src/bgworkers.rs
```diff
@@ -210,10 +212,130 @@ unsafe extern "C" fn worker_spi_sigterm(_signal_args: i32) {
...
+    notify_pid: pg_sys::pid_t,
+}
+
+impl TerminatingBackgroundWorkerHandle {
+    /// Block until the background worker exits, or postmaster dies. When the background worker exits, the return value is unit,
+    /// if postmaster dies it will return error with `BackgroundWorkerStatus::PostmasterDied` status
+    ///
+    /// Requires `BackgroundWorkerBuilder.bgw_notify_pid` to be set to `pg_sys::MyProcPid`, otherwise it'll
+    /// return [`BackgroundWorkerStatus::Untracked`] error
+    pub fn wait_for_shutdown(&self) -> Result<(), BackgroundWorkerStatus> {
```
workingjubilee, line 319: It is erroneous to do anything but reap the worker. We should enforce that.
```suggestion
    pub fn wait_for_shutdown(self) -> Result<(), BackgroundWorkerStatus> {
```

pgx/src/bgworkers.rs
```diff
@@ -210,10 +212,130 @@ unsafe extern "C" fn worker_spi_sigterm(_signal_args: i32) {
...
+        TerminatingBackgroundWorkerHandle { handle: self.handle, notify_pid: self.notify_pid }
+    }
+
+    /// Block until the postmaster has attempted to start the background worker,
+    /// or until the postmaster dies. If the background worker is running, the successful return value
+    /// will be the worker's PID. therwise, the return value will be an error with the worker's status.
+    ///
+    /// Requires `BackgroundWorkerBuilder.bgw_notify_pid` to be set to `pg_sys::MyProcPid`, otherwise it'll
+    /// return [`BackgroundWorkerStatus::Untracked`] error
+    pub fn wait_for_startup(&self) -> Result<Pid, BackgroundWorkerStatus> {
```
workingjubilee, line 280: This may not be mutating state that Rust owns, but it is still mutating state, so it's likely only correct to do this if Rust has `&mut` to the worker. If there is some kind of atomic multithreaded or multiprocess shenanigans this blocks, I am very eager to hear about what they are.
```suggestion
    pub fn wait_for_startup(&mut self) -> Result<Pid, BackgroundWorkerStatus> {
```

pgx/src/bgworkers.rs
```diff
@@ -210,10 +212,130 @@ unsafe extern "C" fn worker_spi_sigterm(_signal_args: i32) {
...
+            _ => Err(status),
+        }
+    }
+
+    /// Block until the background worker exits, or postmaster dies. When the background worker exits, the return value is unit,
+    /// if postmaster dies it will return error with `BackgroundWorkerStatus::PostmasterDied` status
+    ///
+    /// Requires `BackgroundWorkerBuilder.bgw_notify_pid` to be set to `pg_sys::MyProcPid`, otherwise it'll
+    /// return [`BackgroundWorkerStatus::Untracked`] error
+    pub fn wait_for_shutdown(self) -> Result<(), BackgroundWorkerStatus> {
```
workingjubilee, line 300: What happens when this is called on a worker that hasn't been started yet?

workingjubilee (changes requested):
Thank you, the presence of tests makes this much easier to review. I have a few nits that aren't terribly important, but I also have a number of design questions about how this is being handled and some edge-case-related concerns before this can be merged. However it looks like we're getting close.

workingjubilee:
A design question that came up in my mind that I don't know enough about BackgroundWorkers to remark on: Can a BackgroundWorker, if registered and loaded dynamically, move to any other lifecycle point? Or does it always progress in a fairly linear fashion (from "not yet started" to "active" to "ended")? In other words: is this enum truly an enum or is this trying to represent a state machine? We can enforce a much more rigid progression if so, if we want to. This would make the design more complex but would also make it harder to misuse the interface.

yrashk:
> A design question that came up in my mind that I don't know enough about BackgroundWorkers to remark on: Can a BackgroundWorker, if registered and loaded dynamically, move to any other lifecycle point? Or does it always progress in a fairly linear fashion (from "not yet started" to "active" to "ended")? In other words: is this enum truly an enum or is this trying to represent a state machine? We can enforce a much more rigid progression if so, if we want to. This would make the design more complex but would also make it harder to misuse the interface.

I can't say I am an expert in the intricacies of PostgreSQL. Reading this:

typedef enum [BgwHandleStatus](https://doxygen.postgresql.org/bgworker_8h.html#a656505f64167769c666e675242236f74)
 {
     [BGWH_STARTED](https://doxygen.postgresql.org/bgworker_8h.html#a656505f64167769c666e675242236f74a3e3fc34dc2a7279b36775c29f2e2b0eb),               /* worker is running */
     [BGWH_NOT_YET_STARTED](https://doxygen.postgresql.org/bgworker_8h.html#a656505f64167769c666e675242236f74a5ee6cb19589705fe694894780cb1c717),       /* worker hasn't been started yet */
     [BGWH_STOPPED](https://doxygen.postgresql.org/bgworker_8h.html#a656505f64167769c666e675242236f74a696916034f597a517c5f9cc8b447b728),               /* worker has exited */
     [BGWH_POSTMASTER_DIED](https://doxygen.postgresql.org/bgworker_8h.html#a656505f64167769c666e675242236f74a3a4ed23c0ba2d0e0be1ed57e36275371)        /* postmaster died; worker status unclear */
 } [BgwHandleStatus](https://doxygen.postgresql.org/bgworker_8h.html#a656505f64167769c666e675242236f74);

and the documentation suggests that it is linear. not yet started -> started -> stopped and unclear.

What kind of state machine did you have in mind that it might represent?

pgx-tests/src/tests/bgworker_tests.rs
```diff
@@ -0,0 +1,105 @@
+/*
+Portions Copyright 2019-2021 ZomboDB, LLC.
+Portions Copyright 2021-2022 Technology Concepts & Design, Inc. <support@tcdi.com>
+
+All rights reserved.
+
+Use of this source code is governed by the MIT license that can be found in the LICENSE file.
+*/
+use pgx::prelude::*;
+use pgx::{pg_guard, pg_sys, FromDatum, IntoDatum, PgOid, Spi};
```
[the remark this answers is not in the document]
yrashk: Addressed.

pgx/src/bgworkers.rs
```diff
@@ -210,10 +212,130 @@ unsafe extern "C" fn worker_spi_sigterm(_signal_args: i32) {
     pg_sys::SetLatch(pg_sys::MyLatch);
 }
 
+/// Dynamic background worker handle
+pub struct BackgroundWorkerHandle {
```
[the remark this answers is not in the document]
yrashk: Addressed

pgx/src/bgworkers.rs
yrashk: Reading the [code](https://doxygen.postgresql.org/bgworker_8h.html#a29892d414583bf41a83a3e4cf6a4f4c3) suggests that this is fine. It'll simply wait until it is stopped.

pgx/src/bgworkers.rs
yrashk: I am not confident this should be `&mut`. Reading the code suggests that it simply waits for an event to occur. Do you have any other findings?

Are you suggesting that any code that calls into PG code that uses `WaitLatch` should imply a mutating state?

pgx/src/bgworkers.rs
yrashk: I am not 100% sure this would be the correct handling. Waiting for an event that may not succeed shouldn't necessarily reap the handle, should it? I agree that this is rather a corner case, but let's say it returns, um, `Untracked` error status, and you still have the `pid` to work with (of course, you can save it beforehand, but I am just trying to explore the idea why do we need to reap it).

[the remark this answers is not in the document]
yrashk: Well, if it is, for example, `Untracked,` I do want to distinguish between success and failure here. We can perhaps extract narrower error types to avoid status/error confusion.

pgx/src/bgworkers.rs
```diff
@@ -210,10 +212,130 @@ unsafe extern "C" fn worker_spi_sigterm(_signal_args: i32) {
...
+        unsafe {
+            if self.notify_pid != pg_sys::MyProcPid {
+                return Err(BackgroundWorkerStatus::Untracked { notify_pid: self.notify_pid });
+            }
+        }
+        let status: BackgroundWorkerStatus =
+            unsafe { pg_sys::WaitForBackgroundWorkerShutdown(self.handle) }.into();
+        match status {
+            BackgroundWorkerStatus::Stopped => Ok(()),
+            _ => Err(status),
```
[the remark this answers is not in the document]
yrashk: See below. I am open to introducing narrow error types and removing this confusion of status and error.

pgx/src/bgworkers.rs
```diff
@@ -210,10 +212,130 @@ unsafe extern "C" fn worker_spi_sigterm(_signal_args: i32) {
...
+}
+
+/// Handle of a dynamic background worker that is being terminated with
+/// [`BackgroundWorkerHandle::terminate`]. Only allows waiting for shutdown.
+pub struct TerminatingBackgroundWorkerHandle {
+    handle: *mut pg_sys::BackgroundWorkerHandle,
+    notify_pid: pg_sys::pid_t,
+}
+
+impl TerminatingBackgroundWorkerHandle {
```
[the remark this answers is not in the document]
yrashk: Addressed

pgx/src/bgworkers.rs
```diff
@@ -210,10 +212,130 @@ unsafe extern "C" fn worker_spi_sigterm(_signal_args: i32) {
...
+    pub fn terminate(self) -> TerminatingBackgroundWorkerHandle {
+        unsafe {
+            pg_sys::TerminateBackgroundWorker(self.handle);
+        }
+        TerminatingBackgroundWorkerHandle { handle: self.handle, notify_pid: self.notify_pid }
+    }
+
+    /// Block until the postmaster has attempted to start the background worker,
+    /// or until the postmaster dies. If the background worker is running, the successful return value
+    /// will be the worker's PID. therwise, the return value will be an error with the worker's status.
```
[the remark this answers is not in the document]
yrashk: Fixed, thanks.

pgx/src/bgworkers.rs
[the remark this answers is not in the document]
workingjubilee: Ahh re-reviewing the commits I spotted the one where you removed the Unknown variant as useless. Like I said, I think it's fine to do whatever as long as we explain to our future selves why we thought it was a good idea at the time.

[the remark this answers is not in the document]
yrashk: I think the idea in the earlier discussion was that pgx effectively gate-keeps the changes to variants, and we know that variants won't change.

I am happy to revisit this whole thing as a follow-up PR, just want to unblock this one so that we have a smaller surface to iterate on.

pgx/src/bgworkers.rs
workingjubilee: Oh, I misunderstood something while I was reading and got the polling and the activation command confused.

>   * The caller *must* have set our PID as the worker's bgw_notify_pid,
  * else we will not be awoken promptly when the worker's state changes.

...hm, what happens if we don't, exactly?

yrashk: It'll wait for a very, very long time (until postmaster's death). Hence the check to ensure we're actually tracking the worker.

pgx/src/bgworkers.rs
workingjubilee: Hm. If it was `Untracked`, then what valid, safe actions remain with that `pid` that require it to remain attached to its source `TerminatingBackgroundWorker` type?

pgx-tests/src/tests/bgworker_tests.rs
```diff
@@ -0,0 +1,105 @@
...
+    use pgx::bgworkers::*;
+    use pgx::prelude::*;
+    use pgx::{pg_guard, pg_sys, IntoDatum};
+
+    #[pg_test]
+    fn test_dynamic_bgworker() {
+        let worker = BackgroundWorkerBuilder::new("dynamic_bgworker")
+            .set_library("pgx_tests")
+            .set_function("bgworker")
+            .set_argument(123i32.into_datum())
```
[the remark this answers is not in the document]
workingjubilee: Oh, I mostly mean that this basically goes `Option<Datum> -> NullableDatum -> Option<Datum>` (via Postgres) and we should make sure, as simple as it sounds, that if we say the `Datum` is "SQL NULL" (i.e. `None`) that it also becomes a `None` on the other side.

pgx/src/bgworkers.rs
yrashk: Going outside of PG's envelope, sending a signal to that PID?

workingjubilee: Hm. Isn't that fine though because that means we're doing another form of IPC then (and quite possibly `unsafe` code! or at least code that isn't in PGX) and we still have dispensed with the idea of it being some kind of BackgroundWorker, we're just trying to murder what we now see as a rogue process with whatever sky-lasers the OS has to offer us?

pgx/src/bgworkers.rs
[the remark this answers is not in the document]
workingjubilee: Yeah we can revisit this a bit later then.

pgx-tests/src/tests/bgworker_tests.rs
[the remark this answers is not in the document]
workingjubilee: ( I have recently become suspicious of these kinds of transitions after finding out that sometimes Postgres will pass values that are basically `nullptr_t, true` and so I am wondering if there are other hijinx in store. )

I am also kind of wondering what happens if a BackgroundWorker, er, fucks up massively, and if PGX can actually recover from that with our current API surface.

yrashk:
> I figured something like that would be the case. I was wondering if using a generic-enforced state machine (as in https://hoverbear.org/blog/rust-state-machine-pattern/ 3.3 and 3.4) might be better as a model (with the obvious clause that we can go from the "ending" back to the start again if we restart the worker, I think?) or if that would be too cumbersome. We kind of already have that half-way here.

My current feeling about this is that it is a) an overkill for now b) Postgres handles restarts itself so one can't really go "stopped->not_started->started" on their own.

I propose to break this out into a separate discussion. I am more than willing to commit some time to improving these small details in followups.

pgx/src/bgworkers.rs
yrashk: `wait_for_shutdown` isn't a request for murder; it simply waits until the worker dies. Very Zen.

workingjubilee: Hm, right, but interacting with the PID "directly" isn't that.

workingjubilee: I would somewhat prefer to optimistically assume that consuming is okay (especially since a consuming-oriented approach is what's used by the rest of this new API surface) and revisit this assumption if we find a use case that is blocked by this.

workingjubilee:
Hm. I think it's fine to defer some of this to followups and I think we'll need a little time experimenting with this before we can be sure of what the right API is, anyways, so a major break on it is inevitable, but in the general case we should try to tend towards stricter (within reason) APIs -> less strict APIs. It's harder to take back relaxations of the rules when you've already smoothed a given path for something.

yrashk:
I'm 100% on this with you. Good APIs take time and stricter are better, but
only when not taken to an extreme. It should still be usable :)

I'll be happy to help improving this over time.

Do we have anything blocking this PR that I can address before merging as a
first version?

yrashk:
@workingjubilee I've addressed the issue with `TerminatingDynamicBackgroundWorker::wait_for_shutdown` signature as we've discussed.

workingjubilee (approved):
Then away we go!

yrashk:
Woohoo! ❤️

PR #712, opened 2022-09-26
Fixing release scripts 0.5.0 beta.1

workingjubilee:
So these don't quite work yet because when I run them

```
error: no matching package found
searched package name: `pgx`
prerelease package needs to be specified explicitly
pgx = { version = "0.5.0-beta.1" }
location searched: ~/tcdi/pgx/pgx
required by package `pgx-tests v0.5.0-beta.1 (~/tcdi/pgx/pgx-tests)`
```

Any suggestions?

### DIFF
--- a/update-versions.sh
+++ b/update-versions.sh
@@ -62,13 +62,16 @@ SEMVER_REGEX="(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*
 
 for cargo_toml in ${CARGO_TOMLS_TO_SED[@]}; do
     for dependency in ${DEPENDENCIES_TO_UPDATE[@]}; do
-        rg --passthru --no-line-number --multiline --multiline-dotall "(?P<prefix>^${dependency}.*\")(?P<pin>=?)${SEMVER_REGEX}(?P<postfix>\".*$)" -r "\${prefix}=${VERSION}\${postfix}" ${cargo_toml} > ${cargo_toml}.tmp || true
+        rg --passthru --no-line-number \
+        "(?P<prefix>^${dependency}.*\")(?P<pin>=?)${SEMVER_REGEX}(?P<postfix>\".*$)" \
+        -r "\${prefix}=${VERSION}\${postfix}" \
+        ${cargo_toml} > ${cargo_toml}.tmp || true
         mv ${cargo_toml}.tmp ${cargo_toml}
     done
 done
 
 for cargo_toml in ${CARGO_TOMLS_TO_BUMP[@]}; do
-    rg --passthru -N "(?P<prefix>^version = \")${SEMVER_REGEX}(?P<postfix>\"$)" -r "\${prefix}${VERSION}\${postfix}" ${cargo_toml} > ${cargo_toml}.tmp || true
+    rg --passthru --no-line-number "(?P<prefix>^version = \")${SEMVER_REGEX}(?P<postfix>\"$)" -r "\${prefix}${VERSION}\${postfix}" ${cargo_toml} > ${cargo_toml}.tmp || true
     mv ${cargo_toml}.tmp ${cargo_toml}
 done
 
--- a/update-versions.sh
+++ b/update-versions.sh
@@ -21,44 +21,50 @@ set -ex
 HEAD=$(git rev-parse HEAD)
 VERSION=$1
 
+# ordered in a topological fashion starting from the workspace root Cargo.toml
+# this isn't always necessary, but it's nice to use a mostly-consistent ordering
 CARGO_TOMLS_TO_BUMP=(
     ./Cargo.toml
-    ./pgx/Cargo.toml
+    ./pgx-pg-config/Cargo.toml
     ./pgx-utils/Cargo.toml
-    ./pgx-macros/Cargo.toml
-    ./pgx-tests/Cargo.toml
     ./cargo-pgx/Cargo.toml
+    ./pgx-macros/Cargo.toml
     ./pgx-pg-sys/Cargo.toml
+    ./pgx/Cargo.toml
+    ./pgx-tests/Cargo.toml
 )
 
 CARGO_TOMLS_TO_SED=(
+    ./Cargo.toml
+    ./pgx-pg-config/Cargo.toml
+    ./pgx-utils/Cargo.toml
+    ./cargo-pgx/Cargo.toml
+    ./pgx-macros/Cargo.toml
+    ./pgx-pg-sys/Cargo.toml
+    ./pgx/Cargo.toml
+    ./pgx-tests/Cargo.toml
+    ./pgx-examples/*/Cargo.toml
     ./cargo-pgx/src/templates/cargo_toml
     ./nix/templates/default/Cargo.toml
-    ./pgx/Cargo.toml
-    ./pgx-utils/Cargo.toml
-    ./pgx-macros/Cargo.toml
-    ./pgx-tests/Cargo.toml
-    ./cargo-pgx/Cargo.toml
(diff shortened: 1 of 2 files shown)


update-versions.sh
```diff
@@ -21,50 +21,57 @@ set -ex
 HEAD=$(git rev-parse HEAD)
 VERSION=$1
 
+# ordered in a topological fashion starting from the workspace root Cargo.toml
+# this isn't always necessary, but it's nice to use a mostly-consistent ordering
 CARGO_TOMLS_TO_BUMP=(
     ./Cargo.toml
-    ./pgx/Cargo.toml
+    ./pgx-pg-config/Cargo.toml
```
BradyBonnette, line 28: :( Sorry about that.

update-versions.sh
```diff
@@ -21,50 +21,57 @@ set -ex
...
+    ./pgx/Cargo.toml
+    ./pgx-tests/Cargo.toml
 )
 
 CARGO_TOMLS_TO_SED=(
-    ./cargo-pgx/src/templates/cargo_toml
-    ./nix/templates/default/Cargo.toml
-    ./pgx/Cargo.toml
+    ./Cargo.toml
+    ./pgx-pg-config/Cargo.toml
```
BradyBonnette, line 39: :( Sorry!

BradyBonnette (approved):
Looks good! I've never ran the script before but I do remember there needs to be an order of crate publication.

update-versions.sh
workingjubilee: It's okay, really! I shouldn't have missed it in review.

workingjubilee:
hella nice!

update-versions.sh
```diff
@@ -18,55 +19,89 @@ fi
...
+    # this isn't always necessary, but it's nice to use a mostly-consistent ordering
+    CARGO_TOMLS_TO_BUMP=(
+        ./Cargo.toml
+    )
+
+    CARGO_TOMLS_TO_SED=(
+        ./Cargo.toml
+        ./pgx-examples/*/Cargo.toml
+        ./cargo-pgx/src/templates/cargo_toml
+        ./nix/templates/default/Cargo.toml
```
workingjubilee, line 72:
```suggestion
        ./cargo-pgx/src/templates/cargo_toml
        ./nix/templates/default/Cargo.toml
```

Maybe we should invert `./nix/templates` into `./templates/nix/` and have `./cargo-pgx/src/templates` be its own template as well, and include them properly in the workspace and all that. Future PR, though, since now the release script at least works.

BradyBonnette: Very possible, though I am not too familiar with nix and any repercussions of moving the nix template file somewhere else.

update-versions.sh
```diff
@@ -18,55 +19,89 @@ fi
...
+    # Ignore all packages in pgx-examples/
+    for file in $(find pgx-examples/ -type f -name "Cargo.toml"); do
+        EXCLUDE_PACKAGES+=($(cat $file | rg --multiline '\[package\](.*\n)(?:[^\[]*\n)*name\s?=\s?"(?P<name>[-_a-z]*)"(.*\n)*' -r "\${name}"))
+    done
+
+    echo "Excluding the following packages:"
+    for p in "${EXCLUDE_PACKAGES[@]}"; do
+        echo " * $p"
+    done
+    echo ""
```
workingjubilee, line 53: Don't need to loop the list:
```suggestion
    echo "Excluding the following packages:"
    echo "${EXCLUDE_PACKAGES[@]}"
```

update-versions.sh
```diff
@@ -18,55 +19,89 @@ fi
...
+
+    cargo set-version $(echo "${EXCLUDE_PACKAGES[@]/#/--exclude }") --workspace $version
```
workingjubilee, line 55: This is trying to substitute an array into a bunch of strings and then feed it through echo to flatten it and then reexpand it by embedding it as a command evaluation and... it would be simpler to allow the shell to do the thing it does.
```suggestion
    # shellcheck disable=2068 # allow the shell to split --exclude from each EXCLUDE_PACKAGES value
    cargo set-version ${EXCLUDE_PACKAGES[@]/#/--exclude } --workspace "$version"
```

update-versions.sh
```diff
@@ -18,55 +19,89 @@ fi
...
+            -r "\${prefix}=${VERSION}\${postfix}" \
+            ${cargo_toml} > ${cargo_toml}.tmp || true
+            mv ${cargo_toml}.tmp ${cargo_toml}
+        done
+    done
+
+    for cargo_toml in ${CARGO_TOMLS_TO_BUMP[@]}; do
+        rg --passthru --no-line-number "(?P<prefix>^version = \")${SEMVER_REGEX}(?P<postfix>\"$)" -r "\${prefix}${VERSION}\${postfix}" ${cargo_toml} > ${cargo_toml}.tmp || true
         mv ${cargo_toml}.tmp ${cargo_toml}
     done
```
workingjubilee, line 100:
```suggestion
    for cargo_toml in ${CARGO_TOMLS_TO_BUMP[@]}; do
        rg --passthru --no-line-number "(?P<prefix>^version = \")${SEMVER_REGEX}(?P<postfix>\"$)" -r "\${prefix}${VERSION}\${postfix}" ${cargo_toml} > ${cargo_toml}.tmp || true
        mv ${cargo_toml}.tmp ${cargo_toml}
    done
```

Dead code.

update-versions.sh
```diff
@@ -18,55 +19,89 @@ fi
...
+        "pgx-tests"
+    )
+
+    SEMVER_REGEX="(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?"
+
+    for cargo_toml in ${CARGO_TOMLS_TO_SED[@]}; do
+        for dependency in ${DEPENDENCIES_TO_UPDATE[@]}; do
+            rg --passthru --no-line-number \
+            "(?P<prefix>^${dependency}.*\")(?P<pin>=?)${SEMVER_REGEX}(?P<postfix>\".*$)" \
+            -r "\${prefix}=${VERSION}\${postfix}" \
```
workingjubilee, line 91: otherwise this reads from the global

```suggestion
            -r "\${prefix}=${version}\${postfix}" \
```

update-versions.sh
```diff
@@ -18,55 +19,89 @@ fi
...
+# This is a legacy holdover for updating extra toml files throughout various crates
+function update_extras() {
+    local version=$1
+
+    # ordered in a topological fashion starting from the workspace root Cargo.toml
+    # this isn't always necessary, but it's nice to use a mostly-consistent ordering
+    CARGO_TOMLS_TO_BUMP=(
+        ./Cargo.toml
+    )
+
```
workingjubilee, line 67: Not doing anything.

```suggestion
```

update-versions.sh
```diff
@@ -18,55 +19,89 @@ fi
...
+    echo "Command `rg` (ripgrep) was not found. Please install it and try again."
+    exit 1
+fi
+
+if ! cargo set-version --help &> /dev/null; then
+    echo "Cargo extension `cargo-edit` is not installed. Please install it by running: cargo install cargo-edit"
+    exit 1
+fi
+
 HEAD=$(git rev-parse HEAD)
```
workingjubilee, line 32: Dead code since a while ago, I think.
```suggestion
```

PR #714, opened 2022-09-27
Implement SPI_connect_ext and related functionality

yrashk:
This allows pgx users to use SPI in a non-atomic fashion.

### DIFF
--- a/pgx/src/spi.rs
+++ b/pgx/src/spi.rs
@@ -245,6 +245,21 @@ impl Spi {
     >(
         f: F,
     ) -> Option<R> {
+        Self::connect_ext(false, f)
+    }
+
+    /// execute SPI commands via the provided `SpiClient` and return a value from SPI which is
+    /// automatically copied into the `CurrentMemoryContext` at the time of this function call
+    ///
+    /// Setting `nonatomic` to `true` allows transaction control calls (SPI_commit, SPI_rollback).
+    /// Otherwise, calling those functions will result in an immediate error.
+    pub fn connect_ext<
+        R: FromDatum + IntoDatum,
+        F: FnOnce(SpiClient) -> std::result::Result<Option<R>, SpiError>,
+    >(
+        nonatomic: bool,
+        f: F,
+    ) -> Option<R> {
         let outer_memory_context =
             PgMemoryContexts::For(PgMemoryContexts::CurrentMemoryContext.value());
 
@@ -252,9 +267,9 @@ impl Spi {
         struct SpiConnection;
         impl SpiConnection {
             /// Connect to Postgres' SPI system
-            fn connect() -> Self {
+            fn connect_ext(flags: u32) -> Self {
                 // connect to SPI
-                Spi::check_status(unsafe { pg_sys::SPI_connect() });
+                Spi::check_status(unsafe { pg_sys::SPI_connect_ext(flags as _) });
                 SpiConnection
             }
         }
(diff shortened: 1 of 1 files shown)


workingjubilee:
I'm expecting the tests will break on pg10.

pgx/src/spi.rs
```diff
@@ -412,6 +432,40 @@ impl SpiClient {
...
+    ///
+    /// Can only be executed if the SPI connection has been set as `nonatomic` in the call to `Spi::connect`.
+    pub fn rollback(&self) {
+        unsafe { pg_sys::SPI_rollback() }
+    }
+
+    /// Rolls back the current transaction same as `rollback`, but the new transaction is started with the same
+    /// transaction characteristics as the just finished one, like with the SQL command ROLLBACK AND CHAIN.
+    pub fn rollback_and_chain(self: Self) {
+        unsafe { pg_sys::SPI_rollback_and_chain() }
```
workingjubilee, line 467: These need `// SAFETY` explanations.

pgx/src/spi.rs
```diff
@@ -244,6 +244,21 @@ impl Spi {
...
         f: F,
+    ) -> Option<R> {
+        Self::connect_ext(false, f)
+    }
+
+    /// execute SPI commands via the provided `SpiClient` and return a value from SPI which is
+    /// automatically copied into the `CurrentMemoryContext` at the time of this function call
+    ///
+    /// Setting `nonatomic` to `true` allows transaction control calls (SPI_commit, SPI_rollback).
+    /// Otherwise, calling those functions will result in an immediate error.
```
workingjubilee, line 255: There's no reason to rely on global mutable state for this if we are going to expose a safe wrapper: these kinds of predicates should be enforced at the type level.

pgx/src/spi.rs
```diff
@@ -252,9 +267,9 @@ impl Spi {
         struct SpiConnection;
         impl SpiConnection {
             /// Connect to Postgres' SPI system
-            fn connect() -> Self {
+            fn connect_ext(flags: u32) -> Self {
                 // connect to SPI
-                Spi::check_status(unsafe { pg_sys::SPI_connect() });
+                Spi::check_status(unsafe { pg_sys::SPI_connect_ext(flags as _) });
```
workingjubilee, line 272: I do not believe this function exists in Postgres 10, this requires `#[cfg]`.

pgx/src/spi.rs
```diff
@@ -268,7 +283,12 @@ impl Spi {
         }
 
         // connect to SPI
-        let _connection = SpiConnection::connect();
+        let flags = if nonatomic {
+            pg_sys::SPI_OPT_NONATOMIC
```
workingjubilee, line 287: I do not believe this exists in Postgres 10. This requires `#[cfg]`.

workingjubilee (changes requested):
I don't really want all of these to be safely wrapped like this at the moment, see this conversation:
- https://github.com/tcdi/pgx/issues/696

However, if the implementation was sufficiently high quality (i.e. enforced tx/rollback correctness by lifting that constraint into the type level), I would consider it.

I would also accept an implementation of the resolution I was inclined to try (namely, saying "we may or may not use `SPI_connect_ext` at our option" and setting flags implicitly on an experimental basis, so as to see how that pans out for users).

yrashk:
Thank you for the feedback; I will address your commentary in the coming days. This is a lower-priority development for me at the moment, but I'll see what I can do.

workingjubilee:
Another solution to the issues regarding Postgres 10 may just be waiting for a bit: we're going to be dropping Postgres 10 sometime soon.

yrashk:
I'll retire this for now, as it has become stale. May reopen later. WIth sub-transactions available through pgx-contrib-spiext, a bit less of a pressing concern.

PR #716, opened 2022-09-28
Fix invalid SQL generated by `#[pg_extern(create_or_replace)]`

JamesGuthrie:
When using the `create_or_replace` attribute, the generated SQL would contain CREATE OR REPLACE twice, e.g.:

```
CREATE OR REPLACE FUNCTION "my_function"() RETURNS INT /* i32 */
CREATE OR REPLACE STRICT
LANGUAGE c /* Rust */
AS 'MODULE_PATHNAME', 'my_function_wrapper';
```

This is because the `create_or_replace` attribute is not removed from the array of attributes before being transformed to SQL.

### DIFF
--- a/pgx-utils/src/sql_entity_graph/pg_extern/entity/mod.rs
+++ b/pgx-utils/src/sql_entity_graph/pg_extern/entity/mod.rs
@@ -438,6 +438,7 @@ impl ToSql for PgExternEntity {
             } else {
                 let mut retval = extern_attrs
                     .iter()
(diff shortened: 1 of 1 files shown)


willmurnane:
Ideally, it'd be nice to add a test that exercises this functionality, so that this problem doesn't crop up again in the future. Otherwise looks like a good fix.

workingjubilee:
:sweat_smile:

workingjubilee (approved):
Whoops! Thanks for catching that.

PR #722, opened 2022-09-29
Fix full_path/module_path mixup.

willmurnane:
This fixes the `OPERATOR FAMILY` declared before `OPERATOR <` problem I discovered. Here's a screenshot of `cargo pgx schema --dot out.dot && dot -T svg out.dot > out.svg` before:
Screen Shot 2022-09-29 at 1 11 50 PM
and after:
Screen Shot 2022-09-29 at 1 12 06 PM

### DIFF
--- a/pgx-utils/src/sql_entity_graph/pgx_sql.rs
+++ b/pgx-utils/src/sql_entity_graph/pgx_sql.rs
@@ -1171,9 +1171,16 @@ fn connect_ords(
             enums,
         );
 
+        // Make PostgresOrdEntities (which will be translated into `CREATE OPERATOR CLASS` statements) depend
+        // on the operators which they will reference. For example, a pgx-defined Postgres type `parakeet`
+        // which has `#[derive(PostgresOrd)]` will emit a `parakeet_btree_ops` operator class, which references
+        // a definition of a < operator (among others) on the `parakeet` type. This code should ensure that the
+        // < operator (along with all the others) is emitted before the `OPERATOR CLASS` itself.
+
         for (extern_item, &extern_index) in externs {
-            let fn_matches =
-                |fn_name| item.module_path == extern_item.full_path && extern_item.name == fn_name;
+            let fn_matches = |fn_name| {
+                item.module_path == extern_item.module_path && extern_item.name == fn_name
+            };
             let cmp_fn_matches = fn_matches(item.cmp_fn_name());
             let lt_fn_matches = fn_matches(item.lt_fn_name());
             let lte_fn_matches = fn_matches(item.le_fn_name());


pgx-utils/src/sql_entity_graph/pgx_sql.rs
```diff
@@ -1172,20 +1172,14 @@ fn connect_ords(
...
-            let gt_fn_matches = fn_matches(item.gt_fn_name());
-            let gte_fn_matches = fn_matches(item.ge_fn_name());
-            if cmp_fn_matches
-                || lt_fn_matches
-                || lte_fn_matches
-                || eq_fn_matches
-                || gt_fn_matches
-                || gte_fn_matches
+            let fn_matches = |fn_name| extern_item.name == fn_name;
+            if item.module_path == extern_item.module_path
```
workingjubilee, line 1176: Could use a comment reminding future editors of this or something like it:
```suggestion
            // It is important that this closure and resulting conditions
            // correctly order Visibility to depend on all visibility_{ops}
            let fn_matches = |fn_name| extern_item.name == fn_name;
            if item.module_path == extern_item.module_path
```

willmurnane: Well, the problem wasn’t that this code was “incorrectly ordering” things, but rather that it was comparing `module_path` to `full_path`. So the `_gt` function, for example, wouldn’t be detected as being related to the Ord, no edge between them would be generated in the graph, and they might be generated in the wrong order.

workingjubilee: Hm. So how would you prefer to explain the relevant context in positive rather than merely corrective/admonishing form?

Hoverbear (commented):
Ahh I think this might have been a regression that we fixed recently but this PR changed the area. :(

Fix looks good

pgx-utils/src/sql_entity_graph/pgx_sql.rs
willmurnane: I added a test, rather than an explanation. It's kind of an annoying test to value proposition (adds 120 lines of code to test a small interaction) but it's cheap to run and prevents a regression. I could add a similar check for hash functions like what #637 fixed if this looks like a good thing to have.

workingjubilee: If adding another check will make the correctness:lines_of_test_code ratio better, yes.

willmurnane: I modified the test to cover all of the extant comparison functions (not just the `_gt` one), and added the check for hash that I mentioned earlier. This required me to re-define some members as `String` rather than `&'static str`, which expands the footprint a bit.

If this is getting too complicated, I can split the actual meat of the fix from the additional test so we can bikeshed that part sufficiently :)

workingjubilee: Mmm, yes I do think we should extend that discussion into another PR. ^^;

pgx-utils/src/sql_entity_graph/pgx_sql.rs
```diff
@@ -1171,9 +1171,16 @@ fn connect_ords(
             enums,
         );
 
+        // Make PostgresOrdEntities (which will be translated into `CREATE OPERATOR CLASS` statements) depend
+        // on the operators which they will reference. For example, a pgx-defined Postgres type `parakeet`
+        // which has `#[derive(PostgresOrd)]` will emit a `parakeet_btree_ops` operator class, which references
+        // a definition of a < operator (among others) on the `parakeet` type. This code should ensure that the
+        // < operator (along with all the others) is emitted before the `OPERATOR CLASS` itself.
+
```
workingjubilee, line 1179: 💖

workingjubilee (approved):

PR #724, opened 2022-09-30
Allow type structs to be filtered before serialization

mpalmer:
In (very) rare circumstances, structs that can be successfully CBOR serialized shouldn't be serialized with everything that was sent in the query, but should instead be stored with only a subset of that data. Instead of having to write an entire custom inoutfuncs set, this patch allows a type to declare that it needs to "filter" the struct before it's turned into a `Datum`, and then implement a trait to do so.

### DIFF
--- a/pgx-macros/src/lib.rs
+++ b/pgx-macros/src/lib.rs
@@ -693,11 +693,13 @@ Optionally accepts the following attributes:
 
 * `inoutfuncs(some_in_fn, some_out_fn)`: Define custom in/out functions for the type.
 * `pgvarlena_inoutfuncs(some_in_fn, some_out_fn)`: Define custom in/out functions for the `PgVarlena` of this type.
+* `filter_datum`: Mark that the type requires filtration before being serialized into a Datum.  You
+  must then `impl IntoDatumFilter for <type>` yourself.
 * `sql`: Same arguments as [`#[pgx(sql = ..)]`](macro@pgx).
 */
 #[proc_macro_derive(
     PostgresType,
-    attributes(inoutfuncs, pgvarlena_inoutfuncs, requires, pgx)
+    attributes(inoutfuncs, pgvarlena_inoutfuncs, filter_datum, requires, pgx)
 )]
 pub fn postgres_type(input: TokenStream) -> TokenStream {
     let ast = parse_macro_input!(input as syn::DeriveInput);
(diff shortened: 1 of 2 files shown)


pgx-macros/src/lib.rs
```diff
@@ -735,6 +737,14 @@ fn impl_postgres_type(ast: DeriveInput) -> proc_macro2::TokenStream {
         impl #generics pgx::PostgresType for #name #generics { }
     });
 
+    // provide the blanket datum filter implementation unless the type
+    // needs custom filtration
+    if !args.contains(&PostgresTypeAttribute::FilterDatum) {
+        stream.extend(quote! {
+            impl IntoDatumFilter for #name {}
```
workingjubilee, line 744:
```suggestion
            impl #generics pgx::IntoDatumFilter for #name #generics {}
```

pgx-macros/src/lib.rs
```diff
@@ -890,6 +900,7 @@ fn impl_guc_enum(ast: DeriveInput) -> proc_macro2::TokenStream {
 enum PostgresTypeAttribute {
     InOutFuncs,
     PgVarlenaInOutFuncs,
+    FilterDatum,
```
workingjubilee, line 903: Is there a reason this would only apply as a modified to the Default behavior and not possibly a modifier on all InOutFuncs?

mpalmer: My thinking was that, if you were implementing your own InOutFuncs, you'd be able to do the filtering in that codeyourself.  However, I have very limited experience with using InOutFuncs, so if that's not the case, I can see about adding filtering in all cases.

workingjubilee:
I'm not sure about this patch. I'm not sure that the `impl InOutFuncs` is that laborious, or that if it is, that this alternative doesn't do _enough_ to be convenient. It seems like a half-way measure, as it's not as good as just elaborating the original `#[derive(PostgresType)]` slightly, and it's not as controlled as just implementing `trait InOutFuncs` yourself. It seems like it would be inflexible in the face of API changes, also. I think this kind of convenience should look more like [`#[serde::skip]`](https://serde.rs/field-attrs.html#skip), which is much more convenient, self-explanatory, and also says less about the underlying implementation of CBOR/JSON translations.

mpalmer:
I understand your reservations -- I wasn't 100% enthused by my approach, either, but it was the best I could come up with at the time.  I rather like the idea of a `#[serde(skip)]` type tag, although my Rust-fu isn't up to implementing such a thing at the moment.  Using `[serde(skip)]` itself unfortunately isn't feasible, because (a) I wanted to filter conditionally on other parts of the structure, and (b) I couldn't see a way to say "only skip serializing this when going into CBOR".

In any event, as we're both a bit "meh" about this PR, I'll close it off for now, as I do have an alternate approach that I can use for the specific case I'm implementing at present, and I'll open up something else a bit less kludgy if/when I need it in the future.

Thanks for your feedback!

workingjubilee:
Understandable, and thank you for offering! I will look forward to another go-around, and until then feel free to ask any questions you may have about the library or implementation woes.

PR #725, opened 2022-09-30
Let bindgen know which traits Datum implements

JamesGuthrie:
In #545 we replaced `Datum`'s autogenerated bindings with an opaque pointer. Because bindgen was not generating the type, it could not know which traits `Datum` implements. This changed how containers of `Datum` were emitted. In particular containers of `Datum` no longer derived `Debug`, `Copy`, or `Clone`. This also affected how unions containing these containers were emitted: bindgen only uses Rust unions if all members are `Copy`, falling back to a struct of `__BindgenUnionField<_>` fields otherwise.

This change tells bindgen which traits our `Datum` implementation implements, fixing all of the above issues.

### DIFF
--- a/pgx-pg-sys/build.rs
+++ b/pgx-pg-sys/build.rs
@@ -9,7 +9,7 @@ Use of this source code is governed by the MIT license that can be found in the 
 
 extern crate build_deps;
 
-use bindgen::callbacks::MacroParsingBehavior;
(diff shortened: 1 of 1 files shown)


JamesGuthrie:
FYI apparently bindgen only supports calling `.parse_callbacks()` once.

workingjubilee (approved):
Thank you!

PR #729, opened 2022-10-03
Postgres 15 support

Smittyvb:
Adds support for the PostgreSQL 15 release candidate. If the `PGX_USE_PG15_RC1` environment variable is set, the release candidate is used, otherwise pgx tries (and currently fails) to use the final release.

Main changes:
* Existing extension crates need to add a `pg15` feature to indicate compatibility with PostgreSQL 15. `cargo pgx new` automatically adds the feature for new extensions.
* `pg_sys::String` now exists, which conflicts with the std `String` when importing `pg_sys::*`
* There is now a `abi_extra` field in the `Pg_magic_func` that is always `"PostgreSQL"`, but might be different on PostgreSQL forks with an incompatible ABI
* `pg_sys::Value` doesn't exist anymore


pgx/src/lib.rs
```diff
@@ -245,6 +245,21 @@ macro_rules! pg_magic_func {
...
+                len: size_of::<pgx::pg_sys::Pg_magic_struct>() as i32,
+                version: pgx::pg_sys::PG_VERSION_NUM as i32 / 100,
+                funcmaxargs: pgx::pg_sys::FUNC_MAX_ARGS as i32,
+                indexmaxkeys: pgx::pg_sys::INDEX_MAX_KEYS as i32,
+                namedatalen: pgx::pg_sys::NAMEDATALEN as i32,
+                float8byval: pgx::pg_sys::USE_FLOAT8_BYVAL as i32,
+                abi_extra: [
+                    b'P' as i8, b'o' as i8, b's' as i8, b't' as i8, b'g' as i8, b'r' as i8,
+                    b'e' as i8, b'S' as i8, b'Q' as i8, b'L' as i8, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
Smittyvb, line 259:
```rust
unsafe { std::mem::transmute(b"PostgreSQL\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0") };
```
would be nicer, but [is technically unsound](https://users.rust-lang.org/t/how-to-convert-i8-to-u8/16308/11?u=smitop)

workingjubilee: If we're seriously going to quibble over this, then I would prefer my own flavor of incomprehensible conversion nonsense:
```suggestion
{
    // array::from_fn isn't const yet, boohoo, so const-copy a bstr
    let magic = b"PostgreSQL";
    let mut abi = [0 as libc::c_char; 32];
    let mut i = 0;
    while i < magic.len() {
        abi[i] = magic[i] as _;
        i += 1;
    }
    abi
}
```

At least then we can read the string.

Smittyvb: I changed it to use your implementation, using `i8` instead of `libc::c_char` since `libc` isn't always available where the macro is expanded.

workingjubilee:
> `pg_sys::String` now exists, which conflicts with the std `String` when importing `pg_sys::*`

`use pg_sys::{String as PgString};` is an option for this case. A glob import of `pgx-pg-sys` is... [not recommended](https://github.com/tcdi/pgx/issues/730).

Smittyvb:
PostgreSQL 15 is released! 🎉

I updated the PR so that it runs PostgreSQL 15 in CI now that APT packages are available.

eeeebbbbrrrr:
> PostgreSQL 15 is released! 🎉
> 
> I updated the PR so that it runs PostgreSQL 15 in CI now that APT packages are available. I've also removed the code that used the release candidate version so that it always uses the actual release now.

amazing!  Thanks a billion.  We'll get this merged next week.

I think we're kinda already heading towards a 0.5.1, so this will be awesome to get included.

Also, it's probably close enough for us to go ahead and remove v10 support.  I'm planning on volunteering @BradyBonnette for that.

thomcc:
> `use pg_sys::{String as PgString};` is an option for this case. A glob import of `pgx-pg-sys` is... [not recommended](https://github.com/tcdi/pgx/issues/730), at least in user code. We may have to handle it here, of course...

Alternatively, doing nothing here would push folks not to glob import pg_sys, which is kind of a good thing... I suppose that kind of change might be better done over a major bump, though.

workingjubilee:
That extension will require more updates to build in any case, unfortunately, since we changed a fair number of API details it relies on.

Smittyvb:
Any update on when we might be able to get this PR in?

eeeebbbbrrrr:
this looks pretty good.

I hacked @zombodb to use this and after a few tweaks for `feature = "pg15"` here and there it compiled and passed all its tests.  If there was a problem, ZDB would have found it!

PR #732, opened 2022-10-04
Implement access to column meta information in query results

yrashk:


yrashk:
I have a use case where I need to figure out column meta-information (count, names, types) at runtime.

Please let me know if you have any suggestions.

pgx/src/spi.rs
```diff
@@ -487,6 +488,48 @@ impl SpiTupleTable {
             }
         }
     }
+
+    /// Returns the number of columns
+    pub fn columns(&self) -> Option<i32> {
```
willmurnane, line 493: Couldn't this return `usize` unconditionally? If there's no `tupdesc` then there are 0 columns.

pgx/src/spi.rs
```diff
@@ -487,6 +488,48 @@ impl SpiTupleTable {
...
+
+    /// Returns column name
+    pub fn column_name(&self, ordinal: i32) -> Option<String> {
+        match self.tupdesc {
+            Some(tupdesc) => unsafe {
+                let nattrs = (*tupdesc).natts;
+                if ordinal < 1 || ordinal > nattrs {
+                    None
+                } else {
+                    let name = pg_sys::SPI_fname(tupdesc, ordinal);
```
willmurnane, line 524: Would it be less surprising to adjust this to 0-indexed?

willmurnane (approved):
Overall looks good!

pgx/src/spi.rs
yrashk: I am following that's already in Spi and Postgres itself.

pgx/src/spi.rs
yrashk: I guess this is reasonable. I've pushed an update.

pgx/src/spi.rs
thomcc: It's probably worth mentioning in the documentation comment.

yrashk: @thomcc I've updated the documentation to reflect this.

workingjubilee:
Hmm. Why `i32` instead of `usize` or `u32`? If it's because you wanted to make it easier for interop with various Postgres FFI functions, it should probably be `libc::c_int` to make that clear. Is it only because Postgres tends to use `c_int` for indexing?

yrashk:
I honestly don't remember anymore :) But I'm happy to change this to `usize` tomorrow, or if it is the only thing blocking this one, feel free to change it to `usize` in preparation for the merge.

yrashk:
I've updated the PR to use `usize`

workingjubilee (approved):
This PR is probably OK but Eric wants to take a gander at it first because SPI is a bit of a touchy interface.

pgx/src/spi.rs
```diff
@@ -487,6 +488,52 @@ impl SpiTupleTable {
...
+        match self.tupdesc {
+            Some(tupdesc) => unsafe { (*tupdesc).natts as usize },
+            None => 0,
+        }
+    }
+
+    /// Returns column type OID
+    ///
+    /// The ordinal position is 1-based
+    pub fn column_type_oid(&self, ordinal: usize) -> Option<PgOid> {
```
eeeebbbbrrrr, line 503: `usize` should probably be `NonZeroUsize`.

Not very ergonomic, but no surprising behaviors around zero.

eeeebbbbrrrr (changes requested):

yrashk:
I just did it consistently with the rest of SPI interface. Should those
interface be updated eventually, too?

eeeebbbbrrrr (approved):

PR #733, opened 2022-10-04
Stop binding system headers and a few other things

workingjubilee:
Use an allowlist that covers the various Postgres source folders, and blocklist a number of things that only C would be interested in. This reduces the generated bindings by about half a megabyte each.

This partially follows up on tcdi/pgx#730.
Is this what you meant, Thom?

### DIFF
--- a/pgx-pg-sys/Cargo.toml
+++ b/pgx-pg-sys/Cargo.toml
@@ -37,7 +37,7 @@ pgx-utils = { path = "../pgx-utils/", version = "=0.5.0-beta.1" }
 sptr = "0.3"
 
 [build-dependencies]
-bindgen = { version = "0.59.2", default-features = false, features = ["runtime"] }
+bindgen = { version = "0.60.1", default-features = false, features = ["runtime"] }
 build-deps = "0.1.4"
 owo-colors = "3.5.0"
 num_cpus = "1.13.1"
--- a/pgx-pg-sys/build.rs
+++ b/pgx-pg-sys/build.rs
@@ -549,7 +549,8 @@ fn run_bindgen(pg_config: &PgConfig, include_h: &PathBuf) -> eyre::Result<syn::F
         .blocklist_function(".*(set|long)jmp")
         .blocklist_function("pg_re_throw")
         .blocklist_item("CONFIGURE_ARGS") // configuration during build is hopefully irrelevant
-        .blocklist_item("_*(HAVE|have)_.*") // these are for C's use
+        .blocklist_item("(_*HAVE|__have)_.*") // header tracking metadata
+        .blocklist_item("_[A-Z_]+_H") // more header metadata
         .blocklist_item("__[A-Z].*") // these are reserved and unused by Postgres
         .blocklist_item("__darwin.*") // this should always be Apple's names
         .blocklist_function("float[48].*") // Rust has plenty of float handling
--- a/pgx-pg-sys/build.rs
+++ b/pgx-pg-sys/build.rs
@@ -553,6 +553,8 @@ fn run_bindgen(pg_config: &PgConfig, include_h: &PathBuf) -> eyre::Result<syn::F
         .blocklist_item("_[A-Z_]+_H") // more header metadata
         .blocklist_item("__[A-Z].*") // these are reserved and unused by Postgres
         .blocklist_item("__darwin.*") // this should always be Apple's names
+        .blocklist_function("pq(Strerror|Get.*)") // wrappers around platform functions: user can call those themselves
+        .blocklist_item(".*pthread.*)") // shims for pthreads on non-pthread systems, just use std::thread
         .blocklist_function("float[48].*") // Rust has plenty of float handling
         .blocklist_item(".*[vV][aA]_(list|LIST|start|START|end|END|copy|COPY).*") // do not need va_list anything!
         .blocklist_function("(pg_|p)v(sn?|f)?printf")
--- a/pgx-pg-sys/build.rs
+++ b/pgx-pg-sys/build.rs
@@ -555,6 +555,10 @@ fn run_bindgen(pg_config: &PgConfig, include_h: &PathBuf) -> eyre::Result<syn::F
         .blocklist_item("_*(HAVE|have)_.*") // these are for C's use
         .blocklist_item("__[A-Z].*") // these are reserved
         .blocklist_function("float[48].*") // Rust has plenty of float handling
+        .blocklist_item(".*[vV][aA]_(list|LIST|start|START|end|END|copy|COPY).*") // do not need va_list anything!
+        .blocklist_function("(pg_|p)v(sn?|f)?printf")
+        .blocklist_function("appendStringInfoVA")
+        .blocklist_file("stdarg.h")
         .size_t_is_usize(true)
         .rustfmt_bindings(false)
         .derive_debug(true)
--- a/pgx-pg-sys/build.rs
+++ b/pgx-pg-sys/build.rs
@@ -537,27 +537,27 @@ fn run_bindgen(pg_config: &PgConfig, include_h: &PathBuf) -> eyre::Result<syn::F
         .clang_arg(&format!("-I{}", includedir_server.display()))
         .parse_callbacks(Box::new(PgxOverrides::default()))
         .allowlist_file(".*confdefs.h")
-        .allowlist_file(".*([fF]uncs|mgr).h")
-        .allowlist_file(".*htup(_details)?.h")
+        .allowlist_file(".*(?:[fF]uncs|mgr).h")
+        .allowlist_file(".*htup(?:_details)?.h")
         .allowlist_file(".*syscache.h")
         .allowlist_file(".*mcxt.h")
-        .allowlist_file(".*(storage|catalog|access|commands|executor|adt|optimizer|rewrite|postmaster|tcop|replication|nodes|postgres|parse|pg_|item|heap).*")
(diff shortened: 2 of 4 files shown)


thomcc:
I think so, but would need to look at the generated bindings.

pgx-pg-sys/build.rs
```diff
@@ -536,14 +536,26 @@ fn run_bindgen(pg_config: &PgConfig, include_h: &PathBuf) -> eyre::Result<syn::F
...
         .blocklist_function("varsize_any") // pgx converts the VARSIZE_ANY macro, so we don't want to also have this function, which is in heaptuple.c
         .blocklist_function("query_tree_walker")
         .blocklist_function("expression_tree_walker")
-        .blocklist_function("sigsetjmp")
-        .blocklist_function("siglongjmp")
+        .blocklist_function(".*setjmp")
+        .blocklist_function(".*longjmp")
         .blocklist_function("pg_re_throw")
+        .blocklist_item("CONFIGURE_ARGS") // configuration during build is hopefully irrelevant
+        .blocklist_item("_*(HAVE|have)_.*") // these are for C's use
```
workingjubilee, line 556: These are basically metadata that the CPP logic uses to track which headers have been included and all that. I would like to add all of the `# define HEADER_NAME_H 1` cases as well, which are much the same, but the regex for covering all of those feels less obvious so I didn't take a swing at it.

pgx-pg-sys/build.rs
```diff
@@ -536,14 +536,26 @@ fn run_bindgen(pg_config: &PgConfig, include_h: &PathBuf) -> eyre::Result<syn::F
...
+        .blocklist_item("__[A-Z].*") // these are reserved
+        .blocklist_function("float[48].*") // Rust has plenty of float handling
```
workingjubilee, line 558: As far as I can tell, we bind a lot of functions that are actually meant to be called into by Postgres only to handle SQL runtime requests, and thus are of no interest for PGX users because they can just use Rust code to handle them. This is a fairly conservative start on omitting those functions because I deemed floats as least likely to be needed to be handled by C (and Rust passing floats to C can, in rare cases, on architectures we currently don't intentionally support, actually lead to busted behavior).

pgx-pg-sys/build.rs
```diff
@@ -536,14 +536,26 @@ fn run_bindgen(pg_config: &PgConfig, include_h: &PathBuf) -> eyre::Result<syn::F
...
+        .allowlist_var("SIG.*")
         .blocklist_type("Datum") // manually wrapping datum types for correctness
         .blocklist_type("NullableDatum")
         .blocklist_function("varsize_any") // pgx converts the VARSIZE_ANY macro, so we don't want to also have this function, which is in heaptuple.c
         .blocklist_function("query_tree_walker")
         .blocklist_function("expression_tree_walker")
-        .blocklist_function("sigsetjmp")
-        .blocklist_function("siglongjmp")
+        .blocklist_function(".*setjmp")
+        .blocklist_function(".*longjmp")
```
workingjubilee, line 553: This was obviously too conservative.

pgx-pg-sys/build.rs
```diff
@@ -536,14 +536,26 @@ fn run_bindgen(pg_config: &PgConfig, include_h: &PathBuf) -> eyre::Result<syn::F
         .header(include_h.display().to_string())
         .clang_arg(&format!("-I{}", includedir_server.display()))
         .parse_callbacks(Box::new(PgxOverrides::default()))
+        .allowlist_file(".*confdefs.h")
+        .allowlist_file(".*([fF]uncs|mgr).h")
+        .allowlist_file(".*htup(_details)?.h")
+        .allowlist_file(".*syscache.h")
+        .allowlist_file(".*mcxt.h")
+        .allowlist_file(".*(storage|catalog|access|commands|executor|adt|optimizer|rewrite|postmaster|tcop|replication|nodes|postgres|parse|pg_|item|heap).*")
```
workingjubilee, line 544: This effectively includes all of the source directories that Postgres uses (since many files will be `#include`d and thus regexed as "path/file.h") and a handful of keywords that are common in Postgres source filenames.

Note that we cannot simply use `blocklist_file` on libc header names, I tried that. They're too :sparkles: Special :sparkles:. But also we cannot simply use `allowlist_file` on `.*include.*` because then all of glibc appears again.

pgx-pg-sys/Cargo.toml
```diff
@@ -37,7 +37,7 @@ pgx-utils = { path = "../pgx-utils/", version = "=0.5.0-beta.1" }
 sptr = "0.3"
 
 [build-dependencies]
-bindgen = { version = "0.59.2", default-features = false, features = ["runtime"] }
+bindgen = { version = "0.60.1", default-features = false, features = ["runtime"] }
```
workingjubilee, line 40: any 0.60 will do, actually, but might as well ask for the latest, I guess?

pgx-pg-sys/build.rs
```diff
@@ -536,14 +536,26 @@ fn run_bindgen(pg_config: &PgConfig, include_h: &PathBuf) -> eyre::Result<syn::F
...
         .blocklist_function("query_tree_walker")
         .blocklist_function("expression_tree_walker")
-        .blocklist_function("sigsetjmp")
-        .blocklist_function("siglongjmp")
+        .blocklist_function(".*setjmp")
+        .blocklist_function(".*longjmp")
         .blocklist_function("pg_re_throw")
+        .blocklist_item("CONFIGURE_ARGS") // configuration during build is hopefully irrelevant
+        .blocklist_item("_*(HAVE|have)_.*") // these are for C's use
+        .blocklist_item("__[A-Z].*") // these are reserved
```
workingjubilee, line 557: These names are seriously special super duper reserved names we should never be looking at.

pgx-pg-sys/build.rs
```diff
@@ -536,14 +536,26 @@ fn run_bindgen(pg_config: &PgConfig, include_h: &PathBuf) -> eyre::Result<syn::F
         .header(include_h.display().to_string())
         .clang_arg(&format!("-I{}", includedir_server.display()))
         .parse_callbacks(Box::new(PgxOverrides::default()))
+        .allowlist_file(".*confdefs.h")
+        .allowlist_file(".*([fF]uncs|mgr).h")
+        .allowlist_file(".*htup(_details)?.h")
+        .allowlist_file(".*syscache.h")
+        .allowlist_file(".*mcxt.h")
+        .allowlist_file(".*(storage|catalog|access|commands|executor|adt|optimizer|rewrite|postmaster|tcop|replication|nodes|postgres|parse|pg_|item|heap).*")
+        .allowlist_function("memcpy")
```
workingjubilee, line 545: I... have no idea why it really needs us to also bind memcpy but it does. Go figure.

pgx-pg-sys/build.rs
```diff
@@ -536,14 +536,26 @@ fn run_bindgen(pg_config: &PgConfig, include_h: &PathBuf) -> eyre::Result<syn::F
         .header(include_h.display().to_string())
         .clang_arg(&format!("-I{}", includedir_server.display()))
         .parse_callbacks(Box::new(PgxOverrides::default()))
+        .allowlist_file(".*confdefs.h")
```
workingjubilee, line 539: This is created at build time for Postgres.

pgx-pg-sys/build.rs
```diff
@@ -536,14 +536,26 @@ fn run_bindgen(pg_config: &PgConfig, include_h: &PathBuf) -> eyre::Result<syn::F
...
+        .allowlist_file(".*([fF]uncs|mgr).h")
```
workingjubilee, line 540: The Postgres source tree loves `.*funcs` files and names all its managers `.*mgr`.

pgx-pg-sys/build.rs
```diff
@@ -536,14 +536,26 @@ fn run_bindgen(pg_config: &PgConfig, include_h: &PathBuf) -> eyre::Result<syn::F
...
+        .allowlist_file(".*htup(_details)?.h")
+        .allowlist_file(".*syscache.h")
+        .allowlist_file(".*mcxt.h")
```
workingjubilee, line 543: I think that the upcoming regex takes care of most of these actually, but this helps make double-sure we bind these.

workingjubilee (commented):
I think that to reduce the amount of sheer data saved in the repo we should also add a rustfmt.toml with [`use_small_heuristics`] cranked up but that should be in the same PR that runs the bindgen again for 0.5.0. It will significantly claw back pure "empty space" lines, and the rust-lang/rust repository also uses similar settings.

[`use_small_heuristics`]: https://github.com/rust-lang/rustfmt/blob/ef91154250977b3b5d05448dafbca524a1168b47/Configurations.md#use_small_heuristics

thomcc (commented):
This looks good to me but I suspect there are a couple more regexes we'll need in order to filter out the stuff in the macOS headers (unless you've done that). It's usually pretty easy and I don't mind taking it in a bit.

thomcc:
(Thanks a ton also -- this is great)

pgx-pg-sys/build.rs
workingjubilee: Well, there's a number of `_[A-Z]` and `__.*` that seem to actually be in use in our bindings, so I didn't omit all of them, even though I wanted to. I think Postgres uses `_[A-Z]` in its names even though it's not supposed to. But we can add a blocklist for `__darwin.*`?

thomcc (approved):
Looks great! Thanks so much for doing this!

Issue #730, opened 2022-10-03
Prevent symbols from system headers ending up in generated bindings


thomcc:
(Sorry if this issue exists already, I couldn't find it)

`pgx-pg-sys`'s generated bindings seem have a lot of symbols that come from system headers which have been included by postgres headers, rather instead of coming from postgres itself. This causes some significant perf issues, but also some correctness/safety concerns having to do with how system headers do symbol aliasing.

Unfortunately on system libraries symbols frequently[^1] get conditionally aliased in various ways, often to handle stuff like 64 vs 32 bit ino_t/off_t/time_t/etc. As an example, the various `DIR` functions (`readdir`, `seekdir`, etc) are exported from the generated  bindings on x86_64-apple-darwin, but the actual symbols 64 bit processes should be calling on that target are [somewhat more complicated](https://github.com/rust-lang/libc/blob/9d4e7802112e3d3186bdd0191a036281506c28b7/src/unix/bsd/mod.rs#L669-L677). Because these get aliased in the system headers via an [asm label attribute](https://clang.llvm.org/docs/AttributeReference.html#asm) and not a define, bindgen doesn't understand it and gets the wrong symbol name.

The outcome is that we might be exposing functions with slightly wrong ABIs on some targets. While this is bad... in practice it's farily unlikely anybody is going to use `pgx_pg_sys::seekdir` rather than `libc::seekdir` (although, perhaps it could happen if editor autocomplete guesses wrong, or in a glob-import situation). So in practice it's perhaps more a compile-perf and usability issue than anything else.

The easiest way to fix this is the `allowlist_file` fork of bindgen <https://github.com/daviddrysdale/rust-bindgen/tree/allowlist-file>, which would let us only generate bindings that are from certain files. I haven't checked in on this in a while, but I guess it... still hasn't gotten upstreamed (this could probably be pushed on harder -- openssl-sys seems to still use it in their repo...). I suspect there's some other workaround for this, but it's annoying that it's needed.

Alternatively, we could collect the symbols from the postgres dynamic libraries (perhaps when building with `cargo-pgx`), and use that to produce a bindgen allowlist. This is more annoying, and sadly doesn't handle type definitions (unlike `allowlist_file`), but OTOH I *suspect* we want to know what symbols these are anyway.

[^1]: There are 304 instances of `link_name` in `rust-lang/libc`'s `src` folder, for example.

workingjubilee:
[...]

PR #736, opened 2022-10-04
Pass `-isysroot` to bindgen on macOS

### LINKED ISSUE #151: (Mac) error: failed to run custom build command for `pgx-pg-sys v0.1.21`
I following this [Getting Started](https://github.com/zombodb/pgx#getting-started) guide, when running command `cargo pgx run pg13` got this errors:

(macOS 11.3.1, Xcode 12.5, rustc 1.54.0-nightly)

```
error: failed to run custom build command for `pgx-pg-sys v0.1.21`

Caused by:
  process didn't exit successfully: `/Users/hello/work/lab/myfile_fdw/target/debug/build/pgx-pg-sys-7e5dcc9e92761c9a/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-changed=/Users/hello/.pgx/config.toml
  cargo:rerun-if-changed=include/pg10.h
  cargo:rerun-if-changed=include/pg11.h
  cargo:rerun-if-changed=include/pg12.h
  cargo:rerun-if-changed=include/pg13.h
  cargo:rerun-if-changed=cshim/pgx-cshim.c
  cargo:rerun-if-changed=cshim/Makefile

  --- stderr
  manifest_dir=/Users/hello/.cargo/registry/src/github.com-1ecc6299db9ec823/pgx-pg-sys-0.1.21
  shim_dir=/Users/hello/.cargo/registry/src/github.com-1ecc6299db9ec823/pgx-pg-sys-0.1.21/cshim
  Generating bindings for pg13
  Generating bindings for pg10
  Generating bindings for pg11
  Generating bindings for pg12
  /Users/hello/.pgx/13.3/pgx-install/include/postgresql/server/c.h:59:10: fatal error: 'stdio.h' file not found
  /Users/hello/.pgx/13.3/pgx-install/include/postgresql/server/c.h:59:10: fatal error: 'stdio.h' file not found, err: true
  thread '<unnamed>' panicked at 'Unable to generate bindings for pg13: ()', /Users/hello/.cargo/registry/src/github.com-1ecc6299db9ec823/pgx-pg-sys-0.1.21/build.rs:486:13
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
  /Users/hello/.pgx/11.12/pgx-install/include/postgresql/server/c.h:59:10: fatal error: 'stdio.h' file not found
  /Users/hello/.pgx/11.12/pgx-install/include/postgresql/server/c.h:59:10: fatal error: 'stdio.h' file not found, err: true
  thread '<unnamed>' panicked at 'Unable to generate bindings for pg11: ()', /Users/hello/.cargo/registry/src/github.com-1ecc6299db9ec823/pgx-pg-sys-0.1.21/build.rs:486:13
  /Users/hello/.pgx/12.7/pgx-install/include/postgresql/server/c.h:59:10: fatal error: 'stdio.h' file not found
  /Users/hello/.pgx/12.7/pgx-install/include/postgresql/server/c.h:59:10: fatal error: 'stdio.h' file not found, err: true
  thread '<unnamed>' panicked at 'Unable to generate bindings for pg12: ()', /Users/hello/.cargo/registry/src/github.com-1ecc6299db9ec823/pgx-pg-sys-0.1.21/build.rs:486:13
  /Users/hello/.pgx/10.17/pgx-install/include/postgresql/server/c.h:81:10: fatal error: 'stdio.h' file not found
  /Users/hello/.pgx/10.17/pgx-install/include/postgresql/server/c.h:81:10: fatal error: 'stdio.h' file not found, err: true
  thread '<unnamed>' panicked at 'Unable to generate bindings for pg10: ()', /Users/hello/.cargo/registry/src/github.com-1ecc6299db9ec823/pgx-pg-sys-0.1.21/build.rs:486:13
      [error] failed to build extension
```

The command `cargo pgx init` was successful:

```
  Discovered Postgres v13.3, v12.7, v11.12, v10.17
  Downloading Postgres v10.17 from https://ftp.postgresql.org/pub/source/v10.17/postgresql-10.17.tar.bz2
  Downloading Postgres v11.12 from https://ftp.postgresql.org/pub/source/v11.12/postgresql-11.12.tar.bz2
  Downloading Postgres v12.7 from https://ftp.postgresql.org/pub/source/v12.7/postgresql-12.7.tar.bz2
  Downloading Postgres v13.3 from https://ftp.postgresql.org/pub/source/v13.3/postgresql-13.3.tar.bz2
     Removing /Users/hello/.pgx/13.3
     Removing /Users/hello/.pgx/12.7
    Untarring Postgres v13.3 to /Users/hello/.pgx/13.3
    Untarring Postgres v12.7 to /Users/hello/.pgx/12.7
  Configuring Postgres v12.7
  Configuring Postgres v13.3
     Removing /Users/hello/.pgx/11.12
    Untarring Postgres v11.12 to /Users/hello/.pgx/11.12
  Configuring Postgres v11.12
     Removing /Users/hello/.pgx/10.17
    Untarring Postgres v10.17 to /Users/hello/.pgx/10.17
  Configuring Postgres v10.17
    Compiling Postgres v13.3
    Compiling Postgres v11.12
    Compiling Postgres v12.7
    Compiling Postgres v10.17
   Installing Postgres v11.12 to /Users/hello/.pgx/11.12/pgx-install
   Installing Postgres v12.7 to /Users/hello/.pgx/12.7/pgx-install
   Installing Postgres v13.3 to /Users/hello/.pgx/13.3/pgx-install
   Installing Postgres v10.17 to /Users/hello/.pgx/10.17/pgx-install
   Validating /Users/hello/.pgx/10.17/pgx-install/bin/pg_config
   Validating /Users/hello/.pgx/11.12/pgx-install/bin/pg_config
   Validating /Users/hello/.pgx/12.7/pgx-install/bin/pg_config
   Validating /Users/hello/.pgx/13.3/pgx-install/bin/pg_config
The last command took 418.053 seconds.
```

thomcc:
I think this will probably fix #151, but can't repro.

### DIFF
--- a/pgx-pg-sys/Cargo.toml
+++ b/pgx-pg-sys/Cargo.toml
@@ -50,3 +50,4 @@ syn = { version = "1.0.101", features = [ "extra-traits", "full", "fold", "parsi
 eyre = "0.6.8"
 color-eyre = "0.6.2"
 rustversion = "1.0"
+shlex = "1.1.0"
--- a/pgx-pg-sys/Cargo.toml
+++ b/pgx-pg-sys/Cargo.toml
@@ -50,4 +50,4 @@ syn = { version = "1.0.101", features = [ "extra-traits", "full", "fold", "parsi
 eyre = "0.6.8"
 color-eyre = "0.6.2"
 rustversion = "1.0"
-shlex = "1.1.0"
+shlex = "1.1.0" # shell lexing, also used by many of our deps
--- a/pgx-pg-sys/build.rs
+++ b/pgx-pg-sys/build.rs
@@ -679,7 +679,7 @@ fn extra_bindgen_clang_args(pg_config: &PgConfig) -> eyre::Result<Vec<String>> {
                     // The logic we'd like ideally is for `cargo pgx init` to
                     // choose a good SDK in the first place, and force postgres
                     // to use it. Then, the logic in this build script would
-                    // Just Work without changes (since we are using it's
+                    // Just Work without changes (since we are using its
                     // sysroot verbatim).
                     //
                     // The value of "Good" here is tricky, but the logic should
@@ -707,7 +707,7 @@ fn extra_bindgen_clang_args(pg_config: &PgConfig) -> eyre::Result<Vec<String>> {
                     // SDK version, but it's not an option for postgres.
                     let major_version = pg_config.major_version()?;
                     println!(
-                        "cargo:warning=postgres version {major_version} was compiled against a \
+                        "cargo:warning=postgres v{major_version} was compiled against an \
                          SDK Root which does not seem to exist on this machine ({}). You may \
                          need to re-run `cargo pgx init` and/or update your command line tools.",
                         pair[1],
--- a/pgx-pg-sys/build.rs
+++ b/pgx-pg-sys/build.rs
@@ -535,6 +535,7 @@ fn run_bindgen(pg_config: &PgConfig, include_h: &PathBuf) -> eyre::Result<syn::F
     let bindings = bindgen::Builder::default()
         .header(include_h.display().to_string())
         .clang_arg(&format!("-I{}", includedir_server.display()))
+        .clang_args(&extra_bindgen_clang_args(pg_config)?)
         .parse_callbacks(Box::new(PgxOverrides::default()))
         .blocklist_type("Datum") // manually wrapping datum types for correctness
         .blocklist_type("NullableDatum")
@@ -632,6 +633,92 @@ fn build_shim_for_version(
     }
 
     Ok(())
+}
+
+fn extra_bindgen_clang_args(pg_config: &PgConfig) -> eyre::Result<Vec<String>> {
+    let mut out = vec![];
+    if std::env::var("CARGO_CFG_TARGET_OS").unwrap_or_default() == "macos" {
+        // On macOS, find the `-isysroot` arg out of the c preprocessor flags,
+        // to handle the case where bindgen uses a libclang isn't provided by
+        // the system.
+        let flags = pg_config.cppflags()?;
+        // In practice this will always be valid UTF-8 because of how the
+        // `pgx-pg-config` crate is implemented, but even if it were not, the
+        // problem won't be with flags we are interested in.
+        let flags = shlex::split(&flags.to_string_lossy()).unwrap_or_default();
+        // Find the `-isysroot` flags -- The rest are `-I` flags that don't seem
+        // to be needed inside the code (and feel likely to cause bindgen to
+        // emit bindings for unrelated libraries)
+        for pair in flags.windows(2) {
+            if pair[0] == "-isysroot" {
+                if std::path::Path::new(&pair[1]).exists() {
+                    out.extend(pair.into_iter().cloned());
+                } else {
+                    // The SDK path doesnt exist. Emit a warning, which they'll
+                    // see if the build ends up failing (it may not fail in all
+                    // cases, so we don't panic here).
+                    //
+                    // There's a bunch of smarter things we can try here, but
+                    // most of them either break things that currently work, or
+                    // are very difficult to get right. If you try to fix this,
+                    // be sure to consider cases like:
+                    //
+                    // - User may have CommandLineTools and not Xcode, vice
+                    //   versa, or both installed.
+                    // - User may using a newer SDK than their OS, or vice
+                    //   versa.
+                    // - User may be using a newer SDK than their XCode (updated
+                    //   Command line tools, not OS), or vice versa.
+                    // - And so on.
+                    //
+                    // These are all actually fairly common. Note that the code
+                    // as-is is *not* broken in these cases (except on OS/SDK
+                    // updates), so care should be taken to avoid changing that
+                    // if possible.
+                    //
+                    // The logic we'd like ideally is for `cargo pgx init` to
+                    // choose a good SDK in the first place, and force postgres
+                    // to use it. Then, the logic in this build script would
+                    // Just Work without changes (since we are using it's
+                    // sysroot verbatim).
+                    //
+                    // The value of "Good" here is tricky, but the logic should
+                    // probably:
+                    //
+                    // - prefer SDKs from the CLI tools to ones from XCode
+                    //   (since they're guaranteed compatible with the user's OS
+                    //   version)
+                    //
+                    // - prefer SDKs that specify only the major SDK version
+                    //   (e.g. MacOSX12.sdk and not MacOSX12.4.sdk or
+                    //   MacOSX.sdk), to avoid breaking too frequently (if we
+                    //   have a minor version) or being totally unable to detect
+                    //   what version of the SDK was used to build postgres (if
+                    //   we have neither).
+                    //
+                    // - Avoid choosing an SDK newer than the user's OS version,
+                    //   since postgres fails to detect that they are missing if
+                    //   you do.
+                    //
+                    // This is surprisingly hard to implement, as the
+                    // information is scattered across a dozen ini files.
+                    // Presumably Apple assumes you'll use
+                    // `MACOSX_DEPLOYMENT_TARGET`, rather than basing it off the
+                    // SDK version, but it's not an option for postgres.
+                    let major_version = pg_config.major_version()?;
+                    println!(
+                        "cargo:warning=postgres version {major_version} was compiled against a \
+                         SDK Root which does not seem to exist on this machine ({}). You may \
+                         need to re-run `cargo pgx init` and/or update your command line tools.",
+                        pair[1],
+                    );
+                };
+                // Either way, we stop here.
+                break;
+            }
+        }
+    }
+    Ok(out)
 }
 
 fn run_command(mut command: &mut Command, version: &str) -> eyre::Result<Output> {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1448,6 +1448,7 @@ dependencies = [
  "quote",
  "rayon",
  "rustversion",
+ "shlex",
  "sptr",
  "syn",
 ]
--- a/pgx-pg-config/src/lib.rs
+++ b/pgx-pg-config/src/lib.rs
@@ -14,6 +14,7 @@ use std::fmt::{self, Display, Formatter};
 use std::process::Stdio;
 use std::{
     collections::HashMap,
+    ffi::OsString,
     io::ErrorKind,
     path::{Path, PathBuf},
     process::Command,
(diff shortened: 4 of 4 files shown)


eeeebbbbrrrr (approved):
Other than the one question, this looks good to me.  I'll go ahead and pre-approve assuming my the answer to my question is "nope"

pgx-pg-sys/build.rs
```diff
@@ -535,6 +535,7 @@ fn run_bindgen(pg_config: &PgConfig, include_h: &PathBuf) -> eyre::Result<syn::F
     let bindings = bindgen::Builder::default()
         .header(include_h.display().to_string())
         .clang_arg(&format!("-I{}", includedir_server.display()))
+        .clang_args(&extra_bindgen_clang_args(pg_config)?)
```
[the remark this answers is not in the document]
thomcc: It doesn't cause issues, no.

That said (as discussed on discord) I'm going to change this to get the version independent path (and also pass that to `make` as `PG_SDKROOT`). That will mean that things continue to work after OS updates without needing to run `cargo pgx init` again.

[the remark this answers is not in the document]
workingjubilee: Specifically, bindgen will just throw all these args onto its own internal `Vec<Args>`: https://github.com/rust-lang/rust-bindgen/blob/a900f8f863d1313ad76603234aaeea22bb9ba7b3/src/lib.rs#L1190-L1200

thomcc:
I've updated it so it also passes PG_SYSROOT into make. This has a few benefits:
- Not breaking if you update the OS (testing currently). (On the downside, this means people won't be forced to periodically run `cargo pgx init`, I suppose)
- Allows building with versions of clang that cannot locate system files without an explicit `-isysroot` or `SDKROOT` env var (for example, `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang` is apparently this way). I don't know how, but maybe this is what the user in #151 hit.

Note: postgres's docs discourage doing this:

> Note that building an extension using a different sysroot version than was used to build the core server is not really recommended; in the worst case it could result in hard-to-debug ABI inconsistencies.

The system APIs should never change in this way, but it's possible postgres's internals have conditionals based on the SDK version. Much more likely is that this is advised out of an abundance of caution, which is fair.

workingjubilee:
My understanding of our current state is that we are in effect always running that risk anyways, and that actually passing this variable will at least try to make things more consistent in the general case?

thomcc:
> My understanding of our current state is that we are in effect always running that risk anyways, and that actually passing this variable will at least try to make things more consistent in the general case?

We currently are yes. This makes things more consistent and works in more cases, and I strongly suspect that this recommended out of an abundance of caution (and because they don't want bug reports from users who do this thoughtlessly).

I don't believe that it would cause actual ABI issues, as the "postgres build sdk version != user sdk version" case is almost always what happens if postgres is installed via a package manager.

thomcc:
Hm, I realized I should double-check with the PG sources. The way they select https://github.com/postgres/postgres/commit/4823621db312a0597c40686c4c94d47428889fef. That commit describes an issue where their configure script indicates presence of a symbol despite it not actually being present, in the case that the OS version is older than their SDK version. This is interesting, and is very different than the ABI issues they hint at in the docs above. The approach they took in this patch is a bit odd but it's likely less complex than the way the toolchain intends[^1] you to handle this, so I get it.

[^1]: The intended way to handle the issue they hit is probably setting `MACOSX_DEPLOYMENT_TARGET` in the environment to the current OS version (and perhaps passing `-mmacosx-version-min` too). This would cause the compiler to weakly link the symbols which are optional in that version, so they'd also need to adjust the configure script to treat "symbol exists but is null" as not having the symbol. That might be considerably more complex than the SDK path wrangling they do in that patch (even assuming there's no further issue), so I see why it's done this way.

I'll think a bit about this tomorrow, but quick thoughts:
- We don't need to handle being run from within XCode (probably), and we're only not using this to drive a configure script, which simplifies things.
- On the other hand our generated bindings make this a bit worse, due to #730 -- that issue means any symbol in the header must be present at load time, since we have code that calls all of them.

I have a few ideas how to handle this without much additional complexity, but will sleep on them -- wanted to get this down while it was in my brain.

thomcc:
Oh, nevermind. There code doesn't work the way they describe (I should have looked at the actual SDK roots for the postgres versions in `~/.pgx`! This is what I get for commenting late at night...).

Fix incoming (and sorry about the confusion here).

thomcc:
What a headache. I wrote a big comment about the state of affairs. This should fix problems from bindgen not using the same SDK that PG was compiled with, but avoiding breakage on OS/SDK updates is harder. I wrote about what we should do, but I don't think it's worth doing right now, since we should work in basically every other case that matters without bothering.

I did add a cargo warning. These are only presented to users if their build fails, but they will show up for us if they're ever hit inside this workspace. Please let me know if it happens to fire unexpectedly, but I think I've tested the cases that matter here.

pgx-pg-sys/build.rs
```diff
@@ -634,6 +635,92 @@ fn build_shim_for_version(
...
+                    //   have a minor version) or being totally unable to detect
+                    //   what version of the SDK was used to build postgres (if
+                    //   we have neither).
+                    //
+                    // - Avoid choosing an SDK newer than the user's OS version,
+                    //   since postgres fails to detect that they are missing if
+                    //   you do.
+                    //
+                    // This is surprisingly hard to implement, as the
+                    // information is scattered across a dozen ini files.
```
workingjubilee, line 704: :nya_a:

pgx-pg-sys/Cargo.toml
```diff
@@ -50,3 +50,4 @@ syn = { version = "1.0.101", features = [ "extra-traits", "full", "fold", "parsi
 eyre = "0.6.8"
 color-eyre = "0.6.2"
 rustversion = "1.0"
+shlex = "1.1.0"
```
workingjubilee, line 53:
```suggestion
shlex = "1.1.0" # shell lexing, also used by many of our deps
```

pgx-pg-sys/build.rs
```diff
@@ -634,6 +635,92 @@ fn build_shim_for_version(
...
+        // to handle the case where bindgen uses a libclang isn't provided by
+        // the system.
+        let flags = pg_config.cppflags()?;
+        // In practice this will always be valid UTF-8 because of how the
+        // `pgx-pg-config` crate is implemented, but even if it were not, the
+        // problem won't be with flags we are interested in.
+        let flags = shlex::split(&flags.to_string_lossy()).unwrap_or_default();
+        // Find the `-isysroot` flags -- The rest are `-I` flags that don't seem
+        // to be needed inside the code (and feel likely to cause bindgen to
+        // emit bindings for unrelated libraries)
```
workingjubilee, line 651: :see_no_evil:

pgx-pg-sys/build.rs
```diff
@@ -634,6 +635,92 @@ fn build_shim_for_version(
...
+        for pair in flags.windows(2) {
```
workingjubilee, line 652: Hm. This may sound silly but are we sure that `-isysroot` will only have one argument? Is it totally incoherent for it to have two?

pgx-pg-sys/build.rs
```diff
@@ -634,6 +635,92 @@ fn build_shim_for_version(
...
+                    //
+                    // - User may have CommandLineTools and not Xcode, vice
+                    //   versa, or both installed.
+                    // - User may using a newer SDK than their OS, or vice
+                    //   versa.
+                    // - User may be using a newer SDK than their XCode (updated
+                    //   Command line tools, not OS), or vice versa.
+                    // - And so on.
+                    //
+                    // These are all actually fairly common. Note that the code
```
workingjubilee, line 674: rei deadpan

pgx-pg-sys/build.rs
```diff
@@ -634,6 +635,92 @@ fn build_shim_for_version(
...
+                    //   you do.
+                    //
+                    // This is surprisingly hard to implement, as the
+                    // information is scattered across a dozen ini files.
+                    // Presumably Apple assumes you'll use
+                    // `MACOSX_DEPLOYMENT_TARGET`, rather than basing it off the
+                    // SDK version, but it's not an option for postgres.
+                    let major_version = pg_config.major_version()?;
+                    println!(
+                        "cargo:warning=postgres version {major_version} was compiled against a \
```
workingjubilee, line 710: Usually PGX has been saying "[pP]ostgres v{major}"
```suggestion
                        "cargo:warning=postgres v{major_version} was compiled against an \
```

pgx-pg-sys/build.rs
```diff
@@ -634,6 +635,92 @@ fn build_shim_for_version(
...
+                         SDK Root which does not seem to exist on this machine ({}). You may \
+                         need to re-run `cargo pgx init` and/or update your command line tools.",
```
workingjubilee, line 712: Hm. Will this be seen in practice after build errors? I mean, displaying it in this case at least is more helpful than not, but we should be clear as to ourselves when we're emitting this message.

> #### [`cargo:warning=MESSAGE`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargowarningmessage)
> The warning instruction tells Cargo to display a warning after the build script has finished running. Warnings are only shown for path dependencies (that is, those you're working on locally), so for example warnings printed out in [crates.io](https://crates.io/) crates are not emitted by default. The -vv "very verbose" flag may be used to have Cargo display warnings for all crates.

pgx-pg-sys/build.rs
```diff
@@ -634,6 +635,92 @@ fn build_shim_for_version(
...
+                    //
+                    // These are all actually fairly common. Note that the code
+                    // as-is is *not* broken in these cases (except on OS/SDK
+                    // updates), so care should be taken to avoid changing that
+                    // if possible.
+                    //
+                    // The logic we'd like ideally is for `cargo pgx init` to
+                    // choose a good SDK in the first place, and force postgres
+                    // to use it. Then, the logic in this build script would
+                    // Just Work without changes (since we are using it's
```
workingjubilee, line 682:
```suggestion
                    // Just Work without changes (since we are using its
```

pgx-pg-sys/build.rs
workingjubilee: which of "an? SDK" you produce is a personal question.

pgx-pg-sys/build.rs
thomcc: Yep, definitely one. No `-isysroot={path}` version either, clang does not accept that (Well, if you wanna provide an equals sign, it's gotta be as `--sysroot={path}`, but the behavior is different, because ~~of course it is~~ clang uses that one for compatibility with something else, apparently).

pgx-pg-sys/build.rs
thomcc: Not everybody installs XCode -- some only get the CLI tools. And people who already *have* XCode installed might never need to install the CLI tools separately. The latter is getting rarer, though.

Both is probably the most common?

pgx-pg-sys/build.rs
thomcc: They show up if the build fails in the build.rs -- Both stdout and stderr of the build script are emitted in that case. I believe cargo also highlights warning messages it got specifically, but I could be mistaken.

workingjubilee (approved):
Eric said it still looks fine!

PR #739, opened 2022-10-05
`#[derive(Deserialize)]` Timestamps

workingjubilee:
PGX should offer proper deserialization from strings at some point, however, this doesn't need to be more complicated because the only user is deserializing from integer data at the moment, and PGX did not previously directly `impl Deserialize`, nor did PGX enable the serde feature of the time crate.

It is mildly unfortunate that this means serde doesn't round trip, currently, but it is preferable to have a working type over not. This is necessary as previously, PGX's type mappings were more compatible with the `pg_sys::Timestamp{,Tz}` types.

### DIFF
--- a/pgx/src/datum/time_stamp.rs
+++ b/pgx/src/datum/time_stamp.rs
@@ -11,9 +11,10 @@ use crate::{pg_sys, FromDatum, FromTimeError, IntoDatum, TimestampWithTimeZone};
 use pgx_utils::sql_entity_graph::metadata::{
     ArgumentError, Returns, ReturnsError, SqlMapping, SqlTranslatable,
 };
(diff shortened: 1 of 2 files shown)


workingjubilee:
At least, I think this is enough.

cc @JamesGuthrie Does this fix your current problem at all? I don't know if this is what you need, seeing as how it means serde doesn't round trip. But I spent a bit looking into a "proper" Deserializer which can properly parse and let us round-trip `TimestampWithTimeZone` and starting to grok serde's data model made my head explode.

workingjubilee:
Currently James thinks this resolves this so I'm going to merge this so that it is currently in with develop and so where our latest fixes are.

PR #740, opened 2022-10-05
Include type when getting argument

### LINKED ISSUE #735: AnyElement oid is always zero
I'm trying to upgrade an extension that uses `AnyElement` to the 0.5 beta, and it looks like the `AnyElement::oid` of the type in the `AnyElement` is always zero, when an `AnyElement` is passed in from Postgres. `AnyElement::oid` worked fine in 0.4.5.

Here's a simplified test case with a failing test: https://gist.github.com/Smittyvb/b1ee0bffcf413b85cefe1351caedfaf6

Smittyvb:
This fixes #735 by passing the type of an argument when getting an argument.

### DIFF
--- a/pgx/src/datum/anyarray.rs
+++ b/pgx/src/datum/anyarray.rs
@@ -34,6 +34,8 @@ impl AnyArray {
 }
 
 impl FromDatum for AnyArray {
+    const GET_TYPOID: bool = true;
+
     #[inline]
     unsafe fn from_datum(
         datum: pg_sys::Datum,
--- a/pgx/src/datum/anyarray.rs
+++ b/pgx/src/datum/anyarray.rs
@@ -42,6 +42,15 @@ impl FromDatum for AnyArray {
         is_null: bool,
         typoid: pg_sys::Oid,
     ) -> Option<AnyArray> {
+        FromDatum::from_datum_with_typid(datum, is_null, typoid)
+    }
+
+    #[inline]
+    unsafe fn from_datum_with_typid(
+        datum: pg_sys::Datum,
+        is_null: bool,
+        typoid: pg_sys::Oid,
+    ) -> Option<AnyArray> {
         if is_null {
             None
         } else {
--- a/pgx/src/datum/anyarray.rs
+++ b/pgx/src/datum/anyarray.rs
@@ -42,11 +42,11 @@ impl FromDatum for AnyArray {
         is_null: bool,
         typoid: pg_sys::Oid,
     ) -> Option<AnyArray> {
-        FromDatum::from_datum_with_typid(datum, is_null, typoid)
+        FromDatum::from_polymorphic_datum(datum, is_null, typoid)
     }
 
     #[inline]
-    unsafe fn from_datum_with_typid(
+    unsafe fn from_polymorphic_datum(
         datum: pg_sys::Datum,
         is_null: bool,
         typoid: pg_sys::Oid,
--- a/pgx/src/datum/anyelement.rs
+++ b/pgx/src/datum/anyelement.rs
@@ -42,11 +42,11 @@ impl FromDatum for AnyElement {
         is_null: bool,
         typoid: pg_sys::Oid,
     ) -> Option<AnyElement> {
-        FromDatum::from_datum_with_typid(datum, is_null, typoid)
+        FromDatum::from_polymorphic_datum(datum, is_null, typoid)
     }
 
     #[inline]
-    unsafe fn from_datum_with_typid(
+    unsafe fn from_polymorphic_datum(
         datum: pg_sys::Datum,
         is_null: bool,
         typoid: pg_sys::Oid,
--- a/pgx/src/datum/anyelement.rs
+++ b/pgx/src/datum/anyelement.rs
@@ -34,6 +34,8 @@ impl AnyElement {
 }
 
 impl FromDatum for AnyElement {
+    const GET_TYPOID: bool = true;
+
     #[inline]
     unsafe fn from_datum(
         datum: pg_sys::Datum,
--- a/pgx/src/datum/anyelement.rs
+++ b/pgx/src/datum/anyelement.rs
@@ -42,6 +42,15 @@ impl FromDatum for AnyElement {
         is_null: bool,
         typoid: pg_sys::Oid,
     ) -> Option<AnyElement> {
+        FromDatum::from_datum_with_typid(datum, is_null, typoid)
+    }
+
+    #[inline]
+    unsafe fn from_datum_with_typid(
+        datum: pg_sys::Datum,
+        is_null: bool,
+        typoid: pg_sys::Oid,
+    ) -> Option<AnyElement> {
         if is_null {
             None
         } else {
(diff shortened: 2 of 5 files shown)


workingjubilee (dismissed):
This looks correct to me but I think it could use a 1-second glance from @eeeebbbbrrrr as well before it gets merged.

EDIT: My instincts about needing a second pair of eyes were correct, this is actually too slow for usage. The code needs to special-case AnyElement specifically, but I haven't the faintest idea how.

Smittyvb:
I added an associated constant (`GET_TYPOID`) to `FromDatum` that defaults to false so that `pg_getarg` only gets the type OID for `anyelement`/`anyarray` – that should alleviate the performance problem.

workingjubilee:
Hmm. That should work, but is it possible for us to make it only happen when AnyElement's `oid` actually has a lookup done on it? That way we can move back to stripping the third `from_datum` argument in 0.6.0

Smittyvb:
I don't think it's possible to do the type lookup on demand without making `FunctionCallInfo` an argument to `from_datum`. 

Another option would be to make a separate trait method for when a type OID is needed. That approach would allow for `from_datum` to only have 2 arguments in the future. I updated the PR to do that.

workingjubilee:
Hm. Should we use `try_from_datum` instead and say that `IntoDatum::is_compatible_with` for AnyElement simply returns `true`?

Smittyvb:
`try_from_datum` has a trait bound that `T: 'static` (because it uses `TypeId`, which doesn't work for types with lifetimes) that `pg_getarg` currently doesn't have. Requiring that `T: 'static` for `pg_getarg` would mean that all `#[pg_extern]` functions with lifetime parameters would need to have them be `'static`, which wouldn't be ideal – it results in 309 errors on my pgx extension.

workingjubilee:
...yeaaaah, that's not going to work, then.

pgx/src/datum/from.rs
```diff
@@ -54,6 +57,22 @@ pub trait FromDatum {
     where
         Self: Sized;
 
+    /// Like `from_datum` but the typoid must have a value.
+    ///
+    /// ## Safety
+    ///
+    /// Same caveats as `FromDatum::from_datum(...)`.
+    unsafe fn from_datum_with_typid(
```
workingjubilee, line 65:
```suggestion
    /// Like `from_datum` for instantiating polymorphic types
    /// which require preserving the dynamic type metadata.
    ///
    /// ## Safety
    ///
    /// Same caveats as `FromDatum::from_datum(...)`.
    unsafe fn from_polymorphic_datum(
```

pgx/src/fcinfo.rs
```diff
@@ -95,8 +95,11 @@ mod pg_10_11 {
         let datum = unsafe { fcinfo.as_ref() }.unwrap().arg[num];
         let isnull = pg_arg_is_null(fcinfo, num);
         unsafe {
-            let typid = pg_sys::InvalidOid;
-            T::from_datum(datum, isnull, typid)
+            if T::GET_TYPOID {
+                T::from_datum_with_typid(datum, isnull, super::pg_getarg_type(fcinfo, num))
```
workingjubilee, line 99:
```suggestion
                T::from_polymorphic_datum(datum, isnull, super::pg_getarg_type(fcinfo, num))
```

pgx/src/fcinfo.rs
```diff
@@ -134,8 +137,15 @@ mod pg_12_13_14 {
     pub fn pg_getarg<T: FromDatum>(fcinfo: pg_sys::FunctionCallInfo, num: usize) -> Option<T> {
         let datum = get_nullable_datum(fcinfo, num);
         unsafe {
-            let typid = pg_sys::InvalidOid;
-            T::from_datum(datum.value, datum.isnull, typid)
+            if T::GET_TYPOID {
+                T::from_datum_with_typid(
```
workingjubilee, line 141:
```suggestion
                T::from_polymorphic_datum(
```

pgx/src/datum/anyelement.rs
```diff
@@ -34,11 +34,22 @@ impl AnyElement {
...
     unsafe fn from_datum(
         datum: pg_sys::Datum,
         is_null: bool,
         typoid: pg_sys::Oid,
+    ) -> Option<AnyElement> {
+        FromDatum::from_datum_with_typid(datum, is_null, typoid)
+    }
+
+    #[inline]
+    unsafe fn from_datum_with_typid(
```
workingjubilee, line 49:
```suggestion
    unsafe fn from_polymorphic_datum(
```

pgx/src/datum/anyelement.rs
```diff
@@ -34,11 +34,22 @@ impl AnyElement {
...
 impl FromDatum for AnyElement {
+    const GET_TYPOID: bool = true;
+
     #[inline]
     unsafe fn from_datum(
         datum: pg_sys::Datum,
         is_null: bool,
         typoid: pg_sys::Oid,
+    ) -> Option<AnyElement> {
+        FromDatum::from_datum_with_typid(datum, is_null, typoid)
```
workingjubilee, line 45:
```suggestion
        FromDatum::from_polymorphic_datum(datum, is_null, typoid)
```

pgx/src/datum/anyarray.rs
```diff
@@ -34,11 +34,22 @@ impl AnyArray {
...
     unsafe fn from_datum(
         datum: pg_sys::Datum,
         is_null: bool,
         typoid: pg_sys::Oid,
+    ) -> Option<AnyArray> {
+        FromDatum::from_datum_with_typid(datum, is_null, typoid)
+    }
+
+    #[inline]
+    unsafe fn from_datum_with_typid(
```
workingjubilee, line 49:
```suggestion
    unsafe fn from_polymorphic_datum(
```

workingjubilee (commented):
Agreeing with my naming sense is not required, but I think makes it clearer why this function should exist.

pgx/src/datum/anyarray.rs
```diff
@@ -34,11 +34,22 @@ impl AnyArray {
...
 impl FromDatum for AnyArray {
+    const GET_TYPOID: bool = true;
+
     #[inline]
     unsafe fn from_datum(
         datum: pg_sys::Datum,
         is_null: bool,
         typoid: pg_sys::Oid,
+    ) -> Option<AnyArray> {
+        FromDatum::from_datum_with_typid(datum, is_null, typoid)
```
Smittyvb, line 45:
```suggestion
        FromDatum::from_polymorphic_datum(datum, is_null, typoid)
```

workingjubilee (approved):
We should find a cleaner solution in the future but I think this will do for right this second: it more or less brings us full circle to where we were before, but with at least a direction to go this time.

eeeebbbbrrrr:
Can we get rid of the `typeoid` argument to `FromDatum::from_datum()` now?

Smittyvb:
@eeeebbbbrrrr I think this PR fixed all the cases where `typoid` was needed are gone now, so it should be removable in 0.6 now.

PR #741, opened 2022-10-05
Add example type which implements `PostgresOrd`

willmurnane:
This PR adds a type which implements `PostgresOrd` to the `custom_types` module. This will cause failures if the ordering constraints fixed in #722 regress again, rather than let those problems go unnoticed.


workingjubilee:
@willmurnane Can you update this with latest `develop`? Not clear why this isn't building...

workingjubilee:
Thank you!

workingjubilee (approved):

PR #744, opened 2022-10-06
Prefer `impl From` to `impl Into`

thomcc:
Trivial patch. Just noticed a few of these, and switched them all since it looks odd to me to see `Into` implemented explicitly (and none of them hit the coherence edge-cases where `Into` impl is required).

If you `impl From` you get an `Into` automatically, but the opposite is not true, so this is strictly better (and not a breaking change).

### DIFF
--- a/pgx-utils/src/sql_entity_graph/aggregate/entity.rs
+++ b/pgx-utils/src/sql_entity_graph/aggregate/entity.rs
@@ -161,9 +161,9 @@ impl PartialOrd for PgAggregateEntity {
     }
 }
 
-impl Into<SqlGraphEntity> for PgAggregateEntity {
-    fn into(self) -> SqlGraphEntity {
-        SqlGraphEntity::Aggregate(self)
+impl From<PgAggregateEntity> for SqlGraphEntity {
+    fn from(val: PgAggregateEntity) -> Self {
+        SqlGraphEntity::Aggregate(val)
     }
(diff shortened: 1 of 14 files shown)


pgx-utils/src/sql_entity_graph/aggregate/entity.rs
```diff
@@ -161,9 +161,9 @@ impl PartialOrd for PgAggregateEntity {
...
 }
 
-impl Into<SqlGraphEntity> for PgAggregateEntity {
-    fn into(self) -> SqlGraphEntity {
-        SqlGraphEntity::Aggregate(self)
+impl From<PgAggregateEntity> for SqlGraphEntity {
+    fn from(val: PgAggregateEntity) -> Self {
+        SqlGraphEntity::Aggregate(val)
     }
 }
```
workingjubilee, line 168: These are part of a crate that we consider private internal details (anyone who uses them is essentially like a crate enabling rustc attrs... yeah you can do it, but also lol if you think that's stable) and so any change to them would not be a breaking change anyways.

workingjubilee (approved):

pgx-utils/src/sql_entity_graph/aggregate/entity.rs
thomcc: Good to know.

PR #745, opened 2022-10-06
Remove dependency on `core_cstr` now that `core::ffi::CStr` is stable 

thomcc:
*(Note: This PR was made on top of #744 because it would otherwise conflict, due to both PRs changing a `Into`/`From` implementation. I'll rebase this PR on top of develop after that one lands)*

According to our README, we require latest stable Rust. That's great to hear, since it means we can remove some cruft from dependencies.

This is technically breaking, but I've added a (insta-deprecated) shim module to make most code keep working (although it would still break code that relies on `std::ffi::CStr` and `pgx::core_cstr::CStr` being distinct types). I can remove this shim if it is not needed. If we can't make a breaking change here yet, I can drop this PR, it's not a big deal.

### DIFF
--- a/pgx/src/stringinfo.rs
+++ b/pgx/src/stringinfo.rs
@@ -26,27 +26,13 @@ impl From<StringInfo> for pg_sys::StringInfo {
     }
 }
 
-impl From<StringInfo> for &'static std::ffi::CStr {
+impl From<StringInfo> for &'static core::ffi::CStr {
     fn from(val: StringInfo) -> Self {
         let len = val.len();
         let ptr = val.into_char_ptr();
 
         unsafe {
-            std::ffi::CStr::from_bytes_with_nul_unchecked(std::slice::from_raw_parts(
-                ptr as *const u8,
-                (len + 1) as usize, // +1 to get the trailing null byte
-            ))
-        }
-    }
-}
-
-impl From<StringInfo> for &'static crate::cstr_core::CStr {
-    fn from(val: StringInfo) -> Self {
-        let len = val.len();
-        let ptr = val.into_char_ptr();
-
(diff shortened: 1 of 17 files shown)


workingjubilee:
I think this should be merged after 0.5.0 because while we do reserve the right to use the latest stable, I know this does bump our effective requirement and we promised no functional changes which such a bump arguably would require (so we'll make this part of 0.5.1).

workingjubilee:
I think this should be merged after 0.5.0 because while we do reserve the right to use the latest stable, I know this does bump our effective requirement and we promised no functional changes which such a bump arguably would require (so we'll make this part of 0.5.1).

thomcc:
Hm, okay. What is the version that we promised "no functional changes" on? (So I can make sure to `cargo check` with it)

thomcc:
> What is the version that we promised "no functional changes" on?

Answer: 1.61 or so.

> we'll make this part of 0.5.1

I'll wait until then for the rebase so that I don't have to do it twice.

pgx/src/stringinfo.rs
```diff
@@ -26,27 +26,13 @@ impl From<StringInfo> for pg_sys::StringInfo {
...
-    }
-}
-
-impl From<StringInfo> for &'static crate::cstr_core::CStr {
-    fn from(val: StringInfo) -> Self {
-        let len = val.len();
-        let ptr = val.into_char_ptr();
-
-        unsafe {
-            crate::cstr_core::CStr::from_bytes_with_nul_unchecked(std::slice::from_raw_parts(
```
workingjubilee, line 49: Good.

workingjubilee (approved):

PR #746, opened 2022-10-06
Set locale to C.UTF-8 when creating database clusters

### LINKED ISSUE #691: Locale of test databases should be controllable
The locale of databases created by pgx is whatever the system's locale is (since that's what Postgres does by default). In the context of `pgx-tests`, this can lead to accidentally creating tests that only work in the locale of the test author. For instance, `1-2-2020` gets parsed as Jan. 2 in the US but Feb. 1 in Canada.

Some parts of the database locale [can't be changed after database creation](https://www.postgresql.org/docs/current/locale.html), so one can't work around this by setting the database locale in tests. The only way around this is to change the system locale whenever pgx creates databases, which isn't ideal.

Ideally pgx should always set the locale of test databases to a configurable value (by default probably `C`, `en_US.UTF-8` also works if you want to avoid a breaking change for American English speakers).

Smittyvb:
Fixes #691 by setting the locale to `C.UTF-8` when creating databases.

On macOS the `C.UTF-8` locale doesn't exist so the locale is set to `C` and the `CTYPE` to `UTF-8`, which has the same effect as setting the locale to `C.UTF-8` on other operating systems.

Note that testing this requires creating a new database cluster. I tested this by:
- reinstalling cargo-pgx
- deleting `~/.pgx`
- rerunning `cargo pgx init` to get a new database cluster
- cleaning the target directory of any repos you want to test this on

### DIFF
--- a/cargo-pgx/src/command/init.rs
+++ b/cargo-pgx/src/command/init.rs
@@ -12,7 +12,9 @@ use crate::command::version::pgx_default;
 use crate::CommandExecute;
 use eyre::{eyre, WrapErr};
 use owo_colors::OwoColorize;
-use pgx_pg_config::{prefix_path, PgConfig, PgConfigSelector, Pgx, SUPPORTED_MAJOR_VERSIONS};
+use pgx_pg_config::{
+    prefix_path, PgConfig, PgConfigSelector, Pgx, C_LOCALE_FLAGS, SUPPORTED_MAJOR_VERSIONS,
+};
 use rayon::prelude::*;
 
 use std::collections::HashMap;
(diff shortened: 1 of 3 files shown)


workingjubilee:
@thomcc Do you think this handling of the locale setting is correct for Mac?

thomcc:
> @thomcc Do you think this handling of the locale setting is correct for Mac?

Sorry, I didn't see this. Yes, it looks correct (assuming that those arguments work as I'd expect).

eeeebbbbrrrr (approved):
I approve.

I now wonder about pgx extensions running in a non-UTF8 database.  But that's a different problem for a different day.

PR #747, opened 2022-10-06
Replace `rayon` with `std::thread::scope` in `pgx-pg-sys` build system

thomcc:
Rayon is a pretty large dependency (and comes with many dependencies of its own) especially to have as a build 
dependency, since it could be a bottleneck in the overall build process. In practice, we have enough other deps in the build script for this not to make a difference for full rebuilds, but it could for partial rebuilds (if e.g. `syn` is already built but `rayon` is not). Either way, the concrete outcome of this patch is shaving off about 20 dependencies from list that users of `pgx` will inherit from us. While that seems to only be around 10%, these might not otherwise appear in their dependency list at all (unlike many of the others), so this still a wain. Also, we didn't really need rayon here in the first place anyway[^2] (IMO).

I didn't remove it from `cargo-pgx`, since that isn't one of the deps that people inherit from us in their build tree by using `pgx`, so it doesn't matter as much IMO. It'd also be easy to remove there if we want.

The downside: this doesn't cap the number of threads spawned to `std::thread::available_parallelism()`, unlike rayon. This only matters for releases (likely done on machines with more than 5 threads), and is fairly easy[^1] to fix anyway if we decide we need to. In practice, it's likely we'll actually do better than `rayon` (WRT resource usage) on machines with many cores, since by default it uses a thread pool which is sized based on your number of cores, most of which we won't end up using. (We could also have fixed that by manually configuring `rayon` to use a pool with no more than `pg_config.len()` items, but no need).

I also cleaned things up a little while making this change, so that the bindings generation has its own function, and so that the paths we need to pass in are bundled in a little struct (since otherwise there'd be a lot of `&Path` arguments to that function).

[^2]: We use it to `par_iter` over a slice that has either 1 (normally) or 5 (releases) items a single time. Each one is a heavyweight task that takes roughly the same around of time as the others. This means we don't benefit from work stealing or a thread pool around, so it's mostly about ergonomics compared to manual parallelism, and that's a lot smaller of a benefit with `std::thread::scope`.

[^1]: Something like:
	```rs
	let threads = std::thread::available_parallelism().map(|t| t.get()).unwrap_or(1);
	pg_configs.chunks((pg_configs.len() / threads).max(1)).map(|chunk| {
		scope.spawn(|| chunk.try_for_each(|pg_config| ...)))
	})...
	```
	(I was going to handle it but decided it was not worth it)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1439,7 +1439,6 @@ dependencies = [
  "pgx-utils",
  "proc-macro2",
  "quote",
- "rayon",
  "rustversion",
  "shlex",
  "sptr",
--- a/pgx-pg-sys/Cargo.toml
+++ b/pgx-pg-sys/Cargo.toml
@@ -41,7 +41,6 @@ pgx-pg-config= { path = "../pgx-pg-config/", version = "=0.5.0" }
 pgx-utils = { path = "../pgx-utils/", version = "=0.5.0" }
 proc-macro2 = "1.0.46"
 quote = "1.0.21"
-rayon = "1.5.3"
 syn = { version = "1.0.102", features = [ "extra-traits", "full", "fold", "parsing" ] }
 eyre = "0.6.8"
 color-eyre = "0.6.2"
--- a/pgx-pg-sys/build.rs
+++ b/pgx-pg-sys/build.rs
@@ -137,61 +137,7 @@ fn main() -> color_eyre::Result<()> {
 
     pg_configs
         .par_iter()
-        .map(|pg_config| {
-            let major_version =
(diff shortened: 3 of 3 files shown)


workingjubilee:
I think this should be merged after 0.5.0 because while we do reserve the right to use the latest stable, I know this does bump our effective requirement and we promised no functional changes which such a bump arguably would require (so we'll make this part of 0.5.1).

thomcc:
`thread::scope` was 1.63 FWIW, so it does not require latest stable (which is 1.64).

workingjubilee:
Yes, I believe our current version is 1.59 and we want to not move past 1.61 for at least the 0.5.0 release.

PR #748, opened 2022-10-07
Golf Display for Node types

workingjubilee:
This may not seem like much but it removes ~50 KiB from each build. It also uses more conservative type casts in every case, as casting `&self as *const _ as *mut T` is a bit dubious. It has little effect in the Rust abstract machine, but it's the wrong signal when the decl is actually `extern char *nodeToString(const void *obj);`

This thoroughly refactors the safety boundaries so we can be sure the function is safe, despite handling raw pointers in the impl: encapsulating the `unsafe fn` entirely inside a private function.


pgx-pg-sys/src/lib.rs
```diff
@@ -160,29 +162,38 @@ pub use internal::pg13::*;
...
     }
 }
 
 /// implementation function for `impl Display for $NodeType`
-pub(crate) fn node_to_string_for_display(node: *mut crate::Node) -> String {
+///
+/// # Safety
+/// Don't use this on anything that doesn't impl PgNode, or the type may be off
+#[warn(unsafe_op_in_unsafe_fn)]
+pub(crate) unsafe fn display_node_impl(node: NonNull<crate::Node>) -> String {
```
workingjubilee, line 183: The previous `fn node_to_string_for_display` which this function exploded from the chest of was actually `unsafe`, as a null pointer input was not handled. This is only true in a trivial sense as it was also private, but it's worth noting that e.g. making it truly `pub` would have been unsound.

pgx-pg-sys/src/lib.rs
```diff
@@ -160,29 +162,38 @@ pub use internal::pg13::*;
 #[cfg(feature = "pg14")]
 pub use internal::pg14::*;
 
-/// A trait applied to all of Postgres' `pg_sys::Node` types and its subtypes
+/// A trait applied to all Postgres `pg_sys::Node` types and subtypes
 pub trait PgNode {
-    type NodeType;
+    // TODO(0.6.0): seal this trait or take this trait private?
```
workingjubilee, line 167: It seems unlikely we have any dependents on this but I didn't want to scratch it off right away as someone may be using it as a trait bound or something.

pgx-pg-sys/src/lib.rs
```diff
@@ -160,29 +162,38 @@ pub use internal::pg13::*;
...
-pub(crate) fn node_to_string_for_display(node: *mut crate::Node) -> String {
+///
+/// # Safety
+/// Don't use this on anything that doesn't impl PgNode, or the type may be off
+#[warn(unsafe_op_in_unsafe_fn)]
+pub(crate) unsafe fn display_node_impl(node: NonNull<crate::Node>) -> String {
+    // SAFETY: It's fine to call nodeToString with non-null well-typed pointers,
+    // and pg_sys::nodeToString() returns data via palloc, which is never null
+    // as Postgres will ERROR rather than giving us a null pointer,
+    // and Postgres starts and finishes constructing StringInfos by writing '\0'
```
workingjubilee, line 187: This required peeking "behind the curtain", tbh, so it's not certain this will remain true. However, the code in question has remained largely undisturbed: the `outNode` fn was exposed 7 years ago, it had `const` added 11 years ago, and the rest has sat there for ~24 years.

workingjubilee (commented):
It's plausible this should just all be implemented inside `lib.rs` or something but that would require us to phase it over the different versions correctly.

eeeebbbbrrrr (approved):

PR #749, opened 2022-10-07
Require `FromDatum::from_polymorphic_datum`

workingjubilee:
This also implies defaulting `FromDatum::from_datum` which just calls `FromDatum::from_polymorphic_datum` with an InvalidOid. Polymorphic types like AnyElement should generally override this on implementing, and non-polymorphic types should ignore the third parameter, while others may want a pass-through behavior.

### DIFF
--- a/pgx-examples/bgworker/src/lib.rs
+++ b/pgx-examples/bgworker/src/lib.rs
@@ -44,7 +44,7 @@ pub extern "C" fn _PG_init() {
 #[pg_guard]
 #[no_mangle]
 pub extern "C" fn background_worker_main(arg: pg_sys::Datum) {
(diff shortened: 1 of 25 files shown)


workingjubilee:
We should rework the datum unboxing API from the ground up tbh, but this is where we can get right now without breaking any promises.

eeeebbbbrrrr (approved):

workingjubilee:
Also we can get away with this because I warned we might still rename stuff. :smiling_imp:

Smittyvb:
I can confirm that doing a global find/replace for `from_datum` -> `from_polymorphic_datum` is all that I needed to make my extension work with these changes.

PR #750, opened 2022-10-07
Format the world with 0.5.0

workingjubilee:
This has not been tagged / pushed to crates.io yet and should not be until after this is squashed to develop and merged to master.

I forget if our CI actually works with the repo in this liminal state. Maybe not.


workingjubilee:
Only one way to find out.

workingjubilee:
@eeeebbbbrrrr's remarks:
"Wheeee!"

PR #754, opened 2022-10-11
Only specify `-undefined dynamic_lookup` on macOS (to allow linking with `mold` on linux)

thomcc:
Currently, we pass `-Clink-arg=-Wl,-undefined,dynamic_lookup` on all targets. This only has meaning on macOS, and if you try to use certain alternative linkers, such as `mold`, it will complain that it doesn't understand this flag. This patch changes us to only pass it in on targets that need it.

Note that this does not make `mold` work on macOS -- `mold` does not support the `-undefined` flag. This just allows using it on other targets, since it will currently always complain about it.

That said, users have to mimic this change too to get the benefit on existing projects, since it's a change to the template.

---

While I was here, I also switched us from `.cargo/config` to `.cargo/config.toml`, which is preferred according to the [documentation](https://doc.rust-lang.org/cargo/reference/config.html). This is a cosmetic change, and is separated into the 2nd commit.

> Note: Cargo also reads config files without the .toml extension, such as .cargo/config. Support for the .toml extension was added in version 1.39 and is the preferred form. If both files exist, Cargo will use the file without the extension.


thomcc:
Hm, that's odd.

thomcc:
Ah, I see: <https://github.com/tcdi/pgx/blob/master/.github/workflows/tests.yml#L363-L366>

That's annoying...

BradyBonnette:
@thomcc Yeah it's some weird flaw in the Github Actions mac runner. It does not pick up that file. Although, I now wonder if it's because the runners are looking for `.cargo/config.toml` instead of `.cargo/config`

thomcc:
> Although, I now wonder if it's because the runners are looking for .cargo/config.toml instead of .cargo/config

This and the fact that it's in `target.'cfg(target_os="macos")'.rustflags` rather than `build.rustflags`. A hacky fix is doable but it might be worth figuring out why GHA isn't picking up that file...

BradyBonnette:
I'd love nothing more than to remove those lines from the workflow file. I always felt weird putting them in there, but only because I didn't quite understand why I had to.

thomcc:
The issue is we specify `RUSTFLAGS: -Copt-level=0` [here](https://github.com/tcdi/pgx/blob/master/.github/workflows/tests.yml#L14). Specifying `RUSTFLAGS` via environment variable will override it (nearly) anywhere else, as documented here: <https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags>. 

I actually do have to wonder if this helps, given that locally, -Copt-level=0 makes the builds slower (due to making all the code that runs at build-time unoptimized)

BradyBonnette:
iirc, we added that environment variable specifically to help speed up build times. I can't recall where I saw it however, but it had something to do with the fact that increasing optimizations made build times slower.

BradyBonnette:
That is to also say, we're not completely bound to using it at all costs. If we have reasons that removing/adjusting it makes everything better, then by all means.

thomcc:
I think we'll avoid it on macOS since the alternative is more trouble than it's worth.

Moving `-Copt-level=0` down into the other env configuration for the other jobs is easy, and should allow us to avoid needing to extract the toml from the config, which also avoids needing to install `tomlq`/`yq` at all.

The only things that were built as release should be the ones that get it via [`profile.dev.build-override.opt-level`](https://github.com/tcdi/pgx/blob/master/Cargo.toml#L32-L33), and reducing that to `0` significantly slows down over-all compile-times on my machine, since we do some significant work in proc macros, and they benefit from optimizations.

thomcc:
The other easy option would be to have a `.cargo/config.ci.toml` that also specifies `opt-level=0` and such, and do a `mv .cargo/config.ci.toml .cargo/config.toml` inside CI. This is more fiddly though, but might avoid surprises in the future.

BradyBonnette:
> The other easy option would be to have a .cargo/config.ci.toml

Could work, but my only concern there is that if things change in the future, we may forget about its existence and might miss out on needed updates. Hmm..

eeeebbbbrrrr (approved):

PR #756, opened 2022-10-11
Avoid some dependencies in pgx-pg-sys

thomcc:
1. We were using `color_eyre::Result` despite not configuring `color-eyre`, so this was a pointless dependency.

2. We were using `rustversion` to determine if we're using nightly. This is a proc-macro we were using in `pgx-pg-sys`, which adds an extra build phase to it's build script (to run the proc macro). While switching to the `rustc_version` would have avoided it, it also pulls in a dependency on the `semver` crate -- and really both of these are deps that are just to avoid a small bit of manual parsing of the `rustc --version` output, which is... not very complex to parse.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1438,7 +1438,6 @@ dependencies = [
  "pgx-utils",
  "proc-macro2",
  "quote",
- "rustversion",
  "shlex",
  "sptr",
  "syn",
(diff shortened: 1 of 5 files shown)


thomcc:
I also removed tracing from `pgx-pg-config`. It was only used in one places and pulls in a large pile of its own dependencies.

eeeebbbbrrrr (approved):

PR #757, opened 2022-10-12
Avoid wrapping variadic functions in pgx-pg-sys and `#[pg_guard]`

thomcc:
This solves the issue mentioned here: https://github.com/tcdi/pgx/issues/730#issuecomment-1276295238


workingjubilee (approved):
Error handling, nice. I expect this is liable to turn up other issues soon but that's a Feature of this PR rather than a bug.

thomcc:
Right, any issues this turns up are somewhere between "highly dubious" and "straight up unsound".

PR #764, opened 2022-10-14
prepare for a quick 0.5.1

eeeebbbbrrrr:
This fixes a problem where the SQL name for the Postgres `pgx::pgx_pg_sys::Point` type is incorrect.  It should be `point`. (#763)


BradyBonnette (approved):
👍 LGTM

PR #765, opened 2022-10-14
Problem: working with independent extensions simultaneously

yrashk:
As I am developing a few small extensions, I tend to switch between them, leaving the database in a state that's not necessarily stable, and therefore sharing the same instance of PostgreSQL is a bit problematic.

I can use PGX_HOME to use independent installations of PostgreSQL but they compete for the same set of ports.

Solution: make ports configurable during `pgx init` and in `config.toml`

### DIFF
--- a/cargo-pgx/README.md
+++ b/cargo-pgx/README.md
@@ -107,7 +107,7 @@ As shown by the screenshot above, it downloads the latest versions of Postgres v
 In cases when default ports pgx uses to run PostgreSQL within are not available, one can specify
 custom values for these during initialization using `--base-port` and `--base-testing-port`
 options. One of the use cases for this is using multiple installations of pgx (using `$PGX_HOME` variable)
-when developing multiple extensions at the same time.
+when developing multiple extensions at the same time. These values can be later changed in `$PGX_HOME/config.toml`.
 
 If you want to use your operating system's package manager to install Postgres, `cargo pgx init` has optional arguments that allow you to specify where they're installed (see below).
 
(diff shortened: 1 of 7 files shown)


eeeebbbbrrrr:
This generally looks okay.  If you will, please update the cargo-pgx README.md file with the new options and some discussion around why they exist and how to use them and such.

eeeebbbbrrrr (changes requested):
Please update the cargo-pgx README.md file as well.

yrashk:
@eeeebbbbrrrr I've updated README.

thomcc (approved):
This looks fine to me.

cargo-pgx/README.md
```diff
@@ -104,6 +104,11 @@ As shown by the screenshot above, it downloads the latest versions of Postgres v
 
 `pgx` is designed to support multiple Postgres versions in such a way that during development, you'll know if you're trying to use a Postgres API that isn't common across all versions. It's also designed to make testing your extension against these versions easy. This is why it requires you to have all fully compiled and installed versions of Postgres during development.
 
+In cases when default ports pgx uses to run PostgreSQL within are not available, one can specify
+custom values for these during initialization using `--base-port` and `--base-testing-port`
+options. One of the use cases for this is using multiple installations of pgx (using `$PGX_HOME` variable)
+when developing multiple extensions at the same time.
+
```
eeeebbbbrrrr, line 111: @yrashk I think it'd be good to also mention here that they can be later changed by hand-editing `$PGX_HOME/config.toml`.

Once you do that, this will be good to merge.

Thanks a bunch!

eeeebbbbrrrr (changes requested):

cargo-pgx/README.md
yrashk: @eeeebbbbrrrr done

workingjubilee:
Actually wait, I have eyes. Yeah, this looks correct. Apologies.

PR #772, opened 2022-10-17
Remove lifetimes from FunctionMetadata impl

thomcc:
These prevent something like the following, which I believe we want to support:

```rs
#[pg_extern]
fn example(s: &str) -> String {
    s.into()
}
```

That is, they prevent accepting a non-static input and returning a static output. I think they're intended to prevent code from returning a non-static lifetime that outlives the input (but came from it), but the only way for it to do that would be unsafe code, and unsafe code can already lie about lifetimes.


workingjubilee:
I believe it's more likely they were caused by pure thrashing at the time, since there was a lot to try to get working and technically unnecessary lifetime annotations are a common consequence of churn in Rust code.

workingjubilee (approved):
I'm approving this on the belief that the lifetime annotations are supererogatory.

PR #775, opened 2022-10-17
Added includes for "storage/spin.h"

osawyerr:
Added PR for https://github.com/tcdi/pgx/issues/752 - "Consider making SpinLock methods accessible in PGX".

### DIFF
--- a/pgx-pg-sys/include/pg10.h
+++ b/pgx-pg-sys/include/pg10.h
@@ -88,6 +88,7 @@ Use of this source code is governed by the MIT license that can be found in the 
 #include "storage/itemptr.h"
 #include "storage/lwlock.h"
 #include "storage/procarray.h"
+#include "storage/spin.h"
 #include "tcop/tcopprot.h"
 #include "tcop/utility.h"
 #include "tsearch/ts_public.h"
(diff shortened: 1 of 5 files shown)


thomcc (approved):
Looks good to me if CI comes back green.

osawyerr:
All green

thomcc:
A concern here that I realized is that our setjmp shim code (which we insert here and elsewhere) is not thread-safe. So I'm not really sure about this anymore.

osawyerr:
Ok... what's the way forward with this?

thomcc:
I'm not sure. @workingjubilee may have an idea.

osawyerr:
Ah I see. Since this is already used elsewhere, is any greater risk of adding this PR till a permanent fix can be found?

thomcc:
Well, these functions are explicitly *for* use in multiple threads. Which will always be (quite) unsafe. The existing ones aren't, and have a caveat in the readme to this effect (see https://github.com/tcdi/pgx#caveats--known-issues).

osawyerr:
I thought that they were meant to be used in multiple _processes_ not multiple _threads_. Example is in **nbtree.c** in postgres.

thomcc:
Oh, sorry, my bad. In that case these should be fine.

eeeebbbbrrrr:
> I thought that they were meant to be used in multiple _processes_ not multiple _threads_. Example is in **nbtree.c** in postgres. Functionality we are building is intended to be used like in nbtree.c

Correct!  Postgres doesn't use threads and all its various locking primitives are designed to be cross-*process*.  So this PR is fine.  We should definitely merge it.

In fact, I'm doing that now...

PR #777, opened 2022-10-18
Detect and complain about use of `pgx-pg-sys` functions from multiple threads

thomcc:
Doing this will likely break stuff like #775, although that code is already thread-unsafe (due to our setjmp shim)... That said, it reduces UB from data races (at best) to a panic with a reasonable backtrace.

### DIFF
--- a/pgx-pg-sys/src/submodules/setjmp.rs
+++ b/pgx-pg-sys/src/submodules/setjmp.rs
@@ -45,6 +45,7 @@ being converted into a transaction-aborting Postgres `ERROR` by PGX.
 
 **/
 #[inline(always)]
+#[track_caller]
 pub(crate) unsafe fn pg_guard_ffi_boundary<T, F: FnOnce() -> T>(f: F) -> T {
     // SAFETY: Caller promises not to call us from anything but the main thread.
     unsafe { pg_guard_ffi_boundary_impl(f) }
@@ -52,9 +53,14 @@ pub(crate) unsafe fn pg_guard_ffi_boundary<T, F: FnOnce() -> T>(f: F) -> T {
 
 #[cfg(not(feature = "postgrestd"))]
 #[inline(always)]
+#[track_caller]
 unsafe fn pg_guard_ffi_boundary_impl<T, F: FnOnce() -> T>(f: F) -> T {
     //! This is the version that uses sigsetjmp and all that, for "normal" Rust/PGX interfaces.
     use crate as pg_sys;
+
+    // The next code is definitely thread-unsafe (it manipulates statics in an
+    // unsynchronized manner), so we may as well check here.
+    super::thread_check::check_active_thread();
 
     // SAFETY: This should really, really not be done in a multithreaded context as it
     // accesses multiple `static mut`. The ultimate caller asserts this is the main thread.
--- a/pgx-pg-sys/src/submodules/thread_check.rs
+++ b/pgx-pg-sys/src/submodules/thread_check.rs
@@ -1,0 +1,77 @@
+//! Enforces thread-safety in `pgx`.
+//!
+//! It's UB to call into Postgres functions from multiple threads. We handle
+//! this by remembering the first thread to call into postgres functions, and
+//! panicking if we're called from a different thread.
+//!
+//! This is called from the current crate from inside the setjmp shim, as that
+//! code is *definitely* unsound to call in the presense of multiple threads.
+//!
+//! This is somewhat heavyhanded, and should be called from fewer places in the
+//! future...
+
+use std::num::NonZeroUsize;
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+#[track_caller]
+pub(crate) fn check_active_thread() {
+    static ACTIVE_THREAD: AtomicUsize = AtomicUsize::new(0);
+    let current_thread = nonzero_thread_id();
+    // Relaxed is sufficient as we're only interested in the effects on a single
+    // atomic variable, and don't need synchronization beyond that.
+    return match ACTIVE_THREAD.load(Ordering::Relaxed) {
+        0 => init_active_thread(current_thread, &ACTIVE_THREAD),
+        thread_id => {
+            if current_thread.get() != thread_id {
+                thread_id_check_failed();
+            }
+        }
+    };
+}
+
+#[track_caller]
+fn init_active_thread(tid: NonZeroUsize, atom: &AtomicUsize) {
+    match atom.compare_exchange(0, tid.get(), Ordering::Relaxed, Ordering::Relaxed) {
+        Ok(_) => {
+            // All good!
+        }
+        Err(_) => {
+            thread_id_check_failed();
+        }
+    }
+}
+
+#[cold]
+#[inline(never)]
+#[track_caller]
+fn thread_id_check_failed() -> ! {
+    panic!("`pgx` may not functionality may not be used from multiple threads.");
+}
+
+fn nonzero_thread_id() -> NonZeroUsize {
+    // Returns the `addr()` of a thread local variable.
+    //
+    // For now this is reasonably efficient, but could be (substantially, for
+    // our use) improved by using a pointer to the thread control block, which
+    // can avoid going through `__tls_get_addr`.
+    //
+    // Sadly, doing that would require some fiddley platform-specific inline
+    // assembly, which is enough of a pain that it's not worth bothering with
+    // for now. That said, in the future if this becomes a performance issue,
+    // that'd be a good fix.
+    //
+    // For examples of how to do this, see the how checks for cross-thread frees
+    // on are implemented in thread-pooling allocators, ex:
+    // - https://github.com/microsoft/mimalloc/blob/f2712f4a8f038a7fb4df2790f4c3b7e3ed9e219b/include/mimalloc-internal.h#L871
+    // - https://github.com/mjansson/rpmalloc/blob/f56e2f6794eab5c280b089c90750c681679fde92/rpmalloc/rpmalloc.c#L774
+    // and so on.
+    std::thread_local! {
+        static BYTE: u8 = const { 0 };
+    }
+    BYTE.with(|p: &u8| {
+        // Note: Avoid triggering the `unstable_name_collisions` lint.
+        let addr = sptr::Strict::addr(p as *const u8);
+        // SAFETY: `&u8` is always nonnull, so it's address is always nonzero.
+        unsafe { NonZeroUsize::new_unchecked(addr) }
+    })
+}
--- a/pgx-pg-sys/src/submodules/thread_check.rs
+++ b/pgx-pg-sys/src/submodules/thread_check.rs
@@ -19,14 +19,14 @@ pub(crate) fn check_active_thread() {
     let current_thread = nonzero_thread_id();
     // Relaxed is sufficient as we're only interested in the effects on a single
     // atomic variable, and don't need synchronization beyond that.
-    return match ACTIVE_THREAD.load(Ordering::Relaxed) {
+    match ACTIVE_THREAD.load(Ordering::Relaxed) {
         0 => init_active_thread(current_thread),
         thread_id => {
             if current_thread.get() != thread_id {
                 thread_id_check_failed();
             }
         }
-    };
+    }
 }
 
 #[track_caller]
@@ -50,7 +50,7 @@ fn init_active_thread(tid: NonZeroUsize) {
 #[inline(never)]
 #[track_caller]
 fn thread_id_check_failed() -> ! {
-    panic!("`pgx` may not not be used from multiple threads.");
+    panic!("postgres FFI may not not be called from multiple threads.");
 }
 
 fn nonzero_thread_id() -> NonZeroUsize {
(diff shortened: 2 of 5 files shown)


thomcc:
It looks like I was mislead by the docs I read -- this should always be fine then.

thomcc:
Hmm, actually, maybe I need to set an atfork handler to clear this in the child...

pgx-pg-sys/src/submodules/thread_check.rs
```diff
@@ -0,0 +1,77 @@
...
+            thread_id_check_failed();
+        }
+    }
+}
+
+#[cold]
+#[inline(never)]
+#[track_caller]
+fn thread_id_check_failed() -> ! {
+    panic!("`pgx` may not functionality may not be used from multiple threads.");
```
eeeebbbbrrrr, line 48: grammar

pgx-pg-sys/src/submodules/setjmp.rs
```diff
@@ -45,17 +45,23 @@ being converted into a transaction-aborting Postgres `ERROR` by PGX.
...
 #[cfg(not(feature = "postgrestd"))]
 #[inline(always)]
+#[track_caller]
 unsafe fn pg_guard_ffi_boundary_impl<T, F: FnOnce() -> T>(f: F) -> T {
     //! This is the version that uses sigsetjmp and all that, for "normal" Rust/PGX interfaces.
     use crate as pg_sys;
 
+    // The next code is definitely thread-unsafe (it manipulates statics in an
+    // unsynchronized manner), so we may as well check here.
+    super::thread_check::check_active_thread();
```
eeeebbbbrrrr, line 63: Can we `#[cfg]` this out of `--release` builds?  Or make it an `assert!()` or have it behind a feature flag that's disabled by default or something?

We really don't need additional overhead for every call back into Postgres for something that's most likely not going to happen in the typical pgx extension.

eeeebbbbrrrr (dismissed):

pgx-pg-sys/src/submodules/thread_check.rs
thomcc: Whoops, looks like I only half-rewrote it. Nice catch.

pgx-pg-sys/src/submodules/setjmp.rs
thomcc: Hmm, well, it's still unsound. I think we could have a feature to disable this

That said I could also just make this be a single load from `%fs:0`/`tpidr_el0`, as mentioned in the PR. In which case it should only be a few-cycle load and a (easily predicted) branch in the worst case, likely not much worse than array bounds checks.

I'm more concerned that this might break the `fork` case, which I need to do some testing with.

eeeebbbbrrrr: My view is that the ideal situation is that it's compiled out of `--release` builds.

The reasoning here is that basically all pgx extensions are going to be single-threaded, just because Postgres is.  If an extension does threading, it'd likely be "accidentally" from some 3rd-party crate like rayon or crossbeam.

Keeping the check for debug builds would, in a perfect world, catch these problems during development and tests.

If we want a feature like `pedantic-threadness-checks` (or whatever) that can be enabled by the developer for `--release` builds, that'd be fine.  I do see the usefulness of this in a release build for overly-paranoid runtime environments.

thomcc: I'm not sure. It seems pretty easy to accidentally end up calling into PGX from multiple threads given that so much of rust assumes thread-safety by default (outside of unsafe)...

> EDIT: As an aside, what kind of runtime overhead does #[track_caller] add per call?

Adds a secret argument of an `&'static Location<'static>`, which gets loaded from static data and passed in. Concretely, less than passing an extra `&'static str` argument.

eeeebbbbrrrr: Well.  Then I'd like to see some kind of "benchmark" to see if this shows up.

And I don't really mean anything scientific here.  Like, does calling this function (https://github.com/tcdi/pgx/blob/d1ffdb2969f414652d48548341e9d9ff9d02e672/pgx-examples/srf/src/lib.rs#L15) through SQL as something like `SELECT srf.generate_series(1, 10000000);` show any measurable difference with and without this check?

Where I'm concerned about overhead is in a case like this where a `#[pg_extern]` function is being called 10s-or-100s-of-millions-of-times by some SQL statement.

thomcc: Sure, a request for a benchmark is totally reasonable.

eeeebbbbrrrr: For example, the Timescale folks have mentioned `setjmp` overhead already.  My intuition says this would pop up on their radar too.  ZomboDB too, for that matter.

If it doesn't turn out to be significant, then great!

thomcc: This should be significantly cheaper than the call to sigsetjmp, but I'll take the time to ensure that.

thomcc: I measured this in a few ways, both as standalone benchmarks, benchmarks that include the rest of the pgx sigsetjmp error-handling dance, and in real queries (`SELECT my_generate_series(0, 10000000, 1)`, with `\timing` on). I did this on linux and macOS, which reported similar numbers (although Linux numbers were like 10x slower across the board due to the machine being literally the cheapest laptop I could find 3y ago).

The check in this PR takes around 700ps standalone, and increases the average cost the error handling from ~6ns to ~6.5ns in my benchmarks (there's probably around 200ps of overhead from measurement).

When used in a query, these checks had no reliable overhead that I could detect. With or without the thread safety check, `SELECT my_generate_series(0, 10000000, 1)`, reported times between 950ms and 975ms (well, 4.5-5ms on my linux box). I tested this pretty thoroughly (several DBs/extensions), and am confident that it's not a case of the old code remaining in use.

We could get this down to essentially 0-cost[^1] via the use of inline asm. I have implemented this in the `thomcc/threadsafe-asm` branch (the relevant code is https://github.com/tcdi/pgx/blob/thomcc/threadsafe-asm/pgx-pg-sys/src/submodules/thread_check.rs#L115-L194). It wouldn't be *that* hard for me to maintain (I only implemented versions I understood), but is definitely way worse than the version in this PR, so I don't know that I think this is worth it, since the change is not measurable in queries[^2].

It's definitely your call, though -- I'd be totally fine to pull that into this PR if you'd rather take the optimization.

[^1]: Small enough to be comparable with `black_box(0)` on its own (100ps), and to cause no measurable difference when run as part of the other pg error code.

[^2]: I wouldn't have bothered with this had I realized that, but my initial measurements accidentally compared versions with different compiler flags, so I thought the overhead was much more significant at first.

thomcc:
Because it got mentioned in #780 (and because I realized I was making a dumb mistake in the macOS reloading issue, and needed to look at something else), I finished this up. See https://github.com/tcdi/pgx/pull/777#discussion_r1008613495 for a description of the overhead (very small, not measurable in queries) and a link to an alternate approach that is (much) worse to maintain but is basically 0 cost on the platforms we care about.

workingjubilee:
I'm gonna need to make sure this works with PL/Rust before we merge this.

thomcc:
> I'm gonna need to make sure this works with PL/Rust before we merge this.

The only call to the check is inside a `cfg(not(feature = "postgrestd"))` function, but yeah, worth double-checking. (Actually, there might be a lot of unused code warnings if that feature is on...).

pgx-pg-sys/src/submodules/thread_check.rs
```diff
@@ -0,0 +1,82 @@
...
+    // - https://github.com/microsoft/mimalloc/blob/f2712f4a8f038a7fb4df2790f4c3b7e3ed9e219b/include/mimalloc-internal.h#L871
+    // - https://github.com/mjansson/rpmalloc/blob/f56e2f6794eab5c280b089c90750c681679fde92/rpmalloc/rpmalloc.c#L774
+    // and so on.
+    std::thread_local! {
+        static BYTE: u8 = const { 0 };
+    }
+    BYTE.with(|p: &u8| {
+        // Note: Avoid triggering the `unstable_name_collisions` lint.
+        let addr = sptr::Strict::addr(p as *const u8);
+        // SAFETY: `&u8` is always nonnull, so it's address is always nonzero.
```
workingjubilee, line 79:
```suggestion
        // SAFETY: `&u8` is always nonnull, so its address is always nonzero.
```

pgx-pg-sys/src/submodules/thread_check.rs
```diff
@@ -0,0 +1,82 @@
...
+
+#[cold]
+#[inline(never)]
+#[track_caller]
+fn thread_id_check_failed() -> ! {
+    panic!("`pgx` may not not be used from multiple threads.");
+}
+
+fn nonzero_thread_id() -> NonZeroUsize {
+    // Returns the `addr()` of a thread local variable.
```
workingjubilee, line 57: wow.
cute.

pgx-pg-sys/src/submodules/thread_check.rs
```diff
@@ -0,0 +1,82 @@
...
+use std::num::NonZeroUsize;
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+static ACTIVE_THREAD: AtomicUsize = AtomicUsize::new(0);
+#[track_caller]
+pub(crate) fn check_active_thread() {
+    let current_thread = nonzero_thread_id();
+    // Relaxed is sufficient as we're only interested in the effects on a single
+    // atomic variable, and don't need synchronization beyond that.
+    return match ACTIVE_THREAD.load(Ordering::Relaxed) {
```
workingjubilee, line 22: Unless there's a specific reason that this final expression has an explicit return?
```suggestion
    match ACTIVE_THREAD.load(Ordering::Relaxed) {
```

pgx-pg-sys/src/submodules/thread_check.rs
```diff
@@ -0,0 +1,82 @@
...
+            thread_id_check_failed();
+        }
+    }
+}
+
+#[cold]
+#[inline(never)]
+#[track_caller]
+fn thread_id_check_failed() -> ! {
+    panic!("`pgx` may not not be used from multiple threads.");
```
workingjubilee, line 53: We should be more clear: it is possible for a function used by PGX to have no FFI involved, which is sometimes the case when Rust is used to construct data structures for later transfer into Postgres, and it's fine to do that off the main thread... but you can't then call into Postgres.
```suggestion
    panic!("postgres FFI may not not be called from multiple threads.");
```

pgx-pg-sys/src/submodules/thread_check.rs
```diff
@@ -0,0 +1,82 @@
...
+            }
+        }
+    };
+}
+
+#[track_caller]
+fn init_active_thread(tid: NonZeroUsize) {
+    match ACTIVE_THREAD.compare_exchange(0, tid.get(), Ordering::Relaxed, Ordering::Relaxed) {
+        Ok(_) => unsafe {
+            // We won the race. Register an atfork handler to clear the atomic
```
workingjubilee, line 36: [Here he comes, here comes Speed Racer, he's a demon on wheels...](https://www.youtube.com/watch?v=EChZbd0HeGw)

workingjubilee (approved):
This PR passed PL/Rust's test suite in https://github.com/tcdi/plrust/pull/105

700 _picoseconds_ is obviously worth the improvement in soundness, and unlike the previous version, this isn't likely to break on no-op thread renaming antics.

pgx-pg-sys/src/submodules/thread_check.rs
thomcc: Oh, it used to have the helper functions defined locally in that function, at the bottom after the return. Nice catch.

pgx-pg-sys/src/submodules/thread_check.rs
thomcc: You don't think the `asm!` version was cuter? 🐱

pgx-pg-sys/src/submodules/thread_check.rs
thomcc: I'll add this to the batch commit but it'll deffo fail afterwards since there's a semicolon on the end.

pgx-pg-sys/src/submodules/thread_check.rs
```diff
@@ -0,0 +1,82 @@
...
+    // Relaxed is sufficient as we're only interested in the effects on a single
+    // atomic variable, and don't need synchronization beyond that.
+    return match ACTIVE_THREAD.load(Ordering::Relaxed) {
+        0 => init_active_thread(current_thread),
+        thread_id => {
+            if current_thread.get() != thread_id {
+                thread_id_check_failed();
+            }
+        }
+    };
```
workingjubilee, line 29:
```suggestion
    }
```

pgx-pg-sys/src/submodules/thread_check.rs
workingjubilee: _wow_.

PR #779, opened 2022-10-18
Mention pitfall in safety documentation

Smittyvb:
When writing SRFs that return a function of the input arguments with lifetimes, it's easy to accidentally create a use after free. This PR documents that pitfall in the safety documentation.

e.g. this code

```rust
struct State<'a> { ... }
#[pg_extern]
fn foo<'a>(state: State<'a>) -> TableIterator<'a, (pgx::name!(val, i32))> {
    TableIterator::new(
        state.vals.clone().map(|x| (x + 1,))
    )
}
```

has a use after free. See [this disscusion](https://discord.com/channels/561648697805504526/749454413407584258/1029497569925144626).

### DIFF
--- a/SAFETY.md
+++ b/SAFETY.md
@@ -16,8 +16,8 @@ Quirks specific to Postgres.
 The `palloc*` family of functions may throw a Postgres error but will not return `nullptr`.
 
 You should be careful to avoid using data allocated by Postgres after it has been freed. In particular,
-when writing set returning functions, don't directly return data from the arguments without cloning the
-data into memory owned by Rust.
+when writing set returning functions, don't directly return referenced data from the arguments without
+cloning the data into memory owned by Rust.
 
 ## Rust
 
--- a/SAFETY.md
+++ b/SAFETY.md
@@ -15,6 +15,10 @@ Quirks specific to Postgres.
 
 The `palloc*` family of functions may throw a Postgres error but will not return `nullptr`.
 
+You should be careful to avoid using data allocated by Postgres after it has been freed. In particular,
+when writing set returning functions, don't directly return data from the arguments without cloning the
+data into memory owned by Rust.
+
 ## Rust
 
 Quirks specific to Rust that specifically inform the design of this crate and not, say,


eeeebbbbrrrr:
It definitely sucks that the compiler can't see this.

If you can share, what's the full definition for `State` and what does its `FromDatum` impl look like?

thomcc:
How common is the use case for returning borrowed data from `pg_extern`s? We could make it returned owned data by changing https://github.com/tcdi/pgx/blob/develop/pgx-utils/src/sql_entity_graph/metadata/function_metadata.rs#L87 to `Output: SqlTranslatable + 'static`.

eeeebbbbrrrr:
> How common is the use case for returning borrowed data from pg_externs?

I'd say it's actually impossible for set-returning-functions because whatever is being returned has to be allocated in a special MemoryContext.

But it's perfectly fine for something like:

```rust
#[pg_extern]
fn substring(input: &str, s: i32, e: i32) -> &str {
   &input[s as usize..e as usize]
}
```

SRFs are special because the same function is called repeatedly until it's "done".  The wrapper function pgx generates takes over for that process after the first call `(SRF_FIRST_CALL in Postgres-speak)`.

I was on leave when these Iterator functions were invented.  I'm not sure why they have a lifetime at all, actually.

Smittyvb:
In my case `State` is generated by a macro; [here's the expanded form](https://gist.github.com/Smittyvb/c6541dabacd2513c8360285876fe9a35) (and [here's the unexpanded form](https://github.com/timescale/timescaledb-toolkit/blob/fe8b3861ee12b1cebedfb32a0c4fce2b9ce508df/extension/src/frequency.rs#L376-L388)). I fixed it by cloning all the data: https://github.com/timescale/timescaledb-toolkit/commit/fe8b3861ee12b1cebedfb32a0c4fce2b9ce508df.

eeeebbbbrrrr:
> In my case `State` is generated by a macro; [here's the expanded form](https://gist.github.com/Smittyvb/c6541dabacd2513c8360285876fe9a35) (and [here's the unexpanded form](https://github.com/timescale/timescaledb-toolkit/blob/fe8b3861ee12b1cebedfb32a0c4fce2b9ce508df/extension/src/frequency.rs#L376-L388)). I fixed it by cloning all the data: [timescale/timescaledb-toolkit@fe8b386](https://github.com/timescale/timescaledb-toolkit/commit/fe8b3861ee12b1cebedfb32a0c4fce2b9ce508df).

I see.  

If you could somehow come up with an "owned" version of  `SpaceSavingAggregateData` that you can create in that function you could at least avoid collecting the results into a `Vec` first.

If it were a little simpler, another option would be to deep-copy the `agg` argument into the SRF "multi call memory context".  Like so (trigger warning: gross code):

```rust
#[pg_extern]
fn foo<'a, 'b>(
    s: &'a str,
) -> TableIterator<'b, (name!(i, i32), name!(ch, String))> {
    let funcctx = unsafe { pgx::srf_per_call_setup(fcinfo) };
    let copied_s_ptr = unsafe {
        PgMemoryContexts::For(funcctx.multi_call_memory_ctx)
            .copy_ptr_into(s.as_ptr() as *mut _, s.len())
    };
    let s: &[u8] = unsafe { std::slice::from_raw_parts(copied_s_ptr, s.len()) };
    let s = unsafe { std::str::from_utf8_unchecked(s) };
    TableIterator::new(s.chars().enumerate().map(|(i, c)| (i as i32, String::from(c))))
}
```

I think the answer to the compiler not seeing this problem is for us to remove the lifetime annotations on `TableIterator` and `SetOfIterator`.  They just can't be allowed to borrow if we expect the compiler to help us here.

thomcc:
> I think the answer to the compiler not seeing this problem is for us to remove the lifetime annotations on TableIterator and SetOfIterator

This seems like it would break cases like https://github.com/tcdi/pgx/blob/933fb3527f4e2c3bb2d6d69cc03ab78f9d42d301/pgx-examples/strings/src/lib.rs#L41, although I don't know how much we care about that.

eeeebbbbrrrr:
> This seems like it would break cases like 

It's broken now...

If the input to that function is toasted (ie, it came from disk and is large enough to be stored out-of-line), then the detoasted version, which pgx uses for the `&str` reference, is allocated in the wrong memory context.  Example:

```sql
[v13.5][66988] strings=# \d foo
               Table "public.foo"
 Column | Type | Collation | Nullable | Default 
--------+------+-----------+----------+---------
 s      | text |           |          | 

[v13.5][66988] strings=# select length(s) from foo;
 length  
---------
 1999999
(1 row)

Time: 9.226 ms
[v13.5][66988] strings=# select strings.split_set(s, ' ') from foo;
                                                                                                                                     
-------------------------------------------------------------------------------------------------------------------------------------
 a
 ^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?^?
(2 rows)
```

eeeebbbbrrrr:
> ... is allocated in the wrong memory context.

Maybe what we should do is teach the SRF wrapper function it generates to get us switched into the multi-call-memory-context **before** calling the user's function.  That would fix things as well.

thomcc:
It would be pretty easy to require 'static there, but it would likely make some things that users want to write need to take a less efficient path (although, inefficient is much better than broken!)

eeeebbbbrrrr:
Well, I'm not sure 'static is the right answer.  The more I think about this, the more I think I've changed my mind and that it should be possible for SRF functions to borrow from their input -- we just need to make sure the input is allocated in the proper memory context.  

I'll test this out and put up a PR shortly...

eeeebbbbrrrr:
... by shortly I mean tomorrow!

But I think this is the problem.  We're `FromDatum`-ing the arguments in the wrong memory context for SRF functions.

thomcc:
> we just need to make sure the input is allocated in the proper memory context.

Oh, that sounds much better.

I did take a stab at removing the lifetime and moving it to `'static` (which was fruitful either way, in so far as I got a bit more familiarity with some of the stuff `pgx-{utils,macros}` does...), which is [here](https://github.com/thomcc/pgx/tree/iter-static) if things don't pan out. But I think the approach you suggested sounds much better -- there are a few aspects of forcing 'static here that I don't really love...

eeeebbbbrrrr:
Yeah, using that same SQL example above, it works.  I'll need to adjust the wrapper function for `TableIterator` as well.

What a gross bug!  I'm glad @Smittyvb brought this to our attention!

```sql
[v13.5][68846] strings=# select strings.split_set(s, ' ') from foo;
 split_set 
-----------
 a
 a
 a
 a
 a
 a
 a
 ...
```

eeeebbbbrrrr:
pgx's schema generator doesn't recognize there's a lifetime attached to the the second element of the return tuple in:

```rust
#[pg_extern]
fn split_table<'a>(
    input: &'a str,
    pattern: &'a str,
) -> TableIterator<'a, (name!(i, i32), name!(s, &'a str))> {
    TableIterator::new(input.split_terminator(pattern).enumerate().map(|(i, s)| (i as i32, s)))
}
```

It's not properly handling `name!(s, &'a str)`.  I'll have to study this one.  The parse tree for that isn't very nice.

`Macro(TypeMacro { mac: Macro { path: Path { leading_colon: None, segments: [PathSegment { ident: Ident { ident: "name", span: #0 bytes(1131..1135) }, arguments: None }] }, bang_token: Bang, delimiter: Paren(Paren), tokens: TokenStream [Ident { ident: "s", span: #0 bytes(1137..1138) }, Punct { ch: ',', spacing: Alone, span: #0 bytes(1138..1139) }, Punct { ch: '&', spacing: Alone, span: #0 bytes(1140..1141) }, Punct { ch: '\'', spacing: Joint, span: #0 bytes(1141..1143) }, Ident { ident: "a", span: #0 bytes(1141..1143) }, Ident { ident: "str", span: #0 bytes(1144..1147) }] } })`

SAFETY.md
```diff
@@ -15,6 +15,10 @@ Quirks specific to Postgres.
 
 The `palloc*` family of functions may throw a Postgres error but will not return `nullptr`.
 
+You should be careful to avoid using data allocated by Postgres after it has been freed. In particular,
+when writing set returning functions, don't directly return referenced data from the arguments without
+cloning the data into memory owned by Rust.
```
workingjubilee, line 20:
```suggestion
cloning the data into memory with a sufficient lifetime (either allocated by Rust through a system memory allocator, or via Postgres using `pg_sys::SPI_palloc` (for SPI connections) or PGX's `PgMemoryContexts` interface.
```

workingjubilee (changes requested):
I think this should inform more directly about the principles involved, rather than what feels a bit like "just" an ominous warning... even if it includes what is otherwise basically the same caveat.

You don't have to use my wording here.
Ideally we could use compile-time enforcement, but this document would still stand even if (this document is most definitely intended to be aimed at us, the people who are implementing the safety checks, to make sure the safety checks are valid).

SAFETY.md
eeeebbbbrrrr: FYI. There’s actually a pgx bug here. And I’m working on fixing it.  

Once fixed we won’t need this PR at all.

workingjubilee: Hmm, my idea for this documentation here is that it should be mostly design-level: it should contain whatever principles we need to be mindful of in order to write PGX correctly, which includes the ideas that exist beyond the current implementation.

eeeebbbbrrrr: Most of the suggested changes going back-n-forth are informed by bugged behavior.

workingjubilee: The real name of this document is probably going to wind up something like `PRIMER_ON_POSTGRES_MEMORY_CONTEXTS.md` tbh.

eeeebbbbrrrr: I’ve got the beginnings of a PR that will address this.

eeeebbbbrrrr:
Now that https://github.com/tcdi/pgx/pull/784 is up, I think we can just close this.

I'll do my best to talk about how SRFs are (now) handled in the documentation I'm writing for #780.

PR #780, opened 2022-10-18
Initial stab at a design doc

eeeebbbbrrrr:
This is just an outline.  Lots of details to fill in which I'll be doing over the next few days.

If anyone sees anything that should be discussed that isn't already mentioned, please leave some comments.  Same goes if anyone just wants to expand on any given thing already outlined (ie, do my work for me).


DESIGN.md
```diff
@@ -0,0 +1,57 @@
...
+- No user-discernible difference between a Rust `panic!()` and a Postgres `error!()`
+- `#[pg_guard]` procmacro for wrapping Rust `extern "C"` functions that need to be passed to Postgres as function pointers
+  - Also ensures proper stack unwinding across FFI boundaries
+- `fn pg_try(||) -> PgTryResult` for executing code in a recoverable manner in the face of Postgres ERRORs
+  - Makes use of `sigsetjmp`
+  - Akin to Postgres' `PG_TRY/PG_CATCH/PG_FINALLY/PG_RE_THROW` C macros
+  
+## Thead Safety
+
+- As it relates to any Postgres-thing (calling a function, allocated memory, anything at all from `pgx_pg_sys::`), there is none
```
workingjubilee, line 41: By definition, calling a Postgres `extern "C"` function is `unsafe`. Thus thread safety in a Rust crate is mmmostly a question about Rust code, doubly so because the Postgres runtime is mostly non-reentrant.

DESIGN.md
```diff
@@ -0,0 +1,57 @@
...
+- We (will) detect FFI calls into Postgres in non-the-main-thread and immediately panic
```
workingjubilee, line 42: I believe our main weapon for enforcing thread safety in the long term might actually be moving towards not providing `Send` or `Sync` for PGX types that expose calls into Postgres and requiring constructors to create them, then hypothetically making all the other functions either `unsafe` or requiring them to be called from a type that validates its safety during construction.

DESIGN.md
```diff
@@ -0,0 +1,57 @@
...
+
+- Generated from compiled-Postgres header files (ie, from "-server" or "-server-dev" distro package)
+- "OID" type value definitions converted to Rust enum (`PgOid`)
+- Various trait impls for certain Postgres structs (`impl Display for Node`, `impl PgNode for Node`)
+- Separate bindings for each supported Postgres version, behind feature flags
+
+## Memory Allocation
+
+- Rust allocates as it normally does
+  - No custom Rust GlobalAllocator
```
workingjubilee, line 16: "But why not?"

DESIGN.md
```diff
@@ -0,0 +1,57 @@
...
+## Thead Safety
+
+- As it relates to any Postgres-thing (calling a function, allocated memory, anything at all from `pgx_pg_sys::`), there is none
+- We (will) detect FFI calls into Postgres in non-the-main-thread and immediately panic
+- pgx extensions can use threads so long as they **don't use any** Postgres-thing within them
+
+## Type Conversions
+
+- Postgres Datums are converted to equivalent Rust types via `pgx::FromDatum`
+  - Sometimes this is a built-in from Rust's standard library (ie, `text/varchar` -> `String`)
```
Smittyvb, line 48: Should probably mention that this means pgx can't handle databases with non-UTF8 text encoding.

DESIGN.md
thomcc: I'm still working out some wonkiness around forking in #777, but I expect this will allow cheaply detecting cases where we can't enforce it via Send/Sync bounds, FWIW.

thomcc (commented):
This all makes sense to me, and I don't have a ton to add.

I think perhaps mentioning that it we need to work on stable might be worthwhile (although this is a more surprising thing for plrust/postgrestd than it is for pgx).

DESIGN.md
```diff
@@ -0,0 +1,202 @@
...
+- https://github.com/jeff-davis/postgres-extension.rs
+- https://github.com/bluejekyll/pg-extend-rs
+- https://github.com/wasmerio/wasmer-postgres
+
+None of these appear active today.
+
+## Goals
+
+pgx' primary goal is to make Rust extension development as natural to Rust programmers as possible.  This generally 
+means doing the best we can to avoid violating POLA as it relates to error handling, memory management, and type
```
Smittyvb, line 27: The POLA acronym should probably be expanded, it's not immediately clear what it stands for ([principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)?)

eeeebbbbrrrr: indeed.  Will do.  ;)

DESIGN.md
eeeebbbbrrrr: I tried to explain why...

DESIGN.md
eeeebbbbrrrr: I think I got this included.

eeeebbbbrrrr:
I'm gonna stew on this for a few more hours and probably just go ahead and push it right to `master`.  If anyone has feedback, I'll take it now, or later, either in the form of a ping somewhere or a PR or smoke signals or whatever.

What I'm not wanting to do with this document is cut-n-paste the code.  I really just wanted to do the best I could to detail what pgx is doing in a few important categories and try and explain why.

DESIGN-DECISIONS.md
```diff
@@ -0,0 +1,342 @@
...
+are themselves machine-generated, pgx requires its bindings be generated from these headers.  When pgx is managing 
+Postgres instances itself (via `cargo-pgx`), it gets them from there, otherwise the "-server"/"-server-dev" distribution 
+packages must be locally installed.
+
+## Error Handling
+
+Postgres has a few different methods of raising errors -- the `ERROR`, `FATAL`, and `PANIC` levels.  `ERROR` aborts
+the current transaction, `FATAL` terminates the raising backend process, and `PANIC` restarts the entire cluster.
+
+Rust only has one: `panic!()`, and it terminates the process.
```
thomcc, line 82: nit: Not quite right, it either terminates the process, or unwinds until the nearest catch_unwind, terminating the thread if it reaches the top, and the process if done from the thread that runs main().

DESIGN-DECISIONS.md
```diff
@@ -0,0 +1,342 @@
...
+function call boundary, allowing Rust to unwind its stack and call destructors before relinquishing control to 
+Postgres.
+
+### Protecting Postgres Transactions (read: asking Rust for help)
+
+The above is about how to handle Postgres ERRORs when an internal Postgres function is called from Rust.  This section
+talks about how to handle Rust `panic!()`s generated by a Rust function called from Postgres.
+
+A decision was made during initial pgx development that **all** Rust `panic!()`s will properly abort the active
+Postgres transaction.  While Rust's preference is to abort the **process**, this makes no sense in the context of a
```
thomcc, line 134: As above, this isn't quite right.

DESIGN-DECISIONS.md
```diff
@@ -0,0 +1,342 @@
...
+
+As it relates to any Postgres thing (calling a function, allocated memory, anything at all from `pgx::*::*`), **there is 
+none!**  Postgres itself is purposely not thread safe and neither is pgx.
+
+It's difficult (right now) to ask the Rust compiler to ensure code is not using pgx from a thread.  We have been 
+discussing some ideas.  Until then, there's one simple rule:
+
+- Do **not** use anything from the `pgx` crate or any Postgres allocated memory from a thread
+
+You might be really smart and think you can sneak by doing something, but you're probably wrong.  Even if you're right,
```
thomcc, line 267: It would not play well with our `#[pg_guard]` machinery even if you did -- that code writes to globals without synchronization.

thomcc (commented):
This is a really great reference!

I had one nit (which is stated a few times), and I suspect we'll get some more questions about the longjmp behavior, but it's hard to predict what those would be anyway.

DESIGN-DECISIONS.md
```diff
@@ -0,0 +1,342 @@
...
+
+## Error Handling
+
+Postgres has a few different methods of raising errors -- the `ERROR`, `FATAL`, and `PANIC` levels.  `ERROR` aborts
+the current transaction, `FATAL` terminates the raising backend process, and `PANIC` restarts the entire cluster.
+
+Rust only has one: `panic!()`, and it terminates the process.
+
+pgx, wanting to be as least surprising as possible to Rust developers, provides mechanisms for seamlessly handling
+Postgres `ERROR`s and Rust `panic!()`s.  There's two concerns here.  One is that Rust developers expect proper drop
```
BradyBonnette, line 85: `There are two concerns here`

DESIGN-DECISIONS.md
```diff
@@ -0,0 +1,342 @@
...
+Additionally, pgx auto-generates a Rust `PgBuiltInOids` enum from the set of `#define xxxOID` defines found in the
+Postgres headers.
+
+pgx reserves the right to generate other convenience types, traits, etc in the future.  Such things may come from user
+requests or from the need to ensure more robust compile-time type checking and safety guarantees.
+
+The bindings are generated via the `build.rs` script in the `pgx-pg-sys` crate.  As some of the Postgres header files 
+are themselves machine-generated, pgx requires its bindings be generated from these headers.  When pgx is managing 
+Postgres instances itself (via `cargo-pgx`), it gets them from there, otherwise the "-server"/"-server-dev" distribution 
+packages must be locally installed.
```
BradyBonnette, line 75: Im kinda wondering if that should be re-worded slightly, because some distros use stuff like -server-dev, some use -server-devel, etc. Unless server-dev is the Postgres de-facto way of naming that package

DESIGN-DECISIONS.md
```diff
@@ -0,0 +1,342 @@
...
+both Rust and Postgres and "bounce" back-n-forth through either type.  Assuming `#[pg_guard]` is properly applied a call 
+chain like `Postgres-->Rust-->Rust-->Rust-->Postgres-->Rust(panic!)` will properly unwind the Rust stack, call Rust 
+destructors, and safely abort the active Postgres transaction. So will a call chain like 
+`Postgres-->Rust-->Postgres-->Rust-->Rust-->Postgres(ERROR)`.
+
+## Memory Allocation
+
+Postgres uses a system it calls `MemoryContext`s to manage memory.  A `MemoryContext` lives for a certain amount of time
+(not to be confused with Rust's concept of lifetimes), and typically this is no longer than the life of the current
+transaction.  When Postgres decides a `MemoryContext` is no longer needed, it is freed en mas. Combined with Postgres
```
BradyBonnette, line 171: it is freed `en masse`

DESIGN-DECISIONS.md
eeeebbbbrrrr: I changed it above, but I'm just gonna rewrite this as "terminate the **process**".  I don't really want to have to explain threads again.  The document having a bit of a conversational tone, I think folks reading it will understand.

DESIGN-DECISIONS.md
eeeebbbbrrrr: Well, I didn't want to give examples, but I was thinking someone might think they're smart enough to use something like Postgres' `pub static mut InterruptPending: bool;` to see if they ought to coax a thread into terminating.  Like, maaaaaaybe that'd work, but yeah, I ain't wanting to suggest it.

thomcc: It would probably not work on ARM, unless there's some additional fencing.

DESIGN-DECISIONS.md
eeeebbbbrrrr: You

PR #784, opened 2022-10-19
SRF functions need to "fetch" (call `FromDatum`) each argument while …

eeeebbbbrrrr:
…operating within the function's "multi-call-memory-context".  This ensures any allocation `FromDatum` might do lives as along as the entire SRF calling process does.

This is the proper fix for PR #779, which pointed out that it wasn't possible for SRFs to return data borrowed from their input.  Tests are included.

@Smittyvb:  This PR should resolve the problems that #779 was trying to explain away with documentation.  Prior to this PR, pgx was potentially detoasting argument datums in the wrong memory context, causing the UAF you were seeing.  I'd love it if you could rejigger your code against this PR and make sure it now works as you'd expect.  I updated the `strings` example to show that a SRF can borrow from its input.

### DIFF
--- a/pgx-utils/src/lib.rs
+++ b/pgx-utils/src/lib.rs
@@ -7,8 +7,8 @@ All rights reserved.
 Use of this source code is governed by the MIT license that can be found in the LICENSE file.
 */
 
-use crate::sql_entity_graph::PositioningRef;
-use proc_macro2::{TokenStream, TokenTree};
+use crate::sql_entity_graph::{NameMacro, PositioningRef};
+use proc_macro2::{Ident, TokenStream, TokenTree};
 use quote::{format_ident, quote, ToTokens, TokenStreamExt};
 use std::collections::HashSet;
 use syn::{GenericArgument, PathArguments, Type, TypeParamBound};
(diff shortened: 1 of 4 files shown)


Smittyvb (approved):
I tested my code against this PR and it fixed the use-after-free I was getting. (the tests now pass before and after reverting https://github.com/timescale/timescaledb-toolkit/commit/fe8b3861ee12b1cebedfb32a0c4fce2b9ce508df).

pgx-utils/src/sql_entity_graph/pg_extern/mod.rs
```diff
@@ -478,8 +471,19 @@ impl PgExtern {
                                 funcctx.user_fctx = pgx::PgMemoryContexts::For(funcctx.multi_call_memory_ctx).palloc_struct::<IteratorHolder<#retval_ty_resolved>>() as *mut ::core::ffi::c_void;
                                 iterator_holder = pgx::PgBox::from_pg(funcctx.user_fctx as *mut IteratorHolder<#retval_ty_resolved>);
 
-                                #( #arg_fetches )*
-                                #result_handler
+                                // function arguments need to be "fetched" while in the function call's
+                                // multi-call-memory-context to ensure that any detoasted datums will
+                                // live long enough for the SRF to use them over each call
```
workingjubilee, line 476: This is what needs to also be documented in a place outside the context of this function's comments. Especially so because this is something that will always need to remain true: we always must satisfy this sort of lifetime-binding condition, and it doesn't matter whether the code is burned and rewritten a million times: it remains a constraint. It should be easier to reference this sort of information than reading thousands of lines of implementations line by line, because it determines how fast someone can begin to understand the concepts required to write correct code for implementing PGX.

pgx-utils/src/lib.rs
```diff
@@ -408,6 +408,22 @@ pub fn staticize_lifetimes(value: &mut syn::Type) {
...
+            let archetype = mac.path.segments.last().unwrap();
+            match archetype.ident.to_string().as_str() {
+                "name" => {
+                    let mut out: NameMacro = mac.parse_body().expect("name!() in wrong format");
+                    staticize_lifetimes(&mut out.used_ty.resolved_ty);
+
+                    // rewrite the name!() macro so that it has a static lifetime, if any
+                    let ident = syn::parse_str::<Ident>(&out.ident).unwrap();
+                    let ty = out.used_ty.resolved_ty;
+                    type_macro.mac = syn::parse_quote! {name!(#ident, #ty)};
```
workingjubilee, line 422:
```rust
#[pg_extern]
fn split_table_with_borrow<'a>(
    input: &'a str,
    pattern: &'a str,
) -> TableIterator<'a, (name!(i, i32), name!(s, &'a str))>
```
So in this case, the return lifetimes we are expecting will be
```rust
) -> TableIterator<'static, (name!(i, i32), name!(s, &'static str))>
```
?

workingjubilee: Hm. I think this can be done in a better way in the future, but I can't find the solution immediately and we're staticizing other lifetimes, so I think this is acceptable for the current moment.

PR #786, opened 2022-10-19
derive Debug for PgLogLevel and PgSqlErrorCode

yrashk:
Problem: inability to use PgLogLevel and PgSqlErrorCode in debug
This means that it's not feasible to `unwrap` on types containing these.

Solution: derive Debug for both


workingjubilee:
Ah, so not implementing Debug would discourage using `.unwrap()`, an inadvisable function...!

Just kidding, this looks fine.

PR #788, opened 2022-10-19
Handle a wider variety of version strings

jteplitz:
Some custom builds of postgres use more complex
version strings. For example, yugabyte builds postgres
with the version string 'PostgreSQL {MAJOR}.{MINOR}-YB-{YB_VERSION}'.

This PR modifies the version string parsing to allow arbitrary strings
after the {MAJOR}.{MINOR} part. It also refactors the parsing
logic so that major and minor version parsing don't have as
much duplicate code

With this PR `cargo pgx package -c /path/to/yugabyte/postgres/bin/pg_config` works. So you can package / install a rust postgres plugin directly against yugabyte.

### DIFF
--- a/pgx-pg-config/src/lib.rs
+++ b/pgx-pg-config/src/lib.rs
@@ -119,49 +119,48 @@ impl PgConfig {
         self.path().unwrap().parent().unwrap().to_path_buf()
     }
 
+    fn parse_version_str(version_str: &str) -> eyre::Result<(u16, u16)> {
+        let version_parts = version_str.split_whitespace().collect::<Vec<&str>>();
+        let version = version_parts
+            .get(1)
+            .ok_or_else(|| eyre!("invalid version string: {}", version_str))?
+            .split('.')
+            .collect::<Vec<&str>>();
+        if version.len() < 2 {
+            return Err(eyre!("invalid version string: {}", version_str));
+        }
+        let major = u16::from_str(version[0])
+            .map_err(|e| eyre!("invalid major version number `{}`: {:?}", version[0], e))?;
+        let mut minor = version[1];
+        let mut end_index = minor.len();
+        for (i, c) in minor.chars().enumerate() {
+            if !c.is_ascii_digit() {
+                end_index = i;
+                break;
+            }
+        }
+        minor = &minor[0..end_index];
+        let minor = u16::from_str(minor)
+            .map_err(|e| eyre!("invalid minor version number `{}`: {:?}", minor, e))?;
+        return Ok((major, minor));
+    }
+
+    fn get_version(&self) -> eyre::Result<(u16, u16)> {
+        let version_string = self.run("--version")?;
+        Self::parse_version_str(&version_string)
+    }
+
     pub fn major_version(&self) -> eyre::Result<u16> {
         match &self.version {
             Some(version) => Ok(version.major),
-            None => {
-                let version_string = self.run("--version")?;
-                let version_parts = version_string.split_whitespace().collect::<Vec<&str>>();
-                let version = match version_parts.get(1) {
-                    Some(v) => v,
-                    None => {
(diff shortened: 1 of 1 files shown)


thomcc (changes requested):
Do you mind adding tests for this? It might be easier to pull parse_version into something that takes a `&str` arg and add some tests for that functionality.

(Other than that I suspect the formatter will want some changes as well)

jteplitz:
Sure thing. Just pushed the `cargo fmt` change

pgx-pg-config/src/lib.rs
```diff
@@ -119,49 +119,45 @@ impl PgConfig {
...
+        if version.len() < 2 {
+            return Err(eyre!("invalid version string: {}", version_string));
+        }
+        let major = u16::from_str(version[0])
+            .map_err(|e| eyre!("invalid major version number `{}`: {:?}", version[0], e))?;
+        let mut minor = version[1];
+        let digits = HashSet::from(['0', '1', '2', '3', '4', '5', '6', '7', '8', '9']);
+        let mut end_index = minor.len();
+        for (i, c) in minor.chars().enumerate() {
+            if !digits.contains(&c) {
```
thomcc, line 139: `if !c.is_ascii_digit()` (https://doc.rust-lang.org/nightly/std/primitive.char.html#method.is_ascii_digit) and removing `digits` is cleaner IMO

thomcc (commented):
One other thing

pgx-pg-config/src/lib.rs
jteplitz: Didn't know about that stdlib function. That is cleaner. Thanks!

jteplitz:
I added a unit test and removed `digits`. Let me know what you think. And thanks for the super fast review!

eeeebbbbrrrr:
(FYI... I've enabled CI builds for this PR)

pgx-pg-config/src/lib.rs
```diff
@@ -528,3 +527,35 @@ fn does_db_exist(pg_config: &PgConfig, dbname: &str) -> eyre::Result<bool> {
...
+        let (major, minor) =
+            PgConfig::parse_version_str(s).expect("Unable to parse version string");
+        assert_eq!(major, major_expected, "Major varsion should match");
+        assert_eq!(minor, minor_expected, "Minor version should match");
+    }
+
+    // Check some invalid version strings
+    let _ = PgConfig::parse_version_str("10.22").expect_err("Parsed invalid version string");
+    let _ =
+        PgConfig::parse_version_str("PostgresSQL 10").expect_err("Parsed invalid version string");
```
workingjubilee, line 554: ...are we entirely sure this one should be invalid?

thomcc: What should it come out as?

thomcc (commented):
This looks good to me, but I'll let @workingjubilee decide if her comment is blocking.

pgx-pg-config/src/lib.rs
jteplitz: @workingjubilee I do not know. I just matched the behavior of the existing code. Currently "PostgresSQL 10" will fail with an invalid minor version error. Happy to change it if you feel this should really be interpreted as 10.0 or something though.

workingjubilee (approved):
Yeah, on consideration I don't actually have a good answer off the top of my head for what it should be either. If we change it, it really would be by making it valid to use a non-answer for a minor version, not assigning a value to the absence of a parsed minor version.

thomcc (approved):
I'm okay with this now.

eeeebbbbrrrr (approved):
This looks good as it is.  It'll be interesting to see what other random version strings pop up in the future.

We internally discussed how to just get the numbers out of the generated bindings, but that's not so straightforward today.  So until it is, we'll roll on with this...

PR #793, opened 2022-10-20
Wrap `SpinLock.*` functions in the C shim

### LINKED ISSUE #752: Consider making SpinLock methods accessible in PGX
The Postgres methods below are not accessible in PGX. Please consider adding them to pgx::pg_sys to make it easier to implement parallel scans in Index Access Methods.

- SpinLockInit
- SpinLockAcquire
- SpinLockRelease

thomcc:
Fixes #752.

@eeeebbbbrrrr Do we usually just expose these as `pgx_MacroNameHere` functions, or do we re-export them somewhere using the names from PG (in which case I've missed a bit)?

### DIFF
--- a/pgx-pg-sys/cshim/pgx-cshim.c
+++ b/pgx-pg-sys/cshim/pgx-cshim.c
@@ -26,6 +26,7 @@ Use of this source code is governed by the MIT license that can be found in the 
 #include "utils/memutils.h"
 #include "utils/builtins.h"
 #include "utils/array.h"
+#include "storage/spin.h"
 
 
 PGDLLEXPORT MemoryContext pgx_GetMemoryContextChunk(void *ptr);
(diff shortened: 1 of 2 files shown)


eeeebbbbrrrr:
> Fixes #752.
> 
> @eeeebbbbrrrr Do we usually just expose these as `pgx_MacroNameHere` functions, or do we re-export them somewhere using the names from PG (in which case I've missed a bit)?

In this case I think they ought to be re-exported as their Postgres macro names.  I suppose there isn't a hard rule that we've followed.

These functions make up parts of an internal Postgres API so it seems like we should follow those names.  Other things like like "pgx_GETSTRUCT" and "pgx_list_nth", for example, really aren't used by user-facing code.

It might even make sense to make a "spin.rs" file in pgx-pg-sys/src/submodules/ and put them there.

thomcc:
The re-exports would be pretty small so I'm not sure it makes that much sense to put in a dedicated submodule.

eeeebbbbrrrr:
> The re-exports would be pretty small so I'm not sure it makes that much sense to put in a dedicated submodule.

I suppose what we want is for them to appear in the `pgx_pg_sys` submodule, so users would get to them via `pgx::pg_sys::SpinLockInit(...)`.

Down the road, I'd expect us to write a safe wrapper around this API, and that's probably when we should make a spin.rs submodule.

thomcc:
> I suppose what we want is for them to appear in the `pgx_pg_sys` submodule, so users would get to them via `pgx::pg_sys::SpinLockInit(...)`.

Yeah, that's what this patch does. Everything in the `all_versions` modules gets glob-reexported, so this should allow `pgx::pg_sys::SpinLockInit(...)` as it is.

> Down the road, I'd expect us to write a safe wrapper around this API, and that's probably when we should make a spin.rs submodule.

👍

eeeebbbbrrrr (approved):

osawyerr:
Hi @thomcc this doesn't appear to work (segafault)
let mutex : *mut slock_t = null_mut();
SpinLockInit(slock_t);

OS: MacOS
PGX: 0.5.6

thomcc:
Yeah, that's not how you use it. You need to pass a pointer to a real slock_t in. See https://github.com/tcdi/pgx/blob/develop/pgx/src/spinlock.rs for an example.

Issue #792, opened 2022-10-20
false is indistinguishable from NULL in a tuple of bools


sumerman:
Consider the snippet below: I would expect `test_foo` to pass, and yet the assertion fails on `0.5.x`.

```rust
    fn from_helper<T: FromDatum + IntoDatum>(d: Datum) -> Option<T> {
        unsafe { T::from_polymorphic_datum(d, false, pg_sys::InvalidOid) }
    }

    #[pg_test]
    fn test_foo() {
        let d = (Some(false), Some(true)).into_datum().unwrap();
        let (a, b) = from_helper::<(Option<bool>, Option<bool>)>(d).unwrap();
        assert_eq!(a.is_some(), true);
    }
```

I discovered this behavior while trying to make `Spi::get_three_with_args` work in our tests. `Spi::get_one_with_args` works just fine, but `get_two` and `get_three` both suffer from this quirk. I managed to narrow it down to `Spi` functions round-tripping datums in order to move allocation into an appropriate memory context and thus relying on a bunch of `IntoDatum` and `FromDatum` instances starting from `(Option<A>, Option<B>)`.

sumerman:
It would seem something is up with `Vec<Option<T>>` Into and From instances:

```rust
    #[pg_test]
    fn test_bar() {
        let d = vec![Some(false).into_datum(), Some(true).into_datum()].into_datum().unwrap();
        let a = from_helper::<Vec<Option<pg_sys::Datum>>>(d).unwrap();
        assert_eq!(a.first().unwrap().is_some(), true); // fails; a is [None, Some(Datum(...))]
    }
```

eeeebbbbrrrr:
Good catch.  I think the FromDatum impls in `tuples.rs` are bugged.  Looks like they've been bugged for a few years.  

I'll put up a PR now.

sumerman:
> Good catch. I think the FromDatum impls in `tuples.rs` are bugged. Looks like they've been bugged for a few years.
> 
> I'll put up a PR now.

Thank you! It's odd to hear, as it used to work on 0.4.x branch

eeeebbbbrrrr:
> Thank you! It's odd to hear, as it used to work on 0.4.x branch

Yeah, apparently it's not quite what I thought it was.  I'm digging into it now...

eeeebbbbrrrr:
It gets more fun... this fails too:

```rust
    #[pg_test]
    fn test_vec_ints() {
[...]

PR #795, opened 2022-10-20
Fix Issue #792

eeeebbbbrrrr:
The crux of the issue here is that `pgx_sys::Datum::from_datum(d, is_null)` can't ask the input Datum if it thinks it's null (via `Datum::is_null(&self)`), as nullness isn't a concept Datums encode.

A Datum is technically only ever considered null if the accompanying `is_null` argument tells us so.

It so happens that if the `Datum` type is pass-by-reference, and its value is zero, then it's pointing to a NULL, but Postgres doesn't even guarantee us that that is the nullness indicator.  Again, only the `is_null` argument does that.

As such, when converting a `Datum` into an `Option<Datum>`, only listen to the arguments and don't try to infer anything from the input datum's underlying value.

This PR includes a few simple tests that are now green with the change to `from.rs`.

Thanks to @sumerman for catching this!

### DIFF
--- a/pgx-tests/src/tests/mod.rs
+++ b/pgx-tests/src/tests/mod.rs
@@ -43,5 +43,6 @@ mod uuid_tests;
 mod variadic_tests;
 mod xact_callback_tests;
 mod xid64_tests;
+mod zero_datum_edge_cases;
 
 pgx::pg_magic_func!();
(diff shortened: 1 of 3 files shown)


thomcc (approved):
Looks good to me, but I do feel like Datum is a bit of a footgun as it is. Oh well, this seems like a strict improvement on that front.

workingjubilee:
That check was added because the prior state of affairs was causing segfaults.

workingjubilee:
> That check was added because the prior state of affairs was causing segfaults.

This specific part of it was probably incorrect, however.

workingjubilee (approved):

PR #799, opened 2022-10-20
Problem: impossible to return anything from Spi::connect

yrashk:
We're currently limited by `Datum` (`IntoDatum` + `FromDatum`) boundaries. Any data that has nothing to do with Postgres, can't cross it.

Solution: copy data to the parent context when it is retrieved

Hat tip to @eeeebbbbrrrr

### DIFF
--- a/pgx/src/memcxt.rs
+++ b/pgx/src/memcxt.rs
@@ -239,6 +239,16 @@ impl PgMemoryContexts {
         }
     }
 
+    /// Returns parent memory context if any
+    pub fn parent(&self) -> Option<PgMemoryContexts> {
+        let parent = unsafe { pg_sys::MemoryContextGetParent(self.value()) };
+        if parent.is_null() {
+            None
+        } else {
+            Some(PgMemoryContexts::For(parent))
+        }
+    }
+
     /// Run the specified function "within" the `MemoryContext` represented by this enum.
     ///
     /// The important implementation detail is that Postgres' `CurrentMemoryContext` is changed
--- a/pgx/src/memcxt.rs
+++ b/pgx/src/memcxt.rs
@@ -241,6 +241,8 @@ impl PgMemoryContexts {
 
     /// Returns parent memory context if any
     pub fn parent(&self) -> Option<PgMemoryContexts> {
+        // SAFETY: We do this instead of simply plucking the .parent field ourselves
+        // mostly to let Postgres check the context validity if --enable-cassert is on
         let parent = unsafe { pg_sys::MemoryContextGetParent(self.value()) };
         if parent.is_null() {
             None
--- a/pgx/src/spi.rs
+++ b/pgx/src/spi.rs
@@ -241,15 +241,9 @@ impl Spi {
 
     /// execute SPI commands via the provided `SpiClient` and return a value from SPI which is
     /// automatically copied into the `CurrentMemoryContext` at the time of this function call
-    pub fn connect<
-        R: FromDatum + IntoDatum,
-        F: FnOnce(SpiClient) -> std::result::Result<Option<R>, SpiError>,
-    >(
+    pub fn connect<R, F: FnOnce(SpiClient) -> std::result::Result<Option<R>, SpiError>>(
         f: F,
     ) -> Option<R> {
-        let outer_memory_context =
-            PgMemoryContexts::For(PgMemoryContexts::CurrentMemoryContext.value());
-
         /// a struct to manage our SPI connection lifetime
         struct SpiConnection;
         impl SpiConnection {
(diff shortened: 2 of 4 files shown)


workingjubilee:
Would you be so kind as to resolve the merge conflicts?

yrashk:
Surely!

yrashk:
@workingjubilee resolved!

pgx/src/spi.rs
```diff
@@ -494,7 +457,10 @@ impl SpiTupleTable {
                         let datum =
                             pg_sys::SPI_getbinval(heap_tuple, tupdesc, ordinal, &mut is_null);
```
eeeebbbbrrrr, line 458: This call and the below `T::from_datum()` call should happen like so:

```rust
let datum = PgMemoryContexts::For(PgMemoryContexts::CurrentMemoryContext.parent()).switch_to(|context| {
   
      pg_sys::SPI_getbinval(heap_tuple, tupdesc, ordinal, &mut is_null);
      T::from_datum(datum, is_null)
});
```

We need Postgres to do the `SPI_getbinval` call in the parent context as well as the from_datum call.  This way, if the Datum doesn't need to be de-toasted but still allocated, it'll be in the right place.

eeeebbbbrrrr (changes requested):

pgx/src/spi.rs
yrashk: So, I tried to write it like this:

```rust
PgMemoryContexts::CurrentMemoryContext
                            .parent()
                            .expect("parent memory context is absent")
                            .switch_to(|_context| {
                                let mut is_null = false;
                                let datum = pg_sys::SPI_getbinval(
                                    heap_tuple,
                                    tupdesc,
                                    ordinal,
                                    &mut is_null,
                                );
                                T::from_polymorphic_datum(
                                    datum,
                                    is_null,
                                    pg_sys::SPI_gettypeid(tupdesc, ordinal),
                                )
                            })
```

The diff is:

```diff
diff --git a/pgx/src/spi.rs b/pgx/src/spi.rs
index c185d92e..262ff50c 100644
--- a/pgx/src/spi.rs
+++ b/pgx/src/spi.rs
@@ -453,18 +453,24 @@ impl SpiTupleTable {
                     } else {
                         let heap_tuple = std::slice::from_raw_parts((*self.table).vals, self.size)
                             [self.current as usize];
-                        let mut is_null = false;
-                        let datum =
-                            pg_sys::SPI_getbinval(heap_tuple, tupdesc, ordinal, &mut is_null);
-
-                        T::from_datum_in_memory_context(
-                            PgMemoryContexts::CurrentMemoryContext
-                                .parent()
-                                .expect("parent memory context is absent"),
-                            datum,
-                            is_null,
-                            pg_sys::SPI_gettypeid(tupdesc, ordinal),
-                        )
+
+                        PgMemoryContexts::CurrentMemoryContext
+                            .parent()
+                            .expect("parent memory context is absent")
+                            .switch_to(|_context| {
+                                let mut is_null = false;
+                                let datum = pg_sys::SPI_getbinval(
+                                    heap_tuple,
+                                    tupdesc,
+                                    ordinal,
+                                    &mut is_null,
+                                );
+                                T::from_polymorphic_datum(
+                                    datum,
+                                    is_null,
+                                    pg_sys::SPI_gettypeid(tupdesc, ordinal),
+                                )
+                            })
                     }
                 },
                 None => panic!("TupDesc is NULL"),
```

and immediately, tests started failing, showing random data all over the place.

I've looked up [SPI_getbinval](https://www.postgresql.org/docs/current/spi-spi-getbinval.html) documentation, and it states: "This function does not allocate new space for the datum. In the case of a pass-by-reference data type, the return value will be a pointer into the passed row."


Could it be the reason why my code worked fine and the above didn't? I don't necessarily assume that the above code is what you intended me to try. Maybe I didn't get what you wanted me to do.

My reading suggests that calling `SPI_getbinval` is the parent context is wrong if it returns pointers (as states above).

yrashk: @eeeebbbbrrrr any thoughts on this?

eeeebbbbrrrr: I really hate SPI.

After reading the SPI_getbinval docs myself, I don't think it's that calling SPI_getbinval in any context is wrong per se.  I think what it returns points to data that was already allocated for us in CurrentMemoryContext, which is the one created by SPI_connect.

I should have read the docs before my comment here.  I'm sorry.

I now think what you had before using `FromDatum::from_datum_in_memory_context()` is the correct thing to do, as that function is supposed to **copy** the datum into the provided context, which just switching memory contexts like this doesn't actually do.

So yeah, going back to what you originally had is correct.  But now it leaves me with a concern around "what if our current (and future!) `::from_datum_in_memory_context()` impls are bugged.

<short pause>

Geez, I think most of them are bugged right now by virtue of the fact the use the default implementation.

Let me think on this a bit more...

eeeebbbbrrrr (changes requested):

pgx/src/spi.rs
eeeebbbbrrrr: > Geez, I think most of them are bugged right now by virtue of the fact they use the default implementation.

Well, wait.  No, I think they're okay.  We only have non-default impls for (rust) reference types like `&str` and `&[u8]` (and a few others).

For a minute I was second-guessing the code and thinking it'd be broken for `String` or `Vec<u8>` since those aren't first copying the Datum, but in fact they are -- straight into Rust-allocated memory, which is totally fine.

So yeah, lets go back to your original code and I think we're good.  Thanks for your tenacity!

yrashk: I didn't commit the suggested update, so it's at what I wrote initially. Anything I can do to  make this mergeable?

eeeebbbbrrrr:
> I didn't commit the suggested update, so it's at what I wrote initially. Anything I can do to make this mergeable?

I think it's good as it.  Lets have @workingjubilee give it a quick glance and they can smash the merge button.

Thanks a bunch, @yrashk!

eeeebbbbrrrr (approved):

workingjubilee:
Will take a look at this shortly.

pgx/src/memcxt.rs
```diff
@@ -239,6 +239,16 @@ impl PgMemoryContexts {
         }
     }
 
+    /// Returns parent memory context if any
+    pub fn parent(&self) -> Option<PgMemoryContexts> {
+        let parent = unsafe { pg_sys::MemoryContextGetParent(self.value()) };
```
workingjubilee, line 244:
```suggestion
        // SAFETY: We do this instead of simply plucking the .parent field ourselves
        // mostly to let Postgres check the context validity if --enable-cassert is on
        let parent = unsafe { pg_sys::MemoryContextGetParent(self.value()) };
```

workingjubilee (approved):
Sorry, it took me a while to screw in the brain that can review memory context code. Took a look and it seems fine, only have a tiny remark on the addition of this `unsafe` (the alternative is "safe", but ironically less checked code, so it seems like a logical transition... but we shouldn't actually do it right away lmao).

yrashk:
@workingjubilee applied your suggestion! Thank you!

workingjubilee:
Thank you!

PR #804, opened 2022-10-21
Add SpiHeapTupleDataEntry::oid getter

### LINKED ISSUE #688: Need pub visibility to SpiHeapTupleDataEntry.type_oid
My crate issues dynamically generated queries against schemas unknown at compile time. This means we need to handle casting SpiHeapTupleDataEntry's to the correct Rust type at run time. We're currently querying information_schema.columns to get type information, but it would simplify things if we had access to the OID in the query result. Is there a reason SpiHeapTupleDataEntry.type_oid is not `pub`?

workingjubilee:
This fixes #688.

### DIFF
--- a/pgx/src/spi.rs
+++ b/pgx/src/spi.rs
@@ -635,6 +635,10 @@ impl SpiHeapTupleDataEntry {
             None => None,
         }
     }
+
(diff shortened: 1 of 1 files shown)


thomcc (approved):
Seems extremely harmless

PR #809, opened 2022-10-21
Add safe wrappers for Postgres SpinLock functionality.

thomcc:
Ended up doing this now rather than later (getting very different perf numbers today on the thread safety stuff than I was yesterday so I'm taking a short break on it to see if it gives whatever is causing that time to cool off...).

This also fixes PgLwLock's send/sync bounds (in the first commit), which were overly broad.

### DIFF
--- a/pgx/src/lwlock.rs
+++ b/pgx/src/lwlock.rs
@@ -34,8 +34,8 @@ pub struct PgLwLock<T> {
     name: OnceCell<&'static str>,
 }
 
-unsafe impl<T> Send for PgLwLock<T> {}
-unsafe impl<T> Sync for PgLwLock<T> {}
+unsafe impl<T: Send> Send for PgLwLock<T> {}
+unsafe impl<T: Send + Sync> Sync for PgLwLock<T> {}
 
 impl<T> PgLwLock<T> {
     /// Create an empty lock which can be created as a global with None as a
--- a/pgx/src/spinlock.rs
+++ b/pgx/src/spinlock.rs
@@ -55,6 +55,7 @@ impl<T> PgSpinLock<T> {
     #[inline]
     #[doc(alias = "SpinLockFree")]
     pub fn is_locked(&self) -> bool {
+        // SAFETY: Doesn't actually modify state, despite appearances.
         unsafe { !pg_sys::SpinLockFree(self.lock.get()) }
     }
 
--- a/pgx/src/spinlock.rs
+++ b/pgx/src/spinlock.rs
@@ -73,13 +73,13 @@ impl<T> PgSpinLock<T> {
 /// structure falls out of scope (is dropped), the lock will be unlocked.
 ///
 /// See the documentation of [`PgSpinLock`] for more information.
-pub struct PgSpinLockGuard<'a, T> {
+pub struct PgSpinLockGuard<'a, T: 'a> {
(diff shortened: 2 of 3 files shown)


pgx/src/lwlock.rs
```diff
@@ -34,8 +34,8 @@ pub struct PgLwLock<T> {
     name: OnceCell<&'static str>,
 }
 
-unsafe impl<T> Send for PgLwLock<T> {}
-unsafe impl<T> Sync for PgLwLock<T> {}
+unsafe impl<T: Send> Send for PgLwLock<T> {}
+unsafe impl<T: Send + Sync> Sync for PgLwLock<T> {}
```
workingjubilee, line 38: Oh wow yes, that was extremely overly broad.

pgx/src/spinlock.rs
```diff
@@ -0,0 +1,105 @@
...
+/// structure falls out of scope (is dropped), the lock will be unlocked.
+///
+/// See the documentation of [`PgSpinLock`] for more information.
+pub struct PgSpinLockGuard<'a, T> {
+    lock: &'a PgSpinLock<T>,
+    // For !Send, variance, etc.
+    _marker: PhantomData<*mut T>,
+}
+
+unsafe impl<'a, T: Sync> Sync for PgSpinLockGuard<'_, T> {}
```
workingjubilee, line 81: You know, I'm wondering to myself now. Is this capturing the lifetime appropriately? Does it need to specify `T: 'a`? Is that even valid syntax? It shouldn't have to, right? Because the lifetime should inherently handle the concerns about safety regarding the `Sync` bound, yes? And the lifetime should handle the bound for the entire thing, because the lifetime implicitly bounds the entire `PgSpinLockGuard` as a bounding parameter?

Alright, yes, I've convinced myself Rust is correct now. Looks good!

pgx/src/spinlock.rs
```diff
@@ -0,0 +1,105 @@
...
+            pg_sys::SpinLockInit(slock.as_mut_ptr());
+            Self { item: UnsafeCell::new(value), lock: UnsafeCell::new(slock.assume_init()) }
+        }
+    }
+
+    /// Returns true if the spinlock is locked, and false otherwise.
+    #[inline]
+    #[doc(alias = "SpinLockFree")]
+    pub fn is_locked(&self) -> bool {
+        unsafe { !pg_sys::SpinLockFree(self.lock.get()) }
```
workingjubilee, line 58:
```suggestion
        // SAFETY: Doesn't actually modify state, despite appearances.
        unsafe { !pg_sys::SpinLockFree(self.lock.get()) }
```

workingjubilee (approved):

pgx/src/spinlock.rs
thomcc: Seriously. You'd think they could have come up with a better name for that function.

pgx/src/spinlock.rs
thomcc: I've tweaked it to no longer use the implicit bound in the type declaration, and not have the unused `'a` in the `unsafe impl`. I think this should be clearer now.

If it helps, this is now the same as the [`std::sync::MutexGuard` `Sync` impl](https://doc.rust-lang.org/nightly/src/std/sync/mutex.rs.html#204) (although it was before too, just less obviously so)

PR #813, opened 2022-10-21
Preserve attributes through pg_guard

thomcc:
Let's try this again. Mostly pushing this up to get a CI run, might not quite work.

### DIFF
--- a/pgx-pg-sys/build.rs
+++ b/pgx-pg-sys/build.rs
@@ -793,16 +793,17 @@ fn apply_pg_guard(items: &Vec<syn::Item>) -> eyre::Result<proc_macro2::TokenStre
     for item in items {
         match item {
             Item::ForeignMod(block) => {
+                let abi = &block.abi;
                 for item in &block.items {
                     match item {
                         ForeignItem::Fn(func) => {
                             // Ignore other functions -- this will often be
                             // variadic functions that we can't safely wrap.
-                            if let Ok(tokens) = PgGuardRewriter::new().foreign_item_fn(func) {
+                            if let Ok(tokens) = PgGuardRewriter::new().foreign_item_fn(func, abi) {
                                 out.extend(tokens);
                             }
                         }
-                        other => out.extend(quote! { extern "C" { #other } }),
+                        other => out.extend(quote! { #abi { #other } }),
                     }
                 }
             }
--- a/pgx-utils/src/rewriter.rs
+++ b/pgx-utils/src/rewriter.rs
@@ -30,7 +30,7 @@ impl PgGuardRewriter {
         let mut stream = proc_macro2::TokenStream::new();
 
         for item in block.items.into_iter() {
-            stream.extend(self.foreign_item(item));
+            stream.extend(self.foreign_item(item, &block.abi));
         }
 
         stream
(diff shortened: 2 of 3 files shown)


thomcc:
Seems to work this way, and also forces a compiler error if there are cases where our rewriter would otherwise mangle a function name (e.g. if we have any other cases like #757).

May cause some breakage as a result of that though, but should only be in unsound cases.

pgx-utils/src/rewriter.rs
```diff
@@ -106,9 +106,7 @@ impl PgGuardRewriter {
         Ok(quote! {
             pub unsafe fn #func_name ( #arg_list_with_types ) #return_type {
                 crate::submodules::setjmp::pg_guard_ffi_boundary(move || {
-                    extern "C" {
-                        fn #func_name( #arg_list_with_types ) #return_type ;
-                    }
+                    extern "C" { #func }
```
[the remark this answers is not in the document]
eeeebbbbrrrr: if that's easy enough, might as well just do it now.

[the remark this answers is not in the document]
thomcc: > on unix it should basically always be `extern "C"`

I realized I wasn't that confident about this given that we wrap arbitrary functions from headers so I ended up addressing it.

pgx-utils/src/rewriter.rs
```diff
@@ -106,9 +114,7 @@ impl PgGuardRewriter {
         Ok(quote! {
             pub unsafe fn #func_name ( #arg_list_with_types ) #return_type {
                 crate::submodules::setjmp::pg_guard_ffi_boundary(move || {
-                    extern "C" {
-                        fn #func_name( #arg_list_with_types ) #return_type ;
-                    }
+                    #abi { #func }
```
workingjubilee, line 117: Excellent!

pgx-pg-sys/build.rs
```diff
@@ -793,16 +793,17 @@ fn apply_pg_guard(items: &Vec<syn::Item>) -> eyre::Result<proc_macro2::TokenStre
...
                         ForeignItem::Fn(func) => {
                             // Ignore other functions -- this will often be
                             // variadic functions that we can't safely wrap.
-                            if let Ok(tokens) = PgGuardRewriter::new().foreign_item_fn(func) {
+                            if let Ok(tokens) = PgGuardRewriter::new().foreign_item_fn(func, abi) {
                                 out.extend(tokens);
                             }
                         }
-                        other => out.extend(quote! { extern "C" { #other } }),
+                        other => out.extend(quote! { #abi { #other } }),
```
workingjubilee, line 806: Yes, it's entirely possible for a Unix system to have non-`extern "C"` bindings (albeit less common than on Windows... but with Postgres we're deep into the land of doing things that are less common, e.g. we mess with `sigsetjmp` at all). It also makes the changes forward-resilient against any changes necessary for e.g. `extern "C-unwind"`.

thomcc: > with Postgres we're deep into the land of doing things that are less common

😭

> It also makes the changes forward-resilient against any changes necessary for e.g. extern "C-unwind".

Oh, support for C-unwind isn't even a case I thought about! Good point!

PR #814, opened 2022-10-21
Blocklist several needless C functions

thomcc:
Should fix the ~~warnings~~ errors @BradyBonnette was getting. These shouldn't be re-exported anyway since they're from system headers (#730).


workingjubilee (approved):
Let's go!

PR #815, opened 2022-10-21
Dedup version string update logic

workingjubilee:
I was examining the code for improvements related to releases and found this simplification.
    
We can exploit a closure to handle some update repetition, and then use `dyn TableLike` to handle the different tables.

### DIFF
--- a/pgx-version-updater/src/main.rs
+++ b/pgx-version-updater/src/main.rs
@@ -6,7 +6,7 @@ use std::io::{BufRead, Write};
 use std::path::Path;
 use std::process::{Command, Stdio};
 use std::{env, path::PathBuf};
-use toml_edit::{value, Document};
+use toml_edit::{value, Document, Entry, Item};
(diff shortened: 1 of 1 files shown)


pgx-version-updater/src/main.rs
```diff
@@ -190,60 +190,42 @@ fn main() {
...
-                            if let Some(current_version_specifier) = item.as_str() {
-                                updatable_table[package] = value(parse_new_version(
-                                    current_version_specifier,
-                                    &args.update_version.as_str(),
-                                ))
-                            }
-                        }
-
-                        // Don't care!
-                        None => {}
```
BradyBonnette, line 245: So glad to get rid of this. It always bothered me, but didn't really know how to remove it or omit it without making it look weird. :)

BradyBonnette (approved):
Looks so much better! I ran it locally, and still works fine. Thanks!

Issue #798, opened 2022-10-20
Investigate ways of making pgx-pg-sys more modular


thomcc:
In an ideal world, we would be able to expose the symbols from each header in their own submodule, rather than everything in a global namespace.

This would require changes to bindgen, but in principal adding information about the header it comes from to the symbols/types in https://docs.rs/bindgen/latest/bindgen/callbacks/trait.ParseCallbacks.html would be possible, and quite useful for us -- doing so should also allow us to solve https://github.com/tcdi/pgx/issues/730 in a cleaner and more accurate way (probably).

workingjubilee:
We should improve `rust-bindgen` in as many ways as we can, but we should also seriously entertain the idea of writing our own modular C preprocessor because of https://github.com/tcdi/pgx/issues/316 also being a concern.

PR #818, opened 2022-10-24
Generate bindings for functions provided by the `cshim` using bindgen.

thomcc:
Depends on #813, so ignore the 3 patches that come from that.

Need to think a little bit about the approach here, since this is kinda hacky. It might be better done after #798, but I'm not 100% sure yet.

### DIFF
--- a/pgx-pg-sys/build.rs
+++ b/pgx-pg-sys/build.rs
@@ -632,23 +632,20 @@ fn build_shim_for_version(
     eprintln!("shim_dst={}", shim_dst.display());
 
     std::fs::create_dir_all(shim_dst).unwrap();
-
-    if !std::path::Path::new(&format!("{}/Makefile", shim_dst.display())).exists() {
-        std::fs::copy(
-            format!("{}/Makefile", shim_src.display()),
-            format!("{}/Makefile", shim_dst.display()),
-        )
-        .unwrap();
+    for entry in std::fs::read_dir(shim_src)? {
+        let entry = entry?;
+        let dest_file = shim_dst.join(entry.file_name());
+        if !dest_file.exists() {
+            std::fs::copy(entry.path(), &dest_file).unwrap_or_else(|e| {
+                panic!(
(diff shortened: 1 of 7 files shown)


pgx-pg-sys/build.rs
```diff
@@ -161,11 +161,14 @@ fn generate_bindings(
     is_for_release: bool,
 ) -> eyre::Result<()> {
     let major_version = pg_config.major_version().wrap_err("could not determine major version")?;
-    let mut include_h = build_paths.manifest_dir.clone();
-    include_h.push("include");
-    include_h.push(format!("pg{}.h", major_version));
-
-    let bindgen_output = run_bindgen(&pg_config, &include_h)
+    let header_src = format!(
```
eeeebbbbrrrr, line 164: cute.  

I wonder if our c-shim should be last?  It probably doesn't matter, but I guess bindgen would start by picking up its #includes first.

thomcc: > I wonder if our c-shim should be last?

I did it first in order to catch any cases where it is using files without including them.

thomcc: Admittedly we'd probably catch that when building the cshim, so I don't mind putting it second really.

thomcc:
Not sure why it passes locally but it seems that this needs to have the path to the cshim added to the include when building the cshim (this is as much a note for future me as anybody else). That's not hard, I'll do that in a bit (this will need rebasing anyway after #813 lands).

thomcc:
I'm going to say we can't actually do this one until we drop pg10 support, due to the bool differences.

workingjubilee:
Blocked on:
- https://github.com/tcdi/pgx/pull/830

BradyBonnette:
@workingjubilee @thomcc #830 has been merged into develop.

eeeebbbbrrrr:
@thomcc if you have any interest in moving this forward, please feel free to update and re-open.

PR #819, opened 2022-10-25
Reduce Array assert! in release builds

workingjubilee:
Due to the extremely touchy nature of Array's implementation, I left this debugging code in to make sure that anyone screamed if this exploded in production. It has not, and I received my first complaint about Array's slowness and overhead in benchmarks. It's time to cfg this out and investigate other possible speed improvements in addition.


thomcc (approved):
This is seems pretty heavyweight even for `debug_assertions`, so definitely during release it's a lot.

workingjubilee:
It's definitely a candidate for total removal sometime eventually. The Array code was just in a sufficiently execrable state, and sufficiently deoptimized already, that this sadly is not as bad a hit as it might seem: all of these bytes will still be in cache in many cases.

PR #820, opened 2022-10-26
Update clap and related dependencies

thomcc:
Fixes the issue with textwrap in https://github.com/tcdi/pgx/pull/629#issuecomment-1292304458. Although, after I did this, I realized there's a semver-compatible textwrap, and this could have just been pushing up the result of `cargo update -p textwrap --precise 0.15.2`. Still, we probably do want to make these changes, so here they are.

(Note that this still builds fine in 1.61)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -277,7 +277,7 @@ dependencies = [
  "atty",
  "cargo_metadata",
  "cargo_toml",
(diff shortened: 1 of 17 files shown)


eeeebbbbrrrr (approved):
LGTM!  ship it!

PR #822, opened 2022-10-27
Multiple Linux distros nightly run -- re-publishing content from #796

BradyBonnette:
This is a re-issue of content from #796 

This is because ultimately we wanted to merge 796 directly into master, but it already had upstream changes from the `develop` branch that we did not want to merge in pre-emptively.

PR #796, opened 2022-10-20
Adds various Linux distro tests to CI

BradyBonnette:
Originally, this branch was named `fedora-ci`, but as I started developing it I realized that we could be trying out builds in other various distros. With some input from @eeeebbbbrrrr we decided to add as much as we could here for the time being. There is still some stuff we plan on adding in the future, but for now this should cover a lot of use cases.

PR #823, opened 2022-10-27
Re-publishing content from #796 (#822)

BradyBonnette:
This is a re-issue of content from #796 

This is because ultimately we wanted to merge 796 directly into master, but it already had upstream changes from the `develop` branch that we did not want to merge in pre-emptively.


eeeebbbbrrrr:
LGTM.  Merge it

PR #826, opened 2022-10-28
Make default template use UTC timezone

Smittyvb:
Sets the timezone to UTC by default. This removes the need to do a `SET TIME ZONE 'UTC'` in every test that does stuff with dates, which can occur a lot in some extensions. It can be confusing when the tests you write pass on your machine but fail for other people (and CI machines) in other timezones.

I also updated the CI to use the [Pacific/Chatham (UTC+12:45)](https://en.wikipedia.org/wiki/Chatham_Standard_Time_Zone) so that the timezone-invariance of the tests is verified.

### DIFF
--- a/pgx-tests/src/tests/datetime_tests.rs
+++ b/pgx-tests/src/tests/datetime_tests.rs
@@ -409,8 +409,6 @@ mod tests {
         .try_into()
         .unwrap();
 
-        // prevents PG's timestamp serialization from imposing the local servers time zone
-        Spi::run("SET TIME ZONE 'UTC'");
         let json = json!({ "time stamp with timezone test": time_stamp_with_timezone });
 
         // but we serialize timestamps at UTC
@@ -419,9 +417,6 @@ mod tests {
 
     #[pg_test]
     fn test_timestamp_serialization() {
-        // prevents PG's timestamp serialization from imposing the local servers time zone
-        Spi::run("SET TIME ZONE 'UTC'");
-
         let datetime = PrimitiveDateTime::new(
             time::Date::from_calendar_date(2020, time::Month::try_from(1).unwrap(), 1).unwrap(),
             time::Time::from_hms(12, 34, 54).unwrap(),
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -87,6 +87,13 @@ jobs:
         sudo chmod a+rwx `/usr/lib/postgresql/$PG_VER/bin/pg_config --pkglibdir` `/usr/lib/postgresql/$PG_VER/bin/pg_config --sharedir`/extension /var/run/postgresql/
         echo ""
 
+        echo "----- Setting timezone -----"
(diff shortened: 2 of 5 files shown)


pgx-tests/src/tests/datetime_tests.rs
```diff
@@ -410,7 +410,6 @@ mod tests {
         .unwrap();
 
         // prevents PG's timestamp serialization from imposing the local servers time zone
```
workingjubilee, line 412:
```suggestion
```

workingjubilee (changes requested):
Would you be so kind as to document the reason for such a deliberately eccentric timezone in the file? It doesn't seem like it will remain obvious in the future.

workingjubilee:
We talked about this and think that we want to make sure it is easily configurable by the users through PGX for setting the timezone for tests rather than setting it like this for everyone. It's not really clear how to override this.

Smittyvb:
@workingjubilee You can override this by doing this in the `pg_test` module:
```rust
#[cfg(test)]
pub mod pg_test {
    ...
    pub fn postgresql_conf_options() -> Vec<&'static str> {
        // return any postgresql.conf settings that are required for your tests
        vec![
            "America/Blah"
        ]
    }
}
```

since later config entries overrides earlier ones.

Actually, another way of implementing this PR would be to make the default project files specify a timezone in the `postgresql_conf_options`, and letting the user remove that line if they wish. Would you prefer that way?

eeeebbbbrrrr:
> since later config entries overrides earlier ones.

This sounds like relying on a Postgres implementation detail that's subject to change.

I'm not overly excited about defaulting tests to run in UTC when it's easy enough for extension authors to set it.

> Another way of implementing this PR would be to make the default project files specify a timezone in the postgresql_conf_options, and letting the user remove that line if they wish. Would you prefer that way?

Maybe something like changing the function definition to:

```rust
fn postgres_conf_options(timezone: &mut Option<String>) -> Vec<String> {

}
```

Where pgx calls it with either `Some("UTC")` or even `None`.  If the user wants to change it, they can `*timezone = Some("America/Blah");`

workingjubilee:
I'm okay with generating the `postgresql_conf_options` with a default timezone in them. An explicit timezone arg seems dandy too. It definitely is something we should be aiming at making clear and user-visible it is happening, whatever way we do it.

As someone who's better at Rust than SQL or Postgres implementation details (though the last one is catching up frighteningly fast...), I shouldn't have to know about the intricacies of how configuring the timezone goes down. Ideally I would rather not even know about the `SET TIME ZONE` SQL command! I should just be able to see a line of configuration code and change it however I need it to be.

Smittyvb:
I updated the PR to generate the `postgresql_conf_options` with a default UTC timezone.

eeeebbbbrrrr:
Hey. I’m just gonna close this. I’m not convinced this is something pgx should be doing. 

It’s really simple for an extension to set the time zone it needs during testing and pgx forcing it to be something other than the server default is really surprising. 

I do appreciate your efforts here, @Smittyvb.

PR #827, opened 2022-10-29
Add bindings for typedefs within replication/logical.h to pg_sys

agamble:
Specifically I'm interested in the fields of the LogicalDecodingContext struct. At the moment the struct is empty in pg_sys.

Having access to the LogicalDecodingContext allows a couple new plugin use cases around manipulating the Postgres WAL.

### DIFF
--- a/pgx-pg-sys/include/pg11.h
+++ b/pgx-pg-sys/include/pg11.h
@@ -81,6 +81,7 @@ Use of this source code is governed by the MIT license that can be found in the 
 #include "parser/parser.h"
 #include "parser/parsetree.h"
 #include "postmaster/bgworker.h"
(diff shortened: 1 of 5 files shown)


workingjubilee:
Hello, are there any C macros in this header that you would like to use in Rust?

Smittyvb:
This PR needs to be updated to add the bindings to `pg15.h` as well now that PostgreSQL 15 support has been added.

workingjubilee:
No rush, it will also risk conflicting with #830 probably.

BradyBonnette:
#830 has landed

eeeebbbbrrrr:
Thank you!

PR #828, opened 2022-10-30
Problem: PostgresType can't be derived on an enum

### LINKED ISSUE #726: Any reason why PostgresType couldn't be an enum?
At the moment, it seems one can derive `PostgresType` from struct only, so in cases where I want an enum as a Postgres type, I have to wrap it into a structure and derive from it.

I wonder if there's any reason why this is impossible/undesirable to have this derivation allowed? However, if there's no reason besides lack of interest/prior need, I'd be happy to prepare a PR.

yrashk:
In some cases, a Rust enum is not meant to be Postgres enumerated type but just a type that can be managed and serialized using Rust and its ecosystem.

A workaround for this is to wrap an enum into a struct and derive `PostgresType` on it, but it seems like a lot of ceremony for no benefit.

Solution: allow PostgresType to be derived on an enum

Again, this is not to be confused with `PostgresEnum` which derives Postgres enumerated type. This one does not.

Closes #726

### DIFF
--- a/pgx-macros/src/lib.rs
+++ b/pgx-macros/src/lib.rs
@@ -671,6 +671,11 @@ struct Dog {
     treats_recieved: i64,
     pets_gotten: i64,
 }
+
+#[derive(Debug, Serialize, Deserialize, PostgresType)]
+enum Animal {
+    Dog(Dog),
+}
 ```
 
 Optionally accepts the following attributes:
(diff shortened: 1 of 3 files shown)


eeeebbbbrrrr (approved):
I think this looks good.  I'm surprised this is such a small change.

I'm going to go ahead and merge this, but if you will, please circle back with another PR that adds an example to "custom_types" and also updates that crate's readme (https://github.com/tcdi/pgx/tree/master/pgx-examples/custom_types)

yrashk:
I will prepare a follow-up PR in the coming days!

yrashk:
@eeeebbbbrrrr sent #861 as requested

PR #831, opened 2022-11-01
Remove `pgx::datum::Array::over`

workingjubilee:
This drops the highly problematic function. It doesn't follow up on all code improvements that can be gained, because some of those also relate to the `as_slice` story, but I wanted to make sure that this much, at least, happened at the earliest opportunity, since the next pgx version must be 0.6.0 and thus a breaking version (and honestly it's already a significant amount).

### DIFF
--- a/pgx-tests/src/tests/array_tests.rs
+++ b/pgx-tests/src/tests/array_tests.rs
@@ -142,21 +142,6 @@ fn get_arr_ndim(arr: Array<i32>) -> libc::c_int {
     unsafe { RawArray::from_array(arr) }.unwrap().dims().len() as _
 }
 
-#[pg_extern]
-#[allow(deprecated)]
-fn over_implicit_drop() -> Vec<i64> {
-    // Create an array of exactly Datum-sized numbers.
-    let mut vec: Vec<i64> = vec![1, 2, 3, 4, 5];
-    let mut nulls = vec![false, false, true, false, false];
-    // Verify we uphold the length contract.
-    assert_eq!(vec.len(), nulls.len());
-    let len = vec.len();
-    // Create an Array...
-    let _arr = unsafe { Array::<'_, i64>::over(vec.as_mut_ptr().cast(), nulls.as_mut_ptr(), len) };
-    vec
-    // Implicit drop of _arr
-}
-
 // This deliberately iterates the Array.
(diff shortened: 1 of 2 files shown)


pgx/src/datum/array.rs
```diff
@@ -20,7 +20,7 @@ use std::marker::PhantomData;
 use std::{mem, ptr, slice};
 
 pub struct Array<'a, T: FromDatum> {
-    ptr: Option<NonNull<pg_sys::varlena>>,
+    ptr: NonNull<pg_sys::varlena>,
     raw: Option<RawArray>,
```
workingjubilee, line 24: I want to have both of these stop being Options, but due to some other functions and the Drop implementation, that becomes a much larger cleanup job (the Option makes it easier to Drop the Array, basically). So, I opened this PR first because it can go separately.

pgx/src/datum/array.rs
```diff
@@ -498,14 +445,13 @@ impl<'a, T: FromDatum> FromDatum for Array<'a, T> {
         is_null: bool,
         _typoid: u32,
     ) -> Option<Array<'a, T>> {
-        if is_null || datum.is_null() {
+        if is_null {
             None
         } else {
-            let ptr = datum.cast_mut_ptr();
+            let ptr = NonNull::new(datum.cast_mut_ptr())?;
```
workingjubilee, line 451: This avoids repeating the null check.

pgx-tests/src/tests/array_tests.rs
```diff
@@ -142,21 +142,6 @@ fn get_arr_ndim(arr: Array<i32>) -> libc::c_int {
...
-    let mut vec: Vec<i64> = vec![1, 2, 3, 4, 5];
-    let mut nulls = vec![false, false, true, false, false];
-    // Verify we uphold the length contract.
-    assert_eq!(vec.len(), nulls.len());
-    let len = vec.len();
-    // Create an Array...
-    let _arr = unsafe { Array::<'_, i64>::over(vec.as_mut_ptr().cast(), nulls.as_mut_ptr(), len) };
-    vec
-    // Implicit drop of _arr
-}
```
workingjubilee, line 158: This test kinda existed solely to demonstrate `Array::over` is busted, really.

workingjubilee (commented):
I could probably golf this more, but most interesting reductions quickly run into `as_slice` also existing.

PR #834, opened 2022-11-01
Working on fixing our story around `pg_try()` and errors and such

eeeebbbbrrrr:
This commit makes structural changes like moving `log.rs` to the pgx-pg-sys create and splitting out the (hand written) `PgSqlErrorCode` enum into its own module.

Additionally, `setjmp.rs` was renamed to `ffi.rs` and `guard.rs` renamed to `panic_handling.rs`.

From here a number of (re)exports from pgx-pg-sys, and subsquently pgx and pgx::prelude, needed to be adjusted, and the code updated appropriately.

---

The big change to our error handling is that we **only** call `pgx_ereport()` in **one** spot now -- the spot where the stack has completely unwound from our Rust panic and sigsetjmp shenanigans (ie, our top-level `#[pg_guard]` handler).  
The previously existing `ereport()` function has been changed to a macro, and no longer directly calls `pgx_ereport()`, but instead instantiates a `PgxPanicData` and does a `panic_any()` with it.

Other touchups include...
   - no longer leaking (Rust) heap allocated CStrings during error handling
   - `error!(), warning!(), log!(), etc` macros all delegate to the new `ereport!()` macro
   - the C `pgx_ereport()` function has been rewritten to allow pgx to pass in its file:line:column information and have Postgres use it correctly


eeeebbbbrrrr:
This is still a draft.  More work is to follow to take another look at `pg_try()`.  I really want `pg_try()` and `PgTryResult` to be a nicer (and safe/correct) wrapper around Postgres' C `PG_TRY/PG_CATCH` macros.  I have an idea and am going to look at it next.

eeeebbbbrrrr:
As the code stands right now...

All the basic panic handling and error reporting we have now no longer jumps the sharkstack, always includes location when one of our `ereport!` macros are used (`log!(), error!(), warning!()`, etc), and doesn't leak error messages in Rust-allocated heap.  

And we can now properly report SQL error codes via the (now correct) ereport!() macro (was previously a terrible function)

I am also going to audit pgx for our uses of panic!() and swap them out for pgx::error!().  There's no functional difference, but from the error reporting side, error!() will include function name too.

`pg_try()` still needs some TLC, but all of this had to be done first to clean things up.

yrashk:
I'll keep an eye on this development as it may affect https://github.com/supabase/pgx-contrib-spiext

Have you had a chance to look what my sub-transaction and catch implementation there?

eeeebbbbrrrr:
I wanted to pull the latest commit message forward as a comment:

Introduce a new PgTryBuilder type for properly mimicking Postgres' `PG_TRY/PG_CATCH/PG_FINALLY` C macro construct. This completely replaces the old `pg_try()` function (which is now gone and was severely bugged).

`PgTryBuilder` lets users execute code and respond to any `PgSqlErrorCode` (which now includes a `::RustPanic` variant!) to either do cleanup work and rethrow the error, or otherwise ignore it and return a different value.  Documentation, examples, and tests for PgTryBuilder will be coming in another commit.  Rust drop semantics work properly as well.

There's a lot of other nice changes in here as well...

- The `CONTEXT` ("errcontext_msg") slot of a raised Postgres error now contains as much stack information we can gather while pgx is propagating errors.  Additionally, the `LOCATION` slot contains the `file:line` of whatever raised the first error -- either a Rust panic or a Postgres error.  This did require pushing down a `#[track_caller]` attribute on **every** Postgres FFI we wrap.

Overall, error handling has been cleaned up such that the **only** place we call back to postgres to raise an ERROR (or rethrow one) is the top-level `guard(|| ...)` function, which is used by `#[pg_guard]` and friends.

(note to self: other users of that function throughout the code need to be audited.  I think they need to instead use PgTryBuilder).

- The `ereport!()` macro has been expanded to support a shorthand syntax where the first argument can be the error level name.  Example:  `ereport!(ERROR, PgSqlErrorCode::ERRCODE_INTERNAL_ERROR, "oh no!");"`.  And the `ERROR/PANIC/FATAL` versions know they don't return and insert an `unreachable!()`.

- The `error!()` macro now also supprots specifying an error code:  Example: `error!(PgSqlErrorCode::ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE, "too small")`

eeeebbbbrrrr:
> Introduce a new PgTryBuilder type for properly mimicking Postgres' `PG_TRY/PG_CATCH/PG_FINALLY` C macro construct. This completely replaces the old `pg_try()` function (which is now gone and was severely bugged).

As an example, driving this code via psql:

```rust
#[pg_extern]
fn try_it(input: i32) -> Option<i32> {
    PgTryBuilder::new(|| {
        if input < 100 {
            error!(PgSqlErrorCode::ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE, "a number is too small")
        }

        // .. do some work ...
        // cause a different error to happen
        if input == 666 {
            unsafe {
                pg_sys::relation_open(0, 0);
            }
        } else if input >= 1000 {
            panic!("rust panic with input of: {}", input)
        }
        Some(input)
    })
    .catch_when(PgSqlErrorCode::ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE, |_caught_error| None)
    // can add multiple .catch_when() calls to handle different errors differently
    // .finally(|| { }) as well
    .execute()
}
```

... gives us some output like:

```sql
[v13.8][46196] strings=# select strings.try_it(42);
 try_it 
--------
       
(1 row)

Time: 105.914 ms
[v13.8][46196] strings=# select strings.try_it(101);
 try_it 
--------
    101
(1 row)

Time: 3.866 ms
[v13.8][46196] strings=# select strings.try_it(666); \errverbose
ERROR:  could not open relation with OID 0
CONTEXT:  pgx-examples/strings/src/lib.rs:23:17
Time: 3.015 ms
ERROR:  XX000: could not open relation with OID 0
CONTEXT:  pgx-examples/strings/src/lib.rs:23:17
LOCATION:  relation_open, relation.c:62
[v13.8][46196] strings=# select strings.try_it(5000); \errverbose
ERROR:  rust panic with input of: 5000
Time: 2.633 ms
ERROR:  XX000: rust panic with input of: 5000
LOCATION:  lib.rs:16
[v13.8][46196] strings=# 
```

I asked psql to show me the error's details (`\errverbose`).

For the first error (caused by a bad call to `relation_open()`) you can see that the LOCATION points to somewhere in the Postgres sources, because that's what raised the ERROR, but the CONTEXT points to the line in the extension Rust sources that called `relation_open()`!  

In the second error case, there's no CONTEXT because we only have one entry on the stack, and that LOCATION is line 16 of our extenion's `lib.rs` file.  Sadly for LOCATION, Postgres purposely strips all path entries up to the last forward slash.  Bummer.

Anyways, it's 2am here and I need sleep, but I wanted to share what this is looking like.

@yrashk, I have not had a chance to look at what you were working on for subtransactions.  The goal with this is to fix edge cases in our error handling and to get rid of the terrible `pg_try()` function.

pgx-pg-sys/cshim/pgx-cshim.c
```diff
@@ -43,11 +43,90 @@ void pgx_elog_error(char *message) {
...
-    ereport(level,
-            (errcode(code),
-                    errmsg("%s", message), errcontext_msg("%s:%d:%d", file, lineno, colno)));
+PGDLLEXPORT void pgx_errcontext_msg(char *message);
+void pgx_errcontext_msg(char *message) {
+    errcontext_msg("%s", message);
+}
+
+PGDLLEXPORT void pgx_ereport(const int level, int code, char *message, char *detail, char *funcname, char *file, int lineno, char *context);
+void pgx_ereport(const int level, int code, char *message, char *detail, char *funcname, char *file, int lineno, char *context) {
```
eeeebbbbrrrr, line 52: this is incredibly gross, I know. I was playing C preprocessor in my head to unroll the #define ereport() macro.

I think I have a cleaner approach and will take a look at that shortly.

eeeebbbbrrrr:
One consideration here is what does a `PgTryBuilder` implementation look like under postgrestd (ping, @workingjubilee).

I think I know -- basically it'll have to do its own `sigsetjmp` handling just like the C PG_TRY/PG_CATCH macros do.  And instead of doing it over each individual FFI boundary (like our `#[pg_guard]` macro does), it'll just be over the whole PgTryBuilder block.

We definitely need to support this under postgrestd -- I think it'll be incredibly useful for pl/rust.  Being able to write custom functions that handle errors as "easily" as pl/pgsql does will be a great feature there.

pgx/src/callbacks.rs
```diff
@@ -328,7 +328,7 @@ where
                 for hook in hooks.iter() {
                     let hook = hook.borrow();
                     if let Some(hook) = hook.as_ref() {
-                        crate::guard::guard(|| (hook.0)(my_subid, parent_subid));
+                        pg_sys::guard(|| (hook.0)(my_subid, parent_subid));
```
thomcc, line 331: This will longjmp in the case of errors over a bunch of rust stuff, which will probably be leaked at best. I think what we want here is this whole fn to have the guard

thomcc (commented):
This largely looks like an improvement. I'd like to do another pass over the final version (I think you're updating it?), and I only skimmed the examples. Dumping what I have now while I go for lunch.

pgx-pg-sys/src/submodules/panic.rs
```diff
@@ -0,0 +1,389 @@
...
+    }
+
+    /// Report this [PgErrorReport], which will ultimately be reported by Postgres at the specified [PgLogLevel]
+    #[track_caller]
+    pub fn report(self, level: PgLogLevel) {
+        ErrorReportWithLevel { level, ereport: self, stack: Default::default() }.report()
+    }
+}
+
+thread_local! { static PANIC_LOCATION: Cell<Option<ErrorReportLocation >> = Cell::new(None) }
```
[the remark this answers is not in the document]
eeeebbbbrrrr: TIL

pgx-pg-sys/src/submodules/panic.rs
```diff
@@ -0,0 +1,389 @@
...
+                        }
+
+                        pgx_errcontext_msg(errdata.location.to_string().as_pg_cstr());
+                        for location in errdata.stack {
+                            pgx_errcontext_msg(location.to_string().as_pg_cstr());
+                        }
+                        pg_re_throw();
+
+                        // SAFETY: Postgres **will not** return after a call to `pg_re_throw()`
+                        unreachable_unchecked()
```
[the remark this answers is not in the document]
eeeebbbbrrrr: `pg_re_throw()` isn't ours, it belongs to Postgres.

pgx/src/callbacks.rs
eeeebbbbrrrr: That thought had crossed my mind.  I'm glad you pointed it out.

pgx-pg-sys/src/submodules/panic.rs
```diff
@@ -0,0 +1,389 @@
...
+                    PgSqlErrorCode::ERRCODE_INTERNAL_ERROR,
+                    message,
+                    take_panic_location(),
+                ),
+                stack: Default::default(),
+            },
+            payload: e,
+        }
+    } else {
+        // not a type we understand, so it gets raised as an INTERNAL_ERROR at the ERROR level
```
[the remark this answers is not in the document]
thomcc: If you want. Be sure to write a test that ensures you get the right `Box<dyn Any + Send>` from the downcast. By "it's a pain" I actually mean "it's subtle and easy to accidentally write it in such a way that always returns true due to autoref". (I found this out the hard way by writing a version that recursed, got it wrong, and always blew the stack if that code was hit)

I think in practice just checking once would be fine, (and if it gets hit we should consider maybe warning that the user has a `panic_any` somewhere they should have a `resume_unwind` or something).

pgx-pg-sys/src/submodules/panic.rs
```diff
@@ -0,0 +1,423 @@
...
+        let line = ereport.ereport.location.line as i32;
+        MemoryContextSwitchTo(old_cxt);
+
+        // before calling `pgx_ereport` it's imperative we drop everything Rust-allocated we possibly can.
+        // `pgx_ereport` very well might `longjmp` to somewhere else, either in pgx or Postgres, and
+        // we'd rather not be leaking memory during error handling
+        //
+        // the few `.as_pg_cstr()`s do their allocation in Postgres' `CurrentMemoryContext`, so they'll
+        // be cleaned up by Postgres at the right time
+        let _ = ManuallyDrop::into_inner(ereport);
```
thomcc, line 389: I'm pretty sure this needs to use [`ManuallyDrop::drop(&mut ereport)`](https://doc.rust-lang.org/stable/std/mem/struct.ManuallyDrop.html#method.drop) instead. IIUC, having the drop automatically run in this function causes it no longer be a "plain old frame".

pgx/src/bgworkers.rs
```diff
@@ -200,12 +203,12 @@ impl BackgroundWorker {
             pg_sys::StartTransactionCommand();
             pg_sys::PushActiveSnapshot(pg_sys::GetTransactionSnapshot());
         }
-        let result = pg_sys::guard(|| transaction_body());
         unsafe {
+            let result = pg_sys::guard(|| transaction_body());
```
thomcc, line 207: Is this right? I think it should be something like `PgxTryBuilder::new(|| transaction_body()).execute()` (possibly via a `pg_try` helper). `guard` pretty much should only be used in `extern "C"` fn`s, I think.

pgx/src/callbacks.rs
```diff
@@ -191,7 +192,7 @@ where
                 // effectively 'take' the hook from the internal RefCell
                 if let Some(hook) = hook.replace(None) {
                     // and execute it under guard for proper panic/elog(ERROR) handling
-                    crate::guard::guard(hook.0);
+                    pg_sys::guard(hook.0);
```
thomcc, line 195: This is an `extern "C"` fn so it does need to guard, but I think the guard should be over the entire callback, e.g. starting on line 163 and going to 198.

It looks like that's what `#[pg_guard]` does... but it also [applies `#[no_mangle]`](https://github.com/tcdi/pgx/blob/077c447543e863e67f2d774be4dfbf05372b26d9/pgx-utils/src/rewriter.rs#L72), which is super duper wrong for something like this!!

pgx/src/callbacks.rs
```diff
@@ -310,6 +311,7 @@ where
 {
     static mut SUB_HOOKS: Option<SubCallbackMap> = None;
 
+    #[pg_guard]
```
thomcc, line 314: As mentioned above this applies `#[no_mangle]` -- totally wrong for callbacks. This will collide with any function named `callback` in C, or any other `#[pg_guard]` function named `callback`, etc. I think we should not apply `#[no_mangle]` in `#[pg_guard]`, but we could also just not use it here.

pgx-pg-sys/src/submodules/panic.rs
eeeebbbbrrrr: I read the docs and they said if you have ownership of the thing then to use `::into_inner`.

I'll defer to you tho on that...

pgx/src/bgworkers.rs
eeeebbbbrrrr: hmm.  you're right.

pgx/src/callbacks.rs
eeeebbbbrrrr: Yup.  I've fixed up the no_mangle issues as well

pgx-pg-sys/src/submodules/panic.rs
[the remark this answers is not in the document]
eeeebbbbrrrr: I'm not gonna bother with it right now.  If we get some user report about it we'll know what to do.

pgx-pg-sys/src/submodules/panic.rs
thomcc: It's not needed, you're right. Actually, ManuallyDrop is not -- it turns out that it is allowed for us to jump over stack frames with landing pads, so long as those landing pads are no-ops. This is surprising since it defeats the main benefit of making that UB, but honestly its *extremely* good news.

pgx-pg-sys/src/submodules/panic.rs
```diff
@@ -0,0 +1,424 @@
...
+    }
+}
+
+impl From<&Location<'_>> for ErrorReportLocation {
+    fn from(location: &Location<'_>) -> Self {
+        Self { file: location.file().to_string(), line: location.line(), col: location.column() }
+    }
+}
+
+impl From<Option<&Location<'_>>> for ErrorReportLocation {
```
thomcc, line 49: I'd probably not impl this and just ask users to `opt.map(ErrorReportLocation::from).unwrap_or_default()` if that's *really* what they want. But I don't feel strongly about it -- it's semantically optional information.

pgx-pg-sys/src/submodules/pg_try.rs
```diff
@@ -0,0 +1,94 @@
...
+                        let sqlerrcode = errdata.sqlerrcode;
+                        let panic = CaughtError::PostgresError(errdata);
+                        (sqlerrcode, panic)
+                    }
+                };
+
+                // Postgres source docs says that a PG_TRY/PG_CATCH/PG_FINALLY block can't have
+                // both a CATCH and a FINALLY.
+                //
+                // We'll allow it, but if a catch handler (re)throws then the finally block will
```
thomcc, line 72: We should be sure to document this, because it's somewhat surprising that it's not run. Although, I guess I don't know what happens if you throw inside `finally` in like, Java/JavaScript/C# or whatever.

That said, I'm a fan of this behavior. If users *do* want a `finally` that always runs, they are free to do so by putting some type with a Drop impl on the stack (or via `scope_guard` for example).

pgx-pg-sys/src/submodules/errcodes.rs
```diff
@@ -0,0 +1,1291 @@
+use std::fmt::{Display, Formatter};
+
+/// This list of SQL Error Codes is taken directly from Postgres 12's generated "utils/errcodes.h"
```
thomcc, line 3: At some point we should find a way to generate this. Will think about it when I do some bindgen work.

pgx-pg-sys/src/submodules/errcodes.rs
```diff
@@ -0,0 +1,1291 @@
...
+#[allow(non_camel_case_types)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash, Ord, PartialOrd)]
+pub enum PgSqlErrorCode {
```
thomcc, line 6: Some musing to think about later that occurred to me when reading this:

I do wonder about having this be an enum rather than a `#[repr(transparent)] struct PgSqlErrorCode(c_int)`, along with a bunch of associated constants (you could still `match` on that so long as we use `#[derive]` to get PartialEq and friends).

That would let us avoid needing to translate all unknown values to `ERRCODE_INTERNAL_ERROR`.  would also potentially let extensions add their own error codes, but I don't know if that's a thing that makes sense at all (given that we defined `RustPanic` it *feels* like it might make sense, but I don't really know.

But yeah, I could easily be wrong, and even if I'm not -- we probably shouldn't do that in the same PR anyway :p.

pgx-pg-sys/src/submodules/errcodes.rs
```diff
@@ -0,0 +1,1291 @@
...
+    None = -1,
```
thomcc, line 7: Along those lines, ISTM like we could probably ditch this, and push users to use `Option<PgSqlErrorCode>`.

I'm also not 100% sure when one would want to use `PgSqlErrorCode::None` over `PgSqlErrorCode::ERRCODE_SUCCESSFUL_COMPLETION` -- I guess the first is closer to `Option::None`, and the second represents error code from C.

thomcc (approved):
This is really great! Couple non-blocking comments, but really this has ended up better than I had hoped!

pgx-pg-sys/src/submodules/pg_try.rs
eeeebbbbrrrr: Indeed.  I'll be putting up some doc comments today that will address this.

pgx-pg-sys/src/submodules/errcodes.rs
eeeebbbbrrrr: I'll take that `None` out now.  You're right that it SUCCESSFUL_COMPLETION is the correct choice in that case.

thomcc (approved):
This looks great! Really thrilled to see this cleaned up. Have a few comments but nothing major

pgx-pg-sys/src/lib.rs
```diff
@@ -459,6 +505,16 @@ mod all_versions {
         #[link_name = "pgx_SpinLockFree"]
         pub fn SpinLockFree(lock: *mut pg_sys::slock_t) -> bool;
     }
+
+    #[inline(always)]
+    pub fn MemoryContextSwitchTo(context: crate::MemoryContext) -> crate::MemoryContext {
```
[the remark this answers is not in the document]
eeeebbbbrrrr: ha, yes!

pgx-pg-sys/src/submodules/panic.rs
```diff
@@ -0,0 +1,380 @@
...
+/// to Postgres, and the `#[pg_guard]` macro takes care of this for you.
+///
+/// In other words, this isn't the function you're looking for.
+///
+/// You're probably looking for the `#[pg_guard]` macro.
+///
+/// Alternatively, if you're trying to mimic Postgres' C `PG_TRY/PG_CATCH` API, then you instead
+/// want [`crate::pg_try::PgTryBuilder`].
+///
+/// # Safety
```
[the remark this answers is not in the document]
eeeebbbbrrrr: thanks!

yrashk:
I've updated `pgx-contrib-spiext` to use this for checked commands: https://github.com/supabase/pgx-contrib-spiext/commit/45ccd6822e0b89d895a90bbe2fc3f9209f9be955

PR #838, opened 2022-11-03
Add traverse function

willmurnane:
Automatically generates a set of functions that traverse all `Node` objects reachable from whatever origin point is given.

I tested this on my private extension by adding a planner hook which simply calls `traverse` on the planned statement, and running my thorough regression suite. This hit a bunch of edge cases (see lines 385, 626, and 639 for examples), which I dealt with. I haven't added any kind of testing to this repository, though, since I'm not sure where it would make sense to do so. Some of those special-case lists almost certainly need more entries in them.

This is currently broken into two commits: one which adds the functionality with a bunch of debug prints, and one which removes all the debug prints. Feel free to squash.

### DIFF
--- a/pgx-pg-sys/build.rs
+++ b/pgx-pg-sys/build.rs
@@ -322,26 +322,14 @@ fn extract_oids(code: &syn::File) -> proc_macro2::TokenStream {
         }
     }
 }
-fn is_node_struct(name: &str, node_types: &HashSet<String>, nodes: &Vec<StructDescriptor>) -> bool {
-    let node = nodes.iter().filter(|n| n.struct_.ident.to_string() == name).next();
-    match node {
-        None => false,
-        Some(n) => {
-            let first_field = n.struct_.fields.iter().next();
-            match first_field {
-                None => false,
-                Some(f) => match &f.ty {
-                    Type::Path(p) => {
-                        let first_field_type_name =
-                            p.path.segments.first().unwrap().ident.to_string();
-                        if first_field_type_name == "NodeTag" {
-                            return true;
-                        }
-                        is_node_struct(first_field_type_name.as_ref(), node_types, nodes)
-                    }
-                    _ => false,
-                },
-            }
+
+/// Produces code which calls `walker_fn` on the given identifier-like field name.
+fn walk_field_definition(field_name: proc_macro2::TokenStream) -> proc_macro2::TokenStream {
+    quote! {
+        if !#field_name.is_null() {
+            let item = unsafe { &mut *(#field_name as *mut Node) };
+            walker_fn(item, context);
+            Node::traverse::<T>(item, walker_fn, context);
         }
     }
 }
@@ -397,7 +385,7 @@ fn impl_pg_node(
     }
 
     // the set of types which subclass `Node` according to postgres' object system
-    let mut node_set = HashSet::new();
+    let mut node_set = HashMap::new();
     // fill in any children of the roots with a recursive DFS
     // (we are not operating on user input, so it is ok to just
     //  use direct recursion rather than an explicit stack).
@@ -405,14 +393,35 @@ fn impl_pg_node(
         dfs_find_nodes(root, &struct_graph, &mut node_set);
     }
 
-    let mut nodes = node_set.into_iter().collect::<Vec<_>>();
+    let mut nodes = node_set.values().collect::<Vec<_>>();
     if is_for_release {
         // if it's for release we want to sort by struct name to avoid diff churn
         // otherwise we don't care and want to avoid the CPU overhead of sorting
         nodes.sort_by_key(|s| &s.struct_.ident);
     }
+    // Provide a special implementation of PgNode for the Node itself: depending on the `type_` field,
+    // call the override for that type. That way, when we get a `*mut Node` out of a List object, we
+    // can call this function and let it route appropriately.
+    let mut node_traverse_body = proc_macro2::TokenStream::new();
+    let mut node_display_body = proc_macro2::TokenStream::new();
 
-    let node_types = items
+    let mut join_fields_traverse = proc_macro2::TokenStream::new();
+    join_fields_traverse.extend(
+        if std::env::var("CARGO_FEATURE_PG11").is_ok()
+            || std::env::var("CARGO_FEATURE_PG12").is_ok()
+        {
+            vec![]
+        } else if std::env::var("CARGO_FEATURE_PG13").is_ok() {
+            vec![quote! { joinleftcols }, quote! { joinrightcols }]
+        } else {
+            vec![quote! { joinleftcols }, quote! { joinrightcols }, quote! { join_using_alias }]
+        }
+        .into_iter()
+        .map(|f| quote! { self.#f })
+        .map(walk_field_definition),
+    );
+
+    let nodetag_t_values = items
         .iter()
         .filter_map(|i| match i {
             Item::Const(c) if c.ident.to_string().starts_with("NodeTag_T_") => {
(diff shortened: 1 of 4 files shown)


workingjubilee:
Adding testing in pgx-tests under node-tests.rs is fine.

workingjubilee:
traverse machine :b:roke?

willmurnane:
I think I can get somewhere useful, but there are a lot of special cases with arrays stored in structures where a `List` proper is not used. Handling those takes a good bit more code, and I’ve been busy with other things at work. I’ll try to push on this some more Monday.

workingjubilee:
understandable, have a nice day.

willmurnane:
I think this is ready for review. There's a lot of code change here, and I bet there are still a couple good rounds of cleanup to do, but it does all work for the use cases I've tested.

workingjubilee:
Will try to review this next week.

pgx-pg-sys/src/lib.rs
```diff
@@ -167,6 +167,9 @@ pub trait PgNode {
         // this is only implemented for things known to be "Nodes"
         unsafe { display_node_impl(NonNull::from(self).cast()) }
     }
+
+    /// Traverse the object graph, calling the callback at any `Node`s found as children of this object.
+    fn traverse<T>(&mut self, walker_fn: fn(*mut Node, &mut T) -> (), context: &mut T);
```
workingjubilee, line 172: I'm not terribly excited by there not being a default implementation that's valid, and by the `walker_fn` using `*mut Node`. I would prefer that we rely less on macro-generated code for implementations, as it's harder to audit, not more.

pgx-pg-sys/build.rs
```diff
@@ -392,23 +416,632 @@ fn impl_pg_node(
...
+            impl std::fmt::Display for List {
+                fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                    write!(f, "[")?;
+                    if self.length != 0 {
+                        let mut cell = unsafe { *self.head };
+                        for index in 0..self.length {
+                            if index != 0 {
+                                write!(f, ", ")?;
+                            }
+                            let item = unsafe { cell.data.ptr_value as *mut Node };
```
workingjubilee, line 541: What's the precondition that is allowing us to access this as `.ptr_value`?

pgx-pg-sys/build.rs
```diff
@@ -392,23 +416,632 @@ fn impl_pg_node(
...
+                    write!(f, "]")
+                }
+            }
+        }
+    } else {
+        quote! {
+            impl pg_sys::PgNode for List {
+                fn traverse<T>(&mut self, walker_fn: fn(*mut Node, &mut T) -> (), context: &mut T) {
+                    for index in 0..self.length {
+                        let item = unsafe { (*self.elements.offset(index as isize)).ptr_value as *mut Node };
```
workingjubilee, line 561: What's the precondition that is allowing us to access this as `.ptr_value`?

pgx-pg-sys/build.rs
```diff
@@ -392,23 +416,632 @@ fn impl_pg_node(
...
+                }
+            }
+            impl std::fmt::Display for List {
+                fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                    write!(f, "[")?;
+                    for index in 0..self.length {
+                        if index != 0 {
+                            write!(f, ", ")?;
+                        }
+                        let item = unsafe { (*self.elements.offset(index as isize)).ptr_value as *mut Node };
```
workingjubilee, line 576: It looks like this code assumes that all of the entire length of `self.elements` is valid to access, and then does so, even if every single element it encounters is itself a null pointer. It should be stating that precondition once, using [`slice::from_raw_parts`](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html) and then iterating that directly.

workingjubilee (changes requested):
I did my best, but I don't think I can properly review it until after some cleanup around the `unsafe` code. My apologies.

pgx-pg-sys/build.rs
```diff
@@ -392,23 +416,632 @@ fn impl_pg_node(
...
+                                            Node::traverse::<T>( unsafe { &mut *(*item as *mut Node) }, walker_fn, context);
+                                        }
+                                    }
+                                });
+                            }
+                        } else {
+                            ptr_problems.push(format!(
+                                "Unexpected type inside array:ptr {} -> field {}: {:?}",
+                                struct_name, field_name, ptr.elem
+                            ));
```
workingjubilee, line 978: I would kiiinda prefer it if this used a structured type and the formatting was done at panic time. But this is a fairly weak preference.

pgx-pg-sys/build.rs
willmurnane: This (and the other instance) were guarded by the "ignore `IntList` and `OidList` clause elsewhere in the code, but that check should be inside this function instead.

pgx-pg-sys/build.rs
```diff
@@ -392,23 +416,632 @@ fn impl_pg_node(
...
+                    walker_fn(self.eref as *mut Node, context);
+                    Node::traverse::<T>(unsafe { &mut *(self.eref as *mut Node) }, walker_fn, context);
+                }
+                // WRITE_ENUM_FIELD(rtekind, RTEKind);
+
+                match self.rtekind {
+                    RTEKind_RTE_RELATION => {
+                        // WRITE_OID_FIELD(relid);
+                        // WRITE_CHAR_FIELD(relkind);
+                        // WRITE_INT_FIELD(rellockmode);
```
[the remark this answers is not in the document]
workingjubilee: Fair enough, adding a note at the top that the stubbed/commented code sections are because this adheres fairly directly to the Postgres code layout might be useful in assuring they are preserved for posterity.

pgx-pg-sys/build.rs
```diff
@@ -392,23 +416,632 @@ fn impl_pg_node(
...
+    };
 
+    pgnode_impls.extend(quote! {
+        unsafe fn list_length(list: *mut List) -> i32 {
+            if list.is_null() {
+                0
+            } else {
+                (*list).length
+            }
+        }
```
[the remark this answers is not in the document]
willmurnane: It isn't, removed. Good catch.

pgx-pg-sys/build.rs
```diff
@@ -392,23 +416,632 @@ fn impl_pg_node(
...
+                    for _index in 0..self.length {
+                        let item = unsafe { cell.data.ptr_value as *mut Node };
+                        if !item.is_null() {
+                            walker_fn(item, context);
+                            Node::traverse::<T>( unsafe { &mut *(item as *mut Node) }, walker_fn, context);
+                        }
+                        if !cell.next.is_null() {
+                            cell = unsafe { *cell.next };
+                        }
+                    }
```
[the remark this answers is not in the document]
willmurnane: This is now implemented as the `walk_field_definition` function. Open to a better name ;)

pgx-pg-sys/build.rs
willmurnane: Same as above

pgx-pg-sys/build.rs
willmurnane: This is now fixed, specifically for the `List` traversal function.

pgx-pg-sys/build.rs
```diff
@@ -392,23 +416,632 @@ fn impl_pg_node(
...
+                }
+            }
+            impl std::fmt::Display for List {
+                fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                    write!(f, "[")?;
+                    for index in 0..self.length {
+                        if index != 0 {
+                            write!(f, ", ")?;
+                        }
```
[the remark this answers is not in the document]
willmurnane: I'm game to refactor this, but I couldn't find an example of this kind of formatting. The stdlib (reasonably) doesn't define a `Display` for vectors more generally, but I couldn't find an example of what an efficient list formatter looks like.

I think that:
* printing a human-readable version of a list is unlikely to be something that should happen in production, so this code won't actually be executed in a performance critical context?
* if there's a bug in this code (say it walks off the end of a list, or somehow winds up dealing with a `self` which isn't really a `List`) then printing as much of the list as we can might help in diagnosing that problem, so generating a bunch of partial results in memory would not be good
* the underlying calls to `display_node` are gonna be pretty expensive anyways

Thus, I think this implementation is reasonable, and we shouldn't get too fastidious about performance. I'm open to suggestions about how to implement this well, and would be happy to rewrite with performance in mind, but for the reasons above I didn't want to venture down that rabbit hole.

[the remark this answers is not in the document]
workingjubilee: Hmm. I'll take a crack at this detail myself later, then.

pgx-pg-sys/build.rs
workingjubilee: Fair enough.

workingjubilee (approved):
This looks much better now, thank you!

workingjubilee:
So unfortunately this PR had weird bugs that only broke during trying to do the pgx release, causing the pgx-pg-sys build.rs to break. On reflection, we think that this was probably the wrong approach. We may be interested in seeing this again if you can find a version that requires no or almost no macro code.

PR #839, opened 2022-11-03
Problem: pg_shmem_init doesn't work on pg15

yrashk:
It results in the following error:

```
FATAL:  cannot request additional shared memory outside shmem_request_hook
```

Solution: ensure use of that hook in pg15


yrashk:
I suspect GitHub malfunction as my branch has more commits now:

https://github.com/yrashk/pgx/commits/pg15-shmem

This PR only shows the first two so far.

PR #840, opened 2022-11-03
Warns about unsupported PG versions during bindgen

BradyBonnette:
References issue https://github.com/tcdi/pgx/issues/835


BradyBonnette:
Open to suggestions regarding warning output copy. The one thing I was looking for but wasn't sure how to address is: "The user sees the warning, but what can they do to resolve it?"

BradyBonnette:
Example of this in action:
image

pgx-pg-sys/build.rs
```diff
@@ -115,7 +115,17 @@ fn main() -> eyre::Result<()> {
...
     {
         pgx.iter(PgConfigSelector::All)
-            .map(|v| v.wrap_err("invalid pg_config"))
+            .map(|r| r.expect("invalid pg_config"))
+            .map(|c| (c.major_version().expect("invalid major version"), c))
+            .filter(|c| {
+                if SUPPORTED_MAJOR_VERSIONS.contains(&c.0) {
+                    true
+                } else {
+                    println!("cargo:warning=Your PGX configuration is set up to use an unsupported version of Postgres:{}", c.0);
```
eeeebbbbrrrr, line 124: I might write this as:  "{PGX_HOME} is configured for pg{c.0}, which pgx does not support"

And spell out the full path to $PGX_HOME.  Not enough context in this to see where it comes from.

Also, you can use `.filter_map()` instead of `.filter().map()`.  It's a little cleaner and you'd just return `None` or `Some(c.1).

And you probably don't need that `eyre::Result` wrapper down below on line 129 b/c none of this iter chain is returning a Result.

eeeebbbbrrrr (changes requested):

pgx-pg-sys/build.rs
workingjubilee: The `eyre::Result` needs to stay there for `#[cfg]` reasons.

eeeebbbbrrrr (commented):
LGTM.  Thanks!

PR #841, opened 2022-11-03
Updates update-versions.sh script to use pgx-version-updater

BradyBonnette:
This implements the usage of the new `pgx-version-updater` tool in the `update-versions.sh` script.


update-versions.sh
```diff
@@ -10,87 +10,95 @@
...
+# Exclude all pgx-examples Cargo.toml files from version bumping
+for file in pgx-examples/**/Cargo.toml; do
+  EXCLUDE_FROM_VERSION_BUMP+=("$file")
+done
+
+# Ensure that pgx-version-updater is installed and/or updated to the most recent version
+if command -v pgx-version-updater &> /dev/null; then
+  echo "pgx-version-updater found. Checking to see if update is necessary."
+  installed_version=$(pgx-version-updater --version | awk '{print $2}')
+  cargo_toml_version=$(cargo read-manifest --manifest-path pgx-version-updater/Cargo.toml | jq -r .version)
```
workingjubilee, line 75: Do we need to add a dependency on jq here?

update-versions.sh
```diff
@@ -10,87 +10,95 @@
...
+
+# Exclude all pgx-examples Cargo.toml files from version bumping
+for file in pgx-examples/**/Cargo.toml; do
+  EXCLUDE_FROM_VERSION_BUMP+=("$file")
+done
+
+# Ensure that pgx-version-updater is installed and/or updated to the most recent version
+if command -v pgx-version-updater &> /dev/null; then
+  echo "pgx-version-updater found. Checking to see if update is necessary."
+  installed_version=$(pgx-version-updater --version | awk '{print $2}')
```
workingjubilee, line 74: Do we need to add a dependency on awk here?

update-versions.sh
BradyBonnette: Not sure how else to parse out the version number from the Cargo.toml file. I don't think `cargo read-manifest` has a way to query into a manifest file, unless there's another `cargo`-y way of doing it that I am not aware of.

update-versions.sh
BradyBonnette: Unless there's a cool little Bash shortcut here to do this, the output of `pgx-version-updater --version` looks like this:

```
$ pgx-version-updater --version
pgx-version-updater 0.1.0
```

Unless there's a way to tell `clap` to only display the version and not `<program> <version>` in the output.

update-versions.sh
workingjubilee: if pgx-version-updater exists, as it does in this version of events, can't it query the toml itself?

update-versions.sh
workingjubilee:
```suggestion
  installed_version=$(pgx-version-updater --version)
  installed_version=${installed_version##pgx-version-updater } # strips the prefix
```

update-versions.sh
```diff
@@ -9,7 +9,6 @@
 # the LICENSE file.
 
 # requires:
-# * ripgrep
```
workingjubilee, line 12: RIP

BradyBonnette: 💀

update-versions.sh
BradyBonnette: I am not sure I follow here -- I feel like I am missing something.

My intention behind this was to have this script build+install a newer `pgx-version-updater` in the event that the currently-installed version does not match whatever version is specified in the pgx-version-updater Cargo.toml file within the pgx repo. I figured that pgx-version-updater would always have a different version than pgx-proper, which probably wouldn't update that much, and I wanted to prevent pgx-version-updater from being compiled every time this script is ran on a PGX release (or if someone did a `cargo clean`).

Are you suggesting that `pgx-version-updater` have some functionality built in that parses out `pgx-version-updater/Cargo.toml` instead of the `cargo read-manifest | jq` dance?

workingjubilee: Correct, under some flag argument likely.

update-versions.sh
```diff
@@ -21,24 +21,17 @@ fi
 
 set -e
 
-if ! command -v jq &> /dev/null; then
-  echo "Command \`jq\` was not found. Please install it and try again."
-  exit 1
-fi
-
```
BradyBonnette, line 19: RIP

workingjubilee (approved):
Yay, no new dependencies!

PR #842, opened 2022-11-04
Allow installing cargo-pgx using rustls instead of native-tls

felipe-vaultree:
This adds a feature flag `rustls-tls`, switching the `ureq` feature flags from the default `native-tls` to `tls`.
Unless explicitly disabling default features and enabling the new flag, the default behavior is unchanged.

### DIFF
--- a/cargo-pgx/Cargo.toml
+++ b/cargo-pgx/Cargo.toml
@@ -30,7 +30,7 @@ proc-macro2 = { version = "1.0.47", features = [ "span-locations" ] }
 quote = "1.0.21"
 rayon = "1.5.3"
 regex = "1.6.0"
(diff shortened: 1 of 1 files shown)


thomcc (approved):
Are you on a platform where native-tls doesn't work, or is this just for aesthetics?

Either way, seems fine, since it's off by default.

felipe-vaultree:
> Are you on a platform where native-tls doesn't work, or is this just for aesthetics?
This is to simplify building on older distributions, for older glibc compatibility, where `libssl>=1.1` isn't available.
Since libssl is only used by the CLI here, it seems reasonable to just use rustls instead of hacking a manual libssl installation.

thomcc:
Fair enough

Issue #825, opened 2022-10-28
A way to define var args functions


montanalow:
The PostgresML [pgml.predict](https://postgresml.org/user_guides/predictions/overview/) function needs to take a variable number of variously typed arguments, depending on the model the prediction is being run for. Currently, it only accepts a `FLOAT4[]`/`Vec<f32>`, which covers the base cases of bools/ints/floats, since Postgres can convert them all to FLOAT4. This is basically the same as variadic FLOAT4 args, except it doesn't have the 100 argument limit, since they are all passed as a single array. 

Now, we'd like to add `TEXT` to the mix, and we'd like to do it as efficiently as possible. Unfortunately we don't know the possible arg types at compile time, so we can't have a Rust tuple to represent a custom Postgres type, although we can dynamically define appropriate types in Postgres at runtime. We know the expected actual type for each position in the variadic args. 

I'd like to know if anyone has other ideas for us to benchmark:
- Use `TEXT[]` instead of `FLOAT4[]`. Downcast to bool/int/float as appropriate. Slightly less flexible, but I suspect more performant than full JSON.
- Use a single JSON arg instead of `FLOAT4[]`. Upside is perfect flexibility. The downside is performance and syntax clutter.
- Dynamically define a custom plpgsql method with a type signature specific for every trained model, that does the casting before calling a generic FLOAT4[] rust function. I suspect the plgpgsql function call overhead may be worse than parsing JSON, but I won't know until we benchmark it.

Vague idea I can't quite figure out:
- Pass some Postgres complex type and get some sort of array of `Datum` in Rust?

PR #844, opened 2022-11-08
Add `record` data type

levkk:
### The What

Add support for defining functions with data type `record` as an argument or return value.

```postgresql
CREATE FUNCTION my_function(arg record) RETURNS void
AS $$
BEGIN
END
$$ LANGUAGE 'plpgsql';
```

### The Why

I found myself needing to pass arbitrary rows from Postgres into my extension. Parsing those rows isn't very hard because we already have `PgHeapTuple::from_composite_datum()`, of which `record` is the [anonymous](https://www.postgresql.org/docs/current/rowtypes.html) kind.

### The How

This just creates a function argument that's validated by Postgres before being called so I don't have to use `AnyElement` and check for the oid manually myself. You can probably tell this is a copy/pasta of `anyelement.rs` with just the oid and the name changed.

### More info

#825


eeeebbbbrrrr:
As written, this isn't going to work.

The problem is that the backing `Datum` is effectively borrowed from Postgres -- RECORD Datums are pass-by-reference.

`Record` has no idea where the the thing the Datum points to is allocated or how long it'll live.  Using `Record` implies there's some Rust ownership, or at least Rust management of the backing pointer.

I think what we ought to do to support the RECORD type is simply map that sql type name over to pgx' existing `PgCompositeType`, which (barring any bugs), properly maintains ownership of the pointed-to-data, namely the backing `HeapTupleData`.

If the original type oid is record, then'll still need `struct RECORD`,  but we'd probably want it to Deref and DeRefMut over PgCompositeType, so it essentially looks like the same thing to the user.

eeeebbbbrrrr:
As a followup, this isn't much different than the existing `AnyElement` type pgx already has.  So this could be implemented as

```rust
#[repr(transparent)]
pub struct Record(AnyElement);
```

plus all the sql type mapping stuff.  

But the same problem exists... no idea how long the pointed-to-memory is going to live and no real ability to actually use the record

levkk:
Hold on a second, I'm lost.

I think ownership comes in two flavors (I'm about to just repeat what you said to make sure I understand it): owned by Postgres & owned by Rust.

If `record` comes as an argument to a function, it's owned by Postgres because it was created by Postgres. In that case (happy coincidence, that's my use case), it'll work because Postgres will dealloc it at the end right?

> Using Record implies there's some Rust ownership, or at least Rust management of the backing pointer.

The pointer sure, but as long as it's used inside a `#[pg_extern]`, it's assumed to exist only for the duration of that function call. But you're right, it's not a guarantee, so `Record` is `unsafe` (much like `AnyElement` is now). I think I get it now a little bit. Github comment textarea is my rubber duck.

If `record` is returned as a result of a function, Rust created it and Postgres needs to dealloc it, so that's a case that doesn't work currently (all my sentences finishing with periods are actually questions, but putting a question mark everywhere is silly...).

TLDR (putting TLDR at the end of a  long post lol...): `AnyElement` is not a safe struct, and `Record` isn't in this case either, and that's fine for me because I read the warning, but to make it more usable, we need to establish ownership of the data underneath.

workingjubilee:
Yeah, to chime in a bit: I feel this doesn't offer enough on top of AnyElement to be worth the API expansion for a new type. We probably don't want to offer type translations if the translation is of low quality. Yes, AnyElement and AnyArray exist, but PGX needed an answer for handling polymorphic data, since their bottom case already exists in the Postgres type system, and their existence without a better interface already causes us problems, the way I see it. At least they are clearly marked as polymorphic.

For this, it's not obvious that you need to handle this the same way without Postgres/SQL/PGX domain knowledge on how to use the "record" type, and in that case, it's better to ask users who know what they are doing to handle things in an `unsafe` and ad-hoc manner rather than canonizing a bad interface. I would rather all polymorphic types offered solutions that work with or at least feel more like Rust's existing "dynamic dispatch" mechanisms.

I don't know if offering Deref and DerefMut to PgHeapTuple would actually be a good idea: maybe it would actually be the other way around? But definitely I would start from "it should be approximately as good an interface as PgHeapTuple is".

workingjubilee:
And I... think the code you linked is acceptable, because the `PgHeapTuple` manages things? But I won't swear on that.

Anyways, closing this since this PR isn't likely to be accepted in its current form, sorry!

PR #845, opened 2022-11-08
panic! at the polymorphism failure

Smittyvb:
Always `panic!` in the `from_datum` implementation for `AnyElement`/`AnyArray` instead of only panicking when debug assertions are enabled.

These functions should never be called anyways, so there's no real performance benefit by only panicking in debug mode.

### DIFF
--- a/pgx/src/datum/anyarray.rs
+++ b/pgx/src/datum/anyarray.rs
@@ -38,8 +38,7 @@ impl FromDatum for AnyArray {
 
     #[inline]
     unsafe fn from_datum(_datum: pg_sys::Datum, _is_null: bool) -> Option<AnyArray> {
-        debug_assert!(false, "Can't create a polymorphic type using from_datum, call FromDatum::from_polymorphic_datum instead");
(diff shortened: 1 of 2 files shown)


workingjubilee:
I'm surprised all our tests pass.

Can you update the documentation appropriately so people know they should **never** call these?

Smittyvb:
I added some documentation for `Any{Element,Array}`. The documentation for them is very similar, but I figured it wasn't worth it to split the documentation into a separate file to be reused by both.

In the future it might be worth considering using a macro to generate the `any*` modules since they're almost identical, and it would allow us to support the [other polymorphic types](https://www.postgresql.org/docs/current/extend-type-system.html#EXTEND-TYPES-POLYMORPHIC) easily.

workingjubilee:
Thank you! Yeah, I think this is good then.

workingjubilee (approved):
It's true that we should basically never be using `from_datum` for this and should be using `from_polymorphic_datum` always.

PR #847, opened 2022-11-08
Fix typo: indicies -> indices

jteplitz:
NB: This is a breaking change since indices is in the public API


workingjubilee:
That's fine, our next release will be 0.6.0 anyways.

PR #851, opened 2022-11-10
Numeric support

eeeebbbbrrrr:
What else is there to say?  Proper support for Postgres' `NUMERIC` support has finally arrived.


pgx-examples/numeric/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+name = "numeric"
+version = "0.0.0"
+edition = "2021"
+rust-version = "1.58"
```
Smittyvb, line 5: I don't think this line should be here since we don't actually have a MSRV?
```suggestion
```

pgx/src/datum/numeric.rs
```diff
@@ -7,218 +7,225 @@ All rights reserved.
...
-impl From<u8> for Numeric {
-    fn from(val: u8) -> Self {
-        Numeric(val.to_string())
-    }
+/// Which way does a [`AnyNumeric`] sign face?
+pub enum Sign {
+    Negative,
+    Positive,
+    Zero,
 }
```
Smittyvb, line 104: `AnyNumeric::sign` calls the `numeric_sign` function, [which can return NaN](https://github.com/postgres/postgres/blob/36e545cd051ffe58064c1f38faeb64f4feef4aec/src/backend/utils/adt/numeric.c#L1433-L1437). That variant should be in the enum.

pgx/src/datum/numeric.rs
```diff
@@ -7,218 +7,225 @@ All rights reserved.
...
+            let sign: AnyNumeric =
+                direct_function_call(pg_sys::numeric_sign, vec![self.as_datum()]).unwrap();
+            let zero: AnyNumeric = 0.try_into().unwrap();
+            if sign < zero {
+                return Sign::Negative;
+            } else if sign > zero {
+                return Sign::Positive;
+            } else {
+                return Sign::Zero;
+            }
```
Smittyvb, line 119: I think we can just do something like this (with NaN handling added) and remove the need for a call into Postgres?
```suggestion
            let zero: AnyNumeric = 0.try_into().unwrap();
            if self < zero {
                return Sign::Negative;
            } else if self > zero {
                return Sign::Positive;
            } else {
                return Sign::Zero;
            }
```

pgx/src/datum/numeric_support/cmp.rs
```diff
@@ -0,0 +1,126 @@
...
+    #[inline]
+    fn ge(&self, other: &Self) -> bool {
+        unsafe {
+            direct_function_call(pg_sys::numeric_ge, vec![self.as_datum(), other.as_datum()])
+                .unwrap()
+        }
+    }
+}
+
+impl Ord for AnyNumeric {
```
Smittyvb, line 63: Postgres `numeric`s can be NaN so I don't think should implement `Ord`?

```
user=> select 'nan'::numeric < 'nan'::numeric;
 ?column?
----------
 f
(1 row)

user=> select 'nan'::numeric > 'nan'::numeric;
 ?column?
----------
 f
(1 row)
```

pgx/src/datum/numeric_support/cmp.rs
```diff
@@ -0,0 +1,126 @@
...
+                .unwrap()
+        }
+    }
+}
+
+impl Eq for AnyNumeric {}
+
+impl PartialOrd for AnyNumeric {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
```
Smittyvb, line 27: This should handle the case of one value being NaN

pgx/src/datum/numeric_support/cmp.rs
```diff
@@ -0,0 +1,126 @@
...
+    #[inline]
+    fn ne(&self, other: &AnyNumeric) -> bool {
+        unsafe {
+            direct_function_call(pg_sys::numeric_ne, vec![self.as_datum(), other.as_datum()])
+                .unwrap()
+        }
+    }
+}
+
+impl Eq for AnyNumeric {}
```
Smittyvb, line 23: This should document that NaN compares equal to NaN (unlike floats), which makes this implementation valid.

pgx-pg-sys/src/submodules/panic.rs
```diff
@@ -96,6 +96,11 @@ impl ErrorReportWithLevel {
         }
     }
 
+    /// Returns the error message of this error report
+    pub fn message(&self) -> &str {
+        &self.inner.message()
```
thomcc, line 101: nit: doesn't seem to need `&` (the other `message()` fn already returns `&str`)

pgx/src/datum/numeric.rs
```diff
@@ -7,218 +7,225 @@ All rights reserved.
...
-impl From<i8> for Numeric {
-    fn from(val: i8) -> Self {
-        Numeric(val.to_string())
+impl Display for AnyNumeric {
+    fn fmt(&self, fmt: &mut Formatter<'_>) -> fmt::Result {
+        let numeric_out = unsafe {
+            direct_function_call::<&CStr>(pg_sys::numeric_out, vec![self.as_datum()]).unwrap()
+        };
+        let s = numeric_out.to_str().expect("numeric_out is not a valid UTF8 string");
+        fmt.write_str(s)
```
thomcc, line 74: nit: It's minor but `s.fmt(fmt)` (or equivalently `fmt.pad(s)`) is better if you have a string around already, since then the size/alignment stuff gets respected.

pgx/src/datum/numeric.rs
```diff
@@ -7,218 +7,225 @@ All rights reserved.
...
+        (p, s) => ((p << 16) | s) + pg_sys::VARHDRSZ as i32,
     }
 }
 
-impl From<u8> for Numeric {
-    fn from(val: u8) -> Self {
-        Numeric(val.to_string())
-    }
+/// Which way does a [`AnyNumeric`] sign face?
+pub enum Sign {
```
thomcc, line 100: Nit: probably some derives?

pgx/src/datum/numeric.rs
```diff
@@ -7,218 +7,225 @@ All rights reserved.
...
-}
 
-impl IntoDatum for Numeric {
-    fn into_datum(self) -> Option<pg_sys::Datum> {
-        let cstring =
-            std::ffi::CString::new(self.0).expect("failed to convert numeric string into CString");
-        let cstr = cstring.as_c_str();
+    /// Calculate the greatest common divisor of this an another [`AnyNumeric`]
+    #[cfg(not(any(feature = "pg11", feature = "pg12")))]
+    pub fn gcd<const P: u32, const S: u32>(&self, n: &AnyNumeric) -> AnyNumeric {
```
thomcc, line 158: I think the `P` and `S` params here are unnecessary.

pgx/src/datum/numeric_support/convert_anynumeric.rs
```diff
@@ -0,0 +1,121 @@
...
+macro_rules! anynumeric_from_oversized_primitive {
+    ($ty:ty, $signed:ty) => {
+        impl From<$ty> for AnyNumeric {
+            #[inline]
+            fn from(value: $ty) -> Self {
+                match <$signed>::try_from(value) {
+                    Ok(value) => AnyNumeric::from(value),
+                    Err(_) => unsafe {
+                        // SAFETY:  conversion from value.to_string() will always work
+                        AnyNumeric::try_from(value.to_string().as_str()).unwrap_unchecked()
```
thomcc, line 39: Unsure the `unwrap_unchecked()` here is worth the risk, given that savings are quite small compered to the cost of string conversion. But it's your call.

thomcc (approved):
Overall looks great, just a few minor nits.

pgx/src/datum/numeric.rs
eeeebbbbrrrr: I'm not sure.  `numeric_sign()` has some handling for -/+Infinity too.  

I'll definitely fix this to account for NaN.

pgx/src/datum/numeric.rs
eeeebbbbrrrr: Good catch.  Looks like that was leftover from a previous iteration I was working on.

pgx-examples/numeric/Cargo.toml
eeeebbbbrrrr: Was copy-paste from another example.  I'll look at cleaning them all up in a different PR>

pgx/src/datum/numeric_support/convert_anynumeric.rs
eeeebbbbrrrr: I suppose that's fair.

pgx/src/datum/numeric_support/cmp.rs
eeeebbbbrrrr: Postgres provides total ordering for NUMERIC, so we can blindly accept what `pg_sys::numeric:cmp()` tells us.  I've cleaned it up in eeffdfb9f3c294914ed58608c181fae096d81805.

eeeebbbbrrrr:
For for the reviews, @thomcc and @Smittyvb.

@thomcc and I talked offline and Postgres does total ordering for NUMERIC, so there's no need for us to do anything above and beyond what `pg_sys::numeric_cmp()` tells us.  Which means we can impl Ord and Eq and not need to special case NaN.

NaN is interesting as `NaN < NaN == false`, `NaN > NaN == false`, but `NaN = NaN == true`.  Lets just let Postgres figure that out for us!  If it returned NULL for any of those, then I could see how we couldn't impl Ord and Eq, but it doesn't.

PR #852, opened 2022-11-10
Reduce some overhead during the various "direct_function_call" functions

eeeebbbbrrrr:
Roughly a 10%-12% improvement on general overhead.  Did some quick evaluations with a `SELECT func_that_does_direct_function_call() FROM generate_series(1, 1000000);`.  Seems pretty consistent across all supported Postgres versions.

Postgres itself is able to do some C trickery to allocate the `pg_sys::FunctionCallInfo(Base)Data` on the stack, but I don't see how to do that with Rust -- so we have to allocate it on the heap via `pg_sys::palloc()`.

I spotted this as an issue while profiling the new numeric support.


thomcc (approved):
Looks good to me. I'm assuming the C trickery is alloca or something -- we can't do really. We also already allocate the args vec on the heap anyway, so...

PR #853, opened 2022-11-11
Fixes how pgx-version-updater is ran via update-versions.sh

BradyBonnette:


workingjubilee (approved):
Less code! Generally an improvement.

We can figure out how to cache as an optimization if it ever becomes important.

PR #855, opened 2022-11-11
Drop pg 10 from Nix

steve-chavez:
Using the flake template introduced on https://github.com/tcdi/pgx/pull/148. I get the following error:

```
$ nix develop
error: postgresql_10 has been removed from nixpkgs, as this version went EOL on 2022-11-10
```

With this change the error goes away.

```
$ nix develop
$ cargo pgx -V
cargo-pgx 0.5.6
```

I think this goes in line with https://github.com/tcdi/pgx/pull/830.

(Also removed some Nix deps from `cargo-pgx` that look unused)

### DIFF
--- a/cargo-pgx/default.nix
+++ b/cargo-pgx/default.nix
@@ -1,13 +1,8 @@
 { lib
 , naersk
 , hostPlatform
-, fetchFromGitHub
-, postgresql_11
-, postgresql_12
-, postgresql_13
 , pkg-config
 , openssl
-, rustfmt
 , libiconv
 , llvmPackages
 , gitignoreSource
--- a/cargo-pgx/default.nix
+++ b/cargo-pgx/default.nix
@@ -2,7 +2,6 @@
 , naersk
 , hostPlatform
 , fetchFromGitHub
-, postgresql_10
 , postgresql_11
 , postgresql_12
 , postgresql_13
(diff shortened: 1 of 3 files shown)


workingjubilee (approved):
Thank you!

cargo-pgx/default.nix
```diff
@@ -1,14 +1,8 @@
...
 , naersk
 , hostPlatform
-, fetchFromGitHub
-, postgresql_10
-, postgresql_11
-, postgresql_12
-, postgresql_13
 , pkg-config
 , openssl
-, rustfmt
```
workingjubilee, line 11: Actually, wait. I know they look unused but can you reduce this to the most conservative change?

cargo-pgx/default.nix
```diff
@@ -1,14 +1,8 @@
 { lib
 , naersk
 , hostPlatform
-, fetchFromGitHub
```
workingjubilee, line 4: This one is actually unused, yes, it can go.

cargo-pgx/default.nix
steve-chavez: Sure, should I just remove 0f0a1618e2dca0f4b9c3f4057da841c6f4df3358 from this PR or add another commit that only removes `fetchFromGitHub`?

workingjubilee: Pushing another commit is fine.

workingjubilee: Apologies, I am on a timer.

PR #861, opened 2022-11-13
Provide an example of PostgresType derivation on enums

yrashk:
It is not clear if PaostgresType can be derived on enums


workingjubilee:
I enjoy the verbose messaging but please frame PR messages in terms of

imperative (so, your solution, usually)
problem
how imperative resolves

Leading with the imperative makes it easier to parse out the PR names and git messages later in the log in terms of what each does (which I do as part of release notes).

yrashk:
@workingjubilee will do!

workingjubilee:
Thank you!

PR #862, opened 2022-11-14
If pgx detects a Postgres FFI was called on a not-the-main-thread...

eeeebbbbrrrr:
... include the caller's location information in that panic message.

This necessitates adding `#[track_caller]` to every FFI function we wrap.

FWIW, this helped me determine where ZDB was unknowingly calling FFI in a background thread.  Without it, I'm not sure I'd ever have found out what was going on.


thomcc (approved):
LGTM

PR #863, opened 2022-11-14
Problem: BackgroundWorker::transaction `R: Copy` bound

yrashk:
It's not clear why it is necessary and is fairly restrictive.

Solution: drop the bound

The tests are still passing. No other changes were required.

### DIFF
--- a/pgx/src/bgworkers.rs
+++ b/pgx/src/bgworkers.rs
@@ -192,10 +192,7 @@ impl BackgroundWorker {
 
     /// Once connected to SPI via `connect_worker_to_spi()`, begin a transaction to
     /// use the `pgx::Spi` interface. Returns the return value of the `F` function.
-    pub fn transaction<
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr:
Best I can figure I did that as a suggestion from @thomcc  in 2ed92164222b51a4d00662570fa1dcae34b6b2d4 on 11/3 (days before PR was merged).

I don't have any specific memory as to why...

eeeebbbbrrrr:
it had to have come from here:  https://github.com/tcdi/pgx/pull/834#discussion_r1013314774

but that doesn't mention R needing to be Copy.  I wonder if CLion did something automatically?  I'm stumped.

thomcc:
I did suggest a Copy bound elsewhere, but not here. For this, It's not needed.

thomcc (approved):

PR #865, opened 2022-11-15
Use `CaughtError` in `PgTryBuilder` docs

yrashk:
There's no `CatchError` type; use `CaughtError` instead


workingjubilee (approved):
Thank you!

PR #866, opened 2022-11-15
Avoid adding the error (and thread safety) wrapper for SetLatch 

thomcc:
I don't know if this is the route we want to go, but it's technically not very difficult if this is actually safe.

### DIFF
--- a/pgx-pg-sys/build.rs
+++ b/pgx-pg-sys/build.rs
@@ -814,6 +814,11 @@ fn run_command(mut command: &mut Command, version: &str) -> eyre::Result<Output>
     Ok(rc)
 }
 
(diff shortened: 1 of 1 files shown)


thomcc:
I think I'd prefer this route to a global "disable all thread safety checking" flag (at least for this case[^1]). But ideally only if it's actually safe to do this, e.g. if postgres actually can not error, and this function is actually .

[^1]: I'd be willing to add the global flag if I got it the thread safety check wrong in a big way, like if somehow the atfork handler didn't work right on some weird BSD or something.

eeeebbbbrrrr:
We don't want to go down this road.  As we talked about on discord, it's just not inside pgx' world to try and guess the thread safetyness of individual functions.

thomcc:
Yep, I agree.

PR #867, opened 2022-11-15
Make us work under 1.61

thomcc:
I'd prefer we didn't keep `core_cstr` in 0.6.0, since removing it is a breaking change, but...

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -631,6 +631,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "cstr_core"
+version = "0.2.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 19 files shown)


yrashk:
I was recently just switching to `cstr_core` (as I needed the `cstr!` macro), but I am wondering what's the rationale for going back on this change, any particular reason for supporting 1.61?

I guess a deeper question here is "when are we going to allow ourselves to use GAT from 1.6.5"?

eeeebbbbrrrr:
And just continue to use Rayon for now?  I lost track of the Revert Revert Revert Revert path...!!!!

eeeebbbbrrrr:
yeah, my comment:  https://github.com/tcdi/pgx/pull/868#issuecomment-1315691082

I'll approve and merge this

eeeebbbbrrrr (approved):

PR #868, opened 2022-11-15
Document `pg_extern(name)`

yrashk:
This helps with the discovery of this option.

### DIFF
--- a/pgx-macros/src/lib.rs
+++ b/pgx-macros/src/lib.rs
@@ -437,6 +437,7 @@ Optionally accepts the following attributes:
 * `parallel_restricted`: Corresponds to [`PARALLEL RESTRICTED`](https://www.postgresql.org/docs/current/sql-createfunction.html).
 * `no_guard`: Do not use `#[pg_guard]` with the function.
 * `sql`: Same arguments as [`#[pgx(sql = ..)]`](macro@pgx).
+* `name`: Specifies target function name. Defaults to Rust function name.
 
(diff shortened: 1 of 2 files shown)


thomcc (approved):
This looks good to me, but I'll just punt to @workingjubilee to ensure there isn't a reason this was undocumented (although that seems unlikely).

thomcc:
Actually, it's pretty annoying to do this conditionally, since we emit uses of core::ffi::CStr/core_cstr::CStr from macros. I'd probably rather just stick with using core_cstr for another version than try to handle that conditionally (since we can't `cfg` it on that).

thomcc:
It is still possible though, but it would definitely increase the complexity more than just accepting the dependency. So it's your call.

thomcc:
TBH, I don't mind the core_cstr dep that much -- of these, the rayon one is the one I care about more. It's just kinda jank when we've moved CStr into libcore on its own.

eeeebbbbrrrr:
> It is still possible though, but it would definitely increase the complexity more than just accepting the dependency. So it's your call.

fucking cstring.  Okay, lets leave it as you have it here.

thomcc:
Oh god, all those comments were meant for another PR!

eeeebbbbrrrr:
whee... we're in the wrong PR!

workingjubilee (approved):
Yeah I don't think there was a reason for this to be undocumented.

PR #873, opened 2022-11-16
Add a Rust version check to cargo-pgx

### LINKED ISSUE #774: `cargo-pgx` should verify compiler parity requirement
pgx requires that the compiler used to compile `cargo-pgx` is the same compiler used to compile extension crates (https://github.com/tcdi/pgx/pull/687#discussion_r970745123), but doesn't verify that requirement in `cargo-pgx`. It should.

This requirement is not made obvious to users and results in user confusion since a compiler mismatch can result in weird segfaults and errors.

thomcc:
I had done part of this a while ago, and it came up recently, so I cleaned it off.

Fixes #774, although in the future I would like to relax this requirement and remove this check.

It does not detect mismatches between 1.x.0 and 1.x.1 (which should be fine), or between different betas/nightlies (which is probably strictly required, but would be personally pretty annoying for me).

It tends to ignore errors rather than failing, in case something in the output of one of the tools changes (unlikely, given how much of the ecosystem uses checks like this). There's also an escape-hatch environment variable called `PGX_IGNORE_RUST_VERSIONS` to disable the check.

I only check in the schema generation code, since that's the only code that needs to care about this.

### DIFF
--- a/cargo-pgx/README.md
+++ b/cargo-pgx/README.md
@@ -55,6 +55,7 @@ SUBCOMMANDS:
 - `PGX_BUILD_FLAGS` - If set during `cargo pgx run/test/install`, these additional flags are passed to `cargo build` while building the extension
 - `PGX_BUILD_VERBOSE` - Set to true to enable verbose "build.rs" output -- useful for debugging build issues
 - `HTTPS_PROXY` - If set during `cargo pgx init`, it will download the Postgres sources using these proxy settings. For more details refer to the [env_proxy crate documentation](https://docs.rs/env_proxy/*/env_proxy/fn.for_url.html).
+- `PGX_IGNORE_RUST_VERSIONS` - Set to true to disable the `rustc` version check we have when performing schema generation (schema generation requires the same version of `rustc` be used to build `cargo-pgx` as the crate in question).
 
 ## First Time Initialization
 
--- a/cargo-pgx/README.md
+++ b/cargo-pgx/README.md
@@ -803,3 +803,30 @@ Postgres, so loading two versions of the shared library will cause trouble.
 These scenarios are:
 - when using shared memory
 - when using query planner hooks
+
+## Compiler Version Dependence
+
+The version of the Rust compiler and toolchain used to build `cargo-pgx` must be
+the same as the version used to build your extension.
+
+Several subcommands (including `cargo pgx schema`, `cargo pgx test`, `cargo pgx
+install`, ...) will produce an error message if these do not match.
+
+Although this may be relaxed in the future, currently schema generation involves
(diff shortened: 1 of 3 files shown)


cargo-pgx/src/command/schema.rs
```diff
@@ -136,6 +136,53 @@ impl CommandExecute for Schema {
...
+///
+/// To waive this, you may set `PGX_IGNORE_RUST_VERSIONS` in the environment (to
+/// any value other than `"0"`). Also, note that this check is best-effort only,
+/// and is expected to error only if there is a definite mismatch.
+///
+/// It also cannot detect versions of `cargo-pgx` and `pgx` differing, which
+/// could cause similar issues (in the future this may be detected).
+fn check_rust_version() -> eyre::Result<()> {
+    const COMPILE_TIME_MINOR_VERSION: Option<&str> = option_env!("MINOR_RUST_VERSION");
+    if matches!(std::env::var("PGX_IGNORE_RUST_VERSIONS"), Ok(s) if s != "0") {
```
eeeebbbbrrrr, line 171: can yo document this new envar in cargo-pgx/README.md?  I suspect it's only for you, but it'd be nice to have a record of why it exists.

eeeebbbbrrrr (commented):
Perhaps a little note in cargo-pgx/README.md about this general change is also necessary.  When an issue gets reported about this (and it will), I'd like to be able to point that person to our docs that says, "see, this check is on purpose".

cargo-pgx/src/command/schema.rs
thomcc: Sure.

thomcc: Done.

eeeebbbbrrrr (approved):
Thanks for the README words.  Perfect!

cargo-pgx/README.md
```diff
@@ -802,3 +803,30 @@ Postgres, so loading two versions of the shared library will cause trouble.
...
+Several subcommands (including `cargo pgx schema`, `cargo pgx test`, `cargo pgx
+install`, ...) will produce an error message if these do not match.
+
+Although this may be relaxed in the future, currently schema generation involves
+`dlopen`ing the extension and calling `extern "Rust"` functions on
+`#[repr(Rust)]` types. Generally, the appropriate way to fix this is reinstall
+`cargo-pgx`, using a command like the following
+
+```shell script
+$ cargo install --locked cargo-pgx
```
Smittyvb, line 821: Cargo will only reinstall an already installed binary with the same version if `--force` is passed.
```suggestion
$ cargo install --force --locked cargo-pgx
```

thomcc: Oh, nice catch!

PR #875, opened 2022-11-16
It's not necessary for pgx to add a CONTEXT message to the ereport

eeeebbbbrrrr:
pgx doesn't (currently) have any more context than `funcname/file:line:col info`, which we already store in the LOCATION slot.  If in the future pgx learns how to build a backtrace (and it might be interesting for it to do that for level >=ERROR messages), the context slot is where that information should go.

@yrashk brought this up on Discord.


thomcc (approved):
LGTM!

PR #878, opened 2022-11-18
Use dlmopen to load the extension on linux

thomcc:
Hopes to address https://github.com/tcdi/plrust/issues/119, but I haven't checked yet. Pretty hairy, and needs more testing than it has gotten

As I've mentioned, I'd really like us to use a different method than `dlm?open` for this (we should put info we need for schema generation in special link-sections, and then grab them using `object`, so we never need to actually load the extension), but that's a way larger project.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -283,6 +283,7 @@ dependencies = [
  "env_proxy",
  "eyre",
  "fork",
+ "libc",
  "libloading",
  "num_cpus",
  "object 0.28.4",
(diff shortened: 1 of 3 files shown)


thomcc:
FWIW even if this fixes the issue (and it might not, although I'm not sure what the actual fix would be if it doesn't...), it's quite hairy and may need some time to bake.

So I'd probably like us to give it a little more time rather than crash-landing it into the 0.6.0 release.

thomcc:
Turns out this is unrelated, and wouldn't fix the issue anyway.

PR #880, opened 2022-11-18
No needless palloc in `Array<'a, T>`

workingjubilee:


workingjubilee:
Sigh. Looks like nothing will do except uplifting the `deconstruct_array` logic wholesale to guarantee the array is actually walked correctly.

workingjubilee:
Superceded by #1116

PR #882, opened 2022-11-22
Introduce `pgx::slice::PallocSlice`

workingjubilee:
This is an abstraction designed to simplify handling of pallocs when those pallocs are also slices. Its semantics differ from `PgBox<T>` a little bit on a few points so it isn't the same type.

This allows cleaning up `datum::Array` a lot internally, minimizing the potential for UB, and a lot less futzing about in `Drop`, even though it's still not perfect. When https://github.com/tcdi/pgx/pull/881 lands there's some things that can be done with these two PRs in combination, so as to clean stuff up further.


pgx/src/slice.rs
```diff
@@ -0,0 +1,43 @@
...
+        }
+    }
+
+    /// # Safety
+    /// You must know the underlying type at that index is validly initialized in Rust.
+    #[inline]
+    pub unsafe fn get(&self, index: usize) -> Option<&T> {
+        index
+            .le(&self.pallocd.len())
+            .then(|| self.pallocd.as_ptr().cast::<T>().add(index).as_ref().unwrap_unchecked())
```
eeeebbbbrrrr, line 27: just call into unwrapped unchecked here: `.then(self.unwrap_unchecked)`.  It's not necessarily complicated pointer math, but no need to duplicate it

pgx/src/datum/array.rs
```diff
@@ -99,22 +99,13 @@ impl<'a, T: FromDatum + serde::Serialize> serde::Serialize for Array<'a, T> {
...
-
-            // Detoasting the varlena may have allocated: the toasted varlena cloned as a detoasted ArrayType
-            // Checking for pointer equivalence is the only way we can truly tell
-            let raw = raw.take().map(|r| r.into_ptr());
-            if let Some(raw) = raw {
+        // First drop the slice that references the RawArray
+        let slice = mem::take(&mut self.datum_slice);
+        mem::drop(slice);
+        if self.needs_pfree {
+            if let Some(raw) = self.raw.take().map(|r| r.into_ptr()) {
```
eeeebbbbrrrr, line 106: What's the point of `self.needs_pfree` then?

pgx/src/slice.rs
workingjubilee: Oh, you mean `get_unchecked`? Yes, good idea.

pgx/src/datum/array.rs
workingjubilee: It indicates the "was cloned on detoasting" case, the `RawArray` resides in an `Option<_>` to allow `mem::take` on it (basically, Advanced Rust Antics), this `if let` will always pass. I have been thinking maybe I should use something more like the pattern for `Cow` here, but I want to take another shot, later, at removing the need for `Option` first before embracing this as code that will remain here.

workingjubilee:
@thomcc I added a lifetime out of an abundance of caution at first but on reviewing my code I don't think it actually needs one since it's technically owning, right?

thomcc:
It seems to be conceptually owned, so removing lifetime should be fine (at least, as fine as PgBox -- IIUC technically this is borrowed from the current memory context but there's no great way for us to express this at the moment).

>  I don't think anything breaks with `PhantomData<T>`...

I'd use `PhantomData<Box<[T]>>` or something like that, so that you don't inherit `!Unpin` from `T` (which would happen for `PhantomData<T>`), but it doesn't really matter for our use.

workingjubilee:
Good catch.

Yeah, in the current usage context for Array, it's bound to not "escape", so it's fine even if it's technically "naughty". I'd like for us to be able to do allocations using something like `Heap<'current>` but that's a ways of work off.

PR #883, opened 2022-11-22
Reference-counted SpiClient

yrashk:
Problem: Spi::connect is rather limiting

It restricts the signature of the return type. Furthermore, it makes it much more difficult to re-use an existing connection safely as the only way to do so is to create an instance of `SpiClient` manually. However, once we're out of the closure, the connection will be dropped, and code that relies on that other instance of `SpiClient` will fail.

Solution: introduce reference-counted client instances

Without breaking _documented_ API, this change adds `Spi::client()` function that returns a reference-counted instance of `SpiClient` that will finalize the connection when there are no more clones of it left, making it much easier to have greater control over the control flow.

It _does_ break the API because one can no longer instantiate `SpiClient` themselves (without using `Spi` API surface). However, this is an error-prone technique that should rather be avoided.

### DIFF
--- a/pgx-examples/spi/src/lib.rs
+++ b/pgx-examples/spi/src/lib.rs
@@ -42,16 +42,11 @@ fn spi_return_query(
     #[cfg(feature = "pg15")]
     let query = "SELECT oid, relname::text || '-pg15' FROM pg_class";
 
-    let mut results = Vec::new();
-    Spi::connect(|client| {
-        client
-            .select(query, None, None)
-            .map(|row| (row["oid"].value(), row[2].value()))
-            .for_each(|tuple| results.push(tuple));
-        Ok(Some(()))
-    });
+    let client = Spi::client().unwrap();
 
-    TableIterator::new(results.into_iter())
+    TableIterator::new(
+        client.select(query, None, None).map(|row| (row["oid"].value(), row[2].value())),
+    )
 }
 
 #[pg_extern(immutable, parallel_safe)]
--- a/pgx-tests/src/tests/spi_tests.rs
+++ b/pgx-tests/src/tests/spi_tests.rs
@@ -14,6 +14,7 @@ mod tests {
     use crate as pgx_tests;
 
     use pgx::prelude::*;
+    use pgx::SpiSingletonClient;
 
     #[pg_test(error = "syntax error at or near \"THIS\"")]
     fn test_spi_failure() {
@@ -219,7 +220,6 @@ mod tests {
 
     #[pg_test]
     fn test_client() {
-        use pgx::SpiError;
         let client = Spi::client().unwrap();
         assert!(matches!(client.select("SELECT 1", None, None).first().get_one::<i32>(), Some(1)));
         let client1 = client.clone();
@@ -227,5 +227,24 @@ mod tests {
 
         // Can create another client
         assert!(Spi::client().is_ok());
+    }
+
+    #[pg_test]
+    fn test_singleton_client() {
+        let client = Spi::client().unwrap();
+        let client_clone = client.clone();
+        let singleton_attempt: Result<SpiSingletonClient, _> = client.try_into();
+
+        match singleton_attempt {
+            Ok(_) => {
+                // client_clone is still alive
+                assert!(false, "Singleton should not be created");
+            }
+            Err(client) => {
+                drop(client_clone);
+                let mut singleton: SpiSingletonClient = client.try_into().unwrap();
+                let _ = singleton.update("SELECT 1", None, None);
+            }
+        }
     }
 }
--- a/pgx-tests/src/tests/spi_tests.rs
+++ b/pgx-tests/src/tests/spi_tests.rs
@@ -216,4 +216,16 @@ mod tests {
         struct T;
         assert!(matches!(Spi::connect(|_| Ok(Some(T))).unwrap(), T));
     }
+
+    #[pg_test]
+    fn test_client() {
+        use pgx::SpiError;
+        let client = Spi::client().unwrap();
+        assert!(matches!(client.select("SELECT 1", None, None).first().get_one::<i32>(), Some(1)));
+        let client1 = client.clone();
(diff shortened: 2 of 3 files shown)


workingjubilee:
As we're in an alpha release, we're definitely game to accept breaking changes atm anyways.

eeeebbbbrrrr:
I'm not convinced this is a good idea.  The primary reason is that `pg_sys::SPI_connect()` puts the caller into a MemoryContext that is free'd during `pg_sys::SPI_finish()`.

While `pg_sys::SPI_connect()` is documented as being re-entrant, having multiple of these "open" at the same time seems incredibly dangerous to surrounding Rust code since it's no longer clear where the boundaries are.

I haven't thought through how to solve the problems this PR wants to address, but I don't believe the answer is a reference-counted "connection" object.

Also of note, this only works in the face of a `pgx::error!()` due to the recent cleanups to error handling.

yrashk:
To note, I don't suggest having multiple open connections. This reference-counted "connection" is just one, and we clone a token value.

eeeebbbbrrrr:
> To note, I don't suggest having multiple open connections. This reference-counted "connection" is just one, and we clone a token value.

That's just as surprising, just in the opposite way.

While the user has a an open SPI connection, they're inside a special MemoryContext.  This is Postgres' design, not ours, and we have to work with that, carefully, to make sure that surrounding code doesn't go UB.  This PR doesn't help with that, at all.  

In the original comment you said:

> ... However, once we're out of the closure, the connection will be dropped, and code that relies on that other instance of SpiClient will fail.

That's on purpose.  It might be inconvenient, but doing things "right" almost always is.  Maybe if you can expand a bit on why it's inconvenient we can come up with a better solution.  Working with SPI from C is incredibly inconvenient, and that makes me super motivated to make it better with pgx/rust, but we do have some constraints.

yrashk:
On the last case. There are a few things:

* The "inconvenient" part. It's minor, but `connect` restricts the return type signature to `Option<R>` for reasons that unclear
* The "wrong" part. `SpiClient` can be constructed manually at any time, and used just as successfully while the connection is open. Once it's closed, it'll be erroneous. The easiest way to get rid of this is to make it non-constructible for the users of the library, but there's code out there that relies on this behavior to get `SpiClient` when it's consumed.

I based this PR on a brief discussion with @workingjubilee where a suggestion of something like a reference-counted Client was floated.

As a part of the broader picture, one of the reasons I prepared this PR is a part of a larger conversation about sub-transaction API (https://github.com/supabase/pgx-contrib-spiext/blob/main/src/subtxn.rs) that I wanted to contribute but no longer sure I should be pursuing this if you're intent on figuring these parts internally.

yrashk:
Also to note, I understand the desire to encode these difficult effects (change of memory context) into the API in some shape or form and am happy to discuss it further, unless you're set on how you want this to be and there's nothing to discuss.

workingjubilee:
> I'm not convinced this is a good idea. The primary reason is that `pg_sys::SPI_connect()` puts the caller into a MemoryContext that is free'd during `pg_sys::SPI_finish()`.
> 
> While `pg_sys::SPI_connect()` is documented as being re-entrant, having multiple of these "open" at the same time seems incredibly dangerous to surrounding Rust code since it's no longer clear where the boundaries are.
> 

I mean, currently someone can just make as many connections as they want, with no assistance in Rust from that. :upside_down_face:

eeeebbbbrrrr:
> I mean, currently someone can just make as many connections as they want, with no assistance in Rust from that.

If they call `pg_sys::SPI_connect()` directly, yes.  

But if they call `pgx::spi::Spi::connect(||)`, which is what we'd prefer they do, then it's just one "outer" connection at a time.

I guess the fact that `pgx::spi::SpiClient` can be constructed out of thin air is a design/implementation flaw.  When I wrote the Spi wrapper a few years ago, that wasn't my intention.  I feel like that happened accidentally because I didn't understand Rust as well as I do know.  Looking at that now, I'm quite surprised.  😱 

So with that, it sounds like the other complaint is the shape of `Spi::connect()`'s return type.  We can rejigger to be `T`.  There was some problems awhile back around the memory context in which a returned value was allocated in, but we solved that in #799, I believe.

My vision for "being connected to SPI" is this:

```rust
let v = Spi::connect(|client| { /* **only** way to get a SpiClient */
    // use SPI somehow
    // return a value
});
```

yrashk:
I am unsure how useful this visual delineation of `Spi::connect` is.

```rust
mod spi {
  pub struct SpiClient(());

  pub fn connect<T, F: FnOnce(SpiClient) -> T>(f: F) -> T {
      f(SpiClient(()))
  }  
}

fn main() {
    let client = spi::connect(|client| client);
}
```

Arguably, this is worse than the reference-counted client, as `client` in `main` is no longer usable as the connection has been dropped. The reference-counted one is.

yrashk:
At the very least, we can consider moving `SPI_finish` to `Drop` of `SpiClient` instead of that `Connection`. It will make the above example safer.

eeeebbbbrrrr:
We can fix this by making `client` a reference, borrowed from `SpiConnection`.  Then you wouldn't be able to return it.  That's a good catch.

My goal here is that one ought not be able to do anything with Spi unless they're in the `Spi::connect(|| ...)` closure.  Static methods on Spi, like Spi::get_one(), notwithstanding, of course.  That's the only way we can ensure things happen in a predictable manner as it relates to memory contexts and general connected-to-spi-ness.

yrashk:
@eeeebbbbrrrr I've implemented your suggestion in #896

workingjubilee:
Marking this as draft since we probably do want some improvement here but this needs to both address the merge conflicts that #896 introduced and also have its design revisited until we can establish something with more of a consensus.

yrashk:
We can only discuss this if there is an interest in reference-counted clients. @eeeebbbbrrrr seems to have a case that this should not be permitted. If this is the case, we should probably just close this PR.

yrashk:
Let's close this for now, there's probably not enough justification for something like this at the moment.

PR #884, opened 2022-11-22
Fallback to C locale on systems without C.UTF-8

Smittyvb:
CentOS 7 doesn't support the `C.UTF-8` locale, so cargo-pgx doesn't work on CentOS 7 since it tries to set a databases locale to `C.UTF-8`. Rocky Linux >= 8 does have the `C.UTF-8` locale and works fine.

This PR makes pgx fallback to the `C` locale on systems without the `C.UTF-8` locale (according to `locale`, and defaulting to `C` if `locale` fails to run).


workingjubilee:
Hm. Please document this fallback logic in one of our READMEs somewhere? It's getting complicated enough that users should be aware of the system differences.

Smittyvb:
I documented the locale behavior for the `cargo pgx test` and `cargo pgx start/stop` commands.

workingjubilee (commented):
Thank you. Hopefully no one will have to care, but inevitably someone is going to want to, say, port software from CentOS 7 to Rocky Linux 8.

Just one other nit.

PR #885, opened 2022-11-23
Prepared statements and Query trait

yrashk:
Problem: having to re-parse SQL queries every time

Solution: implement basic support for prepared statements

This change introduces the `Query` trait that generalizes what can be seen as a query. For practical purposes, this means strings are one-off queries (as they were used before), and `SpiClient::prepare` can be used to prepare statements.

`select`, `update`, and `open_cursor` all take `Query`, making it a uniform interface.

`get_one/two/three` do not take `Query` to keep their signature simple (as it's used for typing)


eeeebbbbrrrr:
This isn't very complex, which is great, but I think we want a different API here.  Something more along the lines of:

```rust
    let statements: &mut HashMap<String, PreparedStatement> = MY_GLOBAL_STATEMENT_CACHE;

    let stmt = statements.entry("fast_query").or_insert_with(|| {
        // Spi::prepare() automatically calls SPI_keep()
        // no reason not to assume we want it kept around when impl Drop will free it anyways
        Spi::prepare(
            "SELECT $1, $2, $3",
            Some(vec![
                PgOid::BuiltIn(PgBuiltInOids::INT4OID),
                PgOid::BuiltIn(PgBuiltInOids::TEXTOID),
                PgOid::BuiltIn(PgBuiltInOids::TEXTOID),
            ]),
        )
    });

    //
    // various things a PreparedStatement can do
    //
    
    let args = Some(vec![42i32.into_datum(), "hi".into_datum(), "world".into_datum()]);
    
    // Just execute it
    match stmt.execute(args.clone()) {
        Ok(_) => {}
        Err(spi_error) => panic!("got spi error: {}", spi_error)
    }
    
    // Get one/two/three
    let v:i32 = stmt.get_one(args.clone());
    
    // use it within SPI_connect
    let limit = Some(21);
    let result = stmt.execute_with(args.clone(), limit, |tuptable| {
        // do stuff with the tuptable
        return Ok(some_value)
    });
```

These are just some quick thoughts, but the idea is that each PreparedStatement is really its own self-contained Spi interface that only knows how to do things against its internal statement.

Also, I don't see a need for `PreparedStatement` to have a lifetime.  I don't know how we could ever give it anything but `'static`, and since it implements Drop, we ought to just treat it as an owned instance.

yrashk:
The reason why `PreparedStatement` has a lifetime is motivated by `SPI_prepare`'s documentation:

>The statement returned by SPI_prepare can be used only in the current invocation of the C function, since SPI_finish frees memory allocated for such a statement. But the statement can be saved for longer using the functions SPI_keepplan or SPI_saveplan.

Therefore, to avoid being wasteful and to mimic Postgres API as close as possible, I believe `PreparedStatement` should have a lifetime that does not exceed `SpiClient`. Unless, of course, we want to keep it, then it can be converted into an `OwnedPreparedStatement`.

yrashk:
As for your suggestion, I am not convinced what you've proposed is good.

Firstly, `SPI_prepare` must be called within `SPI_connect` / `SPI_finish` boundaries. If you follow through `SPI_prepare` -> `SPI_prepare_cursor` -> `_SPI_begin_call` you will see that it's checking if there's a current connection and if there's none (`SPI_ERROR_UNCONNECTED`), it'll not prepare a cursor.

Therefore, the `Spi::prepare` API does not account for this unless it hides the connection inside, and why would we want that?

As I mentioned above, prepared statements are (by default) scoped to the current connection, and that's why I believe they should be lifetime-bound to it. Unless they are explicitly kept.

Now, to the idea of whether a prepared statement should implement a query interface similar/matching to that of `Spi`/`SpiClient`. I can see a few holes with this:

* In the case of kept/owned prepared statements, they are not in a connection. Should we then connect every time we execute them? That seems like hiding quite a bit of functionality under the facade.
* The API would actually be different to the degree that there's going to be one less argument (no query), and therefore we can't even abstract it away behind a "query" trait.

What would potentially make more sense would be having a parameterizable query argument, something to the tune of:

```rust
fn select<Q: Query>(&self, query: Q, limit: ..., args: ...)
```

Where `Q` could be a string, or a parameterized query, or anything else that implements that `Query` interface.

Of course, this would also be touching the fact that `args` already have types in prepared statements, so it's no longer the same type signature as we have it today in SPI. We can address it, but this is one of the reasons why I opted for a simpler `execute_prepared_statement` in this PR: to avoid making a huge signature change.

yrashk:
Merged `develop` and resolved conflicts.

yrashk:
If you think the parameterized query is a good answer to the general problem, I can work on that, and that will bring prepared statements into the generalized interface. It might make sense to merge this in first just so that functionality is there but mark it experimental. Just so that when we alter the `query`'s type in the interface, we eliminate it.

yrashk:
I've given more thought to this parameterized query interface, and we can get much further there. We can make the cursor a query as well (as it is effectively a SpiTupleTable-returning interface, too), thus uniting at least three types of query under the same umbrella:

* String
* Cursor
* PreparedStatement

I am itching to try this out. I've merged develop into this PR and will try to iterate on this. I think this can become a nice interface.

yrashk:
I did a quick experiment, and it looks like unifying both strings and prepared statements makes a lot of sense.  Putting cursor under the same umbrella doesn't because of its differences. I've pushed out a change to query the API.

I have not updated `get_one/two/three` helpers to use the same interface just yet. It is certainly possible but makes the invocation of `get_one::<Type, _, _>` look a bit uglier because the signature would look like this:

```rust
    pub fn get_one<
        A: FromDatum + IntoDatum,
        Args,
        Q: Query<Result = SpiTupleTable, Arguments = Option<Args>>,
    >(
        query: Q,
    ) -> Option<A>
```

I am still of the opinion that an overhaul of these helpers is necessary and decided to delay this change concerning these helpers. However, if we will like the approach to unifying query API I've prepared in principle, this is something we can further explore.

As for cursors, I've poked around a bit more, and I think with the prepared statement support as this PR has, we can make cursors an implementation property of `Query` trait implementations. For example, strings will use the currently used `SPI_cursor_open_with_args` method, whereas prepared statements can enjoy `SPI_cursor_open`.

yrashk:
I went ahead and implemented cursors for prepared statements.

yrashk:
I was thinking more about `get_one/two/three` helpers and their adaptation to this change. 

The only reason why `Query` has a `Result` associated type is that originally, prepared statements used a `Result` return type to indicate an error as opposed to just returning `SpiTupleTable` as the regular statement implementation of `Query` does.

If we make them the same type, we can simplify `get_one/two/three` as we'll know their return type. This can be done by panicking on an error for the time being in prepared statement execution and integrating its error into the `Error` type from #902, and #902 is ready and merged.

However, this still leaves us with an extra boundary for the `Arguments` associated type to be an `Option`, which can be mitigated by requiring that `Arguments` has a `Default`.

At the end of this, we're still stuck with the parameter for `Q: Query` itself so the code would look like this:

```
client.get_one<i32, _>(query)
```

where `_` is the `Query` type.

yrashk:
Working on merging with `develop`

yrashk:
Updated.

pgx/src/spi.rs
```diff
@@ -160,6 +160,162 @@ impl SpiConnection {
...
+                    )
+                }
+            }
+            None => unsafe { pg_sys::SPI_execute(src.as_ptr(), read_only, limit.unwrap_or(0)) },
+        };
+
+        SpiClient::prepare_tuple_table(status_code)
+    }
+
+    fn open_cursor<'c: 'cc, 'cc>(
```
eeeebbbbrrrr, line 271: This function and `execute()` are essentially the same.  Need to find a way to eliminate all the code duplication.  And I suspect there's more...

pgx/src/spi.rs
```diff
@@ -591,6 +657,230 @@ impl Drop for SpiCursor<'_> {
...
+        }
+        OwnedPreparedStatement(PreparedStatement { phantom: PhantomData, plan: self.plan })
+    }
+}
+
+impl<'a: 'b, 'b> Query for &'b PreparedStatement<'a> {
+    type Arguments = Option<Vec<Option<pg_sys::Datum>>>;
+    type Result = Result<SpiTupleTable, PreparedStatementError>;
+
+    fn execute(
```
eeeebbbbrrrr, line 754: ... yup.  more of basically the same code here too

pgx/src/spi.rs
```diff
@@ -591,6 +657,230 @@ impl Drop for SpiCursor<'_> {
...
+        self,
+        client: &'cc SpiClient<'c>,
+        args: Self::Arguments,
+    ) -> Result<SpiCursor<'c>, Error> {
+        (&self.0).open_cursor(client, args)
+    }
+}
+
+impl<'a> PreparedStatement<'a> {
+    /// Converts prepared statement into an owner prepared statement
```
eeeebbbbrrrr, line 739: "into an **owned**"

pgx/src/spi.rs
```diff
@@ -591,6 +657,230 @@ impl Drop for SpiCursor<'_> {
...
+                nulls.as_mut_ptr(),
+                read_only,
+                limit.unwrap_or(0),
+            )
+        };
+
+        Ok(SpiClient::prepare_tuple_table(status_code))
+    }
+
+    fn open_cursor<'c: 'cc, 'cc>(
```
eeeebbbbrrrr, line 803: ... and here

eeeebbbbrrrr (changes requested):

pgx/src/spi.rs
yrashk: Resolved

pgx/src/spi.rs
yrashk: They are very similar but not the same. There are subtle differences. At the very minimum, we should extract this:

```rust
   for (argtype, datum) in args {
                    argtypes.push(argtype.value());

                    match datum {
                        Some(datum) => {
                            // ' ' here means that the datum is not null
                            datums.push(datum);
                            nulls.push(' ' as std::os::raw::c_char);
                        }

                        None => {
                            // 'n' here means that the datum is null
                            datums.push(pg_sys::Datum::from(0usize));
                            nulls.push('n' as std::os::raw::c_char);
                        }
                    }
                }
```

However, before we do any of this, let's agree on the surface of the PR first. We can either work on code quality improvement in the same PR after that or in a follow-up PR.

eeeebbbbrrrr: Please consolidate as much of the duplicated code as you can in this PR.

yrashk: Does this mean you agree with the API's shape?

yrashk: Working on the consolidation patch.

yrashk: I've extracted the most gnarly part of the repetitiveness. Because of subtle dissimilarities, there's still some similarly-looking code, but I don't see a beautiful way to make abstract it further, at least not seeing this worth the time at the moment.

eeeebbbbrrrr (changes requested):
The only last thing I'd want to see are some `// SAFETY:` docs around the handful of `unsafe {}` blocks.  I think @workingjubilee would insist on that and we have been trying to do a better job about this ourselves as we modify existing code and add new code.

Do that and I'll smash the merge button.

yrashk:
> The only last thing I'd want to see are some `// SAFETY:` docs around the handful of `unsafe {}` blocks. I think @workingjubilee would insist on that and we have been trying to do a better job about this ourselves as we modify existing code and add new code.
> 
> Do that and I'll smash the merge button.

Done and I've fixed an unsafety I found while commenting.

eeeebbbbrrrr:
Oh, one last thing.  Can you update the initial comment here to describe this PR a bit more than "implement basic support for prepared statements".  I'll include that in the commit message and it'll find its way into release notes as well.

Thanks, @yrashk

yrashk:
> Oh, one last thing. Can you update the initial comment here to describe this PR a bit more than "implement basic support for prepared statements". I'll include that in the commit message and it'll find its way into release notes as well.
> 
> Thanks, @yrashk

Done

eeeebbbbrrrr:
m m m m merged

yrashk:
🥳 🎆

PR #887, opened 2022-11-24
External binary representation support (SEND/RECEIVE for custom types)

yrashk:
Problem: text type representation is not always efficient

For this reason, Postgres allows types to have an external binary representation. Also, some clients insist on using binary representation.

Solution: introduce SendRecvFuncs trait and `sendrecvfuncs` attribute

These are used to specify how external binary representation encoding is accomplished.


workingjubilee:
Several traits all reusing the same pattern suggests we should be attempting to create a more generic interface instead of copypasta.

yrashk:
I am not against having a "more generic" interface, but since we don't have any specific proposal for one, can we work on getting this one since it's of very practical use as is? Even if under a feature gate if so desired.

I've spent a good amount of time getting it done following the current approach found in the library (in/out type of traits), and it feels that saying "we should now find a common interface" is an unfair treatment of the PR. So I would propose to get it into a mergeable shape first and getting in it. We can work on generalizing interfaces for these things as a follow-up. What do you think?

workingjubilee:
My concern is not necessarily a blocking concern, just something of note.

yrashk:
As discussed in private, I understand the concern of `SendRecvFuncs` superficial similarity to `PgVarlenaInOutFuncs`. That being said, while they are very similar shape-wise, their distinction lies in the specificity of such a trait (Send/Recv are intended explicitly for SEND/RECEIVE functions of the type).

`PgVarlenaInOutFuncs` is not suitable for some types (where `Copy` is not feasible), and thus I didn't consider adopting that trait (or breaking its API). I understand that you have doubts about whether `Copy` is required and whether it can be relaxed to `Clone`. 

You suggested that perhaps in most cases users will not need different serialization (between varlena and send/receive). While I tend to agree, I don't think this should mean we should not make it possible to specify a send/receive-specific implementation.

I feel like this *can* be a very productive discussion that can help us design the next iteration of API and make this part of pgx much smoother. I'd love to be a part of the conversation.

My only concern is making external binary format (as a feature) depend on the timeline of this discussion. I suggest getting it done first (with the caveat of some potentially changing APIs; but we're pre 1.0 so we can afford *some* refining) and then, once there's a good consensus, transition to a better design.

eeeebbbbrrrr:
> Several traits all reusing the same pattern suggests we should be attempting to create a more generic interface instead of copypasta.

Yes

> My concern is not necessarily a blocking concern, just something of note.

I think it probably should be.  We're at the point where a better approach is necessary.

I've been thinking about this over the holiday and I'd like to take a stab at making all of this better.  I have some ideas, but don't quite yet have the time to focus on it.

yrashk:
If you don't have time for this, perhaps it's a case of "better is the enemy of good," and we should consider something like this PR to provide the functionality first, and improve it later?

workingjubilee:
In practice, almost all implementations and users of `PgVarlenaInOutFuncs` are doing it via macro code, so I am not really worried about breaking users of it who manually touch it.

pgx/src/stringinfo.rs
```diff
@@ -189,6 +191,48 @@ impl StringInfo {
...
+                (*self.sid).cursor += start as c_int;
+                let result = std::slice::from_raw_parts(
+                    (*self.sid).data.add((*self.sid).cursor as usize) as *const u8,
+                    total,
+                );
+                (*self.sid).cursor += total as c_int;
+                result
+            }
+        })
+    }
```
workingjubilee, line 234: We should probably just `impl std::io::Read` using similar underlying code, which can be extracted into another PR. `str` implements a similar [`.get(..)`](https://doc.rust-lang.org/std/primitive.str.html#method.get) but it doesn't modify cursors. I wouldn't mind seeing this function with a different name.

pgx-tests/src/tests/stringinfo_tests.rs
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    #[pg_test]
+    fn test_string_info_read_cap() {
+        let mut string_info = StringInfo::from(vec![1, 2, 3, 4, 5]);
+        assert_eq!(string_info.read(..=1), Some(&[1][..]));
+        assert_eq!(string_info.read(1..=2), Some(&[3][..]));
+        assert_eq!(string_info.read(..), Some(&[4, 5][..]));
+    }
+}
```
workingjubilee, line 40: These tests, of course, can also go in to the other PR.

workingjubilee (changes requested):
There are actually two changesets here, and we can accept one if it takes a slightly different shape.

pgx/src/stringinfo.rs
yrashk: I can totally do this, but I think it would make most sense to do this after #903

etylermoss:
Any update on getting this merged for #1364?

LucaCappelletti94:
Any updates?

PR #892, opened 2022-11-29
Use JSON-serialization to pass things over the FFI, rather than `extern "Rust" fn`

thomcc:
I ended up spending some time last night (and now, because I could not sleep) trying to do the "easy" version of fixing the compiler parity requirement (https://github.com/tcdi/pgx/issues/774#issuecomment-1298783995) -- serializing via JSON, and passing it via a nul-terminated C string allocated in the `malloc` heap.

This currently works (on my machine) for everything other than the custom SQL stuff (`#[pg_extern(sql = generate_function)]` and such -- note that `sql = "blah"` and `sql = format!(...)` are fine, though).

Sadly, I'm unsure how to support that, and not sure it's possible. It's certainly impossible in a full-serialization approach (e.g. `#[link_section]` rather than `dlopen`). My current idea to solve this is to accept it can't be solved in the `to_sql` case, but solve it in all other cases. This would look like using the current approach (perhaps with some tweaks to be less fragile) for the . For example, if we had different symbol naming conventions for the cases where we need full support (`__pgx_internals_ord_foobar` vs `__pgx_internals_json_ord_foobar` for example), that might work. That said, it might be better to accept compiling a separate binary to do this stuff instead.

I'll pull the other miscellaneous improvements in this PR into their own PR later, since I suspect this is not going to land soon due to the issues.

[^1]: Aside from the major issue, I have not adjusted relevant documentation, removed the version check, added good error handling (we should tell users about the version mismatch on deserialization failure), or thought about the performance cost of changing all those `&'static str` to `String` (I'm assuming this is fine since I don't think any of it is at runtime? But I just wanted to get it working).


thomcc:
Actually, in the JSON case (but not the link_section case), the custom SQL generation function could be supported if we remove the `context: &PgxSql` parameter. We don't use it in the examples so IDK what it's use case is.

thomcc:
As expected, everything passes except for the custom SQL generation tests:

```
<snip>
failures:
    tests::schema_tests::tests::pg_custom_to_sql_extern
    tests::schema_tests::tests::pg_custom_to_sql_type

test result: FAILED. 294 passed; 2 failed; 1 ignored; 0 measured; 0 filtered out; finished in 113.61s
```

workingjubilee:
We can probably break those. Sometimes Ana's tests were... overly vigorous, to verify that the code works.

PR #895, opened 2022-11-29
Fix `pgx_utils::sql_entity_graph` Eq/PartialEq/Ord/PartialOrd/Hash impls.

thomcc:
Currently we have a number of incorrect manual implementations of the `PartialEq`/`Eq`/`PartialOrd`/`Ord`/`Hash` traits in `pgx_utils::sql_entity_graph`, and this fixes them to be derived instead.

1. This required using `BTree{Set,Map}` rather than `Hash{Set,Map}` for types that require implementations of `Ord`, `PartialOrd`, or `Hash`. (This is required because `Hash{Set,Map}` do not implement those traits)

2. For the hashing, this potentially comes with a bit of a performance cost (because now many more fields are hashed). To mitigate this, I switched us to use maps that use [`rustc-hash`](https://crates.io/crates/rustc-hash)'s[^why] `Hasher` which is significantly faster as well as being deterministic[^1]

This is cherry-picked from #892, since it's the only part of that that's worth doing on its own.

[^1]: The item order is still arbitrary, but it does not use randomized seeds.

[^why]: It's already in our dep tree, used by rustc, and on typical workloads (from `#[derive]`d structs and small-ish strings), it's much faster than any Hasher that I've tested, aside from ones that don't hash the whole input.

### DIFF
--- a/pgx-utils/src/sql_entity_graph/pg_extern/entity/mod.rs
+++ b/pgx-utils/src/sql_entity_graph/pg_extern/entity/mod.rs
@@ -30,10 +30,9 @@ use crate::sql_entity_graph::{SqlGraphEntity, SqlGraphIdentifier};
 use crate::ExternArgs;
 
 use eyre::{eyre, WrapErr};
-use std::cmp::Ordering;
 
 /// The output of a [`PgExtern`](crate::sql_entity_graph::pg_extern::PgExtern) from `quote::ToTokens::to_tokens`.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Eq, Ord, PartialEq, PartialOrd, Hash)]
 pub struct PgExternEntity {
     pub name: &'static str,
     pub unaliased_name: &'static str,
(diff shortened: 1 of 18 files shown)


workingjubilee:
> This requires . Additionally, it replaces uses of the stdlib's default
hasher inside `pgx-utils` with the hasher from `rustc_hash`, which is
both deterministic and significantly faster.

???

thomcc:
It does look like i messed up that commit message, nice catch.

thomcc:
I just took the message from the PR (lightly edited to work as a commit message) and squashed the commits while I was at it.

workingjubilee:
Not expecting Shakespeare or even useful commit messages, tbh,
I just would rather not have ones that leave me like
"Khé berga?"

workingjubilee (approved):
More determinism is generally a plus!

pgx-utils/src/sql_entity_graph/pg_extern/entity/mod.rs
```diff
@@ -51,32 +50,6 @@ pub struct PgExternEntity {
...
-        self.full_path.cmp(&other.full_path)
-    }
-}
-
-impl PartialOrd for PgExternEntity {
-    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
-        Some(self.cmp(other))
-    }
-}
-
```
eeeebbbbrrrr, line 79: I see how these impls are broken, but I also see the intention here.

The intention, I believe, is for the schema generator to output dependencies on the same level grouped by the source file that contains the entity.

(I'm actually having a hard time even understanding how the above impls worked at all.)

I don't particularly care if some fallout from this PR is that the schema outputs in a different (but still valid, of course) order, but it would be nice if the output can maintain this property of same-file-entities are kept together.  

Sooo... I'm not seeing how instead deriving `Eq, Ord, PartialEq, PartialOrd, Hash` does that?

PR #896, opened 2022-11-29
Ensure SpiClient is only used within boundaries of a connection

yrashk:
Problem: SpiClient can created out of thin air and leaked

One can create `SpiClient` when there's no connection.

One can also return it from `connnect`:

```rust
Spi::connect(|client| Ok(Some(client)))
```

Solution: make its construction private and give out a reference

This way it can only be used within `Spi::connect` and friends.


workingjubilee (approved):
This looks good to me.

eeeebbbbrrrr (approved):
Thanks @yrashk.  This is exactly what we were talking about.  Cool!

PR #897, opened 2022-11-29
SpiClient::update no longer requires &mut self

yrashk:
Problem: &mut SpiClient is not UnwindSafe as a mutable reference

This causes this code to not compile:

```rust
Spi::connect(|client| {
    PgTryBuilder::new(|| {
        client.update("SELECT 1", None, None);
    }).execute();
    Ok(Some(()))
});
```

Solution: let `SpiClient::update` take a non-mutable reference

Strictly speaking, `&mut self` was not necessary there

This is a follow-up to #896

### DIFF
--- a/pgx-tests/src/tests/spi_tests.rs
+++ b/pgx-tests/src/tests/spi_tests.rs
@@ -216,4 +216,15 @@ mod tests {
         struct T;
         assert!(matches!(Spi::connect(|_| Ok(Some(T))).unwrap(), T));
     }
+
+    #[pg_test]
+    fn test_spi_unwind_safe() {
+        Spi::connect(|client| {
+            PgTryBuilder::new(|| {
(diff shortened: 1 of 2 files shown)


eeeebbbbrrrr (approved):

pgx-tests/src/tests/spi_tests.rs
```diff
@@ -216,4 +216,14 @@ mod tests {
         struct T;
         assert!(matches!(Spi::connect(|_| Ok(Some(T))).unwrap(), T));
     }
+
+    #[pg_test]
+    fn test_spi_unwind_safe() {
+        Spi::connect(|client| {
+            PgTryBuilder::new(|| {
+                client.update("SELECT 1", None, None);
+            }).execute();
```
workingjubilee, line 225:
```suggestion
            })
            .execute();
```

PR #898, opened 2022-11-30
Non-leaking, consumable SpiClient

yrashk:
Problem: giving out SpiClient by reference

This makes it non-consumable, which is not desirable for some API surfaces.

Solution: use explicit lifetime bounds to prevent leaking the client

This patch also adds a doctest-based test to ensure leaking client is a compile-time error.

This is a follow-up for #896


eeeebbbbrrrr (approved):
LGTM.  This was definitely the missing link

PR #900, opened 2022-11-30
Ensure the lifetime of `SpiClient` is scoped to the connection 

thomcc:
Small followup to #898. Shouldn't impact safety (or public API) but makes it much harder for us to accidentally expose functions in the future that do the wrong thing here.


eeeebbbbrrrr (approved):
merge it!

PR #902, opened 2022-11-30
Explicit error handling in SPI

yrashk:
Problem: SPI API panics on most errors

This is not great as it doesn't allow for better handling of errors.

Solution: redesign SPI signatures to accommodate for errors

Importantly, this now enforces compatibility checks on Datum. This alone has led me to discover a few bugs in pgx that are now fixed (primarily incorrect compatibility checks and wrong expectations in tests).

---


pgx/src/spi.rs
```diff
@@ -102,6 +109,20 @@ impl TryFrom<libc::c_int> for SpiError {
     }
 }
 
+#[derive(thiserror::Error, Debug)]
+pub enum Error<E: Debug = anyhow::Error> {
```
workingjubilee, line 113: All of the code seems to be within thiserror's power. We do not need to add another dependency.

pgx/src/spi.rs
```diff
@@ -129,69 +150,54 @@ pub struct SpiHeapTupleData {
...
+        Spi::connect(|client| client.select(query, Some(1), None).first().get_two::<A, B>())
     }
 
     pub fn get_three<
-        A: FromDatum + IntoDatum,
-        B: FromDatum + IntoDatum,
-        C: FromDatum + IntoDatum,
+        A: FromDatum + IntoDatum + 'static,
+        B: FromDatum + IntoDatum + 'static,
+        C: FromDatum + IntoDatum + 'static,
```
workingjubilee, line 166: I am not sure we want to add the `'static` constraint to all of these.

pgx-examples/aggregate/src/lib.rs
```diff
@@ -161,15 +161,17 @@ mod tests {
...
         assert_eq!(retval, 2);
     }
     #[pg_test]
     fn test_integer_avg_with_null() {
         Spi::run("CREATE TABLE demo_table (value INTEGER);");
         Spi::run("INSERT INTO demo_table (value) VALUES (1), (NULL), (3);");
         let retval = Spi::get_one::<i32>("SELECT DEMOAVG(value) FROM demo_table;")
-            .expect("SQL select failed");
+            .expect("SQL select failed")
+            .unwrap();
```
workingjubilee, line 174: Adding double-unwraps is really grody, is there any chance we could instead have one of the `Error`'s variants indicate the `Null` case?

yrashk: I guess one can see Null as a type mismatch if the requested type is not `Option<T>`.

pgx/src/spi.rs
yrashk: This is coming from `try_from_datum` which gets it from `is_pass_by_value()`

yrashk: ...which gets it from `TypeIf::of`

pgx/src/spi.rs
yrashk: I've removed the dependency on `anyhow`.

pgx/src/spi.rs
eeeebbbbrrrr: Agree with @workingjubilee, this is definitely not what we want here.

Sounds like the problem is a little deeper in the stack?  I see the problem, but not a solution yet...

Oh wait...

Rather than going by Rust's TypeId seems we ought to be looking at Postgres OIDs in `IntoDatum::is_pass_by_value()`.  Thoughts?

pgx-examples/aggregate/src/lib.rs
eeeebbbbrrrr:
```rust
let i = Spi::get_one::<i32>("SELECT NULL::integer");
```

That ought to return an Err, yes.

eeeebbbbrrrr (changes requested):
At the moment, the addition of another level of unwrapping makes this PR pretty rough to follow.  How can we eliminate that?

And the `'static` lifetime bounds here and there needs to be addressed as well.

yrashk:
> At the moment, the addition of another level of unwrapping makes this PR pretty rough to follow. How can we eliminate that?
> 
> And the `'static` lifetime bounds here and there needs to be addressed as well.

I think we can eliminate extra unwrapping by enforcing the non-optionality of given types.

Static lifetime bounds are a bit of a deeper issue, need to think about it.

yrashk:
> > Static lifetime bounds are a bit of a deeper issue, need to think about it.
> 
> It all comes down to this code:
> 
> ```rust
>     unsafe fn try_from_datum(
>         datum: pg_sys::Datum,
>         is_null: bool,
>         type_oid: pg_sys::Oid,
>     ) -> Result<Option<Self>, TryFromDatumError>
>     where
>         Self: Sized + IntoDatum,
>     {
>         if !Self::is_compatible_with(type_oid) {
>             Err(TryFromDatumError::IncompatibleTypes)
>         } else if !is_null && datum.is_null() && !Self::is_pass_by_value() {
>             Err(TryFromDatumError::NullDatumPointer)
>         } else {
>             Ok(FromDatum::from_polymorphic_datum(datum, is_null, type_oid))
>         }
>     }
> ```
> 
> Like, maybe we can just get rid of that `else if` branch and delete the `is_pass_by_value()` function entirely. This is its **only** usage in pgx. All of the FromDatum::from_polymorphic_datum impls ought to be checking their `is_null` argument anyways, so perhaps this is just a bit of a premature sanity check we don't actually need.

If you think this is safe, I am cool with this.

eeeebbbbrrrr:
> If you think this is safe, I am cool with this.

I think it is.  If you look at the FromDatum impls for pass-by-reference types, you'll see they have something similar to:

```rust
        if is_null || datum.is_null() {
            None
        } else {
           Some(...)
        }
```

Which I suppose is fine.  @workingjubilee, I know we kinda discussed this awhile back offline and basically settled on this pattern is as close to the RightThing as we can be expected to do.

yrashk:
I've removed `'static` lifetime.

workingjubilee:
Correct.

yrashk:
I've converted to not using `Option` and working through updating tests.

yrashk:
I've updated this PR to not have `Option` but just the `Result`.

pgx-examples/aggregate/src/lib.rs
yrashk: Fixed.

pgx/src/spi.rs
yrashk: Fixed

yrashk:
I've improved the type signature for error management, it's less clunky now.

yrashk:
Updated this PR to support an unexpectedly merged #579

yrashk:
Merged `develop` to include #938

yrashk:
Merged `develop` to resolve conflict.

yrashk:
Can I ask for any additional feedback on this PR?

eeeebbbbrrrr:
I've started to take a look at this.  

I really suck at git, but I have now have a local branch here on pgx with all your commits in this.  I dunno what happens if I end up making changes (I probably will), so we'll figure that out.

I expect to spend probably two full days on this.

eeeebbbbrrrr:
@yrashk I've created #969, that is based on this.  I'm not sure on how to commit those changes into this branch.  I'll leave this open for now, but I think we should prefer the direction that #969 is going.  It still has all your commits from here.

yrashk:
> @yrashk I've created #969, that is based on this. I'm not sure on how to commit those changes into this branch. I'll leave this open for now, but I think we should prefer the direction that #969 is going. It still has all your commits from here.

You can't commit to my branch as it is in my personal repository on GitHub.

As for #969 I like some of its changes (notably, `IntoDatum` for `Result` and related changes) but not the reversal of the opinion on `Option`ality of the return as I don't see a good motivation for this. Why do you think "we should prefer" #969? 

I've left extended commentary on #969 as well.

yrashk:
I am still of the opinion that #902 (this PR) has the general interface right and I think we should avoid bundling ALL sorts of changes into one PR like #969 does with the additional introduction of "error as datum". Instead, I propose to continue to keep going with #902 and prepare a separate PR for "error as datum" or any other improvements.

It's not like we have to change the SPI interface at once. We can take steps. Importantly, the larger the PR's surface is, the harder it is to review it. This is why we can merge things into `develop` and improve on them as time passes.

I've added some minor fixes to this branch.

eeeebbbbrrrr:
> It's not like we have to change the SPI interface at once. 

In this case, we kinda do.  I don't want pgx' Spi API to be so volatile that users just don't use it because it changes drastically with every new pgx release.  It's worth it to spend a bit of time getting it as close to right as we can.

> Importantly, the larger the PR's surface is, the harder it is to review it. This is why we can merge things into develop and improve on them as time passes.

I don't see this PR (#902) being merged as-is.  We can continue to talk about it over in #969, but the "Spi::get_xxx()" API deciding that a NULL value is an error, isn't the right answer for SPI.

yrashk:
What I meant is that we don't have to change the SPI API in one PR. It's not like we cut releases after every PR.

yrashk:
I didn't say #902 should be merged as is. We could have discussed concerns and found a resolution, but instead, you simply took it over.

yrashk:
> I don't see this PR (#902) being merged as-is. We can continue to talk about it over in #969, but the "Spi::get_xxx()" API deciding that a NULL value is an error, isn't the right answer for SPI.

It absolutely does NOT decide this. This is simply false.

The user does this by specifying the type it wants out of it.

eeeebbbbrrrr:
@yrashk, if you'll rebase this against develop with no other changes then I'll be happy to merge tomorrow.  Then I'll rejigger #969 on top of develop.  `Option<T>` is the path forward, so some form of #969 will be merged soon.

yrashk:
I am not sure why you need a rebase here if #969 is the way you chose?

eeeebbbbrrrr:
#969 is built on top of this PR, so I'm fine with merging this.  Then when #969 is merged it'll just be the diff against this.  I'd like you to keep the commit credit for the work here since we squash merge.  If you're not interested in that, that's also fine.

yrashk:
You've merged develop a few days ago, and I've updated the tests to work with this. I just checked it's up-to-date.

There is, however, this [commit](fa1cd8dd0d9dd839d45e64d5727eacfe7f713565) that I don't know if you want to get in.

If this is simpler, you can simply give me credit in the merge of #969, or none at all (since you've reworked it substantially anyway), I am cool either way.

eeeebbbbrrrr:
> If this is simpler, you can simply give me credit in the merge of #969, or none at all (since you've reworked it substantially anyway), I am cool either way.

Okay, I'll make sure to do that.  You've got a significant amount of effort/time invested in this and #969.

PR #908, opened 2022-12-01
cleanup cargo-pgx handling of the `--profile` argument

eeeebbbbrrrr:
Especially as it relates to `cargo pgx package --profile foo`.

Fixes the error reported during

```shell
$ cargo pgx package --profile artifacts
Error:
   0: conflicting usage of --profile="artifacts" and --release
```

`cargo pgx package` will now prefer the specified profile, then default to "release"

I put this PR against `master` so that we can get it into a 0.6.1 release sometime.

### DIFF
--- a/cargo-pgx/src/command/install.rs
+++ b/cargo-pgx/src/command/install.rs
@@ -63,7 +63,10 @@ impl CommandExecute for Install {
             Some(config) => PgConfig::new_with_defaults(PathBuf::from(config)),
         };
         let pg_version = format!("pg{}", pg_config.major_version()?);
-        let profile = CargoProfile::from_flags(self.release, self.profile.as_deref())?;
+        let profile = CargoProfile::from_flags(
+            self.profile.as_deref(),
+            self.release.then_some(CargoProfile::Release).unwrap_or(CargoProfile::Dev),
+        )?;
 
         let features =
             crate::manifest::features_for_version(self.features, &package_manifest, &pg_version);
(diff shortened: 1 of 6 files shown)


cargo-pgx/src/command/install.rs
```diff
@@ -63,7 +63,10 @@ impl CommandExecute for Install {
             Some(config) => PgConfig::new_with_defaults(PathBuf::from(config)),
         };
         let pg_version = format!("pg{}", pg_config.major_version()?);
-        let profile = CargoProfile::from_flags(self.release, self.profile.as_deref())?;
+        let profile = CargoProfile::from_flags(
+            self.profile.as_deref(),
+            self.release.then_some(CargoProfile::Release).unwrap_or(CargoProfile::Dev),
```
eeeebbbbrrrr, line 68: I argued with myself about continuing to push down the boolean for "self.release" instead, but I kinda like this as it's a little more explicit about the profile each command wants.  But I have no strong preference.

thomcc (approved):
I think ideally we'd find a way to detect users passing `--profile` at the same time as `--release` or `--debug`, but it's probably not worth the hassle, so this seems fine.

eeeebbbbrrrr:
> I think ideally we'd find a way to detect users passing `--profile` at the same time as `--release` or `--debug`, but it's probably not worth the hassle, so this seems fine.

Yeah, I dunno that it's worth the effort.  The important thing, to me, is if the user gives a `--profile` we use that.  I'll document that in cargo-pgx/README.md

cargo-pgx/src/command/install.rs
workingjubilee: I am happiest when booleans go away, yes.

PR #909, opened 2022-12-01
Add traverse function

willmurnane:
Adds a function which walks a tree of `PgNode` objects, calling a callback at each node. This is useful for things like the planner, where objects of a particular type could occur in many places, and those places change depending on the version of PostgreSQL that's being used. For example, `RTEKind_RTE_RESULT` is only defined and used in version 12+, so the planner would need to handle different layouts of plans depending on the version of PostgreSQL.

### DIFF
--- a/pgx-pg-sys/build.rs
+++ b/pgx-pg-sys/build.rs
@@ -245,7 +245,7 @@ fn generate_bindings(
         .wrap_err_with(|| format!("bindgen failed for pg{}", major_version))?;
 
     let oids = extract_oids(&bindgen_output);
-    let rewritten_items = rewrite_items(&bindgen_output, &oids, is_for_release)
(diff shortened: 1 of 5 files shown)


thomcc:
Can you rebase this against `develop`/change the PR to be against that branch instead? We don't merge new features directly into master. (I could do it for you, but it may break your checkout)

eeeebbbbrrrr:
Hi @willmurnane.  Are you still interested in this PR?  In general I think this looks okay, but it's got some conflicts with `build.rs` now.

willmurnane:
Yes, I am. I’ll try to give it some TLC by early next week, thanks for the reminder.

JelteF:
Instead of implementing this logic directly in pgx I think it would be better to call out to the existing walker functions that are already part of Postgres, like `query_tree_walker`, `expression_tree_walker` and `range_table_entry_walker`.

Issue #655, opened 2022-08-26
Rename `pgx-utils` into something more meaningful, such as `pgx-sqlgen` or equivalent


BradyBonnette:
Some work was done to remove a lot of dependencies out of the critical build path, as a result of https://github.com/tcdi/pgx/pull/645

Essentially `pgx-utils` was stripped of functionality that really could be implemented other spots. At the end of the day, `pgx-utils` now only contains code that handles internal SQL entities parsing/mappings.

Even though this would be a small refactor, it lends itself to something being named appropriately instead of generically.

workingjubilee:
<joke>`pgx-compiler`</joke>

Hoverbear:
@workingjubilee I think it could be billed as a *trans*piler. :smirk:

workingjubilee:
@eeeebbbbrrrr weighed in and thinks that it might be preferable to keep pgx-utils for another round and try to whittle all of the crate down to nothing, moving components into `pgx-macros`, `cargo-pgx`, and `pgx-pg-sys` as appropriate.

PR #911, opened 2022-12-04
`pgx-utils` goes away!

eeeebbbbrrrr:
- Move the `sql_entity_graph` module out of `pgx-utils` and into its own crate.
- Move `pgx-utils/src/rewriter.rs` into `pgx-macros/src` and delete the `pgx-utils` crate entirely
- Some dead code removal

From there, the changes here are largely mechanical to account for the crate changes.  A few 3rd-party dependencies have been eliminated here and there as well.

`pgx-pg-sys/build.rs` now applies `#[pg_guard]` to bindgen `extern "C"` FFI blocks instead of automatically rewriting to its output directly.  In testing this doesn't seem to hurt compilation performance and it removes a dependency in our build chain.

Relates to Issue #655.


workingjubilee:
As you said yourself, we probably don't want to just rename `pgx-utils` if we can remove it instead, and this, despite the name, effectively constitutes renaming the crate because it just moves things around. @thomcc mentioned a plan to move most of the SQL entity graph code into `cargo-pgx`.

pgx-macros/src/rewriter.rs
```diff
@@ -150,9 +140,9 @@ impl PgGuardRewriter {
                             arg_list.extend(quote! { #ident, });
                         }
                     } else {
-                        eyre::bail!(
+                        panic!(
```
thomcc, line 143: Ideally we'd use `syn::Error` here, since panic will give a pretty bad error message inside a macro/build script.

pgx-pg-sys/build.rs
```diff
@@ -812,13 +811,10 @@ fn apply_pg_guard(items: &Vec<syn::Item>) -> eyre::Result<proc_macro2::TokenStre
...
                     match item {
-                        ForeignItem::Fn(func) => {
-                            // Ignore other functions -- this will often be
-                            // variadic functions that we can't safely wrap.
-                            if let Ok(tokens) = PgGuardRewriter::new().foreign_item_fn(func, abi) {
-                                out.extend(tokens);
-                            }
-                        }
+                        ForeignItem::Fn(func) => out.extend(quote! {
+                            #[pgx_macros::pg_guard]
```
thomcc, line 815: This will probably slow user builds down, as now more macro generation has to happen after the build script. That said I'll try and address this in a follow-up, it should actually be pretty easy.

thomcc (approved):
This is basically good -- admittedly I just skimmed stuff that looked like it's part of the rename/move.

A lot of this is stuff I wanted to do anyway after the JSON patch, which this will likely simplify (although it does mean that patch will need rewriting ⚰️).

There's some other follow-up stuff that I think we'll want -- (this will regress macro error messages slightly, and possibly compile-times for pgx users), but none of that is urgent or needs to block this (I'll try to get to those issues as part of a follow-up soon).

eeeebbbbrrrr:
> As you said yourself, we probably don't want to just rename `pgx-utils` if we can remove it instead, and this, despite the name, effectively constitutes renaming the crate because it just moves things around. @thomcc mentioned a plan to move most of the SQL entity graph code into `cargo-pgx`.

It is a rename, but at least it's clear what the code is used for now.  It sounds like @thomcc has further ideas that might let us eliminate the crate entirely, which would be great, but I think this is still important just for clarity.

workingjubilee:
Fair enough, then!

PR #912, opened 2022-12-05
Sub-transactions API

yrashk:
Problem: inability to fine-tune transaction boundaries easily

Solution: implement an interface for sub-transactions

It's a fairly simplistic interface at this point, allowing choosing a mode of release on drop (commit by default).

This is an improved version of the original implementation in pgx-contrib-spiext
(https://github.com/supabase/pgx-contrib-spiext/blob/main/src/subtxn.rs)

---

I welcome feedback as the interface is somewhat opinionated (however, I tried to make it as palatable as possible, and its previous iteration in pgx-contrib-spiext went through a few rounds of improvements to be used in real extensions). I've tried to follow our recent work on `SpiClient` in spirit.

It'd be great to get additional scrutiny for the safety and soundness of the implementation. 

---


yrashk:
Any feedback I can get on this from the maintainers? Thanks!

pgx/src/subxact.rs
```diff
@@ -0,0 +1,196 @@
...
+    pub fn rollback(mut self) -> Parent {
+        self.internal_rollback();
+        self.should_release = false;
+        self.parent.take().unwrap()
+    }
+
+    /// Returns the memory context this transaction is in
+    pub fn memory_context(&self) -> PgMemoryContexts {
+        PgMemoryContexts::For(self.memory_context)
+    }
```
workingjubilee, line 63: Is there a need to expose this? The `PgMemoryContexts` API is... more questionable and is in a... reevaluation stage. As trivial as this is, I think this can be acquired `unsafe`ly via querying Postgres globals if someone needs to examine the context while they're in it.

pgx/src/subxact.rs
```diff
@@ -0,0 +1,196 @@
...
+// This allows a SubTransaction of a SubTransaction to be de-referenced to SpiClient
+impl<Parent: SubTransactionExt, const COMMIT: bool> Deref
+    for SubTransaction<SubTransaction<Parent>, COMMIT>
+{
+    type Target = Parent;
+
+    fn deref(&self) -> &Self::Target {
+        self.parent.as_ref().and_then(|p| p.parent.as_ref()).unwrap()
+    }
+}
```
workingjubilee, line 158: Wouldn't it be more general for `SubTransaction<T, COMMIT>` to deref into `T`? Is there a problem with that approach? Too much recursion?

pgx/src/subxact.rs
```diff
@@ -0,0 +1,196 @@
...
+
+/// Trait that allows creating a sub-transaction off any type
+pub trait SubTransactionExt {
+    /// Parent's type
+    ///
+    /// In most common cases, it'll be equal to `Self`. However, in some cases
+    /// it may be desirable to use a different type to achieve certain goals.
+    type T: SubTransactionExt;
```
workingjubilee, line 166: Naming an associated type `T` is unusual, associated types are usually given descriptive names. In this case:
```suggestion
    /// In common cases, it'll be equal to `Self`. However, in some cases
    /// it may be desirable to use a different type to achieve certain goals.
    type Parent: SubTransactionExt;
```

pgx/src/subxact.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    /// Consume `self` and execute a closure with a sub-transaction
+    ///
+    /// If further use of the given sub-transaction is necessary, it must
+    /// be returned by the closure alongside with its intended result. Otherwise,
+    /// the sub-transaction be released when dropped.
+    fn sub_transaction<F: FnOnce(SubTransaction<Self::T>) -> R, R>(self, f: F) -> R
```
workingjubilee, line 173: Follow-on from previous remark on signature.
```suggestion
    fn sub_transaction<F: FnOnce(SubTransaction<Self::Parent>) -> R, R>(self, f: F) -> R
```

pgx/src/subxact.rs
```diff
@@ -0,0 +1,196 @@
...
+    where
+        Self: Sized;
+}
+
+impl<'a> SubTransactionExt for SpiClient<'a> {
+    type T = Self;
+    fn sub_transaction<F: FnOnce(SubTransaction<Self::T>) -> R, R>(self, f: F) -> R
```
workingjubilee, line 180:
```suggestion
    type Parent = Self;
    fn sub_transaction<F: FnOnce(SubTransaction<Self::Parent>) -> R, R>(self, f: F) -> R
```

pgx/src/subxact.rs
```diff
@@ -0,0 +1,196 @@
...
+    where
+        Self: Sized,
+    {
+        f(SubTransaction::new(self))
+    }
+}
+
+impl<Parent: SubTransactionExt> SubTransactionExt for SubTransaction<Parent> {
+    type T = Self;
+    fn sub_transaction<F: FnOnce(SubTransaction<Self::T>) -> R, R>(self, f: F) -> R
```
workingjubilee, line 190:
```suggestion
impl<P: SubTransactionExt> SubTransactionExt for SubTransaction<P> {
    type Parent = Self;
    fn sub_transaction<F: FnOnce(SubTransaction<Self::T>) -> R, R>(self, f: F) -> R
```

pgx/src/subxact.rs
```diff
@@ -0,0 +1,196 @@
...
+impl<Parent: SubTransactionExt, const COMMIT: bool> SubTransaction<Parent, COMMIT> {
+    /// Create a new sub-transaction.
+    fn new(parent: Parent) -> Self {
+        // Remember the memory context before starting the sub-transaction
+        let ctx = PgMemoryContexts::CurrentMemoryContext.value();
+        // Remember resource owner before starting the sub-transaction
+        let resource_owner = unsafe { pg_sys::CurrentResourceOwner };
+        unsafe {
+            pg_sys::BeginInternalSubTransaction(std::ptr::null());
+        }
```
workingjubilee, line 39: Postgres is having one of their silly "nullptr argument means something special" moments, aren't they? Could we get a safety comment explaining the hidden semantics here?

pgx/src/subxact.rs
```diff
@@ -0,0 +1,196 @@
...
+    }
+}
+
+impl<Parent: SubTransactionExt, const COMMIT: bool> SubTransaction<Parent, COMMIT> {
+    /// Create a new sub-transaction.
+    fn new(parent: Parent) -> Self {
+        // Remember the memory context before starting the sub-transaction
+        let ctx = PgMemoryContexts::CurrentMemoryContext.value();
+        // Remember resource owner before starting the sub-transaction
+        let resource_owner = unsafe { pg_sys::CurrentResourceOwner };
```
workingjubilee, line 36: Could we get a note mentioning when this variable is set (i.e. "what is the last write we are expecting to be reading")?

pgx/src/subxact.rs
```diff
@@ -0,0 +1,196 @@
...
+    // on the drop of the original value.
+    should_release: bool,
+    parent: Option<Parent>,
+}
+
+impl<Parent: SubTransactionExt, const COMMIT: bool> Debug for SubTransaction<Parent, COMMIT> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.write_str(std::any::type_name::<Self>())
+    }
+}
```
workingjubilee, line 28: Does `#[derive(Debug)]` break something?

pgx/src/subxact.rs
```diff
@@ -0,0 +1,196 @@
...
+        self.should_release = false;
+        self.parent.take().unwrap()
+    }
+
+    /// Rollback the transaction, returning its parent
+    pub fn rollback(mut self) -> Parent {
+        self.internal_rollback();
+        self.should_release = false;
+        self.parent.take().unwrap()
+    }
```
workingjubilee, line 58: It seems this code is expecting `self.parent` to actually be unconditionally `Some`, so why is it that we use an `Option`? Just for `mem::take` antics later?

pgx/src/subxact.rs
```diff
@@ -0,0 +1,196 @@
+use crate::{pg_sys, PgMemoryContexts, SpiClient};
+use std::fmt::{Debug, Formatter};
+use std::ops::Deref;
+
+/// Sub-transaction
+///
+/// Can be created by calling `SpiClient::sub_transaction`, `SubTransaction<Parent>::sub_transaction`
+/// or any other implementation of `SubTransactionExt` and obtaining it as an argument to the provided closure.
```
workingjubilee, line 8: My internal grammarian is trying to say the function names should be `subtransaction` (which is valid in English), whereas the type names should be `SubTransaction`, but that seems... somehow peculiar, given that normally those would normalize to either `sub_transaction` and `SubTransaction` like you have it here or `subtransaction` and `Subtransaction`.

pgx/src/subxact.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    /// Returns the memory context this transaction is in
+    pub fn memory_context(&self) -> PgMemoryContexts {
+        PgMemoryContexts::For(self.memory_context)
+    }
+
+    fn internal_rollback(&self) {
+        unsafe {
+            pg_sys::RollbackAndReleaseCurrentSubTransaction();
+            pg_sys::CurrentResourceOwner = self.resource_owner;
```
workingjubilee, line 68: None of pgx's existing code manages `CurrentResourceOwner`, so it would be great to link to the Postgres headers that describe how to manage this variable and to comment on why we are doing so.

pgx/src/subxact.rs
```diff
@@ -0,0 +1,196 @@
...
+/// (default) or roll back if it is `false`.
+pub struct SubTransaction<Parent: SubTransactionExt, const COMMIT: bool = true> {
+    memory_context: pg_sys::MemoryContext,
+    resource_owner: pg_sys::ResourceOwner,
+    // Should the transaction be released, or was it already committed or rolled back?
+    //
+    // The reason we are not calling this `released` as we're also using this flag when
+    // we convert between commit_on_drop and rollback_on_drop to ensure it doesn't get released
+    // on the drop of the original value.
+    should_release: bool,
```
workingjubilee, line 20: "on the drop of the original value"... hm.

pgx/src/subxact.rs
```diff
@@ -0,0 +1,196 @@
...
+        let result = SubTransaction {
+            memory_context: self.memory_context,
+            resource_owner: self.resource_owner,
+            should_release: self.should_release,
+            parent: self.parent.take(),
+        };
+        // Make sure original sub-transaction won't commit
+        self.should_release = false;
+        result
+    }
```
workingjubilee, line 109: Oh, I see. We copy everything into a new SubTransaction and then let `Drop` run for the previous one.

pgx/src/subxact.rs
yrashk: This stems from a very early implementation of try/catch where there was a need for it. I think it can safely go. I will remove it.

pgx/src/subxact.rs
yrashk: This is how it was initially https://github.com/supabase/pgx-contrib-spiext/blob/main/src/subtxn.rs#L137-L143

However, this has prevented me from [using `SubTransaction<T: Deref<SpiClient>>`](https://github.com/supabase/pgx-contrib-spiext/blob/develop/src/checked.rs#L30-L33) so I had to resort to boxing: https://github.com/supabase/pgx-contrib-spiext/blob/develop/src/subtxn.rs#L169-L178

However, with this special case, I no longer need a Box

pgx/src/subxact.rs
yrashk: Agree. I will address it.

pgx/src/subxact.rs
yrashk: It just means there's no name. I can leave a comment there.

```
             /*
              * Savepoint names, like the TransactionState block itself, live
              * in TopTransactionContext.
              */
             if (name)
                 s->name = MemoryContextStrdup(TopTransactionContext, name);
             break;
```

pgx/src/subxact.rs
yrashk: I agree. I will include these

https://github.com/postgres/postgres/blob/cae6fc2bc27cdb072693076249ce688f048ca7b7/src/backend/utils/resowner/README
https://github.com/postgres/postgres/blob/cd4e8caaa0eb5c3d3b12a45b29035e7a4bdfa4a8/src/pl/plpython/plpy_spi.c#L77-L80 (example of its use)

Do you think this will be sufficient?

pgx/src/subxact.rs
yrashk: Can you please elaborate on this request? I am not 100% sure I understood what you're asking me to do.

pgx/src/subxact.rs
yrashk: IIRC, this was because I couldn't move it out of the `SubTransaction` when I am changing the COMMIT type parameter because it implements `Drop`.

pgx/src/subxact.rs
yrashk: Something was preventing it from working initially, but it looks like it's no longer the case. Removed.

pgx/src/subxact.rs
yrashk: Fixed.

pgx/src/subxact.rs
yrashk: I named it `sub_transaction` for consistency with `SubTransaction`. But ultimately I don't have a strong opinion about this.

yrashk: And Postgres calls them "SubTransactions" in their API.

pgx/src/subxact.rs
yrashk: Added background information on this.

pgx/src/subxact.rs
yrashk: Left a comment

pgx/src/subxact.rs
yrashk: Yep. Does this resolve https://github.com/tcdi/pgx/pull/912#discussion_r1047828161 or do you want to provide additional feedback on this?

pgx/src/subxact.rs
yrashk: Ultimately, I wouldn't mind getting rid of `Option` (I tried!) but I haven't thought of anything better yet.

pgx/src/subxact.rs
workingjubilee: Yeah, I think the type/trait names should use `SubTransaction`. Maybe I will start using `sub_transaction` in anger and it will bug me enough to want to take the `_` out, and I'll just let it be inconsistent-but-natural then. But otherwise it seems fine.

pgx/src/subxact.rs
workingjubilee: Yes that should be fine.

pgx/src/subxact.rs
workingjubilee: You can split `SubTransaction` into two fields. You can't destructure types that have `impl Drop`, but you can destructure types made of types that `impl Drop`, so you move the feature of the commit-on-drop (and thus thread the const generic) into that inner field. Since users can't access the type's inner workings, it still has `impl Drop` as far as they are concerned.

Then you use destructuring syntax to get at the fields, and instead of using a `should_release` bool, you simply `mem::forget` it. This also lets you elevate the requirements for `internal_commit` and `internal_rollback`.

pgx/src/prelude.rs
```diff
@@ -47,3 +47,6 @@ pub use crate::pg_sys::{
     check_for_interrupts, debug1, debug2, debug3, debug4, debug5, ereport, error, function_name,
     info, log, notice, warning, FATAL, PANIC,
 };
+
+// Sub-transactions
+pub use crate::subxact::*;
```
eeeebbbbrrrr, line 52: Probably better to limit this to just the module:  `pub use crate::subxact;`.

I think we're trying to be a little better about globbing everything into the current namespace.

pgx/src/subxact.rs
eeeebbbbrrrr: I think Jubilee means when might Postgres initially set (or later change) this and how are we certain it even has a valid value at this point in time?

eeeebbbbrrrr (changes requested):
I think in general this is okay.  I've never used any of Postgres subtransaction facilities from C before, and haven't studied their docs, so I'm leaning heavily on the thought that you have.  

That said, before we can merge I'd like to see two things...

1.  A lot of additional doc comment documentation in `subxact.rs`, from the module-level down.  I think this needs some exposition around not only what it is, but how it interacts with the database, and how that might impact the user's code and design decisions.

2. A new example that demonstrates using this, maybe in some kind of fabricated scenario that users could maybe adapt to their own purposes (ie, something a little more polished than, say, our "strings" example).  And with that example, a nice README.md that covers topics around proper subxact handling.

This code itself isn't very complex, which is nice.  But it's a brand new feature and I'd like it to not only be as immediately useful as possible to users, but also serve as an example of what a new feature PR ought to look like.  ;)

### Other Thoughts

- I also wonder if/how this should interop with pgx' existing subact hooks support (see `callbacks.rs/register_subxact_callback()` and friends.  
- What needs to be said in docs (or done in code?) with regards to `PgTryBuilder` and subtransactions?
- Is this also usable as-is by background workers?  Is there an opportunity for expanding the existing bgworker example to show off subxact support?
- I'm sure I'll come up with other stuff as we get this merged.

yrashk:
Thank you for your feedback.

* I will work on more documentation. I didn't want to invest time into that before we (at least somewhat) agreed on the approach to avoid re-documenting things.

* I will write some examples as well. I don't see any reason why this won't be usable in background workers, as long as they are SPI-enabled. I can extend the example there.

* PgTryBuilder thing is the most interesting here. I left this out of this PR but there's a case when you want to recover the parent even if your sub-transaction failed. Currently, the parent would be gone, which is not perfect. However, with a trivial addition of another crate-level `SubTransactionEx`-implementing type (similar to https://github.com/supabase/pgx-contrib-spiext/blob/develop/src/subtxn.rs#L202-L227) one can implement a wrapper around `PgTryBuilder` that would recover the parent in the event of failure. If you'd like that to be a part of this PR, I will be happy to amend it. Easier to show it in code than explain.

* I have not yet considered `register_subxact_callback` interactions/interop. I'll think about it.

pgx/src/prelude.rs
yrashk: I am not sure this is the best. At the very least, I think we should consider `pub use crate::subxact::{self, SubTransactionExt}` for the sub-transaction functionality to be easily available.

yrashk:
Updated subtransactions to work on recent develop. I will now work on docs/examples and other concerns listed above.

eeeebbbbrrrr:
@yrashk are you still planning on taking this over the finish line?

yrashk:
Yes, I've discovered some issues with the API that make its usability reduced (mostly related to mutability) when porting my "contrib" crate to use this core and extending it to essentially allow for error recovery without losing the entire subtransaction stack.

I'm working on rethinking this and attempting to see what I need to change in the core of this PR to make this possible.

eeeebbbbrrrr:
Fair enough.  Subxacts are definitely pretty hard to reason about.

PR #915, opened 2022-12-05
A `no-schema-generation` feature flag

eeeebbbbrrrr:
This adds a `no-schema-generation` feature flag to pgx, which it pushes down into both `pgx-macros` and `pgx-sql-entity-graph` (`pgx-macros` pushes it down as well).

Also "cleans up" pgx-sql-entity-graph a bit by introducing a two new traits, `ToEntityGraphTokens` and `ToRustCodeTokens`, and a new type `CodeEnrichment<T>`.  It forces the various sql interface types, such as `PgExtern`, `PgAggregate`, `PostgresTypes`, etc, to implement these traits individually, and then it provides a blanket `impl ToTokens for CodeEnrichment<T: ToEntityGraphTokens + ToRustCodeTokens>`.  When the new `no-schema-generation` flag is set, then it simply skips generating the tokens necessary to support schema generation.

The `Schema` type can't follow these rules, for reasons, so it also handles the `no-schema-generation` feature flag directly.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -155,6 +155,9 @@ jobs:
     - name: Run custom_types example tests
       run: cargo test --package custom_types --features "pg$PG_VER" --no-default-features
 
+    - name: Run custom_types without schema generation example tests
+      run: cargo test --package custom_types --features "pg$PG_VER no-schema-generation" --no-default-features
+
     - name: Run custom_sql example tests
       run: cargo test --package custom_sql --features "pg$PG_VER" --no-default-features
 
--- a/pgx-examples/custom_types/Cargo.toml
+++ b/pgx-examples/custom_types/Cargo.toml
@@ -14,6 +14,7 @@ pg13 = ["pgx/pg13", "pgx-tests/pg13" ]
 pg14 = ["pgx/pg14", "pgx-tests/pg14" ]
 pg15 = ["pgx/pg15", "pgx-tests/pg15" ]
 pg_test = []
+no-schema-generation = [ "pgx/no-schema-generation", "pgx-tests/no-schema-generation" ]
 
 [dependencies]
 pgx = { path = "../../pgx", default-features = false }
--- a/pgx-examples/custom_types/src/complex.rs
+++ b/pgx-examples/custom_types/src/complex.rs
@@ -59,6 +59,7 @@ mod tests {
     use maplit::*;
     use pgx::prelude::*;
 
+    #[cfg(not(feature = "no-schema-generation"))]
     #[pg_test]
     fn test_known_animals_via_spi() {
         let animals = Spi::get_one::<Animals>("SELECT known_animals();");
--- a/pgx-examples/custom_types/src/ordered.rs
+++ b/pgx-examples/custom_types/src/ordered.rs
@@ -54,6 +54,7 @@ mod tests {
     use crate::ordered::OrderedThing;
     use pgx::prelude::*;
 
+    #[cfg(not(feature = "no-schema-generation"))]
     #[pg_test]
     fn test_ordering_via_spi() {
         let items = Spi::get_one::<Vec<OrderedThing>>(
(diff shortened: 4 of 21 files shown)


thomcc (commented):
This is much cleaner than the status quo (just the refactoring is a big win), and while I'd like to improve it a bit further, that can wait and def shouldn't block this.

I have a weak concern around it being a feature, since if anybody toggles it on accidentally, everything is broken without recourse, compared to a `--cfg`, which can be forced on/off. But that also comes with the cost of a full rebuild, and being more annoying to enable. I think this is fine for now, but if it causes some issues down the road, well, it's unsurprising.

I have two concrete short-term concerns:
- Is it possible to check that this compiles in CI at all? It adds some conditional paths that could bitrot if that doesn't happen.
- If it's possible and you haven't done so yet, you should make sure it works for your use case in plrust before landing. It would not surprise me if something unexpected prevented it from working, like weird feature behavior or unexpected symbol issues.

The first is more important tho for sure, since it could get forgotten (whereas presumably the other will happen one way or another eventually). We should keep the amount of configurations that don't even get seen by `cargo check` as low as possible, IMO.

r=me if that gets done.

eeeebbbbrrrr:
> Is it possible to check that this compiles in CI at all? It adds some conditional paths that could bitrot if that doesn't happen.

Sure, I'll add some kind of test like we discussed in Discord

> If it's possible and you haven't done so yet, you should make sure it works for your use case in plrust before landing. It would not surprise me if something unexpected prevented it from working, like weird feature behavior or unexpected symbol issues.

My use of this with plrust is still WIP, but yes, it is doing what we want it to do there -- pushing down this new `no-schema-generation` feature flag creates user_function.so's without any of the sql entity graph "__pgx__internals" functions.

eeeebbbbrrrr:
Okay, I added support for the "custom_types" example to also be run with "no-schema-generation".  

Compilation of that crate with the feature flag turned on produces some warnings in that crate (not pgx), but that's expected in this case:

```
"cargo" "test" "--features" "no-schema-generation pg_test"
   Compiling custom_types v0.1.0 (/Users/e_ridge/_work/pgx/pgx-examples/custom_types)
warning: unused imports: `Animals`, `known_animals`
  --> pgx-examples/custom_types/src/complex.rs:58:26
   |
58 |     use crate::complex::{known_animals, Animals};
   |                          ^^^^^^^^^^^^^  ^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: unused import: `maplit::*`
  --> pgx-examples/custom_types/src/complex.rs:59:9
   |
59 |     use maplit::*;
   |         ^^^^^^^^^

warning: unused import: `pgx::prelude::*`
  --> pgx-examples/custom_types/src/complex.rs:60:9
   |
60 |     use pgx::prelude::*;
   |         ^^^^^^^^^^^^^^^

warning: unused import: `crate::ordered::OrderedThing`
  --> pgx-examples/custom_types/src/ordered.rs:54:9
   |
54 |     use crate::ordered::OrderedThing;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: unused import: `pgx::prelude::*`
  --> pgx-examples/custom_types/src/ordered.rs:55:9
   |
55 |     use pgx::prelude::*;
   |         ^^^^^^^^^^^^^^^

warning: unused import: `crate::rust_enum::SomeEnum`
  --> pgx-examples/custom_types/src/rust_enum.rs:23:9
   |
23 |     use crate::rust_enum::SomeEnum;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: unused import: `pgx::prelude::*`
  --> pgx-examples/custom_types/src/rust_enum.rs:24:9
   |
24 |     use pgx::prelude::*;
   |         ^^^^^^^^^^^^^^^

warning: `custom_types` (lib test) generated 7 warnings
    Finished test [unoptimized + debuginfo] target(s) in 1.78s
     Running unittests src/lib.rs (/Users/e_ridge/_work/pgx/target/debug/deps/custom_types-342797d76b18bbbb)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

PR #917, opened 2022-12-05
Introduce a "codegen" module 

eeeebbbbrrrr:
Publicly exported as `pgx::prelude::__pgx__codegen__private__` that exports all the various pgx and pgx-related symbols that our code generation (via `pgx-macros` and `pgx-sql-entity-graph`) use.

Going forward, if we modify any code generation or support code generation for some new construct, we will need to make sure any new symbols are added to the `codegen.rs` module and referenced as `__pgx__codegen__private__::SymbolName`.

The primary justification for this is that it'll make it easier for certain types of extensions (read: plrust) to limit the scope of pgx internals they need to publicly expose.  And while it'll be a bit of a maintenance burden for us, it is very nice to have a single module that lists all the symbols our code generation uses -- this will be handy for auditing purposes among others, I'm sure.

Also touch `pgx::prelude` a bit, adding a few more things to help existing code compile.  It also updates the tests (doctest and unit tests) along with examples that needed it to `use pgx::prelude::*`.

I'd expect this commit to be a bit of a breaking change for existing users where perhaps they're not doing `use pgx::prelude::*` in their code.

I have a thought of a follow-up commit to *reduce* the rexports from the top-level `pgx::` module, but we can save that for another day.


pgx-macros/src/lib.rs
```diff
@@ -628,18 +628,18 @@ fn impl_postgres_enum(ast: DeriveInput) -> proc_macro2::TokenStream {
...
-        into_datum.extend(quote! { #enum_ident::#label_ident => Some(::pgx::lookup_enum_by_label(#enum_name, #label_string)), });
+        into_datum.extend(quote! { #enum_ident::#label_ident => Some(__pgx__codegen__private__::lookup_enum_by_label(#enum_name, #label_string)), });
     }
 
     stream.extend(quote! {
-        impl ::pgx::datum::FromDatum for #enum_ident {
+        impl __pgx__codegen__private__::FromDatum for #enum_ident {
             #[inline]
-            unsafe fn from_polymorphic_datum(datum: ::pgx::pg_sys::Datum, is_null: bool, typeoid: ::pgx::pg_sys::Oid) -> Option<#enum_ident> {
+            unsafe fn from_polymorphic_datum(datum: __pgx__codegen__private__::Datum, is_null: bool, typeoid: __pgx__codegen__private__::Oid) -> Option<#enum_ident> {
```
workingjubilee, line 637: This makes reading the `cargo expand` versions of PGX code **much** worse and it doesn't feel like it actually improves anything. In addition, it creates another layer of indirection, instead of just stating the path that something came from, because of the reexports.

eeeebbbbrrrr: we can bikeshed the name, but I'm convinced of real benefits to having all pgx symbols our code generation uses coming from a central place.

thomcc (approved):
This looks fine to me.

pgx-macros/src/lib.rs
thomcc: > This makes reading the cargo expand versions of PGX code much worse and it doesn't feel like it actually improves anything

Possibly, but it's semi-important that we make it clear which things are intended to be user-accessible. `rustc`, `rust-analyzer`, and `clion`, will all happily tell you to use things in `#[doc(hidden)]` modules that are effectively private with no warning. This has caused big issues in the ecosystem before (a largeish crate (dashmap) was using private modules from serde due to this, and there was widespread breakage).

My feeling is that anybody using `cargo expand` on `pgx` is in for a night of horrors, but I hadn't really considered it. Certainly if you'd like it to be slightly less noisy while still being clear that it's private, I don't mind.

PR #918, opened 2022-12-05
Problem: can't use `pg_guard` on parameterized functions

yrashk:
This code won't compile:

```rust
#[pg_guard]
extern "C" fn extern_func_impl<T>() -> bool {
    true
}
```

The error is

```
error[E0282]: type annotations needed
```

Solution: ensure this case is handled

### DIFF
--- a/pgx-tests/src/tests/pg_guard_tests.rs
+++ b/pgx-tests/src/tests/pg_guard_tests.rs
@@ -1,0 +1,28 @@
+/*
+Portions Copyright 2019-2021 ZomboDB, LLC.
+Portions Copyright 2021-2022 Technology Concepts & Design, Inc. <support@tcdi.com>
+
+All rights reserved.
+
+Use of this source code is governed by the MIT license that can be found in the LICENSE file.
+*/
+
+use pgx::prelude::*;
+
+#[pg_extern]
+fn extern_func() -> bool {
+    extern_func_impl::<u8>()
+}
+
+// This ensures that parameterized function compiles when it has `pg_guard` attached to it
+#[pg_guard]
+extern "C" fn extern_func_impl<T>() -> bool {
+    true
+}
+
+#[cfg(any(test, feature = "pg_test"))]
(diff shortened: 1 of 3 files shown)


eeeebbbbrrrr (approved):
oh, good one!

I'll merge soon as CI passes.  Thanks @yrashk!

yrashk:
You can merge it now :)

workingjubilee (changes requested):
Mmm. Using a type generic or const generic in a function, as opposed to a lifetime generic, is incompatible with `#[no_mangle]`. Rejecting type generics for `#[pg_guard]`, which is often used in concert with `#[no_mangle]`, is thus the correct behavior. PGX often does this, in particular. However, this change is still permissible if we have a separate enforcement strategy that specifically catches the case of a function annotated with both `#[no_mangle]` and `#[pg_guard]`. Otherwise I believe we are exposing ourselves to accidentally allowing significant miscompilations.

yrashk:
I'll take care of this case

yrashk:
@workingjubilee great catch; I've addressed your concern. Let me know if this looks good!

pgx-tests/src/tests/pg_guard_tests.rs
```diff
@@ -0,0 +1,46 @@
...
+
+#[pg_extern]
+fn extern_func() -> bool {
+    extern_func_impl::<u8>()
+}
+
+// This ensures that parameterized function compiles when it has `pg_guard` attached to it
+#[pg_guard]
+// Uncommenting the line below will make it fail to compile
+// #[no_mangle]
```
workingjubilee, line 20: We should have some kind of compiletest-like suite, so we can test for compile failures.

workingjubilee (approved):
Excellent, this is now fine!

eeeebbbbrrrr:
I'll merge this into develop and then manually fix it for 0.6.1 -- `rewriter.rs` is in a different place now

PR #919, opened 2022-12-05
Improve error handling a bit, by converting some panics into `syn::Error`s

thomcc:
Quick one-off since @eeeebbbbrrrr mentioned difficulty debugging macros, and much of this is easier after #913, even though #911 made things briefly worse (actually, it made this easier too, even if it introduced a couple panics in the short term).

This should improve the error messages of (several of) our macros, by actually emitting `syn::Error`s with span info. It's not a complete overhaul, basically just what I could get done in an hour or two before deciding "I've gotten side-tracked on this long enough, time to push it up". 

It changes some parsing logic around a bit, just to push things towards emitting the error when parsing the tokens into a SQL entity type (since this is where we have more consistent and accurate span information in most cases), rather than when emitting the `extern "C"` function. While this might look slightly less than mechanical, it was not.

It's also a lot of replacing `unwrap`, `panic!`, etc. with actual handling, and making sure that `syn::Error`s bubbled all the way up.


eeeebbbbrrrr:
I fixed the merge conflicts that merging #925 caused.  Lets see what CI says now.  I'll look at merging tomorrow.

eeeebbbbrrrr (approved):

eeeebbbbrrrr:
Thanks Thom.  I'm merging this now.  It'll definitely make life easier for someone when they encounter an error.

PR #920, opened 2022-12-05
Handle panics in PgMemoryContexts::switch_to

yrashk:
Problem: panics in ctx.switch_to are unhandled

This also means the memory context we're switching from is not going to be restored if a panic occurs.

Solution: ensure previous memory context is always restored


yrashk:
For additional context, the test in this PR will crash (not just fail) if the change in memcxt.rs is not present.

workingjubilee:
That is absolutely wild. Great catch.

workingjubilee (approved):

workingjubilee:
whoops, sorry!

PR #921, opened 2022-12-05
Ensuring owned memory context is not current when droppng it

yrashk:
Problem: dropping owned memory context while it's current

The error will look like this:

```
TRAP: FailedAssertion("context != CurrentMemoryContext", File: "mcxt.c", Line: 224, PID: 71536)
```

Solution: avoid failing this assertion

We can say that it is on the user of `pgx` to ensure that the owned context is dropped at the right time, but it may get tricky to do so.

For example, certain types are not UnwindSafe, so we can't use `switch_to` or `PgTryBuilder` and guard the exceptional situation to ensure the context is no longer current before it is dropped. So, due to these limitations, we're forced to use `set_as_current`, and the reversal to the old context will never happen if there's a panic.

Notes:

* This method restores context's parent, not the previous context, which is not ideal. However, for that to happen, I'd probably need to make `set_as_current`  take `&mut self` to be able to restore to the previous context. I am happy to explore this option, too.
* This does introduce a small amount of overhead on dropping (comparing pointers), but I feel like this is a worthy trade-off to avoid crashing the process. That being said, I agree to a certain extent that this is defensive programming, but, perhaps, justifiable (see the case above).


eeeebbbbrrrr:
I’d like to hold this for a 0.6.1 release. Which I can put together on Tuesday.

pgx/src/memcxt.rs
```diff
@@ -179,6 +179,11 @@ pub struct OwnedMemoryContext(pg_sys::MemoryContext);
 impl Drop for OwnedMemoryContext {
     fn drop(&mut self) {
         unsafe {
+            // In order to prevent crashes, if we're trying to drop
+            // a context that is current, switch to its parent, and then drop it
+            if pg_sys::CurrentMemoryContext == self.0 {
```
eeeebbbbrrrr, line 184: Similar to #922, this ought to use `std::ptr::eq()`

eeeebbbbrrrr (changes requested):

eeeebbbbrrrr:
Sorry to do this to you @yrashk, but can you rebase this off the newly created `0.6.1-prep` branch?

yrashk:
If we agree that #922 is better, let's close this one?

PR #922, opened 2022-12-06
Ensuring owned memory context is not current when dropping it (extension)

yrashk:
(Supersedes #921)

Problem: dropping owned memory context while it's current
    
The error will look like this:
    
```
TRAP: FailedAssertion("context != CurrentMemoryContext", File: "mcxt.c", Line: 224, PID: 71536)
```
    
Solution: avoid failing this assertion
  
We can say that it is on the user of `pgx` to ensure that the owned
context is dropped at the right time, but it may get tricky to do so.
    
For example, certain types are not UnwindSafe, so we can't use
`switch_to` or `PgTryBuilder` and guard the exceptional situation to
ensure the context is no longer current before it is dropped. So, due
to these limitations, we're forced to use `set_as_current` and the
reversal to the old context will never happen if there's a panic.

---

This is a stackable extension to #921. Instead of returning to the parent context, which may be an unexpected behavior, keep track of the previous context and returns to it instead.

I think this behavior is better than the original one in #921 (returns context's parent)

### DIFF
--- a/pgx/src/memcxt.rs
+++ b/pgx/src/memcxt.rs
@@ -178,6 +178,11 @@ pub struct OwnedMemoryContext(pg_sys::MemoryContext);
 impl Drop for OwnedMemoryContext {
     fn drop(&mut self) {
         unsafe {
+            // In order to prevent crashes, if we're trying to drop
+            // a context that is current, switch to its parent, and then drop it
+            if pg_sys::CurrentMemoryContext == self.0 {
+                pg_sys::CurrentMemoryContext = (*self.0).parent;
+            }
             pg_sys::MemoryContextDelete(self.0);
         }
     }
--- a/pgx/src/memcxt.rs
+++ b/pgx/src/memcxt.rs
@@ -183,7 +183,7 @@ impl Drop for OwnedMemoryContext {
         unsafe {
             // In order to prevent crashes, if we're trying to drop
             // a context that is current, switch to its predecessor, and then drop it
-            if pg_sys::CurrentMemoryContext == self.owned {
+            if ptr::eq(pg_sys::CurrentMemoryContext, self.owned) {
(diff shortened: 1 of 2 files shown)


workingjubilee (approved):
On one hand: this behavior is itself still at least a touch error prone and arbitrary, and I would prefer users not rely on it much. It smacks of the Bad Kind of implicitness. We should be thinking about how to rework things here so that memory contexts are much more inherently sound to manipulate and that it's clear as to what's going on.

On the other hand: given the current state of `pgx` w/r/t memory contexts, I think it's fine as an improvement.

yrashk:
At the very least I'd love to have it in until we come up with a better approach to memory contexts.

eeeebbbbrrrr:
I’d like to hold this for a 0.6.1 release. Which I can put together on Tuesday.

pgx/src/memcxt.rs
```diff
@@ -174,28 +174,40 @@ pub enum PgMemoryContexts {
...
+    previous: pg_sys::MemoryContext,
+}
 
 impl Drop for OwnedMemoryContext {
     fn drop(&mut self) {
         unsafe {
-            pg_sys::MemoryContextDelete(self.0);
+            // In order to prevent crashes, if we're trying to drop
+            // a context that is current, switch to its predecessor, and then drop it
+            if pg_sys::CurrentMemoryContext == self.owned {
```
eeeebbbbrrrr, line 187: I think this probably should use [`std::ptr::eq()`](https://doc.rust-lang.org/std/ptr/fn.eq.html).  Just a little more explicit that we are comparing pointers.

eeeebbbbrrrr:
> Rebased and addressed your request for `ptr::eq`

This still has all the recent changes from develop.  You might need to force-push your branch?  I hate git

yrashk:
Cherry-picked against 0.6.1-prep

PR #924, opened 2022-12-06
Prepare v0.6.1

eeeebbbbrrrr:
Contains a handful of commits that should go into a 0.6.1 release.

Waiting on #921 and #922 then I think that's it.

Note to self:  DO NOT SQUASH MERGE THIS INTO MASTER.


yrashk:
#921 is no longer relevant, #922 is rebased and feedback addressed.

PR #925, opened 2022-12-06
Fully qualify all pgx symbol references during code generation

eeeebbbbrrrr:
This changes the code we generate (via `pgx-macros` and `pgx-sql-entity-graph`) to fully qualify any pgx symbols that are injected into user code as `::pgx::path::to::Symbol`.

Note that there's a usage of `pgx::pg_sys::submodules::panic::pgx_extern_c_guard` that is **not** qualified with `::pgx`.  This bit of generated code from `rewriter.rs` happens as the result of a `#[pg_guard]` and pgx happens to use `#[pg_guard]` itself.  In fact, you'll see a few places (`hooks.rs`, `callbacks.rs`, etc) that now also `use crate as pgx` so that they can use `#[pg_guard]`.

It also touches `pgx/src/prelude.rs` a bit and updates doctests and such to `use pgx::prelude::*` since that's generally what we want users to do.

The reason fully qualifying these symbols is important is so that it's easier to find them for code auditing purposes (think `git grep "::pgx::"`-style snooping around) and it'll make it a bit easier to develop other "reexport only"-style crates on top of pgx that only allow a subset of pgx to be used.


thomcc (approved):
This looks good to me, although I wouldn't catch if you missed something.

pgx-tests/src/tests/pgx_module_qualification.rs
```diff
@@ -0,0 +1,129 @@
+//! If this test compiles, and assuming it includes a usage of all the things pgx generates code for
+//! then we know that pgx has ::fully::qualified::all::its::pgx::symbols
+
+#[cfg(feature = "pg_test")]
+mod pgx_modqual_tests {
+    // this is the trick (thanks @thomcc for the idea!).  We redefine the pgx module as (essentially)
+    // empty.  This way, if pgx emits any code that uses `pgx::path::to::Thing` instead of `::pgx::path::to::Thing`
+    // we'll fail to compile
+    mod pgx {
+        // the code #[pg_guard] emits isn't qualified with `::ppgx`
```
workingjubilee, line 10:
```suggestion
        // the code #[pg_guard] emits isn't qualified with `::pgx`
```

pgx/src/prelude.rs
```diff
@@ -1,25 +1,36 @@
...
 };
+pub use crate::inoutfuncs::{InOutFuncs, JsonInOutFuncs, PgVarlenaInOutFuncs};
+
+// Trigger support
+pub use crate::trigger_support::{
+    PgTrigger, PgTriggerError, PgTriggerLevel, PgTriggerOperation, PgTriggerWhen,
+};
+
+// Aggregate support
+pub use crate::aggregate::{Aggregate, FinalizeModify, ParallelOption};
```
workingjubilee, line 33: Hm, not everyone is going to need all of these, so we might want to reevaluate this set in the future. However I am not terribly concerned at this exact moment.

workingjubilee (approved):
Cool, very nice!

PR #927, opened 2022-12-07
MemoryContext Drop for a context that has a child that is current

yrashk:
Problem: MemoryContext-related issue

When an owned context is dropped, but one of its children is current, it will fail or will result in the current context being deleted.

Solution: Ensure we're checking if any children are current, too

If they are, change the current context to the previous context of the context being dropped.

---


pgx/src/memcxt.rs
```diff
@@ -181,12 +181,60 @@ pub struct OwnedMemoryContext {
...
+            pub fn pgx_MemoryContextIsValid(context: pg_sys::MemoryContext) -> bool;
+        }
+
+        unsafe fn switch_context(
+            cxt: pg_sys::MemoryContext,
+            previous: pg_sys::MemoryContext,
+        ) -> bool {
+            if ptr::eq(pg_sys::CurrentMemoryContext, cxt) {
+                // If the previous context is no longer valid, switch to TopMemoryContext
+                pg_sys::CurrentMemoryContext = if pgx_MemoryContextIsValid(previous) {
```
workingjubilee, line 194: We want to try to reduce the involvement of the C shim in PGX so we can make it easier to build for various environments, and hopefully even enable cross-compiling in some limited scenarios. This embeds it in every context switch. I am not sure if we can accept this.

pgx/src/memcxt.rs
```diff
@@ -181,12 +181,60 @@ pub struct OwnedMemoryContext {
...
+                if switch_context(child, previous) {
+                    return true;
+                }
+                if switch_children(child, previous) {
+                    return true;
+                }
+                child = (*child).nextchild;
+            }
+            return false;
+        }
```
workingjubilee, line 220: I find this "`bool`-returning mutation function" idiom to be confusing to reason about when used in a recursive situation. I prefer it when conditionals are "pure" in and of themselves.

pgx/src/memcxt.rs
```diff
@@ -181,12 +181,60 @@ pub struct OwnedMemoryContext {
...
+
         unsafe {
-            // In order to prevent crashes, if we're trying to drop
-            // a context that is current, switch to its predecessor, and then drop it
-            if ptr::eq(pg_sys::CurrentMemoryContext, self.owned) {
-                pg_sys::CurrentMemoryContext = self.previous;
+            // In order to prevent failing Postgres assumption and assertion, if we're trying to drop
+            // a context that is current, switch to its predecessor, and then drop it.
+            if !switch_context(self.owned, self.previous) {
```
workingjubilee, line 225: Especially when they involve a negation.

pgx/src/memcxt.rs
```diff
@@ -181,12 +181,60 @@ pub struct OwnedMemoryContext {
...
+                // or (when assertions are disabled) will leave with an invalid current context, which is not
+                // great either.
+                //
+                // So, in order to avoid this, we scan through children to see if any of them are current, and switch
+                // them to the previous context of this context and proceed with the deletion.
+                switch_children(self.owned, self.previous);
             }
+
+            // Now, proceed with the deletion
             pg_sys::MemoryContextDelete(self.owned);
```
workingjubilee, line 238: This has grown very complex and I am somehow still not actually sold on the idea that the deletion we do is correct.

workingjubilee (changes requested):
At this point I am duty-bound to ask:

Is the `OwnedMemoryContext` abstraction actually ever safe or correct to use, or is this simply adding error-handling to an inherently unsafe type that we should not try any longer to save and instead just scrap completely? Can we save it by making it only returned in our API wrapped in `ManuallyDrop<OwnedMemoryContext>`? Technically `into_inner` is safe so the answer is no, perhaps we could use a custom `UnsafeDrop<OwnedMemoryContext>`.

yrashk:
How about I extract a separate PR for this:

> Ensure double set_as_current doesn't obliterate the reference to "previous" context.

If we are to keep any deletion logic, which is a bigger conversation here. Which I think we should. Otherwise, there are cases when the deletion of a current context _will_ happen, and the results will not be nice.

pgx/src/memcxt.rs
yrashk: Let's first establish this. If we agree on handling the deletion of the current context or a parent of a current context, I am more than happy to rework this PR to address the comments above.

workingjubilee:
> Ensure double set_as_current doesn't obliterate the reference to "previous" context.

Yes, that can be merged separately.

pgx/src/memcxt.rs
workingjubilee: This won't be a concern if https://github.com/tcdi/pgx/pull/947 is merged.

yrashk:
> > Ensure double set_as_current doesn't obliterate the reference to "previous" context.
> 
> Yes, that can be merged separately.

I've extracted it to #956

pgx/src/memcxt.rs
yrashk: Very good point. I've simplified this code. Please let me know if this is better.

yrashk:
To address a comment from above. The deletion logic is as follows:

If the context being dropped _or_ a direct or indirect child of it is current, change the current context to the previous context of the context being dropped, _unless_ that previous context is no longer valid. Then, switch to `TopMemoryContext`.

workingjubilee:
But why is arriving at `TopMemoryContext` after deleting a context desirable? Especially if we panic and start unwinding, a situation very likely to have initiated confusing memory context invalidations, I think, we don't want to risk leaking allocations that we may make while unwinding into Top. The program must contain the damaged state.

yrashk:
I suppose we should arrive at some context in the event of previous context not being available. Or should we,  in this case,  produce some failed state? If so,  what do you think should be the way we manifest it?

yrashk:
I was thinking more about this. What if the previous context is invalid (thus we can't switch to it), then we don't delete the context but reset it? We will remain in the context but at least the memory will be released.

Or, we can neither delete nor reset the context in this situation.

workingjubilee:
This is the kind of illogical situation where we really ought to be panicking, except throwing from a destructor is a hilariously bad idea.

pgx/src/memcxt.rs
yrashk: Removed the shim use as #947 was merged.

yrashk:
> This is the kind of illogical situation where we really ought to be panicking, except throwing from a destructor is a hilariously bad idea.

I agree on both ends. Wrapping something as complicated as PostgreSQL is probably never going to be perfect, but I believe we can find reasonable compromises.

workingjubilee:
Will Postgres naturally GC the memory context if we don't delete it?

workingjubilee:
The other one was an "obvious fix" but I feel the remainder of this patch needs Eric's expertise to weigh in on it because this is definitely a "dark art" decision rather than a "mad science" one.

yrashk:
> The other one was an "obvious fix" but I feel the remainder of this patch needs Eric's expertise to weigh in on it because this is definitely a "dark art" decision rather than a "mad science" one.

What "obvious fix"?

yrashk:
> Will Postgres naturally GC the memory context if we don't delete it?

I don't think so but can't say this with 100% certainty yet. Haven't learned all aspects of memory contexts yet.

workingjubilee:
> What "obvious fix"?

I meant the extracted part, mostly.

workingjubilee:
Due to recent issues regarding [soundness in handling memory contexts correctly](https://github.com/pgcentralfoundation/pgrx/issues/1209), I do not believe handling this implicitly and dynamically is correct.

Instead, the appropriate thing to do is to make it so that the way that Rust naturally wants to drop things will appropriately handle recursive drops, and that this error is impossible to actually commit, no matter what sort of lifetime or type sorcery is required to implement such. Otherwise, it becomes far too easy to create some sort of UAF situation because you thought abstractions like PgBox and OwnedMemoryContext might protect you. The current "noisy error" is borderline acceptable.

workingjubilee:
As described in #1337, we are pursuing a new design for memory contexts which encodes the lifetimes of things correctly, rather than trying to handle purely dynamic object lifetimes in a language designed around scope-based region management. In other words, to use more things like SpiClient, and less things like PgMemoryContexts.

Thus we are soft-deprecating the PgMemoryContexts API, including components of it like OwnedMemoryContext. So it is finished for the perspective of development, and no major behavior changes to it will be landed, except strict soundness-based changes that are easy-enough to review and merge.

Thank you for your contributions.

PR #929, opened 2022-12-07
Fix pgx-sql-entity-graph impls of derivable traits.

thomcc:
Also fix the indentation of some really badly formatted code I happened to have to touch while debugging this.

Replaces https://github.com/tcdi/pgx/pull/895. I also have a couple more PRs for pgx-sql-entity-graph that will be on-top of this inbound (but there's no real urgency).


eeeebbbbrrrr (approved):
LGTM

@thomcc, I think you should merge as it might cause conflicts with some of the other stuff you've got in progress

PR #934, opened 2022-12-08
Cursor API should not require &self mut

yrashk:
Problem: cursor API takes &mut self

Solution: make cursor functions take &self

Just like with other functions, it makes them UnwindSafe

This improves just-merged #579

### DIFF
--- a/pgx-tests/src/tests/spi_tests.rs
+++ b/pgx-tests/src/tests/spi_tests.rs
@@ -289,15 +289,12 @@ mod tests {
     }
 
     #[pg_test]
-    fn test_spi_unwind_safe() {
-        Spi::connect(|client| {
-            PgTryBuilder::new(|| {
-                client.update("SELECT 1", None, None);
-                let cursor = client.open_cursor("SELECT 1", None).detach_into_name();
-                client.find_cursor(&cursor);
-            })
-            .execute();
-            Ok(Some(()))
+    fn test_spi_non_mut() {
+        // Ensures update and cursor APIs do not need mutable reference to SpiClient
+        Spi::execute(|client| {
+            client.update("SELECT 1", None, None);
+            let cursor = client.open_cursor("SELECT 1", None).detach_into_name();
+            client.find_cursor(&cursor);
         });
     }
 }
--- a/pgx-tests/src/tests/spi_tests.rs
+++ b/pgx-tests/src/tests/spi_tests.rs
@@ -189,7 +189,7 @@ mod tests {
 
     #[pg_test]
     fn test_cursor() {
-        Spi::execute(|mut client| {
+        Spi::execute(|client| {
             client.update("CREATE TABLE tests.cursor_table (id int)", None, None);
             client.update(
                 "INSERT INTO tests.cursor_table (id) \
@@ -211,7 +211,7 @@ mod tests {
 
     #[pg_test]
     fn test_cursor_by_name() {
-        let cursor_name = Spi::connect(|mut client| {
+        let cursor_name = Spi::connect(|client| {
             client.update("CREATE TABLE tests.cursor_table (id int)", None, None);
             client.update(
                 "INSERT INTO tests.cursor_table (id) \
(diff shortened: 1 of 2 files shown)


workingjubilee:
I'm not as convinced as to why this should be done? The cursor API makes a much stronger statement about how things are being executed and implies more "ownership" of the resulting execution (it apparently is intended to be able to escape the function and last until the end of the transaction).

yrashk:
My reasoning for this was the same as for update: so that it can be executed in PgTryBuilder.

thomcc:
You can just use [`AssertUnwindSafe`](https://doc.rust-lang.org/stable/std/panic/struct.AssertUnwindSafe.html). I think we should remove the use of `UnwindSafe`/`RefUnwindSafe` from most of `pgx` though.

yrashk:
Should we `AssertUnwindSafe` for all uses of `update` and make `update` take `&mut self` again? I think @eeeebbbbrrrr was cool with dropping `&mut self` before. I may have misunderstood his stance on this?

eeeebbbbrrrr:
I’m fine with dropping &mut, but not to make it unwind safe. Simply because it’s unnecessary. 

If there’s work to do around unwind unsafe in general, this ain’t the PR for it.

yrashk:
I am using UnwindSafe as a proxy (for testing purposes; should have been a simple test of being able to use it without being mutable) here, indeed. My concern is that if `update` takes `&self`, why should `_cursor` functions take `&mut self`?.

yrashk:
If desired, I can rewrite the test in this PR to be more straightforward: ensuring we don't need `mut` for `update` or `_cursor` functions. No confusion over UnwindSafe aspect of it.

yrashk:
I've updated this PR as described above.

eeeebbbbrrrr:
The complaint here seems to be that this code won't compile:

```rust
    Spi::connect(|mut client| {
        let cursor = PgTryBuilder::new(|| {
            let cursor = client.open_cursor("SELECT 1", None);
            cursor
        })
        .execute();

        Ok(None)
    });
```

```
error[E0277]: the type `&mut SpiClient<'_>` may not be safely transferred across an unwind boundary
   --> pgx-examples/spi/src/lib.rs:103:40
    |
103 |           let cursor = PgTryBuilder::new(|| {
    |                        ----------------- ^-
    |                        |                 |
    |  ______________________|_________________within this `[closure@pgx-examples/spi/src/lib.rs:103:40: 103:42]`
    | |                      |
    | |                      required by a bound introduced by this call
104 | |             let cursor = client.open_cursor("SELECT 1", None);
105 | |             cursor
106 | |         })
    | |_________^ `&mut SpiClient<'_>` may not be safely transferred across an unwind boundary
    |
    = help: within `[closure@pgx-examples/spi/src/lib.rs:103:40: 103:42]`, the trait `UnwindSafe` is not implemented for `&mut SpiClient<'_>`
    = note: `UnwindSafe` is implemented for `&pgx::SpiClient<'_>`, but not for `&mut pgx::SpiClient<'_>`
note: required because it's used within this closure
   --> pgx-examples/spi/src/lib.rs:103:40
    |
103 |         let cursor = PgTryBuilder::new(|| {
    |                                        ^^
note: required by a bound in `pgx::PgTryBuilder::<'a, R, F>::new`
   --> /Users/e_ridge/_work/pgx/pgx-pg-sys/src/submodules/pg_try.rs:21:32
    |
21  | impl<'a, R, F: FnOnce() -> R + UnwindSafe> PgTryBuilder<'a, R, F> {
    |                                ^^^^^^^^^^ required by this bound in `pgx::PgTryBuilder::<'a, R, F>::new`
```

I don't think this is a good argument in and of itself.  There's little value in a cursor being returned from a PgTryBuilder block.

That said, I would agree there's no need for `SpiClient::open/find_cursor()` to be `&mut` -- they're not changing any internal state.  So on those grounds, I think this is fine to merge.

> it apparently is intended to be able to escape the function and last until the end of the transaction

@workingjubilee A cursor can't outlive the `SpiClient` that created it, unless `cursor.detach_into_name()` is called.  Later, `SpiClient` can retrieve it via `SpiClient::find_cursor()`.

yrashk:
I agree that the `PgTryBuilder` argument is good. That's why I updated the PR to reflect the intent better.

eeeebbbbrrrr (approved):

yrashk:
That being said, I wouldn't say it is an entirely bad argument, either. There's a point in having PgTryBuilder guard a potential error:

```rust
  Spi::execute(|client| {
            PgTryBuilder::new(|| {
                client.open_cursor("SLECT 1", None);
            })
            .catch_others(|_| {})
            .execute();
        });
```

whereas without `PgTryBuilder` we'd be forced to fail with no recourse.

PR #937, opened 2022-12-08
pgx-pg-sys: cshim: Makefile: user AR variable

jaskij:
Instead of calling ar directly, the AR variable should be used. In a cross-compilation environment, ar might not exist or point to the wrong ar binary.


thomcc (approved):
LGTM

workingjubilee (approved):
I was at first concerned as to whether `$AR` is set, but the POSIX spec for Make defines `$AR` as defaulting to `ar`, so this is conformant with POSIX.

workingjubilee:
Just to be safe I'm running this job before merging:
https://github.com/tcdi/pgx/actions/runs/3652798720

PR #938, opened 2022-12-08
:fix: ensure multiple open SpiTupTables can coexist

EdMcBane:
Ensure `SpiTupleTable` refers to `self.size` and not to the static, shared `pg_sys::SPI_processed`, as to avoid interference between multiple open `SpiTupleTable`s.

Fixes https://github.com/tcdi/pgx/issues/936

### DIFF
--- a/pgx-tests/src/tests/spi_tests.rs
+++ b/pgx-tests/src/tests/spi_tests.rs
@@ -298,4 +298,28 @@ mod tests {
             Ok(Some(()))
         });
     }
+
+    #[pg_test]
+    fn test_open_multiple_tuptables() {
+        Spi::execute(|client| {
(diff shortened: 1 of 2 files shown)


yrashk:
I think this is a good catch. 👍

workingjubilee (approved):
Thank you!

workingjubilee:
```c
 * These global variables are part of the API for various SPI functions
 * (a horrible API choice, but it's too late now).  To reduce the risk of
 * interference between different SPI callers, we save and restore them
 * when entering/exiting a SPI nesting level.
 */
uint64		SPI_processed = 0;
```
YIKES! These are **constantly** and racily mutated, too!
Yes, a **very** good catch.

PR #940, opened 2022-12-09
Blocklist c.h

workingjubilee:
This header is used by Postgres, but is not of real use to PGX, as all it does is include various internally-used interfaces. This also reduces about 1MB from the generated symbols.

It seems to work on our tested Linux distros:
https://github.com/tcdi/pgx/actions/runs/3653997966/jobs/6174023739

@thomcc This affects the bindings significantly.

### DIFF
--- a/pgx-pg-sys/build.rs
+++ b/pgx-pg-sys/build.rs
@@ -584,6 +584,7 @@ fn run_bindgen(pg_config: &PgConfig, include_h: &PathBuf) -> eyre::Result<syn::F
         .blocklist_function("(?:pg_|p)v(?:sn?|f)?printf")
         .blocklist_function("appendStringInfoVA")
         .blocklist_file("stdarg.h")
+        .blocklist_file("c.h")
         // these cause cause warnings, errors, or deprecations on some systems,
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr:
A problem with this, I think is that `struct varlena` and `struct nameData`, along with some of their typedefs, are right there in `c.h`.  Because why wouldn't they be?

Not sure we should merge this...

workingjubilee:
Huh. After generating the bindings for all versions, they continue to include
```rust
#[repr(C)]
#[derive(Debug, Default)]
pub struct varlena {
    pub vl_len_: [::std::os::raw::c_char; 4usize],
    pub vl_dat: __IncompleteArrayField<::std::os::raw::c_char>,
}
// more bindings
#[repr(C)]
#[derive(Debug, Copy, Clone)]
pub struct nameData {
    pub data: [::std::os::raw::c_char; 64usize],
}
```

Not sure what to make of this, aside from maybe "blocklisting files doesn't work".

thomcc:
> blocklisting files doesn't work

Indeed. Bindgen doesn't really have good ideas about files.

PR #946, opened 2022-12-10
Port `c-shim.c`'s `pgx_ereport()` function to rust

eeeebbbbrrrr:
This rewrites the C version of `pgx_ereport()` as Rust, as `pgx-pg-sys/src/submodules/panic.rs:do_ereport()`.  Parts of it are under Postgres version feature flags, because of course Postgres' error reporting system is different.

It blocks a small list of "err*" functions from our bindings (in `build.rs`) and as a drive-by, adds support for the "hint" error line, which has a default of None.

All tests pass in all supported Postgres versions.

(If you don't like how I formatted the function, too bad soo sad.  I do!)


eeeebbbbrrrr:
Highly recommend the "split" view for this diff.

PR #948, opened 2022-12-10
Port various c-shim.c `ARR_*` functions to Rust

eeeebbbbrrrr:
Our c-shim has a few functions that expose Postgres `#define` macros for inspecting `ArrayType`.  This ports these to Rust directly inside `pgx/src/array.rs`.

@workingjubilee I'd appreciate your review since you've spent quite a bit of time with Arrays.  I only bothered porting the small set of macros pgx uses, rather than *every* `#define ARR_*` macro in Postgres' `array.h`.  I think these are honest ports.  Tests pass on all PG versions.

I'm doing this and #946 in an effort to add a `no-chsim` feature flag which I believe will make cross-compilation for plrust easier in the short term.

I don't believe we'll ever be able to entirely remove the c-shim, but using it less is probably useful when the tradeoff to porting to Rust doesn't seem too expensive.

### DIFF
--- a/pgx/src/array.rs
+++ b/pgx/src/array.rs
@@ -1,23 +1,104 @@
 use crate::datum::{Array, FromDatum};
-use crate::pg_sys::{self, bits8, ArrayType};
+use crate::pg_sys;
 use bitvec::prelude::*;
 use bitvec::ptr::{bitslice_from_raw_parts_mut, BitPtr, BitPtrError, Mut};
 use core::ptr::{slice_from_raw_parts_mut, NonNull};
 use core::slice;
 
-extern "C" {
-    /// # Safety
-    /// Does a field access, but doesn't deref out of bounds of ArrayType
-    fn pgx_ARR_DATA_PTR(arrayType: *mut ArrayType) -> *mut u8;
-    /// # Safety
-    /// Does a field access, but doesn't deref out of bounds of ArrayType
-    fn pgx_ARR_DIMS(arrayType: *mut ArrayType) -> *mut libc::c_int;
-    /// # Safety
-    /// Must only be used on a "valid" (Postgres-constructed) ArrayType
-    fn pgx_ARR_NELEMS(arrayType: *mut ArrayType) -> i32;
-    /// # Safety
-    /// Does a field access, but doesn't deref out of bounds of ArrayType
-    fn pgx_ARR_NULLBITMAP(arrayType: *mut ArrayType) -> *mut bits8;
+#[allow(non_snake_case)]
+#[inline(always)]
+const unsafe fn TYPEALIGN(alignval: usize, len: usize) -> usize {
+    // #define TYPEALIGN(ALIGNVAL,LEN)  \
+    // (((uintptr_t) (LEN) + ((ALIGNVAL) - 1)) & ~((uintptr_t) ((ALIGNVAL) - 1)))
+    ((len) + ((alignval) - 1)) & !((alignval) - 1)
+}
+
+#[allow(non_snake_case)]
+#[inline(always)]
+const unsafe fn MAXALIGN(len: usize) -> usize {
+    // #define MAXALIGN(LEN) TYPEALIGN(MAXIMUM_ALIGNOF, (LEN))
+    TYPEALIGN(pg_sys::MAXIMUM_ALIGNOF as _, len)
+}
+
+#[allow(non_snake_case)]
+#[inline(always)]
+unsafe fn ARR_NDIM(a: *mut pg_sys::ArrayType) -> usize {
+    // #define ARR_NDIM(a)				((a)->ndim)
+    a.as_ref().unwrap_unchecked().ndim as usize
+}
+
+#[allow(non_snake_case)]
+#[inline(always)]
+unsafe fn ARR_HASNULL(a: *mut pg_sys::ArrayType) -> bool {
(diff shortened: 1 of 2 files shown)


thomcc (approved):

pgx/src/array.rs
```diff
@@ -1,23 +1,104 @@
...
+    // (((uintptr_t) (LEN) + ((ALIGNVAL) - 1)) & ~((uintptr_t) ((ALIGNVAL) - 1)))
+    ((len) + ((alignval) - 1)) & !((alignval) - 1)
+}
+
+#[allow(non_snake_case)]
+#[inline(always)]
+const unsafe fn MAXALIGN(len: usize) -> usize {
+    // #define MAXALIGN(LEN) TYPEALIGN(MAXIMUM_ALIGNOF, (LEN))
+    TYPEALIGN(pg_sys::MAXIMUM_ALIGNOF as _, len)
+}
```
workingjubilee, line 21: somewhat axiomatically, these only manipulate `usize`s and thus are not `unsafe`

pgx/src/array.rs
```diff
@@ -1,23 +1,104 @@
...
-    fn pgx_ARR_DATA_PTR(arrayType: *mut ArrayType) -> *mut u8;
-    /// # Safety
-    /// Does a field access, but doesn't deref out of bounds of ArrayType
-    fn pgx_ARR_DIMS(arrayType: *mut ArrayType) -> *mut libc::c_int;
-    /// # Safety
-    /// Must only be used on a "valid" (Postgres-constructed) ArrayType
-    fn pgx_ARR_NELEMS(arrayType: *mut ArrayType) -> i32;
-    /// # Safety
-    /// Does a field access, but doesn't deref out of bounds of ArrayType
-    fn pgx_ARR_NULLBITMAP(arrayType: *mut ArrayType) -> *mut bits8;
```
workingjubilee, line 20: The safety comments got lost. The ported functions should still document, mostly, whether they're a deref or only an offset computation (and thus whether the field accessed must be properly initialized, or whether we merely must be pointing at a large enough palloc). These are fairly obvious from reading the code but it's useful to have it registered in the description of the function which will be previewed by tooling, no need for an extensive remark.

pgx/src/array.rs
```diff
@@ -1,23 +1,104 @@
...
+    // #define ARR_DIMS(a) \
+    // ((int *) (((char *) (a)) + sizeof(ArrayType)))
+
+    a.cast::<u8>().add(std::mem::size_of::<pg_sys::ArrayType>()).cast::<i32>()
+}
+
+#[allow(non_snake_case)]
+#[inline(always)]
+unsafe fn ARR_NELEMS(a: *mut pg_sys::ArrayType) -> usize {
+    pg_sys::ArrayGetNItems(a.as_ref().unwrap_unchecked().ndim, ARR_DIMS(a)) as usize
```
workingjubilee, line 49: I agree with @thomcc that we shouldn't create a ref here, just unsafe-read the field.

pgx/src/array.rs
```diff
@@ -1,23 +1,104 @@
...
+    } else {
+        std::ptr::null_mut()
+    }
+}
+
+/// The total array header size (in bytes) for an array with the specified
+/// number of dimensions and total number of items.
+#[allow(non_snake_case)]
+#[inline(always)]
+const unsafe fn ARR_OVERHEAD_NONNULLS(ndims: usize) -> usize {
```
workingjubilee, line 74: Not unsafe, also gained an extra character here while porting.
```suggestion
const fn ARR_OVERHEAD_NONULLS(ndims: usize) -> usize {
```

pgx/src/array.rs
```diff
@@ -1,23 +1,104 @@
...
-    fn pgx_ARR_NELEMS(arrayType: *mut ArrayType) -> i32;
-    /// # Safety
-    /// Does a field access, but doesn't deref out of bounds of ArrayType
-    fn pgx_ARR_NULLBITMAP(arrayType: *mut ArrayType) -> *mut bits8;
+#[allow(non_snake_case)]
+#[inline(always)]
+const unsafe fn TYPEALIGN(alignval: usize, len: usize) -> usize {
+    // #define TYPEALIGN(ALIGNVAL,LEN)  \
+    // (((uintptr_t) (LEN) + ((ALIGNVAL) - 1)) & ~((uintptr_t) ((ALIGNVAL) - 1)))
+    ((len) + ((alignval) - 1)) & !((alignval) - 1)
```
workingjubilee, line 13: These don't need parens around the names since they're not being cast anymore, the rest can stay.
```suggestion
    (len + (alignval - 1)) & !(alignval - 1)
```

workingjubilee (changes requested):
This looks overall very good and I believe this will help me massively, thank you! It was something I often was thinking of but stalled on starting. My only requirement for this set is that everything remain `unsafe`-correct: no false positives where the false positive is obvious, and safety comments remain in some form (mostly, to document whether the requirements of a function are "like [`ptr::read`](https://doc.rust-lang.org/std/ptr/fn.read.html)" or "like [`*mut T::add`](https://doc.rust-lang.org/std/primitive.pointer.html#method.add-1)").

workingjubilee:
I'll probably want to poke at code organization afterwards but you're in a much better position right now to write or rewrite the safety comments.

PR #952, opened 2022-12-11
Optimize our String/&str/Vec<u8>/&[u8] IntoDatum conversions

### LINKED ISSUE #941: Investigate if String::into_datum() and &str::into_datum() can be optimised
Calling String::into_datum() and &str::into_datum() seem to take a lot of time. 

Running a trace in a profiler when selecting 300 million rows, String::into_datum() showed up as taking 25-30% of the total query time. For a "simple" conversion to a Datum this seemed quite excessive.

This issue is to investigate if this conversion can be optimised.

eeeebbbbrrrr:
We now directly palloc the `pg_sys::varlena` buffer, set its size, and essentially just memcpy the Rust bytes to the Postgres-allocated memory.

closes #941

### DIFF
--- a/pgx/src/datum/into.rs
+++ b/pgx/src/datum/into.rs
@@ -12,11 +12,8 @@ Use of this source code is governed by the MIT license that can be found in the 
 //! Primitive types can never be null, so we do a direct
 //! cast of the primitive type to pg_sys::Datum
 
-use crate::{
-    pg_sys, rust_byte_slice_to_bytea, rust_regtypein, rust_str_to_text_p, PgBox, PgOid,
-    WhoAllocated,
-};
-use pgx_pg_sys::Oid;
+use crate::{pg_sys, rust_regtypein, PgBox, PgOid, WhoAllocated};
+use pgx_pg_sys::{Datum, Oid};
 
 /// Convert a Rust type into a `pg_sys::Datum`.
 ///
@@ -236,12 +233,7 @@ impl IntoDatum for PgOid {
 impl<'a> IntoDatum for &'a str {
     #[inline]
     fn into_datum(self) -> Option<pg_sys::Datum> {
-        let varlena = rust_str_to_text_p(&self);
-        if varlena.is_null() {
-            None
-        } else {
-            Some(varlena.into_pg().into())
-        }
+        self.as_bytes().into_datum()
     }
 
     fn type_oid() -> u32 {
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr:
See my comment over on #941 about timing numbers:  https://github.com/tcdi/pgx/issues/941#issuecomment-1345613328

pgx/src/datum/into.rs
```diff
@@ -333,11 +325,31 @@ impl<'a> IntoDatum for &'a crate::cstr_core::CStr {
...
     #[inline]
     fn into_datum(self) -> Option<pg_sys::Datum> {
-        let varlena = rust_byte_slice_to_bytea(&self);
-        if varlena.is_null() {
-            None
-        } else {
-            Some(varlena.into_pg().into())
+        let len = pg_sys::VARHDRSZ + self.len();
+        unsafe {
+            // SAFETY:  palloc gives us a valid pointer if if there's not enough memory it'll raise an error
```
BradyBonnette, line 330: `if if`

pgx/src/datum/into.rs
```diff
@@ -333,11 +325,31 @@ impl<'a> IntoDatum for &'a crate::cstr_core::CStr {
...
+                .cast::<pg_sys::varattrib_4b>()
+                .as_mut()
+                .unwrap_unchecked()
+                .va_4byte
+                .as_mut();
+            varattrib_4b.va_header = <usize as TryInto<u32>>::try_into(len)
+                .expect("Rust string too large for a Postgres varlena datum")
+                << 2u32;
+
+            // SAFETY: src and dest pointers are valid and are exactly `self.len()` bytes long
```
workingjubilee, line 345: hidden implicit assumption:
```suggestion
            // SAFETY: src and dest pointers are valid, exactly `self.len()` bytes long,
            // and the `dest` was freshly allocated, thus non-overlapping
```

pgx/src/datum/into.rs
```diff
@@ -333,11 +325,31 @@ impl<'a> IntoDatum for &'a crate::cstr_core::CStr {
...
+            // allocated thanks to our call to `palloc` above
+            let varattrib_4b = varlena
+                .cast::<pg_sys::varattrib_4b>()
+                .as_mut()
+                .unwrap_unchecked()
+                .va_4byte
+                .as_mut();
+            varattrib_4b.va_header = <usize as TryInto<u32>>::try_into(len)
+                .expect("Rust string too large for a Postgres varlena datum")
+                << 2u32;
```
workingjubilee, line 343: Intended resulting bitmask is `0bxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx00`, correct? And the two LSB will be used by Postgres for the varlena tag?

workingjubilee (approved):

pgx/src/datum/into.rs
eeeebbbbrrrr: This is the same as Postgres' `#define SET_VARSIZE_4B` (which we do have in `pgx/src/varlena.rs`), but I wanted that `.expect()` on the size.  I'll at least comment this

PR #954, opened 2022-12-12
A `no-cshim` feature flag

eeeebbbbrrrr:
Add a `no-cshim` feature flag

This adds a `no-cshim` feature flag to `pgx` and pushes it down to `pgx-pg-sys` and `pgx-tests`.

With this turned on we lose low-level access to a few things:

   - `pgx_list*` functions
   - `pgx_planner_rt_fetch()`
   - `pgx_SpinLock*` functions

From `pgx`' perspective, this means a few modules are just not available under `no-cshim`:

   - `hooks` (uses `PgList`)
   - `list` (impl for `PgList` which uses `pgx_list*` functions)
   - `namespace` (uses `PgList`)
   - `spinlock` (uses `pgx_SpinLock*` functions)

This feature flag will hopefully make it easier to cross-compile pgx with only a small bit of compromises.  This feature flag was initially designed with [plrust](https://github.com/tcdi/plrust) user functions in mind, and they won't need access to any of the above.

This PR includes the cherry-picked commits from #946, #947, #948, #949, #950, and #953.  I needed these to ensure the code wasn't referencing any of the c-shim symbols during linking.

If those PRs get merged individually then this should still be okay to merge.  Regardless, this PR should be merged after all of those.  Unless it's somehow preferred we just merge (**not** squash-merge) this commit.

If it's helpful for review, https://github.com/tcdi/pgx/pull/954/commits/42307ead8648d902bf1d7820b7aaf29010da8886 is the commit that adds the new feature flag.


eeeebbbbrrrr:
Worth noting that:

```shell
~/_work/pgx/ $ cargo test --all --features "pg14 no-cshim" --no-default-features
```

passes all tests.  In fact, all the various `pgXX` versions do.

thomcc (approved):
This looks mostly good to me.

I think the `ptr.as_{ref,mut}().unwrap_unchecked().blah` would probably be better as `(*ptr).blah` (and have flagged a couple places but not all of them), but if you strongly disagree, I probably don't feel that strongly about it.

eeeebbbbrrrr:
> This looks mostly good to me.
> 
> I think the `ptr.as_{ref,mut}().unwrap_unchecked().blah` would probably be better as `(*ptr).blah` (and have flagged a couple places but not all of them), but if you strongly disagree, I probably don't feel that strongly about it.

What do you think our merge plan should be here?  Should we regular-merge just this PR, or merge all the others referenced first, then this one?  That'll dictate where I fix this up.

thomcc:
I think merging the referenced ones works, but it's your call. I did end up reviewing these in the worst possible order, but have re-reviewed the others now.

BradyBonnette:
> Perhaps also worth seeing what we can do about having a CI run that enables this feature? @BradyBonnette?

@eeeebbbbrrrr  Do you want this per-commit, or nightly?

eeeebbbbrrrr:
> Do you want this per-commit, or nightly?

Probably as part of "Will it Blend" that runs against develop.  We can hold off on that, of course, until this and all the referenced PRs get merged.

It's probably unlikely we'll often merge stuff that uses one of the modules that are compiled out, which is why I'd like a sanity check.

eeeebbbbrrrr:
> but have re-reviewed the others now.

@thomcc what level of effort did you put into making sure the rust versions of all these things are actually correct relative to the C macros?

eeeebbbbrrrr:
> I think merging the referenced ones works

I'm down with that.

thomcc:
> @thomcc what level of effort did you put into making sure the rust versions of all these things are actually correct relative to the C macros?

I put some effort in, but it wouldn't be impossible that I missed something.

eeeebbbbrrrr:
closing this one as it's now a gigantic PITA to rebase.  Gonna make a new PR off develop with 42307ead8648d902bf1d7820b7aaf29010da8886

PR #958, opened 2022-12-13
Add a default `cshim` feature flag

eeeebbbbrrrr:
This adds a `cshim` feature flag to `pgx`, enabled by default, and pushes it down to `pgx-pg-sys` and `pgx-tests`.

With this turned off we lose low-level access to a few things:

   - `pgx_list*` functions
   - `pgx_planner_rt_fetch()`
   - `pgx_SpinLock*` functions

From `pgx`' perspective, this means a few modules are just not available without `cshim`:

   - `hooks` (uses `PgList`)
   - `list` (impl for `PgList` which uses `pgx_list*` functions)
   - `namespace` (uses `PgList`)
   - `spinlock` (uses `pgx_SpinLock*` functions)

This feature flag will hopefully make it easier to cross-compile pgx with only a small bit of compromises.  This feature flag was initially designed with [plrust](https://github.com/tcdi/plrust) user functions in mind, and they won't need access to any of the above.


pgx-pg-sys/src/lib.rs
```diff
@@ -450,25 +451,21 @@ mod all_versions {
...
-                root: *mut super::PlannerInfo,
-            ) -> *mut super::RangeTblEntry;
-        }
-
-        pgx_planner_rt_fetch(index, root)
+    #[cfg(not(feature = "no-cshim"))]
+    #[pg_guard]
+    extern "C" {
+        #[link_name = "planner_rt_fetch"]
+        pub fn pgx_planner_rt_fetch(
```
eeeebbbbrrrr, line 458: oops!  These names are backwards!

eeeebbbbrrrr:
I think this is ready for merge.

thomcc (approved):
Strictly speaking this should probably be a `cshim` feature enabled by default (enabling a cargo feature is never supposed to remove functionality).

But in practice I don't think this matters much, and it might be a hassle to disable otherwise.

eeeebbbbrrrr:
> Strictly speaking this should probably be a `cshim` feature enabled by default (enabling a cargo feature is never supposed to remove functionality).
> 
> But in practice I don't think this matters much, and it might be a hassle to disable otherwise.

Hmm.  I that that would be incredibly invasive, code wise.  Would that not mean basically putting a `#[cfg]` on nearly everything, or otherwise a major reorg?

thomcc:
I think it would mostly be a matter of changing `#[cfg(not(feature = "no-cshim"))]` to `#[cfg(feature = "cshim")]`, no?

eeeebbbbrrrr:
> #[cfg(feature = "cshim")]

Would it not mean adding this to everything that currently doesn't have this feature flag at all?  Am I missing something.

workingjubilee:
@eeeebbbbrrrr No, it would involve adding the feature flag only to things enabled specifically by the `cshim`. Currently the flag strategy is a "double negative": "if the `no-cshim` feature is NOT enabled, then include this code". Thus it's fine to simply reduce the double negatives to positives. Then, we'd add `default = ["cshim"]` to the Cargo.toml.

eeeebbbbrrrr:
> `default = ["cshim"]`

Do we like that?  Sometimes using cargo across the entire pgx repo (`cargo test --all`) requires setting specific features and also specifying --no-default-features.

I can do it this way, of course, if it's preferred and otherwise idiomatic.

workingjubilee:
Yeah, I don't mind setting features like that. We can prevent settings creep by making sure at least one feature includes all default features that aren't a Postgres version.

eeeebbbbrrrr:
Okay, I'll rearrange this.

workingjubilee (approved):

eeeebbbbrrrr:
I don't think this PR is going to survive in this state after our discord discussions around `cargo test`.  I'm switching gears to work on that, then I'll come back to this.

eeeebbbbrrrr:
This PR is slightly complicated by https://github.com/rust-lang/cargo/issues/4753.

The "solution", if you can call it that, is for our `pgx-tests/Cargo.toml` to **not** have a `default = [ "cshim" ]` feature.  And if we want to run the tests with the cshim turned on, we'll have to specify it explicitly.  

[9646bdf](https://github.com/tcdi/pgx/pull/958/commits/9646bdf6647d5dc34400f15488657e65e17d2ad0) does that, assuming it passes CI.

eeeebbbbrrrr:
I'm content with this now.  Merging.

PR #965, opened 2022-12-15
Customizable serialization of PostgresType

yrashk:
Problem: CBOR is not the most universal answer to everything

However, all `Serialize/Deserialize` PostgresTypes are forced to use with the blanket implementation of `IntoDatum`/`FromDatum`

Solution: extract a `Serializer` and `Deserializer` traits to abstract this away

By default, it uses CBOR and encodes it into a Varlena, replicating existing behavior.

However, this gives a good amount of flexibility to end-user to determine their encoding needs.

I've also removed JSON-encoding/decoding functions from varlena module, as they don't seem to be used anywhere.

---

### DIFF
--- a/pgx-macros/src/lib.rs
+++ b/pgx-macros/src/lib.rs
@@ -781,7 +781,8 @@ fn impl_postgres_type(ast: DeriveInput) -> syn::Result<proc_macro2::TokenStream>
         de.bounds = bounds;
         lt_generics.params.insert(0, GenericParam::Lifetime(de));
         stream.extend(quote! {
-            impl #lt_generics ::pgx::datum::Serializer<'de> for #name #generics { }
+            impl #generics ::pgx::datum::Serializer for #name #generics { }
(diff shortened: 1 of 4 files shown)


Smittyvb:
This is useful for [the pgx extension I work on](https://github.com/timescale/timescaledb-toolkit), since we use a non-CBOR serialization for our aggregate states (Bincode) and it would be great if we were able to use `pg_aggregate` for them.

eeeebbbbrrrr:
hey @yrashk!  this fell off my radar somehow.  I think it looks good in general.  Would you mind rebasing so we can see it without any conflicts?  I'd like to merge if we can

yrashk:
Absolutely! Will look into this shortly.

yrashk:
Here's my first take on rebasing. It still requires a bit more of a review as things have changed and I may have missed something. The test passes.

yrashk:
@eeeebbbbrrrr did you have a chance to look at this updated PR yet?

eeeebbbbrrrr:
I have not. :( I’ll tackle it next week. Thanks for the ping.

PR #966, opened 2022-12-15
Adds unlocked cargo runs to Will It Blend CI

BradyBonnette:
This migrates the original "will-it-blend" mechanics into the new full nightly suite. So far, the difference between new vs. old "Will it blend" is that the old version would build everything without `--locked` to check for semver issues. Everything in the new "will it blend" uses `--locked` for everything since that's what our documentation dictates.

With this update, Will It Blend will run both:  `--locked` and "unlocked" builds/tests. It also removes the old version since its cron specifier was not correct in the first place.


BradyBonnette:
Note that these changes won't take effect until they're merged into `master`. Once this is squished, it'll be an easily cherry-pickable commit.

eeeebbbbrrrr (approved):

PR #967, opened 2022-12-15
Teach `pgx-tests/src/framework.rs` to detect `cargo test` feature arguments

eeeebbbbrrrr:
Running a plain `cargo test --features "a b c"` never properly worked with pgx' test framework as none of the "feature"-related arguments were being passed down into pgx' test framework.  Now they are.

In the case of a `cargo test --all --features "a b c"`-style execution, `cargo-pgx` will elide passing features to a workspace member that doesn't have that feature.

This necessitated reworking how **all** of the `cargo-pgx` commands detect which Postgres version feature flag to use, along with the general set of features to turn on.  The search order is now:

### schema/run/test/connect

- direct `cargo-pgx` argument
- `--features "pgXX"` flag
- `default = [ "pgXX" ]` from the extension's Cargo.toml

### install/package

- created from a direct `--pg_config` argument
- created from the `pg_config` on the system `$PATH`

### start/stop/status

- direct `cargo-pgx` argument
- `default = [ "pgXX" ]` from the extension's Cargo.toml

As a few drive-bys, teach cargo-pgx to output the type of `pgXX` feature flag it found and the full path to the `pg_config` binary it's using for that operation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -42,6 +42,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.66"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "216261ddc8289130e551ddcd5ce8a064710c0d064a4d2895c67151c92b5443f6"
+
+[[package]]
 name = "arrays"
 version = "0.1.0"
 dependencies = [
@@ -1248,6 +1254,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ntapi"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bc51db7b362b205941f71232e56c625156eb9a929f8cf74a428fd5bc094a4afc"
+dependencies = [
+ "winapi",
+]
+
+[[package]]
 name = "nu-ansi-term"
 version = "0.46.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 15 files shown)


eeeebbbbrrrr:
There's quite a bit of exact and similar code duplication that's been removed in this PR across all the various commands.  All the decision making has been moved into `crate::manifest::{get_package_manifest, pg_config_and_version}`, instead of being duplicated-ish across every-ish command.

workingjubilee:
```yaml
    - name: create new sample extension
      run: cd /tmp/ && cargo pgx new sample

    # hack Cargo.toml to use this version of pgx from github
    - name: hack Cargo.toml
      run: |
       echo "[patch.crates-io]" >> /tmp/sample/Cargo.toml
       echo "pgx        = { path = \"${GITHUB_WORKSPACE}/pgx\"        }" >> /tmp/sample/Cargo.toml
       echo "pgx-macros = { path = \"${GITHUB_WORKSPACE}/pgx-macros\" }" >> /tmp/sample/Cargo.toml
       echo "pgx-tests  = { path = \"${GITHUB_WORKSPACE}/pgx-tests\"  }" >> /tmp/sample/Cargo.toml
    - name: show Cargo.toml
      run: cat /tmp/sample/Cargo.toml

    - name: Test sample for ${{ matrix.version }}
      run: cd /tmp/sample && cargo pgx test pg$PG_VER
```
Need to either adjust the templates used by `cargo pgx new` or adjust this CI script.

eeeebbbbrrrr:
> Need to either adjust the templates used by cargo pgx new or adjust this CI script.

It's weird.  This doesn't happen locally.  The following works just fine:

```shell
cd /tmp
cargo pgx new sample
cd sample
cargo pgx test pg15
```

There must be a bug somewhere in the PR...

eeeebbbbrrrr:
> There must be a bug somewhere in the PR...

Oh!  The "cargo-pgx" for this test was **only** configured with `pg15`, so it doesn't know that the default, `pg13`, is a postgres version feature flag that needs to be removed from the set, which happens here:  

https://github.com/tcdi/pgx/pull/967/files#diff-501d8223e92f9f60da2c06eb375f5284d6a76718e6e3ccae747df3ae31871c3bR87

Let me noodle this...

eeeebbbbrrrr:
I'm pleased with this now and it'll let me finish up #958 so I'm merging.

It's dangerous to leave me to my own devices over a long holiday!

PR #969, opened 2022-12-18
Build upon #902:  Spi Error Handling

eeeebbbbrrrr:
Building upon @yrashk's work in https://github.com/tcdi/pgx/pull/902, this does a few small things, which caused a bunch of files (mostly tests and examples) to be updated.

First off, the various Spi functions like `Spi::get_one()` go back to returning `-> Result<Option<T>, Error>`.  This might seem counter to my comment at https://github.com/tcdi/pgx/pull/902#pullrequestreview-1199966423, but there's an important thing Spi **must** support:  Any Datum **can be NULL**, which means we must return `Option<T>`.

So, the thing that makes this better is we also now have `impl<T, E> IntoDatum for Result<T, E> where T: IntoDatum, E: std::error::Error`.  This means any `#[pg_extern]`-style function can now return a `Result`, and if that Result represents an error, it is raised as a Postgres ERROR.

This is a big win for `#[pg_extern]`-style functions in general, and makes working with an Spi interface that returns `Result<Option<T>, Error>` pretty fluent as now you can just directly return the result from `Spi::get_XXX()` or you can use the `?` operator to nicely unwrap the error and then deal with the returned, possibly NULL, Datum.

Doing this, and updating the Spi API caused all sorts of code to be touched in the tests and examples.  As a result, there's (hacky) support for `#[pg_test]` functions being able to return `-> Result<(), pgx::spi::Error>`, which makes using Spi in the tests nearly the same, but now with a `?` operator and a `return Ok(())` at the end of such functions.

The examples that use SPI needed to be touched as well.  This is a pretty big breaking API change, afterall.

/cc @yrashk


pgx-tests/src/tests/spi_tests.rs
```diff
@@ -306,19 +313,39 @@ mod tests {
             assert!(!a.is_empty());
             assert_eq!(1, a.len());
             assert!(a.get_heap_tuple().is_some());
-            assert_eq!(Some(1), a.get_datum(1));
+            assert_eq!(1, a.get_datum::<i32>(1).expect("a.get_datum::<i32>(1) failed").unwrap());
         });
     }
 
     #[pg_test]
+    #[ignore = "come back to this test"]
```
eeeebbbbrrrr, line 321: I don't understand why this test fails.  I guess I don't understand what it's testing

pgx-tests/src/tests/fcinfo_tests.rs
```diff
@@ -128,46 +126,13 @@ fn fcinfo_not_named_no_arg(fcinfo: pg_sys::FunctionCallInfo) -> i32 {
     todo!()
 }
 
-#[derive(PostgresType, Serialize, Deserialize, Debug, PartialEq)]
-#[inoutfuncs]
-pub struct NullStrict {}
```
eeeebbbbrrrr, line 133: I think this was deleted for reasons that it no longer needs to be.  I'll be bringing it back...

eeeebbbbrrrr:
I’ve marked this as draft because I’m probably not done yet. 

My major complaint now is with Spi::connect() and Spi::execute().  I think we only need Spi::connect() and may make that decision in this PR. 

With that, I don’t like that Spi::connect()’s E type is generic. I think it needs to just be spi::Error.

pgx-tests/src/tests/spi_tests.rs
yrashk: It's testing that we longer have a bug in that we always use `SPI_processed` (#938)

eeeebbbbrrrr:
> If the Datum being returned from Postgres is NULL, and the requested type on Rust's side is not Option, we make this an error

As far as SQL is concerned, it's not an error.  NULL is a perfectly fine Datum value for **any** type and is a fundamental part of the SQL type system.  pgx's "generalized" SPI interface shouldn't decide that a valid Datum value for a given type is an error. In fact, Postgres' SPI api, as flawed as it is, doesn't consider NULL to be an ERROR either -- the various functions like `SPI_getbinval` take a `bool *is_null` out parameter, so that the caller can decide for themself.

If the user, in whatever their use of SPI, decides that NULL isn't valid for some situation, they should encode that decision in their code.  It's a business logic or data domain problem, not a fundamental error from within pgx.  We wouldn't, for example, decide that "42" isn't a valid integer either -- that kind of decision is up to the user.

Sadly, getting a Datum value out of SPI is a pure runtime operation.  We can't lean on the compiler (plus help from Postgres itself to ensure invariants) like we can with, say, the `STRICT` keyword and `#[pg_extern]` where we **know** that if no argument is `Option<T>` then Postgres will never call the function with a NULL argument.

(As an aside, this leads me down the thought process that `IntoDatum::is_compatible_with()` isn't quite robust enough either and we ought to be using Postgres' CAST system for that, and just transparently provide conversions from type A to type B during `FromDatum::from_datum()`, but that can be saved for another day)

> Why have you reverted your opinion on this?

I asked how we could eliminate the double unwrapping to make the PR easier to follow.  I didn't have any particular solution or implementation detail in mind.  SPI viewing a NULL value as an error is an incorrect worldview for pgx, so it makes sense SPI still work with `Option<T>`.

I fully realize the convenience of `impl<T> FromDatum for Option<T>`, which is why that impl is still there (although it's not necessary anymore).  It's just not the proper API for SPI to push that decision to the programmer at compile time.

> I don't suppose this comes from the return of Option

It came out me still wanting to "eliminate the double unwrapping".  Being able to return a Result from the places a user is likely to use SPI (namely `#[pg_extern]`-style functions) seemed to be the next right thing knowing `Option<T>` was going to stick around.  I don't need to extol the awesomeness of Rust's Result type and the `?` operator here!

eeeebbbbrrrr:
> What if we interact with SPI and I/O (reading files, for example)? How do we return an I/O error?

Yeah, I haven't really thought through this particular problem.  I plan on sitting down to write some code with SPI and see what feels right.

It might be that we do want it to be a `Box<dyn std::error::Error>`.  Or perhaps have some "OtherError" variant on `spi::Error` and provide all the proper From impls to get one of those through the `?` operator.  I dunno.  I don't have any specific implementation details yet.

I just know that in updating all the tests that use Spi::connect(), returning `Ok::<_, pgx::spi::Error>(())` is just really awkward.

eeeebbbbrrrr:
> As far as SQL is concerned, it's not an error. NULL is a perfectly fine Datum value for any type and is a fundamental part of the SQL type system

Just to expand on this, @workingjubilee and I have discussed ideas about representing a Datum, instead of `Option<T>` as something more strictly typed.  Something like:

```rust
enum NullableDatum<T> {
   Null,
   Concrete(T)
}
```

It kinda still looks like an `Option<T>` but it would more clearly convey the true nature of a Datum.  Postgres internals actually have a similar thing now too.  I think they probably invented it for other reasons, but nonetheless, I wouldn't be surprised so see their codebase slowly migrate to using it everywhere instead of a pair of `(bool is_null, uintptr_t datum)` that they mostly pass around now.

Right now pgx represents datums as `Option<T>` everywhere, so we should also stick with that in Spi as, if we do go down the road of inventing something like a "NullableDatum enum", then the user-facing API breakage will be limited as I'd expect "NullableDatum" to have a similar set of methods to Option.

What I don't want, and this isn't to say that #902 did this because it didn't, is for NULL to be considered an error like the "postgres" dbi driver crate does (https://docs.rs/postgres/latest/postgres/row/struct.Row.html#method.get).  Hell, it flat out panics if you ask for a concrete value that happens to be null!  I know why they did their API that way -- because it's much more convenient for the user, but it's just incorrect as it relates to SQL.  

An SQL integer, for example can be one of `[NULL, -2147483648.. 2147483647]`.  That's the valid set of values.  pgx needs to be able to properly represent any of those.  Right now, Option is what we can use.

yrashk:
Option itself, or an Option-like type for better readability, doesn't really matter. I'm fine with either.

I think the user should be able to decide if null is an erroneous value to them. That's why they should use type signatures to indicate what's considered correct and what is not.

What am I missing in this picture?

eeeebbbbrrrr:
> What am I missing in this picture?

Maybe this:  https://en.wikipedia.org/wiki/Null_(SQL) ?

Among many other good explanations, including the paragraph on "Closed-world assumption", is:

> SQL null is a marker, not a value. This usage is quite different from most programming languages, where [null value](https://en.wikipedia.org/wiki/Null_pointer) of a reference means it is not pointing to any [object](https://en.wikipedia.org/wiki/Object_(computer_science)).

`Option<T>` is always a perfect representation of such a thing.

Above I may have been a little too loose in saying "NULL Datum value", but the point remains the same:  NULL is a fine and acceptable thing in SQL.  It's not an error condition as it relates to SQL.

> I think the user should be able to decide if null is an erroneous value to them.

I absolutely agree, which is why returning `Option<T>` allows the user to `.unwrap()`, `.expect()`, `.ok_or()`,`match {}`, etc the returned value.  It's not pgx' place to decide any valid Datum representation, which includes NULL, is an error.  In SQL, NULL means "missing" or "unknown", not a "state or condition of being wrong".

From a technical perspective, if a missing/unknown value is an error situation for the user, they need to encode that themselves.  Unfortunately we can't involve the compiler to help us see sql check constraints behind arbitrary SQL queries, so we shouldn't instead use the compiler to mis-represent a fundamental aspect of SQL.

yrashk:
> > What am I missing in this picture?
> 
> Maybe this: https://en.wikipedia.org/wiki/Null_(SQL) ?

I am aware of what is Null (however, the definition at the top is dubious -- `SELECT NULL::text` doesn't mean the value does not exist in the database )

> 
> Among many other good explanations, including the paragraph on "Closed-world assumption", is:
> 
> > SQL null is a marker, not a value. This usage is quite different from most programming languages, where [null value](https://en.wikipedia.org/wiki/Null_pointer) of a reference means it is not pointing to any [object](https://en.wikipedia.org/wiki/Object_(computer_science)).
> 
> `Option<T>` is always a perfect representation of such a thing.

Indeed it is.

> 
> Above I may have been a little too loose in saying "NULL Datum value", but the point remains the same: NULL is a fine and acceptable thing in SQL. It's not an error condition as it relates to SQL.

That's right. However, there are cases where it is possible to indicate that null is _not_ accepted, and they do lead to errors.

> 
> > I think the user should be able to decide if null is an erroneous value to them.
> 
> I absolutely agree, which is why returning `Option<T>` allows the user to `.unwrap()`, `.expect()`, `.ok_or()`,`match {}`, etc the returned value. It's not pgx' place to decide any valid Datum representation, which includes NULL, is an error. In SQL, NULL means "missing" or "unknown", not a "state or condition of being wrong".

That's just misleading to assume they may always get a null value to force them to unwrap in some way. Here's a contrived examples (for brevity):

```
select::<i32>('SELECT 1')
```

should we be forced to unwrap this because _it may be null_? I don't think it may become null. Therefore, the way things are done in #902 is, strictly speaking, better as it doesn't make assumptions about the nullability of the value. If one expects that the value may be `null`, it has to be encoded in the type signature

``
select::<Option<i32>>('SELECT foo()')
```

or 

```
select::<Nullable<i32>>('SELECT foo()`)
```

This way, at no point we need to match or unwrap when there's clarity with regard to the type's nullability.


> 
> From a technical perspective, if a missing/unknown value is an error situation for the user, they need to encode that themselves. Unfortunately we can't involve the compiler to help us see sql check constraints behind arbitrary SQL queries, so we shouldn't instead use the compiler to mis-represent a fundamental aspect of SQL.

You are making an assumption one way or another. You're either forced to unwrap always or specify the correct type in the first place. In reality, this means that a lot of signatures will contain `Option` or `Nullable` and _that_ is **perfectly fine**.

**Comment by yrashk** [comment]
> > What if we interact with SPI and I/O (reading files, for example)? How do we return an I/O error?
> 
> Yeah, I haven't really thought through this particular problem. I plan on sitting down to write some code with SPI and see what feels right.

> 
> It might be that we do want it to be a `Box<dyn std::error::Error>`. Or perhaps have some "OtherError" variant on `spi::Error` and provide all the proper From impls to get one of those through the `?` operator. I dunno. I don't have any specific implementation details yet.

Ugh, Boxes are a kludge.

> 
> I just know that in updating all the tests that use Spi::connect(), returning `Ok::<_, pgx::spi::Error>(())` is really awkward.

If anything, this above shows holes in the `Spi::connect` design you insisted on. It alone is responsible for the need to even have this discussion.

If we simply had a `Spi::client` or `Spi::connect` that returns a client that `SPI_finish` on Drop. And yes, I remember your concern about clearer delineation for when we're connected and are in a memory context for that particular connection. We can discuss that to make it better. But it's fairly clear to me that closures and errors are not a good way to handle this and it shows.

**Comment by eeeebbbbrrrr** [comment]
> should we be forced to unwrap this because it may be null?

Yes.  Should pgx force the user to deal with possible `Error::DatumError` conversion problems?  Yes.  The difference is that NULL is simply not an error condition in SQL and pgx turning it into one, even at the user's request, misrepresents the SQL type from which the value originates.

> If one expects that the value may be null, it has to be encoded in the type signature

One should expect that any datum may be null.  As such, pgx must encode that in the type signature for them.  The meaning of null is context dependent.  pgx should not allow the user to write type signatures that makes a fundamental part of SQL values into an error. It is not the same as actual error conditions pgx might encounter, such as "cannot convert 'hello world' to an i32" or "some internal pointer is invalid".  They're entirely separate concerns.

> You're either forced to unwrap always or specify the correct type in the first place.

We appear to have a disagreement on what "correct type" means.  The correct type is `Option<T>`.

The Postgres Java JDBC driver's `ResultSet.getString()` function, for example, doesn't `throw new NullPointerException()` if the requested column is null.  This is essentially what you're advocating for.  

Take this:

```rust
let i:i32 = Spi::get_one("SELECT non_null_i32()")?;  // Oh no! this actually returned NULL
```

With what you're proposing, the user would get a runtime `Err(spi::Error:NullValue)` (or whatever), and need to go and update their code to:

```rust
let i:i32 = match Spi::get_one("SELECT non_null_i32()") {
    Ok(i) => i, 
    Err(spi::Error::NullValue) => i32::default(),
    Err(e) => panic!("{}", e)
};
```

Whereas what I'm proposing gets us to:

```rust
let i:i32 = Spi::get_one("SELECT non_null_i32()")?.unwrap_or_default();
```

With the latter form, the user can wave away the actual errors they likely can't do anything about and then turn the NULL case into something that's relevant to their application.

And honestly, I don't see a problem with this:

```rust
// SAFETY:  ain't no way Postgres will ever return NULL here
let i:i32 = unsafe { Spi::get_one("SELECT 1")?.unwrap_unchecked() };
```

The nullability of a Datum can only be determined at runtime.  We shouldn't imply that the compiler can check it.  In SQL, all types *can* have a NULL variant.

If there were ways we could do some fancy compile-time sql type checking, constraint checking, or whatever, to ensure the nullability of an SQL value, man, I'd be all over that!

yrashk:
> With what you're proposing, the user would get a runtime `Err(spi::Error:NullValue)` (or whatever), and need to go and update their code to:
> 
> ```rust
> let i:i32 = match Spi::get_one("SELECT non_null_i32()") {
>     Ok(i) => i, 
>     Err(spi::Error::NullValue) => i32::default(),
>     Err(e) => panic!("{}", e)
> };
> ```

That is not what I am proposing at all. I am proposing they spell out what they actually expect:

```
let i: i32 = Spi::get_one("SELECT strictly_non_null_i32()")?; // ? for error
let i: Option<i32> = Spi::get_one("SELECT i_am_not_sure_if_it_may_be_a_null_i32())?; 
// or, for readability
let i: Nullable<i32> = Spi::get_one("SELECT i_am_not_sure_if_it_may_be_a_null_i32())?; 
```

Now if they are SURE that it is a non-null value, and null does appear, yes it will fail with an error, as it should (because they specified they did not expect this)

yrashk:
> One should expect that any datum may be null. As such, pgx must encode that in the type signature for them. The meaning of null is context dependent. pgx should not allow the user to write type signatures that makes a fundamental part of SQL values into an error. It is not the same as actual error conditions pgx might encounter, such as "cannot convert 'hello world' to an i32" or "some internal pointer is invalid". They're entirely separate concerns.

I think this is where our disconnect is. On the database side, NULL can be anywhere and is not part of the type system. In Rust, it effectively is, and we can and should use it – at least, this is how I think about it.

yrashk:
> And honestly, I don't see a problem with this:
> 
> ```rust
> // SAFETY:  ain't no way Postgres will ever return NULL here
> let i:i32 = unsafe { Spi::get_one("SELECT 1")?.unwrap_unchecked() };
> ```

I hope you are not serious. That's not a good way to get around an assumption that "anything can be null."

If you are not sure: use `Option/Nullable<T>`
If you are sure: use `T`

Seems pretty clear to me. In the end, you're getting your Option-like interface where you need it, and none if you don't need it (where you're sure). And even if you were wrongly sure, an error is the worst thing that happens to you. And we have to handle the `Error` side of the `Result` anyway.

yrashk:
> I just know that in updating all the tests that use Spi::connect(), returning `Ok::<_, pgx::spi::Error>(())` is really awkward.

FWIW, this is easily avoidable:

```rust
 client.open_cursor("SELECT * FROM tests.cursor_table", None).map(|mut cursor| {
                assert_eq!(sum_all(cursor.fetch(3)), 1 + 2 + 3);
                cursor.detach_into_name()
            })
```

Sure, the `?`-notation makes code nest less, but I guess one has to pay the price somewhere. Since the error type of the result is not known and `?` is actually using `From`, we either need a type annotation or map the result type.

Then again, this won't be necessary at all if we let go of `Spi::connect` as a function that takes a closure.

pgx-tests/src/tests/spi_tests.rs
EdMcBane: My bad, it was definitely missing a comment.
I was thinking something along these lines:

```rust
    #[pg_test]
    fn test_open_multiple_tuptables() {
        // Regression test to ensure a new `SpiTupTable` instance does not override the
        // effective length of an already open one due to misuse of Spi statics
```

```rust
    #[pg_test]
    #[ignore = "come back to this test"]
    fn test_open_multiple_tuptables_rev() {
        // Regression test to ensure a new `SpiTupTable` instance does not override the
        // effective length of an already open one.
        // Same as `test_open_multiple_tuptables`, but with the second tuptable being empty  
```

eeeebbbbrrrr:
> In the end, you're getting your Option-like interface where you need it, and none if you don't need it (where you're sure).

pgx represents an instantiated Datum as an `Option<T>`.  It has to, because any Datum could be null.  

This is built into `FromDatum::from_datum/try_from_datum` (via the `is_null` argument we get from Postgres), `IntoDatum::into_datum()`, the various `fcinfo::get_arg()` functions, various `PgHeapTuple` methods, and until #902, SPI.

As I've said before, pgx should not allow the programmer to specify expected types that might not be a valid representation of the Datum for any cases where Postgres itself would do a successful conversion.

The only correct representation of an "i32" that comes from a Datum, is `Option<i32>`.  That's it.

> And even if you were wrongly sure, an error is the worst thing that happens to you.

That's a pretty terrible thing to happen because that means we've designed an API that gives the programmer no compiler help along the way to make sure they're handling possibly unexpected values when they should.  And "when the should" is when they're writing their code.

```rust
let i: i32 = Spi::get_one("SELECT strictly_non_null_i32()")?; // ? for error
```

This code should not compile.  It is not correct because an "i32 Datum" cannot be represented as an `i32`.  It must represented as an `Option<i32>`. SQL/Postgres provide ZERO guarantees that something "will never be NULL".  Zero.  And if it is NULL, that is not a `Result::Err()` -- it's just the other form of the value.

It may be undesired by the programmer.  It may be unexpected by the programmer.  It may be impossible to the programmer to imagine.  None of that should change pgx' knowledge that any Datum can be NULL. 

All of this is the indicator that pgx must build its API signatures so that the programmer has to deal with reality order for their code **to even compile**.

--- 

> I hope you are not serious (about gross unsafe block)

If the programmer thinks they're smarter than pgx, they're free to write that.  I'd never encourage it, of course.  I'm not a monster.  But it's a fine way of the programmer making their assumptions clear without pgx silently injecting an Error they didn't handle when that assumption turns out to be wrong.

yrashk:
It seems we're at an impasse here. You have merge/commit rights, and I don't. So, I can't win the argument! :) 

We see it differently. Error for a null where null is absolutely not expected but the program is an absolutely valid and correct API choice, period (if we're speaking in absolutes like "The only correct representation of an "i32" that comes from a Datum, is Option<i32>. That's it.")

eeeebbbbrrrr:
> Error for a null where null is absolutely not expected but the programmer is an absolutely valid and correct API choice, period

The good news is that there's nothing about always working with `Option<T>` that prevents the programmer from making that explicit in their code.

The bad news (or the other good news) is that pgx isn't going to help them do that by creating a runtime situation that they may not have realized they need to account for.

---

I'm going to take a look at the `spi::Error` stuff in the tests and `Spi::connect()` now.  It may be just fine or it may need some usability tweaks.  Not sure yet.

yrashk:
That's not good news. That's just insertion of nullability whether you want/need it or not.

I'm going to stop arguing about this. You have your own vision for pgx. I can't do much about that if you have made your mind up.

pgx-tests/src/tests/spi_tests.rs
eeeebbbbrrrr: Thanks.  I've pushed a few minor changes to these two tests in bc27ff31cd253855163fc066228214eda0bb912a.

eeeebbbbrrrr: ha!  and 9074043891d63cb2d660d650af79dcdcfa03e394.

eeeebbbbrrrr:
I think this is now ready for merge.  I'm gonna let it sit overnight and mull things over.  It's as big as it is because of all the updates and cleanups to the `pgx-tests` crate.

eeeebbbbrrrr:
One last change I *think* I'm going to make in this PR, since it'll update a bunch of tests too, is get rid of `Spi::execute()`.  Now that `Spi::connect()` can return anything, it can also return nothing, so now there's no difference between the two functions.  That'll take care of my comment from up above (https://github.com/tcdi/pgx/pull/969#issuecomment-1356764818).

eeeebbbbrrrr:
I'm gonna merge this now.  @yrashk, I plan on looking at #885 and #963 next, in that order.  I suspect once I click the merge button here those will need to be rebased.  So... heads up!

PR #970, opened 2022-12-19
add doc to pg_test module in template

ChuckHend:
Adding some doc to hopefully make it more clear about what is required for testing pgx extensions. As far as I can tell, tests won't run unless there's a module `pg_test` with BOTH `setup` and `postgresql_conf_options` defined.

### DIFF
--- a/cargo-pgx/src/templates/lib_rs
+++ b/cargo-pgx/src/templates/lib_rs
@@ -19,10 +19,10 @@ mod tests {{
 
 }}
 
+/// This module is required by `cargo pgx test` invocations. 
+/// It must be visible at the root of your extension crate.
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr (changes requested):
I might put this above the `pg_test` module itself as `///` doc comments.  Something like, "This module is required by `cargo pgx test` invocations.  It must be visible at the root of your extension crate"

eeeebbbbrrrr (approved):
Thanks.  I'll merge after CI passes.  Obviously, this ain't gonna break in CI, but I'm doing it so that any future PRs from you will automatically run.

eeeebbbbrrrr:
welcome to the contributor's club, @ChuckHend.

PR #973, opened 2022-12-20
Change bounds on `IntoDatum for Result<T, E>` such that `E: Any + Display`

eeeebbbbrrrr:
Broaden the bounds on our support for converting a Result into a datum a bit so that we can detect if the `Err` variant contains one of our `ErrorReport` types.  If it does, that's the thing we'll raise, at the `PgLogLevels::ERROR` level.  Otherwise we'll just raise a `panic!()` with the to_string() of the error as before.

This enables users to return `Result::Err` variants that contain a more rich set of error reporting information such as a specific `PgSqlErrorCode` and optionally "hint" and "detail" lines.

### DIFF
--- a/pgx-pg-sys/src/submodules/panic.rs
+++ b/pgx-pg-sys/src/submodules/panic.rs
@@ -79,6 +79,19 @@ pub struct ErrorReport {
     pub(crate) location: ErrorReportLocation,
 }
 
+impl Display for ErrorReport {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}: {}", self.sqlerrcode, self.message)?;
+        if let Some(hint) = &self.hint {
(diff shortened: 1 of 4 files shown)


eeeebbbbrrrr:
Example:

```rust
    #[pg_test(error = "raised custom ereport")]
    fn test_custom_ereport() -> Result<(), ErrorReport> {
        Err(ErrorReport::new(
            PgSqlErrorCode::ERRCODE_NO_DATA,
            "raised custom ereport",
            function_name!(),
        ))
    }
```

```sql
[v13.9][88869] pgx_tests=# \errverbose
ERROR:  02000: raised custom ereport
LOCATION:  pgx_tests::tests::result_tests::tests::test_custom_ereport, result_tests.rs:86
```

Note that `02000` is in fact `ERRCODE_NO_DATA`

PR #975, opened 2022-12-21
Support returning `Result<TableIterator/SetOfIterator, E>`

eeeebbbbrrrr:
Add the ability to return a Result of a TableIterator or SetOfIterator.  This just helps to make things more consistent with the other cases where we can now return a Result of something.

Also introduce a new trait, `pg_sys::panic::ErrorReportable` that when its `.report()` method is called will return a value or raise an error depending on the context.  We provide a default implementation for `Result<T, E: Any+Display>` that understands how to downcast the error to look for an `ErrorReport` instance.  This moves that code out of the `impl IntoDatum for Result` code.

### DIFF
--- a/pgx-pg-sys/src/submodules/panic.rs
+++ b/pgx-pg-sys/src/submodules/panic.rs
@@ -22,6 +22,42 @@ use crate::elog::PgLogLevel;
 use crate::errcodes::PgSqlErrorCode;
 use crate::{pfree, AsPgCStr, MemoryContextSwitchTo};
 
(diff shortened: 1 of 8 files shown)


eeeebbbbrrrr:
A small observation:  the sql-entity-graph code has gotten too complex.  Some day one of us needs to sit down and tighten it up.  Parsing `syn` trees is just terrible, I know, but I'm sure there's quite a bit we can do to make our code more manageable.  I don't feel like it should have taken 2.5 hours to put this little PR together.

PR #977, opened 2022-12-22
Proper error handling for SPI

eeeebbbbrrrr:
The long story short is that pretty much everything returns a Result now: the getter methods on PgTupleTable and SpiHeapTupleData go through try_from_datum(), and many methods like `Spi::run`, `SpiClient::update`, `SpiClient::select` now return a `Result<R, spi::Error>`.

~~`Spi::connect()` now requires the user to return a `Result<R, spi::Error>` because, believe it or not, `pg_sys::SPI_connect()` is documented as being able to return an error (even tho in practice it doesn't).~~

Also do some cleanups to the `spi::Error` variants, make a few internal pointers we track `NonNull`, added some comments here and there, and updated all the tests/examples accordingly.

/cc @EdMcBane & @yrashk -- y'all have had a lot of work and interest in Spi here recently, so I'd appreciate your eyes.

There's no functional changes here, just user-facing API changes (hence so many tests being touched) along with internal cleanup. 

~~Probably the most controversial change is that `Spi::connect()` now returns `-> Result<R, spi::Error>`.  It really needs to as `pg_sys::SPI_connect()` is documented as being able to return an error (even tho it doesn't!).~~  
UPDATE:  `Spi::connect()` makes the executive decision to `panic!()` if `pg_sys::SPI_connect()` doesn't return a `SpiOk`.  As such, it is now capable of simply returning `-> R` and everyone can sleep at night.

Prior to this PR spi.rs was silently transforming non SpiOK returns into a panic!, and now those are pushed forward to the user as a `spi::SpiError` (either directly or indirectly as a variant of `spi::Error`).  If a Postgres SPI function returns an error that means its recoverable, so we really need to pass them back to the user.  This includes methods like `Spi::run()`, `SpiClient::update()`, etc.

### DIFF
--- a/pgx-examples/aggregate/src/lib.rs
+++ b/pgx-examples/aggregate/src/lib.rs
@@ -157,18 +157,20 @@ mod tests {
     }
 
     #[pg_test]
-    fn test_integer_avg_state_sql() {
-        Spi::run("CREATE TABLE demo_table (value INTEGER);");
-        Spi::run("INSERT INTO demo_table (value) VALUES (1), (2), (3);");
+    fn test_integer_avg_state_sql() -> Result<(), spi::Error> {
+        Spi::run("CREATE TABLE demo_table (value INTEGER);")?;
+        Spi::run("INSERT INTO demo_table (value) VALUES (1), (2), (3);")?;
         let retval = Spi::get_one::<i32>("SELECT DEMOAVG(value) FROM demo_table;");
         assert_eq!(retval, Ok(Some(2)));
+        Ok(())
     }
     #[pg_test]
-    fn test_integer_avg_with_null() {
-        Spi::run("CREATE TABLE demo_table (value INTEGER);");
-        Spi::run("INSERT INTO demo_table (value) VALUES (1), (NULL), (3);");
+    fn test_integer_avg_with_null() -> Result<(), spi::Error> {
+        Spi::run("CREATE TABLE demo_table (value INTEGER);")?;
+        Spi::run("INSERT INTO demo_table (value) VALUES (1), (NULL), (3);")?;
         let retval = Spi::get_one::<i32>("SELECT DEMOAVG(value) FROM demo_table;");
         assert_eq!(retval, Ok(Some(2)));
+        Ok(())
     }
 }
 
--- a/pgx-examples/bgworker/src/lib.rs
+++ b/pgx-examples/bgworker/src/lib.rs
@@ -64,21 +64,23 @@ pub extern "C" fn background_worker_main(arg: pg_sys::Datum) {
         }
 
         // within a transaction, execute an SQL statement, and log its results
-        BackgroundWorker::transaction(|| {
+        let result: Result<(), pgx::spi::Error> = BackgroundWorker::transaction(|| {
             Spi::connect(|client| {
                 let tuple_table = client.select(
                     "SELECT 'Hi', id, ''||a FROM (SELECT id, 42 from generate_series(1,10) id) a ",
                     None,
                     None,
-                );
-                tuple_table.for_each(|tuple| {
-                    let a = tuple.by_ordinal(1).unwrap().value::<String>().unwrap();
-                    let b = tuple.by_ordinal(2).unwrap().value::<i32>().unwrap();
-                    let c = tuple.by_ordinal(3).unwrap().value::<String>().unwrap();
-                    log!("from bgworker: ({}, {}, {})", a, b, c);
-                });
-            });
+                )?;
+                for tuple in tuple_table {
+                    let a = tuple.get_datum_by_ordinal(1)?.value::<String>()?;
+                    let b = tuple.get_datum_by_ordinal(2)?.value::<i32>()?;
+                    let c = tuple.get_datum_by_ordinal(3)?.value::<String>()?;
+                    log!("from bgworker: ({:?}, {:?}, {:?})", a, b, c);
+                }
+                Ok(())
+            })
         });
+        result.unwrap_or_else(|e| panic!("got an error: {}", e))
     }
 
     log!("Goodbye from inside the {} BGWorker! ", BackgroundWorker::get_name());
--- a/pgx-examples/custom_sql/src/lib.rs
+++ b/pgx-examples/custom_sql/src/lib.rs
@@ -85,13 +85,17 @@ mod tests {
     use pgx::prelude::*;
 
     #[pg_test]
-    fn test_ordering() {
+    fn test_ordering() -> Result<(), spi::Error> {
         let buf = Spi::connect(|client| {
-            client
-                .select("SELECT * FROM extension_sql", None, None)
-                .flat_map(|tup| tup.by_ordinal(1).ok().and_then(|ord| ord.value::<String>()))
-                .collect::<Vec<String>>()
-        });
+            Ok(client
+                .select("SELECT * FROM extension_sql", None, None)?
+                .flat_map(|tup| {
+                    tup.get_datum_by_ordinal(1)
+                        .ok()
+                        .and_then(|ord| ord.value::<String>().ok().unwrap())
+                })
+                .collect::<Vec<String>>())
+        })?;
 
         assert_eq!(
             buf,
@@ -103,7 +107,8 @@ mod tests {
                 String::from("multiple"),
                 String::from("finalizer")
             ]
-        )
+        );
+        Ok(())
     }
 }
 
--- a/pgx-examples/custom_sql/src/lib.rs
+++ b/pgx-examples/custom_sql/src/lib.rs
@@ -87,14 +87,16 @@ mod tests {
     #[pg_test]
     fn test_ordering() -> Result<(), spi::Error> {
         let buf = Spi::connect(|client| {
-            Ok(client
-                .select("SELECT * FROM extension_sql", None, None)?
-                .flat_map(|tup| {
-                    tup.get_datum_by_ordinal(1)
-                        .ok()
-                        .and_then(|ord| ord.value::<String>().ok().unwrap())
-                })
-                .collect::<Vec<String>>())
+            Ok::<_, spi::Error>(
+                client
+                    .select("SELECT * FROM extension_sql", None, None)?
+                    .flat_map(|tup| {
+                        tup.get_datum_by_ordinal(1)
+                            .ok()
+                            .and_then(|ord| ord.value::<String>().ok().unwrap())
+                    })
+                    .collect::<Vec<String>>(),
+            )
         })?;
 
         assert_eq!(
--- a/pgx-examples/schemas/src/lib.rs
+++ b/pgx-examples/schemas/src/lib.rs
@@ -99,20 +99,21 @@ mod tests {
     }
 
     #[pg_test]
-    fn test_my_some_schema_type() {
+    fn test_my_some_schema_type() -> Result<(), spi::Error> {
         Spi::connect(|mut c| {
             // "MySomeSchemaType" is in 'some_schema', so it needs to be discoverable
-            c.update("SET search_path TO some_schema,public", None, None);
+            c.update("SET search_path TO some_schema,public", None, None)?;
             assert_eq!(
                 String::from("test"),
-                c.select("SELECT '\"test\"'::MySomeSchemaType", None, None)
+                c.select("SELECT '\"test\"'::MySomeSchemaType", None, None)?
                     .first()
                     .get_one::<MySomeSchemaType>()
                     .expect("get_one::<MySomeSchemaType>() failed")
                     .unwrap()
                     .0
             );
-        });
+            Ok(())
+        })
     }
 
     #[pg_test]
--- a/pgx-examples/spi/src/lib.rs
+++ b/pgx-examples/spi/src/lib.rs
@@ -29,8 +29,16 @@ INSERT INTO spi_example (title) VALUES ('I like pudding');
 );
 
 #[pg_extern]
-fn spi_return_query(
-) -> TableIterator<'static, (name!(oid, Option<pg_sys::Oid>), name!(name, Option<String>))> {
+fn spi_return_query() -> Result<
+    TableIterator<
+        'static,
+        (
+            name!(oid, Result<Option<pg_sys::Oid>, pgx::spi::Error>),
+            name!(name, Result<Option<String>, pgx::spi::Error>),
+        ),
+    >,
+    spi::Error,
+> {
     #[cfg(feature = "pg11")]
     let query = "SELECT oid, relname::text || '-pg11' FROM pg_class";
     #[cfg(feature = "pg12")]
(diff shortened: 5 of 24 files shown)


yrashk:
I think this conflation of the connection, closure and errors is rather rigid. I understand your reasoning for having the closure scope (whether I ultimately agree with it or not is a subject for another discussion), so here's what I can propose.

Just because the process of connecting may return `spi::Error` doesn't mean the closure should be rigidly constrained by that. 

Imagine if we had this API:

```rust
Spi::connect()?.execute(|client| ...)
```

Then we would be able to lift the return type signature restriction for the closure and still maintain that "little box" that shows SPI users that they are within this scope.

eeeebbbbrrrr:
> I think this conflation of the connection, closure and errors is rather rigid. I understand your reasoning for having the closure scope (whether I ultimately agree with it or not is a subject for another discussion), so here's what I can propose.
> 
> Just because the process of connecting may return `spi::Error` doesn't mean the closure should be rigidly constrained by that.
> 
> Imagine if we had this API:
> 
> ```rust
> Spi::connect()?.execute(|client| ...)
> ```
> 
> Then we would be able to lift the return type signature restriction for the closure and still maintain that "little box" that shows SPI users that they are within this scope.

`.execute()` returns a `Result<R, spi::Error>` too...

Keep in mind that now basically every function in SPI returns a result (interestingly, except for "open_cursor" as it's documented to only elog).  That means within the connect() closure we have now, users are going to want to be using `?` after any `client` usage.  So... they'll be propagating up a `Result<R, spi::Error>` anyways.

Checkout the branch and spend a bit of time playing with this.  It's actually nice.

This is a fine usage of the API:

```rust
let result: spi::Result<String> = Spi::connect(|client| client.select("SELECT 42")?.get_one::<i32>());

assert_eq!(Ok(Some(42), result);
```

It's clear, as much as working with Result can be, that everything worked and they got back the i32 they wanted.

`Spi::connect()` still doesn't quite ensure we can't leak a Rust owned object that's holding onto a Postgres-allocated pointer.  That's a problem in general for pgx, but at least with the current API we can say, "hey, you gotta keep everything in that little box" until we figure out a general way to somehow make rustc understand MemoryContexts (or something, who knows!).

And for what it's worth, I'm no fan of `Spi::connect()` returning `Result<R, spi::Error>`.  I really liked the 1 day where it only returned `R`.  That was a good day!  I spent numerous hours yesterday toying with different ideas to eliminate this.  It just has to happen.  

I'm also no fan of spi.rs converting any kind of Postgres SpiError result into a `panic!()`.  That really stinks.  Those are recoverable things (otherwise Postgres would have raised an ERROR instead), so we really need to let the user deal with them.

yrashk:
You didn't seem to get the idea of the proposed change. I've suggested splitting existing `connect` into two things:

1. `connect` would return SpiConnection that has only one function: `execute` (find a better name?)
2. `execute` takes the closure that was originally passed to `connect`, and consumes `SpiConnection`

This way, if you drop the connection before you call `execute`, no problem, nothing happened. But if you want to run something in it, you can use `execute` which does not impose `Result<_, spi::Error>` onto closure's return type.

eeeebbbbrrrr:
No, I got it.  There's a #3, which is:

3. If Postgres allocates anything between `Spi::connect()` and `SpiConnection::drop()`, that memory will be freed.  
 
Allowing an open `pg_sys::SPI_connect()` to be hanging around without strictly controlling its "lifetime" is dangerous.

yrashk:
Then I can't help. It is rather important for me to be able to pass different error types other than spi::Error in those closures, but if your mind is set, it is set.

yrashk:
You can impose #[must_use] on `SpiConnection` to produce a warning

eeeebbbbrrrr:
> It is rather important for me to be able to pass different error types other than spi::Error in those closures, but if your mind is set, it is set.

Nested Results are fine.  The first `?` on `Spi::connect()` gets you your Result type.

```rust
    enum MyError {A, B, C}

    let result: Result<(), MyError> = Spi::connect(|client| {
        Ok(Err(MyError::B))
    })?;
```

> You can impose #[must_use]

["failing to explicitly consider ... is almost certainly a bug"](https://std-dev-guide.rust-lang.org/code-considerations/design/must-use.html) isn't quite the same as "your program will likely segfault if you don't use this value".  That's what the compiler is for.

> Then I can't help.

Listen, I acknowledge you've got opinions and desires around Spi and, well, all of pgx.  I've spent countless hours the past 4 days working on fixing the holes just in Spi.  If you'd like to help review this PR in the context of what it is, I'd greatly appreciate it.  If you'd like to design a brand new Spi API and put up a PR we can discuss that too.  We bend over backwards to be as accommodating as we possibly can to our users, including you, while still trying to keep a maintainable product we're proud of.  

I appreciate, sincerely, your contributions to pgx, but I can't handle this kind of attitude anymore. I am more than happy to continue working together, but not like this.  It's mentally and emotionally exhausting.

EdMcBane:
I feel like a guard (like `SpiConnection`) would be the most idiomatic thing, but also that something like that can only be implemented once we've tackled the issue of postgres-borrowed memory and early-free of `SpiTupleTable`.

As for returning any error type, could a `connect` overload returning `Result<T,E: From<spi::Error>>` work? I've tested something like this and, apart from when the type inference is not available, I kinda like it:

```rust
    pub fn connect_with_custom_error<R, E: std::error::Error + From<Error>, F: FnOnce(SpiClient<'_>) -> std::result::Result<R, E>>(f: F) -> std::result::Result<R, E> {
        let connection = SpiConnection::connect()?;
        f(connection.client())
    }

    #[derive(thiserror::Error)]
    enum MyError {
        Spi(#[from] spi::Error),
        Io(#[from] io::Error),
    }

    #[pg_test]
    fn test_custom_error() -> Result<(), MyError> {
        Spi::connect_with_custom_error(|mut client| {
            client.select("SELECT 1")?;
            std::fs::File::create("")?;
            Ok(())
        })
    }
```

eeeebbbbrrrr:
>  could a connect overload returning Result<T,E: From<spi::Error>> work?

Oh, that's interesting.  Hadn't thought of that.

A quick check, it lights up all the existing return types, but I suppose that's expected.  I'll play with this and see what happens.

EdMcBane:
> A quick check, it lights up all the existing return types

All the `unwrap`s, yes. Works better for the newer `Result`-returning tests.

yrashk:
Here's a simple idea in continuation of the connect/execute split:

* `connect()`
    * Remember memory context before `SPI_connect`
    * `SPI_connect`
    * Remember the new current context set by `SPI_connect`
   * Restore remembered memory context
* We're now in the same context we were before `connect()`
* `execute(|client| ...)`
    * Remember the context before `execute()` 
    * Restore the context returned by `SPI_connect`
    * Run the closure
    * Consume `SpiConnection` (perhaps explicitly?)
        *  `SPI_finish` will restore the context before `SPI_connect`
    * Restore context before `execute()`

What do you think?

eeeebbbbrrrr:
> > A quick check, it lights up all the existing return types
> 
> All the `unwrap`s, yes. Works better for the newer `Result`-returning tests.

That wasn't working out really well b/c of the type inference as you noted.

What are y'alls thoughts on this signature?

```rust
    pub fn connect<R, F: FnOnce(SpiClient<'_>) -> R>(f: F) -> Result<R>
```

The user's closure can return anything, including their own Result/Error type, and we always wrap it in our own Result b/c we have to deal with spi errors.  Basically we just automatically do the double-Result idea if the user wants to return a Result.

So this is kinda cool:

```rust
#[derive(thiserror::Error, Debug)]
pub enum MyError {
    #[error("spi did bad")]
    Spi(#[from] spi::Error),

    #[error("filesystem did bad")]
    IO(#[from] std::io::Error),
}

#[pg_extern]
fn foo() -> Result<Option<i32>, MyError> {
    let result = Spi::connect(|client| {
        let forty_two = client.select("SELECT 42", None, None)?.first().get_one::<i32>()?;
        std::fs::read("/tmp/does/not/exist")?;
        Ok(forty_two)
    })?;
    result
}
```

Or just shorten it to:

```rust
#[pg_extern]
fn foo() -> Result<Option<i32>, MyError> {
    Spi::connect(|client| {
        let forty_two = client.select("SELECT 42", None, None)?.first().get_one::<i32>()?;
        std::fs::read("/tmp/does/not/exist")?;
        Ok(forty_two)
    })?
}
```

Still need that first `?` to check for the result from `Spi::connect()`, but we're not forcing the user to return their own SpiError.

Fair compromise?

This breaks compilation on a lot of the tests/examples, but that's a lot of what this PR is about.

yrashk:
> What are y'alls thoughts on this signature?
> 
> ```rust
>     pub fn connect<R, F: FnOnce(SpiClient<'_>) -> R>(f: F) -> Result<R, spi::Error>
> ```

Did you have a chance to give the above idea of memory context-saving wrappers I posted above a thought? 

It *may* allow us to proceed with the split of `connect` and `execute` (or `run` for brevity?) without exposing users to the connection's memory context while between `connect` and `run`.

eeeebbbbrrrr:
It's not that simple.  Internally, Postgres' SPI system is managed by at least these things:

```c
static _SPI_connection *_SPI_stack = NULL;
static _SPI_connection *_SPI_current = NULL;
static int	_SPI_stack_depth = 0;	/* allocated size of _SPI_stack */
static int	_SPI_connected = -1;	/* current stack index */
```

It's stack based.  There's likely considerations around SPI and subtransactions I haven't even considered.  

I see trying to fake it out as too much complexity to reason about, highly dependent on Postgres internal implementation details, and I try very hard to lean towards APIs where rustc can help the user just do it right.

Postgres has callbacks for "spi cleanup" at both the subxact and xact levels (which we don't support... yet).  SPI also has the concept of ["atomic or not"](https://www.postgresql.org/docs/current/spi-spi-connect.html) (which we also don't support... yet).  I can't imagine what kind of interactions being connected to SPI in both atomic and non-atomic modes at the same time would have.

What I can imagine is having a user-facing API that gives users too much flexibility is not going to be future proof, against both pgx and postgres.

EdMcBane:
> Fair compromise?
> 
> This breaks compilation on a lot of the tests/examples, but that's a lot of what this PR is about.

It makes dealing with heterogeneous errors easier at the cost of making the other cases a little harder (but not too much). 
Seems the best balance given the constraints. 

Tomorrow I'll take a couple hours to port my existing SPI code to it and see if any other corner case comes to mind.
Nice 👍

eeeebbbbrrrr:
> Tomorrow I'll take a couple hours to port my existing SPI code to it and see if any other corner case comes to mind.

Oh, that'd be nice.  I'll finish up the fallout from changing the Spi::connect() signature and get it pushed up here before I go to bed tonight.  I gotta make some pasta tonight, so that'll give me something to do while I wait for the dough to rise!

eeeebbbbrrrr:
I have one other idea here...

I changed `Spi::connect()` to return a `Result<R, spi::Error>` because `pg_sys::SPI_connect()` is documented as being able to return an error.  

I've traced through the Postgres sources going back to pg11, and it actually does not return an error.  It *always* returns `SPI_OK_CONNECT`.  In fact, the only use of `SPI_ERROR_CONNECT` is converting the constant value into a string.  It's not actually returned anywhere!

So... what do y'all think about going back to `Spi::connect(...) -> R` and just immediately `panic!()` if for some reason we don't see `SPI_OK_CONENCT`?  That's what pgx has done all along, and despite all its problems around SPI, nobody's bitched about that until I try to make the API match the docs.

@yrashk? @EdMcBane?

yrashk:
I am in favor of this!

eeeebbbbrrrr:
> I am in favor of this!

Then [deaeaeb](https://github.com/tcdi/pgx/pull/977/commits/deaeaeb835f28bc1b470fd692283a62a17f57cb8) ought to be the best commit you've ever seen.

EdMcBane:
👍 
BTW: whoa, that `pg_extern` magic in pgx-examples/spi/src/lib.rs:53 caught me by surprise. Returning a iterator over tuples of results 👍

eeeebbbbrrrr:
> 👍
> BTW: whoa, that `pg_extern` magic in pgx-examples/spi/src/lib.rs:53 caught me by surprise. Returning a iterator over tuples of results 👍

Yah. That’s pretty sweet. And makes the whole Result<Option>> thing easy to work with in pure pass-through situations.

eeeebbbbrrrr:
> 👍
> BTW: whoa, that `pg_extern` magic in pgx-examples/spi/src/lib.rs:53 caught me by surprise. Returning a iterator over tuples of results 👍

haha, you can do some pretty dumb things with this:

```rust
#[pg_extern]
fn wtf() -> TableIterator<'static, (name!(id, i32), name!(title, Result<Option<String>, spi::Error>))> {
    let ids = vec![2619, 1247, 4159, 4160, 2830, 2831, 4161, 4162, 2832, 2833, 4143, 4144];

    TableIterator::new(ids.into_iter().map(|id| {
        (id, Spi::get_one(&format!("SELECT relname::text FROM pg_class WHERE oid = {id}")))
    }))
}
```

```sql
[v13.9][7518] srf=# select * from srf.wtf();
  id  |        title        
------+---------------------
 2619 | pg_statistic
 1247 | pg_type
 4159 | pg_toast_2600
 4160 | pg_toast_2600_index
 2830 | pg_toast_2604
 2831 | pg_toast_2604_index
 4161 | pg_toast_3456
 4162 | pg_toast_3456_index
 2832 | pg_toast_2606
 2833 | pg_toast_2606_index
 4143 | pg_toast_826
 4144 | pg_toast_826_index
(12 rows)
```

EdMcBane:
I've found a glitch, I believe; `SpiCursor::fetch` still returns an `std::result::Result<SpiTupleTable, SpiError>` instead of an `spi::Result`.

eeeebbbbrrrr:
> I've found a glitch, I believe; `SpiCursor::fetch` still returns an `std::result::Result<SpiTupleTable, SpiError>` instead of an `spi::Result`.

There’s a few that still do. I argued with myself over which to return. One ones that return SpiError don’t have a code path where it might also encounter an spi::Error. 

I’m happy to change everything to return spi::Error if you think that’s what makes me most sense?

EdMcBane:
A single return type makes calling different methods in the same `connect` closure simpler, i believe, as you don't need to `map_err` or provide your custom error type.

EdMcBane:
One more point, is it intentional for pgx to reexport all of spi, including spi::Error? If so, maybe we could rethink the naming (e.g. SpiError and PgSpiError instead of Error and SpiError).

eeeebbbbrrrr:
> One more point, is it intentional for pgx to reexport all of spi, including spi::Error? If so, maybe we could rethink the naming (e.g. SpiError and PgSpiError instead of Error and SpiError).

I noticed this too. I think the intention is to only report the spi module itself plus spi::Spi.  I’ll look at it. I did that for pgx::prelude but I guess the top-level pgx crate is still too generous.

eeeebbbbrrrr:
> A single return type makes calling different methods in the same `connect` closure simpler, i believe, as you don't need to `map_err` or provide your custom error type unless you actually want to.

Will do

eeeebbbbrrrr:
Thanks @EdMcBane.  I've pushed these changes.

I am considering renaming SpiOk and SpiError to SpiOkCodes and SpiErrorCodes.  What are your thoughts?  Looking for something that's akin to what Postgres sources call them and also something that doesn't necessarily indicate it can/should be used as a Result::Err value...

EdMcBane:
Makes sense to me, achieves both goals 👍 

It has more to do with internal development now, as they are minimally visible from the new API.

eeeebbbbrrrr:
> Makes sense to me, achieves both goals 👍
> 
> It has more to do with internal development now, as they are minimally visible from the new API.

pushed.  Imma let this PR simmer over the weekend and merge on Monday.

Thanks for your participation with this.  If you think of or see anything else, just let us know!  Happy Holidays.

PR #980, opened 2022-12-23
:enh: bail out early in do_ereport if errstart returns false

EdMcBane:
Changes `pg_sys::submodules::panic::do_ereport` to bail out early in case `errstart` returns false, in order to lower the performance cost of calling e.g. `debug2!` in a tight loop when neither the log nor the client will report the event.

As postgres versions up to 12 require to pass some palloced data into `errstart`, while postgres 13 and later postpone this to  `errfinish`, in order to maximize performance _and_ legibility, a **copromise** was required, splitting the implementation by version, introducing some duplication.


eeeebbbbrrrr:
hey, thanks @EdMcBane!  Looks like you need to push up a `cargo fmt` too.  I'll look at this in detail over the weekend.  I was actually thinking about this during my 6h car drive today.

EdMcBane:
The results are visible but not groundbreaking; tests were conducted with a pair of simple functions:

```rust
#[pg_extern]
fn with_debug() -> i32 {
    debug5!("debug");
    42
}

#[pg_extern]
fn without_debug() -> i32 {
    42
}
```

| Query | Original (ms) | Optimized (ms) |
| ---------| ---------|-----------|
| select i from generate_series(1,10000000) as i where without_debug() < 0;| 1069,830 | N/A |
| select i from generate_series(1,10000000) as i where with_debug() < 0;| 3839,472 |  3030,129 |

Profiling the optimized version shows that most of the remaining time is spent in allocating and deallocating the data that goes into `panic::ErrorReport`, which is done beforehand, and regardless of the eventual return value of `errstart`.

I'm not sure pursuing this further is worthwhile, as the benefits are limited and the impact of changes is significant.

EdMcBane:
> hey, thanks @EdMcBane! Looks like you need to push up a `cargo fmt` too. 

I was taken aback, then shown a single missing newline-at-end-of-file. Fat fingers :smile:

eeeebbbbrrrr:
> I'm not sure pursuing this further is worthwhile, as the benefits are limited and the impact of changes is significant.

This is pretty good tho.  And it makes the code correct which is pretty important too.  Thanks a bunch!

Issue #955, opened 2022-12-12
Goal: Limited, configurable cross-compilation


workingjubilee:
- [ ] https://github.com/tcdi/pgx/pull/954
- [ ] Make building the SQL entity graph arch-neutral
- [ ] Introduce `--target` support to cargo-pgx

@jaskij and @thomcc probably have other thoughts, comments, requirements, or requests.

workingjubilee:
This will effectively impose an API requirement that we essentially not need to use the C shim for anything that PGX relies on at the "basic level". It's fine to use it for e.g. a `BackgroundWorker`, but this will mean there are some shenanigans, like `MemoryContext`, that it can't be a requirement in.

jaskij:
First of all, sadly, the way schema is generated right now means we have to build the extension twice - both for host and for target. Or rather, run codegen and linker twice - I'm not overly familiar with Rust's build process, but some parts are reused when compiling within the same tree for multiple targets.

Lack of cshim would be a great help, since it's generating quite some pain. But it being there is actually not a deal breaker from what I'm seeing - it just needs two more flags (or environmental variable, but those prove to be problematic in the long run). What I'd need to support it being there cross-compilatiom is two more flags, for sysroots - for host and target.

I hope to have working 0.5.4 soon, but it will require a large cleanup and rebasing on top of devel, so an actual PR won't happen in the next week or two.

Of course, there's the big ticket item - `pg_config` itself, but this is supplied by whoever's mad enough to cross compile PostgreSQL extensions.

jaskij:
So far, in my experimental branch, I have taken the approach to specify the target (because it's for Cargo, not C) and the sysroots as arguments, while the rest is passed as environment variables - those are well known and have been used for a long time.

The only little snag is terminology: whether to use the prefix `BUILD_` or `HOST_`, there's some historical context to this, and arguments for both. As Rust generally only uses host and target, without using build, for now I'm using the `HOST_` prefix. Although this might come back at some point.

workingjubilee:
Yes, the "traditional" ones are well-understood and, in many cases, have an actual rulebook as to how you should handle them. One you can print out and eat and everything.

thomcc:
Yeah, where possible we should use `HOST_` (this is because I'm working on excising the makefile in favor of `cc-rs`, which uses `HOST_`, as described [here](https://github.com/rust-lang/cc-rs#external-configuration-via-environment-variables)).

jaskij:
[...]

PR #981, opened 2022-12-24
cargo-pgx cross compilation support

jaskij:
Work in progress for adding cross compilation support to `cargo-pgx`, as discussed in #955

### DIFF
--- a/cargo-pgx/src/command/cross_options.rs
+++ b/cargo-pgx/src/command/cross_options.rs
@@ -1,0 +1,51 @@
+use clap::Args;
+use std::path::PathBuf;
+
+#[derive(Args, Clone, Default, Debug)]
+pub(crate) struct CrossBuildArgs {
+    /// Cross-compilation target - passing this option will make cargo-pgx assume we are cross-compiling
+    #[clap(long)]
+    pub(crate) target: Option<String>,
+    /// Cross-compilation sysroot
+    #[clap(long)]
+    pub(crate) sysroot: Option<PathBuf>,
+    /// Host sysroot
+    #[clap(long)]
+    pub(crate) host_sysroot: Option<PathBuf>,
+    /// Host pg_config
+    #[clap(long)]
+    pub(crate) host_pg_config: Option<PathBuf>,
+}
+
+impl CrossBuildArgs {
+    pub(crate) fn is_cross_compiling(&self) -> bool {
+        self.target.is_some()
+    }
+
+    pub(crate) fn to_build(&self) -> CrossBuild {
+        match self.is_cross_compiling() {
+            true => CrossBuild::Target {
+                target: self.target.clone().unwrap(),
+                sysroot: self.sysroot.clone(),
+            },
+            false => CrossBuild::None,
+        }
+    }
+
+    pub(crate) fn to_host_build(&self) -> CrossBuild {
+        match self.is_cross_compiling() {
+            true => CrossBuild::Host {
+                sysroot: self.host_sysroot.clone(),
+                pg_config: self.host_pg_config.clone(),
+            },
+            false => CrossBuild::None,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub(crate) enum CrossBuild {
+    None,
+    Target { target: String, sysroot: Option<PathBuf> },
+    Host { sysroot: Option<PathBuf>, pg_config: Option<PathBuf> },
+}
--- a/cargo-pgx/src/command/install.rs
+++ b/cargo-pgx/src/command/install.rs
@@ -7,6 +7,7 @@ All rights reserved.
 Use of this source code is governed by the MIT license that can be found in the LICENSE file.
 */
 
+use crate::command::cross_options::{CrossBuild, CrossBuildArgs};
 use crate::command::get::{find_control_file, get_property};
 use crate::manifest::{display_version_info, PgVersionSource};
 use crate::profile::CargoProfile;
@@ -84,6 +85,7 @@ impl CommandExecute for Install {
             package_manifest_path,
             &pg_config,
             &profile,
+            &CrossBuildArgs::default(),
             self.test,
             None,
             &self.features,
@@ -104,6 +106,7 @@ pub(crate) fn install_extension(
     package_manifest_path: impl AsRef<Path>,
     pg_config: &PgConfig,
     profile: &CargoProfile,
+    cross_args: &CrossBuildArgs,
     is_test: bool,
     base_directory: Option<PathBuf>,
     features: &clap_cargo::Features,
@@ -124,8 +127,13 @@ pub(crate) fn install_extension(
 
     let versioned_so = get_property(&package_manifest_path, "module_pathname")?.is_none();
 
-    let build_command_output =
-        build_extension(user_manifest_path.as_ref(), user_package, &profile, &features)?;
+    let build_command_output = build_extension(
+        &user_manifest_path.as_ref(),
+        user_package,
+        &profile,
+        &features,
+        &cross_args.to_build(),
+    )?;
     let build_command_bytes = build_command_output.stdout;
     let build_command_reader = BufReader::new(build_command_bytes.as_slice());
     let build_command_stream = cargo_metadata::Message::parse_stream(build_command_reader);
@@ -179,6 +187,7 @@ pub(crate) fn install_extension(
         &package_manifest_path,
         pg_config,
         profile,
+        cross_args,
         is_test,
         features,
         &extdir,
@@ -224,10 +233,11 @@ fn copy_file(
 }
 
 pub(crate) fn build_extension(
-    user_manifest_path: Option<impl AsRef<Path>>,
+    user_manifest_path: &Option<impl AsRef<Path>>,
     user_package: Option<&String>,
     profile: &CargoProfile,
     features: &clap_cargo::Features,
+    cross_options: &CrossBuild,
 ) -> eyre::Result<std::process::Output> {
     let flags = std::env::var("PGX_BUILD_FLAGS").unwrap_or_default();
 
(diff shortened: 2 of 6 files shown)


jaskij:
Now, the questions is, how do I test it? Since `cargo-pgx` must be in the same version as `pgx` required by the plugin being built.

cargo-pgx/src/command/install.rs
```diff
@@ -259,6 +269,54 @@ pub(crate) fn build_extension(
...
+    match cross_options {
+        CrossBuild::None | CrossBuild::Target { .. } => {
+            if let Some(pg_config) = std::env::var_os("PGX_PG_CONFIG_PATH") {
+                command.env("PG_CONFIG", pg_config);
+            }
+        }
+        CrossBuild::Host { pg_config, .. } => {
+            if let Some(ref host_pg_config) = pg_config {
+                command.env("PGX_PG_CONFIG_PATH", host_pg_config);
+                command.env("PF_CONFIG", host_pg_config);
```
thomcc, line 293: Typo?
```suggestion
                command.env("PG_CONFIG", host_pg_config);
```

cargo-pgx/src/command/package.rs
```diff
@@ -89,15 +92,16 @@ impl CommandExecute for Package {
...
             self.test,
             &self.features,
         )
     }
 }
 
 #[tracing::instrument(level = "error", skip_all, fields(
-    pg_version = %pg_config.version()?,
-    profile = ?profile,
+    pg_version = % pg_config.version() ?,
```
thomcc, line 103: Nit: these whitespace changes aren't necessary.

cargo-pgx/src/command/schema.rs
```diff
@@ -538,7 +560,9 @@ fn create_stub(
     let mut so_rustc_invocation = Command::new("rustc");
     so_rustc_invocation.stderr(Stdio::inherit());
 
-    if let Some(rustc_flags_str) = std::env::var("RUSTFLAGS").ok() {
+    let rustflags_var_name = if cross_build { "RUSTFLAGS" } else { "HOST_RUSTFLAG" };
```
thomcc, line 563: Typo?
```suggestion
    let rustflags_var_name = if cross_build { "RUSTFLAGS" } else { "HOST_RUSTFLAGS" };
```

cargo-pgx/src/command/schema.rs
```diff
@@ -230,13 +235,17 @@ pub(crate) fn generate_schema(
             command.arg(user_package);
         }
 
-        if let Some(user_manifest_path) = user_manifest_path {
+        if let Some(ref user_manifest_path) = user_manifest_path {
```
thomcc, line 238: Nit.
```suggestion
        if let Some(user_manifest_path) = &user_manifest_path {
```

cargo-pgx/src/command/install.rs
```diff
@@ -259,6 +269,54 @@ pub(crate) fn build_extension(
...
+    match cross_options {
+        CrossBuild::None => {}
+        CrossBuild::Target { target, sysroot } => {
+            command.arg("--target");
+            command.arg(target);
+
+            apply_sysroot(&mut command, &sysroot)?;
+        }
+        CrossBuild::Host { sysroot, .. } => {
+            let var_names = vec!["CC", "LD", "CFLAGS", "LDFLAGS"];
```
thomcc, line 308: I think this is missing AR (although perhaps it's not needed for some reason?)
```suggestion
            let var_names = vec!["CC", "LD", "CFLAGS", "LDFLAGS", "AR"];
```

cargo-pgx/src/command/install.rs
```diff
@@ -259,6 +269,54 @@ pub(crate) fn build_extension(
         command.arg("--all-features");
     }
 
+    fn apply_sysroot(command: &mut Command, sysroot: &Option<PathBuf>) -> eyre::Result<()> {
+        if let Some(ref sysroot) = sysroot {
```
thomcc, line 273: Nit:
```suggestion
        if let Some(sysroot) = sysroot {
```

thomcc (commented):
This will need some work (well, it's a draft, so I think you know that), but is a really good start. Here's some preliminary feedback.

cargo-pgx/src/command/cross_options.rs
```diff
@@ -0,0 +1,51 @@
...
+                pg_config: self.host_pg_config.clone(),
+            },
+            false => CrossBuild::None,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub(crate) enum CrossBuild {
+    None,
```
thomcc, line 48: Nit: You should remove this variant and pass around Option<CrossBuild> instead.

cargo-pgx/src/command/cross_options.rs
```diff
@@ -0,0 +1,51 @@
...
+            true => CrossBuild::Target {
+                target: self.target.clone().unwrap(),
+                sysroot: self.sysroot.clone(),
+            },
+            false => CrossBuild::None,
+        }
+    }
+
+    pub(crate) fn to_host_build(&self) -> CrossBuild {
+        match self.is_cross_compiling() {
```
thomcc, line 36: µnit: match on bool isn't particularly idiomatic -- `if` is better (here and elsewhere).

thomcc (commented):
Ah, two super minor idiom nits too.

cargo-pgx/src/command/install.rs
jaskij: My working code branched off of 0.5.4 doesn't have it, but generally, yes, it should be passed. I believe that AR is architecture-independent, but there could be some weird edge cases around using different toolchains between host and target that would make this not work.

cargo-pgx/src/command/cross_options.rs
jaskij: Fair, I got into the stupid habit because in some places they're more concise and (to me) more readable. Because `rustfmt` enforces new lines in `if {} else{}` blocks, and I wanted something like. Here it doesn't make sense, as you say.

```rust
if (answer) { foo.answer(42); }
else { foo.ignore(); }
```

cargo-pgx/src/command/package.rs
jaskij: I'll need to double check, but this probably got changed by `rustfmt`. Or maybe by that idiotic CLion formatter (why it doesn't default to `rustfmt` is beyond me).

cargo-pgx/src/command/cross_options.rs
workingjubilee: > use_small_heuristics = "Max"

This line in our rustfmt.toml means that rustfmt will only enforce that if it cannot actually one-line your code. Otherwise it will attempt to put a simple if-else on one line. It will never do so to a `match`. So you actually are defeating rustfmt's attempt to make your code the smallest.

cargo-pgx/src/command/install.rs
jaskij: Fixed

jaskij:
@thomcc I primarily marked this as a draft to discuss the interface - mostly around what is passed as an argument, and what is passed as an environmental variable.

Meanwhile, I'm going to document some caveats here:

- host and target need compatible versions of PostgreSQL (not sure how much they need to much, if it must be the exact same version, or just same minor)
- `PG_CONFIG` needs to be an absolute path
- setting the path to `pg_config` is an absolute mess - there's `PGX_PG_CONFIG`, `PG_CONFIG`, `--pg-config` and I added `--host-pg-config`. My testing so far indicates that `PG_CONFIG` does not need to be set, but it's still early.

jaskij:
I have absolutely no clue why that one test is failing, but it seems to be misconfigured, especiall since on my PC (with PG15 installed via `cargo pgx init`) that same command does not fail.

```
  [shim for PG v15] [stderr] Makefile:26: /usr/lib/postgresql/15/lib/pgxs/src/makefiles/pgxs.mk: No such file or directory
  [shim for PG v15] [stderr] make: *** No rule to make target '/usr/lib/postgresql/15/lib/pgxs/src/makefiles/pgxs.mk'.  Stop.
```

jaskij:
A quick note on how to actually cross-compile with `cargo-pgx`, seems I forgot to post it yesterday.

1. Read [Thom's guide](https://github.com/tcdi/pgx/blob/develop/CROSS_COMPILE.md)
2. Additional requirements:
-- host *and* target PostgreSQL (including server dev stuff and `pg_config`)
-- user space qemu for your target architecture (eg `qemu-aarch64`)
3. Make `pg_config` wrapper (below)
4. Call `PGX_PG_CONFIG_PATH=~/pg_config_wrapper cargo pgx package --sysroot=/usr/local/oecore-x86_64/sysroots/cortexa53-crypto-poky-linux --pg-config=/home/jaskij/pg_config_wrapper --target=aarch64-unknown-linux-gnu`

`pg_config` wrapper:

```sh
#!/usr/bin/env sh
set -x
SDK_DIR="/usr/local/oecore-x86_64"
SYSROOT_DIR="$SDK_DIR/sysroots"
OE_HOST="x86_64-poky-linux"
OE_TARGET="cortexa53-crypto-poky-linux"
HOST_SYSROOT="$SYSROOT_DIR/$OE_HOST"
TARGET_SYSROOT="$SYSROOT_DIR/$OE_TARGET"

$HOST_SYSROOT/usr/bin/qemu-aarch64 -L $TARGET_SYSROOT  $TARGET_SYSROOT/usr/bin/pg_config "$@"
```

jaskij:
As is, schema generation builds the extension for the host so that it can generate the bindings.
This isn't a good solution, but is a workable one. Sadly, schema generation feels like
black magic to me.

The current solution has two, huge and obvious, drawbacks:

- requiring host PostgreSQL, at the very least in similar version to target's
- doubling the already long build times

I'm not sure if it's in-scope for this MR though, and I would need a lot of guidance to implement it.

thomcc (commented):
I think running the schema generation on the host is reasonable. It's not ideal, surely, but it seems unlikely that SQL will vary between targets (this is also something we can document).

Overall this mostly looks good, but I'm testing it out today and that might dig something up.

I think it needs some docs, but those don't have to block things, since this is a pretty niche use case (and I can handle them, when I find the time).

Do you mind resolving the conflicts?

jaskij:
@thomcc I see that your work on cross compilation was merged, that's what created the conflict. Do you think there is something I should be looking into or integrating with? Or can I just simply resolve the conflict and be done with it?

thomcc:
You don't need to integrate with it really. There's some redundancy, but it largely handles a few cases that this completes (specifically the no cshim case where you don't care about cargo-pgx — it's basically for https://github.com/tcdi/plrust).

Resolving the conflict is sufficient, thanks.

thomcc (approved):
This is great. It will need some further documentation, but I'm working on an update/completion to CROSS_COMPILE.md anyway, and I can integrate it afterwards.

jaskij:
@thomcc conflict resolved.

If you're working on update to `CROSS_COMPILE.md` already, my instructions are condensed in [this comment](https://github.com/tcdi/pgx/pull/981#issuecomment-1383113583).

A cleaned up version of `pg_config_wrapper`:

```sh
#!/usr/bin/env sh
set -x
TARGET_SYSROOT="/usr/local/oecore-x86_64/sysroots/cortexa53-crypto-poky-linux"

qemu-aarch64 -L $TARGET_SYSROOT  $TARGET_SYSROOT/usr/bin/pg_config "$@"
```

thomcc:
Can you rustfmt and push an update?

jaskij:
Damnit, thought it was already formatted. Will do as soon as I'm at my PC.

jaskij:
@thomcc done.

thomcc:
Hmm, can you check the CI error? I didn't see this locally when I tested, but I think I was testing on pg14 rather than 15.

workingjubilee:
If this gets rebased I think we can take a look at landing it?

jaskij:
I don't have the time or energy to get on this now, but might have it at the beginning of September, feel free to ping me then if I don't show up.

eeeebbbbrrrr:
if anyone involved here has an interest in bringing this back to life, please update the PR and re-open.

jaskij:
So... as much as I would absolutely love to do it, it's unlikely. I do not have the time outside of work. Meanwhile, the one use case we have for it at work is [`timescaledb-toolkit`](<https://github.com/timescale/timescaledb-toolkit>) which is several versions behind.

PR #985, opened 2022-12-27
Add type/oid details to Datum conversion errors

EdMcBane:
Adds rust type+oid and datum oid to `TryFromDatumError::IncompatibleTypes` variant. Updates error message to report the additional metadata.

### DIFF
--- a/pgx/src/datum/from.rs
+++ b/pgx/src/datum/from.rs
@@ -19,10 +19,11 @@ use std::num::NonZeroUsize;
 /// If converting a Datum to a Rust type fails, this is the set of possible reasons why.
 #[derive(thiserror::Error, Debug, Clone, PartialEq, Eq)]
 pub enum TryFromDatumError {
-    #[error("The specified type of the Datum (oid {got}) is not compatible with the desired Rust type (preferred oid {preferred})")]
+    #[error("The specified type of the Datum (oid {rust_oid}) is not compatible with the desired Rust type ({rust_type}, primary oid {rust_oid})")]
     IncompatibleTypes {
-        preferred: pg_sys::Oid,
-        got: pg_sys::Oid,
+        rust_type: &'static str,
+        rust_oid: pg_sys::Oid,
+        datum_oid: pg_sys::Oid,
     },
 
     #[error("The specified attribute number `{0}` is not present")]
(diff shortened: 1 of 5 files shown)


EdMcBane:
Currently `TryFromDatumError` is also used by `PgHeapTuple::set_by_index`, which actually leverages the `IntoDatum` trait; this might be confusing, but is not addressed as of yet.

pgx/src/datum/from.rs
```diff
@@ -19,8 +19,8 @@ use std::num::NonZeroUsize;
 /// If converting a Datum to a Rust type fails, this is the set of possible reasons why.
 #[derive(thiserror::Error, Debug, Clone, PartialEq, Eq)]
 pub enum TryFromDatumError {
-    #[error("The specified type of the Datum is not compatible with the desired Rust type.")]
-    IncompatibleTypes,
+    #[error("The specified type of the Datum (oid {rust_oid}) is not compatible with the desired Rust type ({rust_type}, primary oid {rust_oid})")]
```
eeeebbbbrrrr, line 22: Looking at the message again with fresh eyes, it might be more clear as:  

"The Postgres Datum type (oid={datum_oid}) is not compatible with the requested Rust type ({rust_type}, oid={rust_oid})" 

Also note that your message here doesn't seem to use `datum_oid` at all.

eeeebbbbrrrr (changes requested):

pgx/src/datum/from.rs
EdMcBane: > Looking at the message again with fresh eyes, it might be more clear as:
> 
> "The Postgres Datum's type (oid={datum_oid}) is not compatible with the requested Rust type ({rust_type}, oid={rust_oid})"

I'm not 100% sure about this; I'm ok with whatever you deem most appropriate, just want to make sure I gave you all the info I've got.

The same error type is also used (improperly?) in `HeapTuple::set_by_index`; in that case input/output are flipped and the Rust type is _expected_, not _requested_ . Maybe we can get rid of the "requested"?

"The Postgres Datum's type (oid={datum_oid}) is not compatible with the ~requested~ Rust type ({rust_type}, oid={rust_oid})"

> Also note that your message here doesn't seem to use `datum_oid` at all.

Most definitely, sorry about that.

eeeebbbbrrrr: oh, I see.  How about just "Postgres type oid {datum_oid} is not compatible with the Rust type {rust_type} (oid={rust_oid})".

It is kinda confusing!  

And perhaps using this to lookup the type names would be helpful too, in addition to the oids?

```rust
fn lookup_type_name(oid: &PgOid) -> String {
    unsafe {
        // SAFETY: nothing to concern ourselves with other than just calling into Postgres FFI
        // and Postgres will raise an ERROR if we pass it an invalid Oid, so it'll never return a null
        let cstr_name = pg_sys::format_type_extended(oid.value(), -1, 0);
        let cstr = CStr::from_ptr(cstr_name);
        let typname = cstr.to_string_lossy().to_string();
        pg_sys::pfree(cstr_name);  // don't leak the palloc'd cstr_name
        typname
    }
}

EdMcBane: I've integrated your recommendations and added a test.
Happy new year! :fireworks:

eeeebbbbrrrr:
nice!.

PR #986, opened 2022-12-28
Life UnwindSafe + RefUnwindSafe bounds in PgMemoryContexts

eeeebbbbrrrr:
These bounds aren't necessary, and neither is running the innards of `::exec_in_context()` within PgTryBuilder.  If the closure panics/ERRORs, Postgres will clean things up for us.


eeeebbbbrrrr:
I've run the tests locally and need this for some other work I'm doing, so just merging now...

PR #989, opened 2022-12-30
PgLwLock: skip releasing the lock when unwinding from an `elog` call in postgres code

### LINKED ISSUE #988: PgLwLock: backend crash when error reported in postgres code while holding lock
When holding a PgLwLock (exclusive or shared), any call into a `pg_guard_ffi_boundary`-protected postgres function that raises an error (e.g. via `elog`) triggers a backend crash due to a failed assertion in `lwlock.c:LwLockRelease`.

```rust
static SHARED_LOCK: PgLwLock<i32> = PgLwLock::new();

#[pg_guard]
pub extern "C" fn _PG_init() {
    pg_shmem_init!(SHARED_LOCK);
}

#[pg_extern]
fn trigger_assertion_error() {
    // Hold lock
    let _lock = SHARED_LOCK.share();
    // Call into `pg_guard_ffi_boundary`-protected postgres function which internally reports an error
    unsafe { pg_sys::format_type_extended(pg_sys::InvalidOid, -1, 0) };
}
```

I've published a reproducer here: https://github.com/EdMcBane/pgx-bug960

The cause seems to be that `elog.c:errfinish()`, if the level is `ERROR`, resets `InterruptHoldoffCount` to zero.
`LwLockRelease` on the other hand assumes to be running in an interrupt holdoff section, and if run after `errfinish` will fail a `InterruptHoldoffCount > 0` assertion in the `RESUME_INTERRUPTS` macro.

So, while holding a lock:
* If we `error!` in rust code
  * the error is converted in a panic
  * the rust stack is unwound
  * the lock released
  * only once we hit the `pgx_extern_c_guard` barrier the panic gets turned into a postgres  `elog` and `errfinish` is called, resetting `InterruptHoldoffCount` without consequences
* If instead we call a `pg_guard_ffi_boundary`-protected postgres function which internally reports an error:
  * `errfinish` is called, resetting `InterruptHoldoffCount` to zero
  * a c forced unwind is triggered, rolling back the stack to the `pg_guard_ffi_boundary` barrier
  * the error is converted in a rust panic
  * the rust stack is unwound
  * the lock is released, but now `InterruptHoldoffCount` is zero and the assertion fails
  * postgres crashes

This was found while investigating #960 , but I'm not sure it is the same issue.

## Solution
`PgLwLockShareGuard::drop` would need to avoid calling LWLockRelease if unwinding  was triggered by `errfinish`.
As postgres docs say, in `src/backend/storage/lmgr/README` says that
>LWLock manager will automatically release held LWLocks during elog() recovery

It should be possible to detect this situation by checking the value of `InterruptHoldoffCount` at lock release time.

EdMcBane:
Avoid backend crash when trying to release the lock on drop due to `elog` resetting the interrupt holdoff section status.
The code now detects this situation and skips releasing the lock.

Fixes #988

### DIFF
--- a/pgx-tests/src/lib.rs
+++ b/pgx-tests/src/lib.rs
@@ -23,6 +23,6 @@ pub mod pg_test {
     }
 
     pub fn postgresql_conf_options() -> Vec<&'static str> {
-        vec![]
+        vec!["shared_preload_libraries='pgx_tests'"]
     }
 }
--- a/pgx-tests/src/tests/shmem_tests.rs
+++ b/pgx-tests/src/tests/shmem_tests.rs
@@ -7,13 +7,49 @@ All rights reserved.
 Use of this source code is governed by the MIT license that can be found in the LICENSE file.
 */
 use pgx::prelude::*;
-use pgx::{pg_shmem_init, PgAtomic, PgSharedMemoryInitialization};
+use pgx::{pg_shmem_init, PgAtomic, PgLwLock, PgSharedMemoryInitialization};
 use std::sync::atomic::AtomicBool;
 
 static ATOMIC: PgAtomic<AtomicBool> = PgAtomic::new();
+static LWLOCK: PgLwLock<bool> = PgLwLock::new();
 
 #[pg_guard]
 pub extern "C" fn _PG_init() {
     // This ensures that this functionality works across PostgreSQL versions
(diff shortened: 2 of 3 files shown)


eeeebbbbrrrr:
Thank you, @EdMcBane, for diving into this one.  Much appreciated.

PR #990, opened 2023-01-03
The `cshim` feature should not be a default in `pgx-pg-sys`.

eeeebbbbrrrr:
Users don't use the pgx-pg-sys crate as a direct dependency, and `pgx` itself has `cshim` as its default.  This makes it easy for users to expose their own feature to toggle the cshim without pgx needing knowledge of pgx-pg-sys' default feature set.


eeeebbbbrrrr:
I stumped across this while working on https://github.com/tcdi/plrust/pull/142.  I think this was just an oversight back from #958.

Issue #983, opened 2022-12-27
`SpiClient::select` not seeing changes from `Spi::run` in same transaction


EdMcBane:
This issue affects the current develop branch.

Queries executed with `SpiClient::select` do not see data changes applied within the same transaction by update queries executed using `Spi::run`. 

E.g. this fails, as `with_select` is zero, instead of 1 as expected:
```rust
    #[pg_test]
    fn test_spi_select_sees_run() -> spi::Result<()> {
        Spi::run("CREATE TABLE asd(id int)")?;
        Spi::run("INSERT INTO asd(id) VALUES (1)")?;
        let with_select = Spi::connect(|client| {
            client.select("SELECT COUNT(*) FROM asd", None, None)?
                .first().get_one::<i64>()
        })?;
        let with_get_one = Spi::get_one::<i64>("SELECT COUNT(*) FROM asd")?;

        assert_eq!(with_select, with_get_one);
        Ok(())
    }
```

Using `SpiClient::update` in the same session works:
```rust
    #[pg_test]
    fn test_spi_select_sees_update() -> spi::Result<()> {
        let with_select = Spi::connect(|mut client| {
            client.update("CREATE TABLE asd(id int)", None, None)?;
            client.update("INSERT INTO asd(id) VALUES (1)", None, None)?;
            client.select("SELECT COUNT(*) FROM asd", None, None)?
                .first().get_one::<i64>()
        })?;
        let with_get_one = Spi::get_one::<i64>("SELECT COUNT(*) FROM asd")?;

        assert_eq!(with_select, with_get_one);
        Ok(())
    }
```

Using `SpiClient::update` in a separate session **does not** work:
```rust
    #[pg_test]
    fn test_spi_select_sees_update_in_other_session() -> spi::Result<()> {
        Spi::connect::<spi::Result<()>, _>(|mut client| {
            client.update("CREATE TABLE asd(id int)", None, None)?;
            client.update("INSERT INTO asd(id) VALUES (1)", None, None)?;
            Ok(())
        })?;
        let with_select = Spi::connect(|client| {
            client.select("SELECT COUNT(*) FROM asd", None, None)?
                .first().get_one::<i64>()
        })?;
        let with_get_one = Spi::get_one::<i64>("SELECT COUNT(*) FROM asd")?;

        assert_eq!(with_select, with_get_one);
        Ok(())
    }
```

EdMcBane:
[...]

PR #992, opened 2023-01-04
Extend Spi's understanding of "readonly" to transaction

eeeebbbbrrrr:
Mucking with the pushing a new Snapshot if we *think* we've modified the database isn't correct.  It can cause problems elsewhere within Postgres where Postgres expects the Portal's snapshot to be the same as when the Portal was created.

The best I can see to do is just track "readonly" (renamed to `MUTABLE_MODE`) for the duration of the current transaction.

This problem appeared while working on plrust, upgrading it to the pgx 'develop' branch.

I've spent almost 2 hours trying to come up with a test case for this, but as best I can tell, without actually implementing an actual language handler in pgx' test suite I can't see how to expose the underlying problem.

For now, the fact plrust works will have to be the test.

### DIFF
--- a/pgx/src/spi.rs
+++ b/pgx/src/spi.rs
@@ -9,7 +9,10 @@ Use of this source code is governed by the MIT license that can be found in the 
 
 //! Safe access to Postgres' *Server Programming Interface* (SPI).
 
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr:
This is more followup to PR #984, which came about from issue #983.  No good deed goes unpunished.

PR #995, opened 2023-01-06
Mark functions `unsafe` that should be

eeeebbbbrrrr:
By no means is this an exhaustive audit.  I'm specifically targeting things that are exposed through the `trusted-pgx` crate over in https://github.com/tcdi/plrust/pull/142

### Note

I deleted `PgBox::new()` and `PgBox::new_in_context()` as I think they were incredibly broken by design and aren't used by pgx itself.  There were two test functions around them, but that's it.

If users were using them, they'll need to find a different way of doing what those did -- and hopefully one that isn't fundamentally broken.

### DIFF
--- a/pgx-pg-sys/src/lib.rs
+++ b/pgx-pg-sys/src/lib.rs
@@ -159,6 +159,16 @@ pub use internal::pg15::*;
 /// A trait applied to all Postgres `pg_sys::Node` types and subtypes
 pub trait PgNode: seal::Sealed {
     /// Format this node
+    ///
+    /// # Safety
+    ///
+    /// While pgx controls the types for which [`PgNode`] is implemented and only pgx can implement
+    /// [`PgNode`] it cannot control the members of those types and a user could assign any pointer
+    /// type member to something invalid that Postgres wouldn't understand.
+    ///
+    /// Because this function is used by `impl Display` we purposely don't mark it `unsafe`.  The
+    /// assumption here, which might be a bad one, is that only intentional misuse would actually
+    /// cause undefined behavior.
     #[inline]
     fn display_node(&self) -> std::string::String {
         // SAFETY: The trait is pub but this impl is private, and
--- a/pgx/src/memcxt.rs
+++ b/pgx/src/memcxt.rs
@@ -15,8 +15,8 @@ Use of this source code is governed by the MIT license that can be found in the 
 //! An enum-based interface (`PgMemoryContexts`) around Postgres' various `MemoryContext`s provides
 //! simple accessibility to working with MemoryContexts in a compiler-checked manner
 //!
+use crate::pg_sys;
 use crate::pg_sys::AsPgCStr;
-use crate::{pg_sys, PgBox};
 use core::ptr;
 use std::fmt::Debug;
 use std::ptr::NonNull;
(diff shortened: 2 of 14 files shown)


thomcc (approved):
Seems good to me!

pgx/src/memcxt.rs
```diff
@@ -369,15 +414,39 @@ impl PgMemoryContexts {
...
+    /// This function will panic if `src` is a null pointer.
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe because we cannot ensure that any of the [`PgMemoryContexts`] variants,
+    /// specifically those with payloads, actually represent valid Postgres [`pg_sys::MemoryContextData`]
+    /// pointers.
+    ///
+    /// We also cannot ensure that the result of this function will stay allocated as long as Rust's
+    /// borrow checker thinks it will.
```
thomcc, line 449: It's not really the borrow checker that's relevant here (since the raw pointer probably will have no lifetime), but I don't have a better suggestion for wording, and this gets the point across

thomcc (approved):
This looks great, thanks.

pgx/src/memcxt.rs
eeeebbbbrrrr: The returned thing is an owned `R`, that presumably would have been allocated in the memory context since the user called `.switch_to()`.

So not really any kind of "raw pointer" is necessarily involved here (unless that's what the returned `R` actually is).

pgx/src/pgbox.rs
```diff
@@ -309,10 +282,15 @@ impl<T, AllocatedBy: WhoAllocated> PgBox<T, AllocatedBy> {
     /// ## Examples
     /// ```rust,no_run
     /// use pgx::{PgBox, pg_sys};
-    /// let create_trigger_statement = PgBox::<pg_sys::CreateTrigStmt>::alloc_node(pg_sys::NodeTag_T_CreateTrigStmt);
+    /// let create_trigger_statement = unsafe { PgBox::<pg_sys::CreateTrigStmt>::alloc_node(pg_sys::NodeTag_T_CreateTrigStmt) };
     /// ```
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe as we cannot ensure that the MemoryContext used to allocate will
```
thomcc, line 290: It also returns a zeroed value (which may be invalid for T), and is probably UB if `T: pg_sys::PgNode` isn't true (can we add that bound?), since it writes the tag to the start.

thomcc (approved):
Mostly good, one comment nit, and `alloc_node` should ideally have a `where T: pg_sys::PgNode`.

pgx/src/pgbox.rs
eeeebbbbrrrr: taken care of in 8bcb127.  Thanks!

pgx-pg-sys/src/lib.rs
```diff
@@ -159,6 +159,16 @@ pub use internal::pg15::*;
...
+    ///
+    /// # Safety
+    ///
+    /// While pgx controls the types for which [`PgNode`] is implemented and only pgx can implement
+    /// [`PgNode`] it cannot control the members of those types and a user could assign any pointer
+    /// type member to something invalid that Postgres wouldn't understand.
+    ///
+    /// Because this function is used by `impl Display` we purposely don't mark it `unsafe`.  The
+    /// assumption here, which might be a bad one, is that only intentional misuse would actually
+    /// cause undefined behavior.
```
workingjubilee, line 171: Violating type invariants is the fault of the code that violates the type invariants, so it would be fully sound for `display_node` to assume a type's invariants have not been violated... if those types did not have public fields and we managed access to them.

Maybe later.

eeeebbbbrrrr: @thomcc and I discussed this a bit on Discord today. “Maybe Later” was the consensus there too.

PR #996, opened 2023-01-06
allow `#![forbid(unsafe_op_in_unsafe_fn)]` to work with pgx-generated…

eeeebbbbrrrr:
This is a followup to #995.


thomcc (approved):
No arguments from me. Would be nice to have a test that verifies we don't regress this.

PR #1001, opened 2023-01-10
Convert `pg_sys::Oid` into a newtype

workingjubilee:
This enforces type safety around `pg_sys::Oid`. Vigorously.

This rewrite was... very opinionated. If this is not suitable for whatever reason I am fairly amenable to changes. It was simply easier to draft a strongly-worded letter to the codebase to open the discussion.

Effects include:
- `pg_sys::Oid` is now a `#[repr(transparent)]` newtype around a `u32`
- More `pg_sys::Oid` constants are correctly classed as an `Oid`, thus are in `BuiltinOid`
- `BuiltinOid` also is now a u32, since it need be no larger (so it's FFI-safe and transmutable into `pg_sys::Oid`), which also means `PgOid` has a smaller repr (not FFI-safe the way the others are, however)
- There are more conversions to/from the various "oidal" types
- Some nonfunctional changes to how bindings are generated happened
- `__pgx_sql_mappings` goes away, as does special-casing of translation to or from SQL, which will make it significantly easier to move schemagen "out" of the PGX library itself.

It is best to focus primarily on the action inside `pgx-pg-sys`, `cargo-pgx`, and `pgx/src/lib.rs`. Everything else is "just" a consequence of API differences.

### DIFF
--- a/pgx-pg-sys/src/submodules/oids.rs
+++ b/pgx-pg-sys/src/submodules/oids.rs
@@ -9,41 +9,188 @@ Use of this source code is governed by the MIT license that can be found in the 
 
 #![allow(non_camel_case_types)]
 use crate as pg_sys;
-use crate::PgBuiltInOids;
+use crate::BuiltinOid;
+use crate::Datum;
+use pgx_sql_entity_graph::metadata::{
+    ArgumentError, Returns, ReturnsError, SqlMapping, SqlTranslatable,
+};
 
-impl PgBuiltInOids {
-    pub fn value(self) -> pg_sys::Oid {
-        self as isize as pg_sys::Oid
+/// An [object identifier][pg_docs_oid] in Postgres.
+///
+/// This is meant to be understood purely by equality. There is no sensible "order" for Oids.
+///
+/// # Notes
+/// `Default` shall return a sensical Oid, not necessarily a useful one.
+/// Currently, this means that it returns the invalid Oid.
+///
+/// [pg_docs_oid]: https://www.postgresql.org/docs/current/datatype-oid.html
+#[repr(transparent)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+#[derive(serde::Deserialize, serde::Serialize)]
+pub struct Oid(pub(crate) u32);
+
+impl Oid {
+    pub const INVALID: Oid = Oid(0);
+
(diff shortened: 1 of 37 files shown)


eeeebbbbrrrr:
This is a lot to review.  I'll go over it first thing in the morning.  Once you sort out the test failures would you mind updating the description here with some words about what this does?

workingjubilee:
The test errors appear to be mostly transient network failures that have sorted themselves out. Sure, I'll happily expand the description a bit.

pgx-pg-sys/src/submodules/oids.rs
```diff
@@ -9,41 +9,188 @@ Use of this source code is governed by the MIT license that can be found in the
...
+use pgx_sql_entity_graph::metadata::{
+    ArgumentError, Returns, ReturnsError, SqlMapping, SqlTranslatable,
+};
 
-impl PgBuiltInOids {
-    pub fn value(self) -> pg_sys::Oid {
-        self as isize as pg_sys::Oid
+/// An [object identifier][pg_docs_oid] in Postgres.
+///
+/// This is meant to be understood purely by equality. There is no sensible "order" for Oids.
```
eeeebbbbrrrr, line 20: There might not be a sensible order for Oids, but Postgres does indeed support `ORDER BY oid` clauses, and they're technically just u32s.

I can see situations where an extension might want to store a bunch of Oids in a BTreeSet or as keys in a BTreeMap.

This is the only thing I see in the PR that we might want to re-consider, so I'm not going to let that stop me from smashing the merge button once I get CI to pass it...

eeeebbbbrrrr (approved):

eeeebbbbrrrr:
Thanks, @workingjubilee.  This is the kind of thing that only you can see how to do correctly.  Making pgx better one type at a time!

I'm also excited about the prospect of getting rid of the pgx-sql-entity-graph crate at some point.

pgx-pg-sys/src/submodules/oids.rs
workingjubilee: If anyone objects I'm happy to see us add the derive, it's just usually adding that impl suggests the comparison is meaningful, where here it's... kinda not.

eeeebbbbrrrr: That's fine.  No problem.

PR #1004, opened 2023-01-11
impl Display for `pg_sys::Oid`

workingjubilee:
Implement Display for `pg_sys::Oid`, adding formatting rules that identify whether or not it  is a "recognized" OID# and giving the appropriate name so that programmers will not have to scour Postgres source code as often when debugging with PGX.

Also adds in another `BuiltinOid` that was missed due to not falling into the existing heuristics. Unfortunately it's harder to generalize this case.


eeeebbbbrrrr (approved):
Wonderful!  Merge it!

PR #1010, opened 2023-01-15
Allow overriding the binding generator and work on a better CROSS_COMPILE.md

thomcc:
Kind of a pain TBH, but if this saves me from having to explain how to doctor cross-compile toolchains output out by https://github.com/crosstool-ng/crosstool-ng so that they work with a PG which was configured against a major linux distribution... I'm for it.

Needs more docs, more testing.


eeeebbbbrrrr (approved):
Thank you, @thomcc.  I know this has been a lot of work!

PR #1012, opened 2023-01-16
WIP: Initial IntArray<T> implementation

mhov:
This is a first swing at a true zero-copy array for `int[], int2[], int8[]` based off the optimizations taken in PG's intarray extension https://github.com/postgres/postgres/tree/master/contrib/intarray

As in `contrib/intarray` only non-null arrays can be supported due to how the elements are packed into `*ArrayType`. However, the performance benefits over converting `Vec<i32> -> Array<int>` (using `pg_sys::accumArrayResult` and `pg_sys::makeArrayResult`) are non-trivial.

Looking for the team's input on whether this feels like a good addition, OR could Array<T> ever be extended made to support such optimizations?


mhov:
I now consider this PR as "for demonstration purposes only"

workingjubilee:
Partially superceded by https://github.com/tcdi/pgrx/pull/1012 for the "passing arrays through Postgres" cases, the only question is how to make array allocation work.

workingjubilee:
Tracking the remaining needs in https://github.com/tcdi/pgrx/pull/1012 so I am closing this.

PR #1013, opened 2023-01-18
Fix version in pg15 help text

rustprooflabs:
The `--pg15` help currently mentions pg14 when running `cargo pgx init --help`.  I think this is the right place to fix that text.


Smittyvb (approved):
Yep, this should fix it – I made a copy paste error when I added this originally.

rustprooflabs:
@rtwalker Fixed that, and a few more.  Thanks!

eeeebbbbrrrr (approved):

PR #1019, opened 2023-01-24
enhance `PgConfig` to be able to construct itself from a set of environment variables

eeeebbbbrrrr:
... that are prefixed with `PGX_PG_CONFIG_`.  Also adjusts how `Pgx::iter(PgConfigSelector::All)` behaves if we detect we have a "pg_config" defined as environment variables.

- Also adds Postgres' `common/config_info.h` header to pgx' bindings.

- Some minor API breakage as we now return owned `PgConfig` instances.


thomcc (approved):
LGTM.

PR #1020, opened 2023-01-24
If there's a "pg_config" described in the environment, use it.  

eeeebbbbrrrr:
Follow-up to PR #1019.

### DIFF
--- a/pgx-pg-config/src/lib.rs
+++ b/pgx-pg-config/src/lib.rs
@@ -138,9 +138,9 @@ impl PgConfig {
     ///
     /// It also requires that the `PGX_PG_CONFIG_AS_ENV` variable be set to some value that isn't
     /// the string `"false"`.
-    pub fn from_env() -> Option<Self> {
+    pub fn from_env() -> eyre::Result<Self> {
         if !Self::is_in_environment() {
-            None
+            Err(eyre::eyre!("`PgConfig` not described in the environment"))
         } else {
             const PREFIX: &str = "PGX_PG_CONFIG_";
 
@@ -153,7 +153,7 @@ impl PgConfig {
                 }
             }
 
-            Some(Self {
+            Ok(Self {
                 version: None,
                 pg_config: None,
                 known_props: Some(known_props),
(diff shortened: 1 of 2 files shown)


pgx-pg-sys/build.rs
```diff
@@ -164,8 +164,19 @@ fn main() -> eyre::Result<()> {
                     .join(", ")
             )
         })?;
-        let specific = pgx.get(&found_feat)?;
-        vec![(found_ver, specific)]
+
+        if PgConfig::is_in_environment() {
```
thomcc, line 168: Might be cleaner to just do `if let Some(...) = PgConfig::from_env()` rather than changing it to return an error only if that's not checked.

eeeebbbbrrrr: `if let Ok(...) = PgConfig::from_env()` you mean?  I can do that.  I changed it to return a Result and ought to treat it as such, I suppose.

eeeebbbbrrrr: Done in [cac55e4](https://github.com/tcdi/pgx/pull/1020/commits/cac55e477996b0492b0941be14e24f4088f14bae)

eeeebbbbrrrr:
I dunno why this commit upset `cargo fmt` so much, but whatever.

pgx-pg-config/src/lib.rs
```diff
@@ -138,9 +138,9 @@ impl PgConfig {
     ///
     /// It also requires that the `PGX_PG_CONFIG_AS_ENV` variable be set to some value that isn't
     /// the string `"false"`.
-    pub fn from_env() -> Option<Self> {
+    pub fn from_env() -> eyre::Result<Self> {
```
thomcc, line 141: Is there still a reason to make this return a Result?

thomcc (approved):
This is basically fine. I have one minor further followup to my previous nit but I don't feel strongly about it.

pgx-pg-config/src/lib.rs
eeeebbbbrrrr: Is asking for it from the environment when it’s not there akin to asking for it from a file that doesn’t exist?  And would we return a Result in that case?

obviously I could go either way as v1 of this was an Option.

eeeebbbbrrrr: I'm gonna stick with Result

PR #1021, opened 2023-01-24
Bump the Nix flake.lock

workingjubilee:
It just requires `nix flake update`...
...and configuring Nix to let you use "experimental features" that are used by like 80% of the Nix userbase. Sure Jan.

cc @bbigras


bbigras:
Seems fine, thanks!

PR #1025, opened 2023-01-30
remove the  dependency on `cstr_core`

eeeebbbbrrrr:
`core::ffi::CStr` and `alloc::ffi::CString` stabilized in rust v1.64, and since we have this loose requirement that 1.65 is pgx' bottom end supported version (b/c of plrust), we can now use these types (along with `std::ffi::CStr(ing)`) and drop the `cstr_core` dependency.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -589,22 +589,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "cstr_core"
-version = "0.2.6"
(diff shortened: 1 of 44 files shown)


workingjubilee (approved):
Nice.

thomcc (approved):
Looks good to me! Tried to do this a few times, but was thwarted by 1.61 at the time.

PR #1026, opened 2023-01-31
Remove `include!` usage from pgx-pg-sys

thomcc:
There's no reason to do this really at the moment, but If we do this and somehow find a workaround for our dependencies usage of `include!` (possibly by forking them? It would just be `syn` as far as I could tell...), we could fully solve the issue mentioned https://github.com/tcdi/postgrestd/pull/23#pullrequestreview-1277971684.

This is also slightly cleaner, IMO.


thomcc:
Hmm, it looks like this didn't get stablized at the same time as some of the other expressions-in-attributes. Unfortunate.

PR #1028, opened 2023-02-02
Support in `FromDatum` for binary coercible types (including domain types)

EdMcBane:
Add support for deserializing domain types into rust types compatible with the base type.
If the initial oid compatibility check fails, we iteratively check the type cache for the domain base type and test against that.

### DIFF
--- a/pgx-pg-sys/include/pg11.h
+++ b/pgx-pg-sys/include/pg11.h
@@ -82,6 +82,7 @@ Use of this source code is governed by the MIT license that can be found in the 
 #include "parser/parse_func.h"
 #include "parser/parse_oper.h"
 #include "parser/parse_type.h"
+#include "parser/parse_coerce.h"
 #include "parser/parser.h"
 #include "parser/parsetree.h"
 #include "postmaster/bgworker.h"
--- a/pgx-pg-sys/include/pg12.h
+++ b/pgx-pg-sys/include/pg12.h
@@ -82,6 +82,7 @@ Use of this source code is governed by the MIT license that can be found in the 
 #include "parser/parse_func.h"
 #include "parser/parse_oper.h"
 #include "parser/parse_type.h"
+#include "parser/parse_coerce.h"
 #include "parser/parser.h"
 #include "parser/parsetree.h"
 #include "postmaster/bgworker.h"
--- a/pgx-pg-sys/include/pg13.h
+++ b/pgx-pg-sys/include/pg13.h
@@ -82,6 +82,7 @@ Use of this source code is governed by the MIT license that can be found in the 
 #include "parser/parse_func.h"
 #include "parser/parse_oper.h"
 #include "parser/parse_type.h"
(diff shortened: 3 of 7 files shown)


eeeebbbbrrrr:
Interesting.  This is pretty simple.  What might any downsides be?

Would we also want to get Postgres' `IsBinaryCoercible()` function involved before checking the typecache?

EdMcBane:
As my patch currently stands, it partially duplicates logic in `isBinaryCoercible`, so there might be maintenance overhead.

I (mistakenly?) assumed from our previous exchange that it was not the correct solution, but looking at it now it basically does, at least for domain types.
I'll explore that further and update this, integrating it might allow to handle more than just domain types.

From a performance standpoint:

- the happy path is unaffected (as long as we first rely on the existing direct "supported oids" check in rust);
- binary-coercible types would pay a slight penalty (FFI + typecache / typecast lookup);
- `try_from_datum` with incompatible types would also pay a penalty, as above, but these are basically programming errors and thus the impact is negligible.

eeeebbbbrrrr:
Just coming back to this.  My question about `isBinaryCoercible` is simply "should we also ask it", such as:

```rust
fn is_output_compatible<T: IntoDatum>(mut type_oid: pg_sys::Oid) -> bool {
     loop {
         if T::is_compatible_with(type_oid) {
             return true;
         } else if pg_sys:: isBinaryCoercible(T::type_oid(), type_oid) {
            return true;
         }

         type_oid = unsafe {
             (*pg_sys::lookup_type_cache(type_oid, pg_sys::TYPECACHE_DOMAIN_BASE_INFO as i32))
                 .domainBaseType
         };
         if type_oid == pg_sys::InvalidOid {
             return false;
         }
     }
 }
 ```

I know we can't only consider it, but I think we can include it in the list.

Thoughts?

**Comment by EdMcBane** [comment]
Sorry it took me so long, but I've finally taken the time to look back at it, and I think that getting rid of the loop and using just `IsBinaryCoercible` is the right thing to do:

- we remove any code duplication / overlap with postgres;
- we need to pay the FFI price at most once;
- we get support for more binary coercible types such as `cidr` to `inet`, and any other custom type registered by an extension.

It ends up looking like this:
```rust
fn is_output_compatible<T: IntoDatum>(type_oid: pg_sys::Oid) -> bool {
    T::is_compatible_with(type_oid) 
     || unsafe { pg_sys:: isBinaryCoercible(type_oid, T::type_oid()) }
}
```
⚠️ `T::type_oid()` needs to be the second argument, as coercion is not symmetric. 

I've updated the branch and added a test for `cidr` to `inet`.
Does this look right? Did I miss anything?

eeeebbbbrrrr:
This seems reasonable.

I hope one day soon I'll have time to look at automatic casting between types.  That'd be the ideal solution here.  Being able to get back any type as a `String` would be sweeeeeet, for example.

PR #1029, opened 2023-02-02
Support all the Rust "math" operations for `Numeric<P, S>` and `AnyNumeric` against primitives

eeeebbbbrrrr:
It was a complete oversight that these didn't get implemented as part of the initial Numeric support back in November 2022 as part of #851.

I was both shocked and awed to learn about this.

So here they are.  All of Rust's "math" traits (Add, Sub, Mul, Div, Rem) are implemented for both `Numeric<P, S>` and `AnyNumeric` against all rust primitive types, with the primitive supported on both the left and right.

In all cases, the resulting datatype is an `AnyNumeric`.

### DIFF
--- a/pgx-examples/numeric/src/lib.rs
+++ b/pgx-examples/numeric/src/lib.rs
@@ -28,6 +28,22 @@ fn math() -> Numeric<10, 3> {
     n %= 19;
     n += 99.42;
 
+    let mut n: AnyNumeric = n * 42;
+    n = n / 42;
+    n = n + 42;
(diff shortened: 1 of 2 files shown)


thomcc (approved):
LGTM. There are probably ways to reduce the boilerplate further (with another layer of macro nonsense), but it's your call if you want to try for that (it's not exactly unclear as-is).

eeeebbbbrrrr:
I'm gonna do the reverse to have the primitive on the lhs and call it a day.

PR #1030, opened 2023-02-02
Update version to 0.7.1

eeeebbbbrrrr:
Welcome to pgx v0.7.1.  It is a small point release that primarily finishes `Numeric` support.  That said, it wouldn't be a proper pgx release without at least one possible breaking API change.

As always, make sure to install cargo-pgx with `cargo install cargo-pgx --version 0.7.1 --locked` and update crate dependencies as well.

## New Trait Impls

* Support all the Rust "math" operations (Add, Sub, Mul, Div, Rem) for `Numeric<P, S>` and `AnyNumeric` against primitives (and each other) by @eeeebbbbrrrr in https://github.com/tcdi/pgx/pull/1029

## Breaking API Change

* remove the  dependency on `cstr_core` by @eeeebbbbrrrr in https://github.com/tcdi/pgx/pull/1025

Our prior usages of `cstr_core` were to work around the fact that Rust's `CStr`/`CString` types required `std`.  As of Rust v1.64, they can be used in `no_std` environments (which pgx, surprisingly!, supports), so there is no longer a need for us to rely on `cstr_core` and to re-export it as our own type.  

You may run into compilation issues if you implement custom Aggregates.

## Something Interesting

* Support extensions on i686-unknown-linux-gnu by @Smittyvb in https://github.com/tcdi/pgx/pull/1014

This ought to be considered experimental.  If you do need to make a 32bit pgx extension and run into problems, please let us know.

**Full Changelog**: https://github.com/tcdi/pgx/compare/v0.7.0...v0.7.1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -58,9 +58,9 @@ dependencies = [
 
 [[package]]
 name = "async-trait"
-version = "0.1.63"
+version = "0.1.64"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "eff18d764974428cf3a9328e23fc5c986f5fbed46e6cd4cdf42544df5d297ec1"
+checksum = "1cd7fce9ba8c3c042128ce72d8b2ddbf3a05747efb67ea0313c635e10bda47a2"
 dependencies = [
  "proc-macro2",
  "quote",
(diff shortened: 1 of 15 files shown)


thomcc (approved):

workingjubilee (approved):
**Very** experimental. But: naisu.

Issue #1032, opened 2023-02-03
NeonDatabase, based on pg15, can't load pgx extensions


vadim2404:
For pg15, ABI is hardcoded to "PostgreSQL" and didn't get it from the source code, whereas C extensions get it and can be compiled.

eeeebbbbrrrr:
I looked at your issue 3535 and, I mean, pgx only knows how to do the “PostgreSQL” ABI, not the custom “Neon PostgreSQL”. 

Now, I can imagine they’re actually the same but pgx cannot guarantee that. 

I might suggest you not change the ABI name if it’s actually the same.

vadim2404:
> I looked at your issue 3535 and, I mean, pgx only knows how to do the “PostgreSQL” ABI, 

It's the issue. Is it possible to get the ABI name from PG sources while compiling extensions?

eeeebbbbrrrr:
This is from the Postgres sources in `src/include/pg_config_manual.h`:

```c
/*
 * When creating a product derived from PostgreSQL with changes that cause
 * incompatibilities for loadable modules, it is recommended to change this
 * string so that dfmgr.c can refuse to load incompatible modules with a clean
 * error message.  Typical examples that cause incompatibilities are any
 * changes to node tags or node structures.  (Note that dfmgr.c already
 * detects common sources of incompatibilities due to major version
 * differences and due to some changed compile-time constants.  This setting
 * is for catching anything that cannot be detected in a straightforward way.)
 *
 * There is no prescribed format for the string.  The suggestion is to include
 * product or company name, and optionally any internally-relevant ABI
 * version.  Example: "ACME Postgres/1.2".  Note that the string will appear
 * in a user-facing error message if an ABI mismatch is detected.
 */
#define FMGR_ABI_EXTRA          "PostgreSQL"
```

Whatever "Neon PostgreSQL" is indicates that this is an ABI that is not compatible with "PostgreSQL".  "PostgreSQL" is the only ABI that pgx understands.

This is a new feature in pg15, that came in this commit:

```
commit d6d1dfcc99e3dd6e70e2a7024924e491bb7a9670
Author: Peter Eisentraut <peter@eisentraut.org>
Date:   Mon Nov 22 07:40:17 2021 +0100

    Add ABI extra field to fmgr magic block
    
    This allows derived products to intentionally make their fmgr ABI
    incompatible, with a clean error message.
    
[...]

PR #1033, opened 2023-02-03
Fix issue #1032

eeeebbbbrrrr:
Blindly use the `FMGR_ABI_EXTRA` constant as part of the magic block. Hardcoding to "PostgreSQL" isn't exactly correct when using bindings from, for example, a Postgres fork that has purposely changed its ABI "name".

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,13 +251,15 @@ dear compiler runner, can make this guarantee for yourself by specifying the `un
 feature flag.  Otherwise, a pgx extension will fail to compile with an error similar to:
 
 ```
-error: Unrecognized Postgres ABI.  Perhaps you need `--features unsafe-postgres`?
-   --> pgx/src/lib.rs:139:9
+error[E0080]: evaluation of constant value failed
+   --> pgx/src/lib.rs:151:5
     |
-139 | /         compile_error!(
-140 | |             "Unrecognized Postgres ABI.  Perhaps you need `--features unsafe-postgres`?"
-141 | |         );
-    | |_________^
+151 | /     assert!(
+152 | |         same_slice(pg_sys::FMGR_ABI_EXTRA, b"xPostgreSQL\0"),
+153 | |         "Unsupported Postgres ABI. Perhaps you need `--features unsafe-postgres`?",
+154 | |     );
+    | |_____^ the evaluated program panicked at 'Unsupported Postgres ABI. Perhaps you need `--features unsafe-postgres`?', pgx/src/lib.rs:151:5
+    |
 ```
 
 ### Experimental Features
--- a/README.md
+++ b/README.md
@@ -242,6 +242,24 @@ this feature is now considered deprecated in favor of a lower-overhead interop.
 You may still request implementations of `TryFrom<time::Type> for pgx::MatchingType`
 and `From<time::Type> for pgx::MatchingType` by enabling the `"time-crate"` feature.
 
+### "unsafe-postgres": Allow compilation for Postgres forks that have a different ABI
+
+As of Postgres v15, forks are allowed to specify they use a different ABI than canonical Postgres.
+Since pgx makes countless assumptions about Postgres' internal ABI it is not possible for it to 
+guarantee that a compiled pgx extension will probably execute within such a Postgres fork.  You,
+dear compiler runner, can make this guarantee for yourself by specifying the `unsafe-postgres` 
+feature flag.  Otherwise, a pgx extension will fail to compile with an error simiar to:
+
+```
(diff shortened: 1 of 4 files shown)


workingjubilee (changes requested):
This violates the compile-time guarantees we have provided until now. That is: we have previously assumed the responsibility for making PGX work against Postgres.

We should not enable this change except conditionally on the user explicitly requesting this be overwritten at compile time, in a way that we can detect when an issue is opened against PGX. It should be the user's responsibility, not ours, for whether their extension works against a bespoke Postgres fork.

eeeebbbbrrrr:
> This violates the compile-time guarantees we have provided until now. That is: we have previously assumed the responsibility for making PGX work against Postgres.
> 
> We should not enable this change except conditionally on the user explicitly requesting this be overwritten at compile time, in a way that we can detect when an issue is opened against PGX. It should be the user's responsibility, not ours, for whether their extension works against a bespoke Postgres fork.

What are you thinking here?  `--features this_isnt_postgres`?

workingjubilee:
I _think_ a feature name can use all of Unicode, so we could go with `"unsafe-postgres-doppelgänger"` so it's annoying to type.

eeeebbbbrrrr:
and do we raise a compilation error if it's not "PostgreSQL" and that feature doesn't exist?

Doing this pushes the burden of supporting Postgres forks onto pgx extension developers.  Maybe that's where it belongs tho.  We can't know everything.

workingjubilee:
Yes, I think so. We could also lift this checking into `cargo pgx`, effectively, and then it's the fault of whoever runs the compiler.

eeeebbbbrrrr:
> Yes, I think so. We could also lift this checking into `cargo pgx`, effectively, and then it's the fault of whoever runs the compiler.

Well, one can compile a pgx extension without `cargo-pgx`, so I don't think that's super valuable.  Lets go with a feature flag of `--features unsafe-postgres` and a compilation error without it.

I'll update the PR and we can see what it looks like.

workingjubilee (approved):

README.md
```diff
@@ -242,6 +242,24 @@ this feature is now considered deprecated in favor of a lower-overhead interop.
 You may still request implementations of `TryFrom<time::Type> for pgx::MatchingType`
 and `From<time::Type> for pgx::MatchingType` by enabling the `"time-crate"` feature.
 
+### "unsafe-postgres": Allow compilation for Postgres forks that have a different ABI
+
+As of Postgres v15, forks are allowed to specify they use a different ABI than canonical Postgres.
+Since pgx makes countless assumptions about Postgres' internal ABI it is not possible for it to 
+guarantee that a compiled pgx extension will probably execute within such a Postgres fork.  You,
+dear compiler runner, can make this guarantee for yourself by specifying the `unsafe-postgres` 
+feature flag.  Otherwise, a pgx extension will fail to compile with an error simiar to:
```
Smittyvb, line 251: typo:
```suggestion
feature flag.  Otherwise, a pgx extension will fail to compile with an error similar to:
```

eeeebbbbrrrr:
this isn't working out like I had hoped.  #grr

pgx/src/lib.rs
```diff
@@ -128,6 +128,20 @@ pub use pg_sys::{
...
+// Postgres v15 has the concept of an ABI "name".  The default is `b"PostgreSQL\0"` and this is the
+// ABI that pgx extensions expect to be running under.  We will refuse to compile if it is detected
+// that we're trying to be built against some other kind of "postgres" that has its own ABI name.
+//
+// Unless the compiling user explicitly told us that they're aware of this via `--features unsafe-postgres`.
+#[cfg(all(feature = "pg15", not(feature = "unsafe-postgres")))]
+const _: () = {
+    if crate::pg_sys::FMGR_ABI_EXTRA.iter().zip(b"PostgreSQL\0".iter()).all(|(a, b)| a == b) {
+        compile_error!(
+            "Unrecognized Postgres ABI.  Perhaps you need `--features unsafe-postgres`?"
```
Smittyvb, line 140: We should error if it doesn't match, not if it does match.
```suggestion
    if !crate::pg_sys::FMGR_ABI_EXTRA.iter().zip(b"PostgreSQL\0".iter()).all(|(a, b)| a == b) {
        compile_error!(
            "Unrecognized Postgres ABI.  Perhaps you need `--features unsafe-postgres`?"
```

eeeebbbbrrrr: Yeah, I'm learning how to program today.  See latest push.

thomcc (approved):
Looks good to me (but I did write some of it 😅)

eeeebbbbrrrr:
You wrote all of it.  In the discord input box!

PR #1034, opened 2023-02-03
Ensure the `libpq{,N}-dev` package is removed in CI

thomcc:
In apt `libpq-dev` (or something like `libpq5-dev`) provides a `pg_config` binary, so that the client libraries can be located. We should remove that with the other stuff, so that we can catch cases wehre we are accidentally using thw wrong `pg_config` (see #981).

I also changed the globs to regexs, since I'm not sure how the globbing was working in the first place (shouldn't the glob expand to paths) and the internet says that it `apt remove` can be given a regex.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -56,7 +56,7 @@ jobs:
         echo ""
 
         echo "----- Remove old postgres -----"
-        sudo apt remove -y postgres*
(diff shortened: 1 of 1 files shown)


BradyBonnette:
Looks good, let's see what CI says

BradyBonnette (approved):
Let's do it.

thomcc:
Crap, that should have been against develop! Well, it's CI only...

PR #1035, opened 2023-02-06
Disable thin lto for dev builds by default in template

JelteF:
When changing something trivial in the hello world project created by
`cargo pgx new`, it would take ~20 seconds to compile a dev build of the
generated crate on my machine. This seemed way slower than it should be.
It turns out that the thin LTO setting greatly inrceases compilation
times. By disabling it, compilation time of the crate is only 1 second.

As far as I can tell this setting was enabled initially as a workaround
for this bug in Rust: https://github.com/rust-lang/rust/issues/50007
Since this bug has been fixed since Rust 1.62, and current stable Rust
is version 1.67, this PR removes the workaround.

### DIFF
--- a/cargo-pgx/src/templates/cargo_toml
+++ b/cargo-pgx/src/templates/cargo_toml
@@ -23,7 +23,6 @@ pgx-tests = "=0.7.1"
 
 [profile.dev]
 panic = "unwind"
-lto = "thin"
 
 [profile.release]
 panic = "unwind"
--- a/nix/templates/default/Cargo.toml
+++ b/nix/templates/default/Cargo.toml
@@ -25,7 +25,6 @@ once_cell = "1.7.2"
 
 [profile.dev]
 panic = "unwind"
-lto = "thin"
 
 [profile.release]
 panic = "unwind"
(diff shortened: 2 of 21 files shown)


eeeebbbbrrrr:
We like PRs against the “develop” branch. Would you mind retargeting this?

otherwise this is perfectly fine. I guess I’m not surprised thin LTO takes so long and we all forgot this was there. Thanks!

eeeebbbbrrrr (approved):

JelteF:
> We like PRs against the “develop” branch. Would you mind retargeting this?

Done

eeeebbbbrrrr:
CI is angry at the universe today, it seems.  It's not this PR.  We'll give it one more chance...

thomcc (changes requested):
You need to change this in the `cargo-pgx` template too, [here](https://github.com/tcdi/pgx/blob/master/cargo-pgx/src/templates/cargo_toml). The one you've changed is part of some nix-specific stuff that most people won't use.

JelteF:
> You need to change this in the cargo-pgx template too, [here](https://github.com/tcdi/pgx/blob/master/cargo-pgx/src/templates/cargo_toml?rgh-link-date=2023-02-06T15%3A50%3A34Z).

Done, I also removed the same line from all the examples in the repo. (Even though it was already commented out there, but the comment above it suggests to uncomment it)

PR #1036, opened 2023-02-07
Fix outdated documentation for `pgx::name!`

Smittyvb:
Since a few versions ago, `TableIterator`s are used instead of an iterator of tuples for table returning functions. The documentation for `pgx::name!` didn't reflect that change though, this PR fixes it.


workingjubilee (approved):
Thank you!

PR #1039, opened 2023-02-11
Made Date::to_unix_epoch_days() public

osawyerr:
Made Date::to_unix_epoch_days() public. @eeeebbbbrrrr confirmed there was no specific reason why its private

### DIFF
--- a/pgx/src/datum/date.rs
+++ b/pgx/src/datum/date.rs
@@ -83,6 +83,7 @@ impl Date {
         self.0
     }
 
+    /// Returns the date as an i32 representing the elapsed time since UNIX epoch in days
     #[inline]
     pub fn to_unix_epoch_days(&self) -> i32 {
         self.0 + POSTGRES_EPOCH_JDATE - UNIX_EPOCH_JDATE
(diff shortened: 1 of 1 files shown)


workingjubilee (changes requested):
Before it is made public it should also clearly and publicly document what a "Unix epoch day" _is_, because it's a totally fictional concept without referent anywhere else, one that only exists in the codebase to bridge some math.

osawyerr:
Unix epoch day is not entirely a fictional / unknown concept it’s used in other date APIs as well like apache arrow - https://docs.rs/arrow/latest/arrow/datatypes/struct.Date32Type.html

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

workingjubilee (approved):
In that case we mostly just have to make sure which fictions we count as canonical in our Extended Universe.

jaskij:
If I'm reading the code right, this is the number of [Julian days](https://en.wikipedia.org/wiki/Julian_day) since the Unix epoch - and should probably be described as such, as "Julian day" is a much better defined and precise concept than just a "day".

PR #1043, opened 2023-02-16
Fixes #1040

### LINKED ISSUE #1040: Make `pgx::Range::from_pg` an `unsafe fn`
https://github.com/tcdi/pgx/blob/24ad866922b312e51e71b568f4a955e0ac74afdd/pgx/src/datum/range.rs#L33-L39

This is only internal, so its safety isn't a critical concern, but this can't actually detect, as-written, if it's pointing at a valid datum. It can only detect nullness and panic in that case.

mhov:


mhov:
@workingjubilee does this look right?

eeeebbbbrrrr (approved):
LGTM.  Thanks for the follow-up, @mhov!

workingjubilee (commented):
Yep, this looks good!

PR #1044, opened 2023-02-16
fix: check root by uid

skyzh:
In environment like docker containers, `USER` environment variable is not set, and therefore it might be a better idea to check root by UID instead of USER env variable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -263,6 +263,7 @@ dependencies = [
  "flate2",
  "fork",
  "libloading",
+ "nix",
  "num_cpus",
  "object 0.28.4",
  "once_cell",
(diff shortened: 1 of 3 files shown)


eeeebbbbrrrr:
Please note that we prefer PRs against our `develop` branch, as noted in our [CONTRIBUTING.md](https://github.com/tcdi/pgx/blob/master/CONTRIBUTING.md) document.  I've retargeted this one for you.

skyzh:
Thanks! Sorry for forgetting reading the contribution guide 🙈

eeeebbbbrrrr (approved):
Thanks for the contribution!

PR #1047, opened 2023-02-16
[WIP] pg-config

ChuckHend:
Following on https://discord.com/channels/561648697805504526/749453981494804590/1075794824243982469

Looking to have a convenient way of accessing the `pg_config` that `pgx` is using. That information lives int `${PGX_HOME}/config.toml`, but I think it makes sense to expose it to the user.

Considered `cargo pgx get`, but `get` does a lot of other things specific to a pgx project. Implementing as a new command keeps things isolated/higher level.

`cargo pgx pg-config pg14 describe` could exec `~/.pgx/14.6/pgx-install/bin/pg_config` , but thats not build into this PR yet.


ChuckHend:
Found this: https://github.com/tcdi/pgx/blob/24ad866922b312e51e71b568f4a955e0ac74afdd/pgx-pg-config/src/lib.rs#L545

Should get me more direct access to the toml.

PR #1050, opened 2023-02-19
overhaul our Postgres Range support

eeeebbbbrrrr:
- We now go ahead and deconstruct the range value in the `FromDatum` implementation.  Immediately, this fixes issues with `pg_sys::RangeBound.val` datums potentially being used after free.

- `RangeData` is gone and replaced with a more idiomatic `RangeBound` enum, to which the internal Postgres `pg_sys::RangeBound` struct can be mapped.

- We provide `From<std::ops::Range*>` impls for all of Rust's built-in Range types.

- Our `pgx::Range` can be constructed from any type that can be converted into a `RangeBound` and we provide direct impls for the set of supported `RangeSubType` types along with `Option<T: RangeSubType>` which maps `Some` to `::Inclusive` and `None` to `::Infinite`.

- The idea that a Postgres range can be "empty" is properly relayed through `Range`'s public API

- add an example (albeit puny)


eeeebbbbrrrr:
Apologizes to @mhov... this isn't very faithful to your original impl from #663, but it has greatly improved ergonomics from the user-facing API side along with better interoperability with Rust's `std::ops::Range*` types.

I'd appreciate it if you'd look it over and let me know what you think.  My intention is to merge this in the next day or two.

mhov:
I wasn't attached to my implementation, no worries. As a serious RangeType end user, I'll try it out and review soon. Thanks @eeeebbbbrrrr

mhov:
Ok, 100% more ergonomic and doesn't look like it was someone first year doing Rust.  My main concern with _my_ implementation was Range<T>->Range<T> was an implicit `From<,>` and so it wasn't obvious how you get the data without reading the code, so it's way easier now. 
My concern with _this_ implementation is having non-trivial conversion work in the `FromDatum`/`IntoDatum` 

@workingjubilee might be able to elaborate better than me, but in the 4.5.0->0.5.0 period there was an effort to move non-trivial conversion logic out of the Into/FromDatum to the user code so it wouldn't cost anything until you actually needed it.
ex:
https://github.com/tcdi/pgx/issues/504
https://github.com/tcdi/pgx/pull/643
https://github.com/tcdi/pgx/pull/622

Corner-case as it may be if we had a `pg_extern` with an input param of array of 10,000  dates/timestamp/daterange/intrange/etc there would 10,000 conversion overheads before line 1 of your actual fn logic.  A few of us were hitting these bumps in our extension migration.

With `*RangeType` there's the deserialization/serialization step PLUS the TypeCache lookup for each datum into/from.   In C-land if i had an array of 10,000 daterange (which i do quite often in my work) I'd get the typecache once and re-use it  for 10k calls to `range_deserialize` and/or `make_range`.  I had meant to add that as an option to pass to my original `Range<T>/RangeData<T>` methods.

I can think of a bunch of other contrived cases which may, understandably, lose out to ergonomics.  Please tell me if I'm over-optimizing.  As a C extension developer of a very performance-sensitive project, I'm in the zero-copy pgx gang.  I'm afraid of losing any time to pgx  wrappers/overhead.  I understand I'm a different user from those who have never done C and their first pg extension is in `pgx` (which should be a target audience).

The great part about the `SqlTranslatable` change a while back, is I can easily add my ugly/optimized version of Range to my extension project without needing pgx to include it.  So if none of the overhead is a concern to you, I have easy options to work around it. 

Side question, would we ever want a `contrib` section of `pgx` for unfriendly, but possibly faster, Type implementations? :)

eeeebbbbrrrr:
> Corner-case as it may be if we had a pg_extern with an input param of array of 10,000 dates/timestamp/daterange/intrange/etc there would 10,000 conversion overheads before line 1 of your actual fn logic. A few of us were hitting these bumps in our extension migration.

Are you using `pgx::Array<T>` which is lazily evaluated, or `Vec<T>` which is not?

Yes, this PR does the FromDatum conversion on both bounds when the Range itself is deconstructed, but these conversions only now only happen once per Range, whereas the current implementation would do that conversion every time the user asked for one of the bound values.

There was also a problem where the Datums behind the lower/upper bounds could have been pfree'd out from underneath the user without them knowing (that's a problem with all "zero copy" type impls, especially the FromDatum impl for &str), leading to UAF.  And this could be a problem for any pass-by-reference range type -- specifically `numrange`.  

As it is now, for pass-by-value ranges such as int4/8range and date/timerange types, I'd expect the compiler to basically optimize all the machinery away such that the bounds become the backing Datum value, as-is.  However, `numrange` requires that we get the Datum's detoasted as soon as possible.

---

Circling back to large arrays of values... the point is well received, and I think it underscores that we need to do more work on making array support fully zero-copy and lazily evaluated where we can.  Although, `pgx::Array<T>` is basically there, Rather than optimizing individual types that then rely on postgres-allocated memory to which we don't have a lifetime attached.

eeeebbbbrrrr:
Thinking about the TypeCache thing...

I'm not sure how we can mitigate that right now.  It would require some thought and changes to `pgx::Array` in order to reduce it to one call per `Array<Range<T>>` (or even `Vec<Range<T>>`).  But it's probably worth investigating how we could push it down.

eeeebbbbrrrr:
I had to recreate this branch with cherry-picks and do a force push because I have no idea.

workingjubilee:
git be like that sometimes.

README.md
```diff
@@ -163,39 +163,45 @@ cargo pgx init
...
+| `"char"`                   | `i8`                                                  |
+| `smallint`                 | `i16`                                                 |
+| `integer`                  | `i32`                                                 |
+| `bigint`                   | `i64`                                                 |
+| `oid`                      | `u32`                                                 |
+| `real`                     | `f32`                                                 |
+| `double precision`         | `f64`                                                 |
+| `bool`                     | `bool`                                                |
+| `json`                     | `pgx::Json(serde_json::Value)`                        |
+| `jsonb`                    | `pgx::JsonB(serde_json::Value)`                       |
```
workingjubilee, line 180: Hm. Is this implementation detail public? Is there a need to spell this out for users?

README.md
```diff
@@ -163,39 +163,45 @@ cargo pgx init
...
+| `timestamp`                | `pgx::Timestamp`                                      |
+| `time with time zone`      | `pgx::TimeWithTimeZone`                               |
+| `timestamp with time zone` | `pgx::TimestampWithTimeZone`                          |
+| `anyarray`                 | `pgx::AnyArray`                                       |
+| `anyelement`               | `pgx::AnyElement`                                     |
+| `box`                      | `pgx::pg_sys::BOX`                                    |
+| `point`                    | `pgx::pgx_sys::Point`                                 |
+| `tid`                      | `pgx::pg_sys::ItemPointerData`                        |
+| `cstring`                  | `&core::ffi::CStr`                                    |
+| `inet`                     | `pgx::Inet(String)` -- TODO: needs better support     |
```
workingjubilee, line 192: See previous remark.

README.md
```diff
@@ -163,39 +163,45 @@ cargo pgx init
...
+| `ARRAY[]::<type>`          | `Vec<Option<T>>` or `pgx::Array<T>` (zero-copy)       |
+| `int4range`                | `pgx::Range<i32>`                                     |
+| `int8range`                | `pgx::Range<i64>`                                     |
+| `numrange`                 | `pgx::Range<Numeric<P, S>>` or `pgx::Range<AnyRange>` |
+| `daterange`                | `pgx::Range<pgx::Date>`                               |
+| `tsrange`                  | `pgx::Range<pgx::Timestamp>`                          |
+| `tstzrange`                | `pgx::Range<pgx::TimestampWithTimeZone>`              |
+| `NULL`                     | `Option::None`                                        |
+| `internal`                 | `pgx::PgBox<T>` where `T` is any Rust/Postgres struct |
+| `uuid`                     | `pgx::Uuid([u8; 16])`                                 |
```
workingjubilee, line 204: See previous remark.

These don't have to be changed since this is preexisting, however they caught my eye since this was diffed.

pgx/src/datum/range.rs
```diff
@@ -9,56 +9,281 @@ Use of this source code is governed by the MIT license that can be found in the
...
+                val: v.into_datum().unwrap(),
+                infinite: false,
+                inclusive: true,
+                lower: false,
+            },
+            RangeBound::Exclusive(v) => pg_sys::RangeBound {
+                val: v.into_datum().unwrap(),
+                infinite: false,
+                inclusive: false,
+                lower: false,
```
workingjubilee, line 65: Considering the repeated logic here, do you think it would make sense for pgx-pg-sys to contain an `impl Default for pg_sys::RangeBound`? `pg_sys::Datum` doesn't have an impl for Default because pointers don't, but in this case it feels correct to have the default be the same as `RangeBound::Infinite` here.

Then these could reuse that like `pg_sys::RangeBound { val: v.into_datum().unwrap(), infinite: false, ..Default::default() }`. Though that's only saving us a few lines and may require cross-reference so... maybe not.

pgx/src/datum/range.rs
```diff
@@ -9,56 +9,281 @@ Use of this source code is governed by the MIT license that can be found in the
...
 {
-    /// ## Safety
-    /// This function is safe, but requires that
-    /// - datum is not null
-    /// - datum represents a PG RangeType datum
     #[inline]
-    unsafe fn from_pg(datum: pg_sys::Datum) -> Option<Self> {
-        unsafe { Self::from_polymorphic_datum(datum, false, T::range_type_oid()) }
+    fn from(value: &RangeBound<T>) -> Self {
+        Clone::clone(value)
```
workingjubilee, line 95: This is an atypical impl, I'd be inclined to simply remove it.

pgx/src/datum/range.rs
```diff
@@ -9,56 +9,281 @@ Use of this source code is governed by the MIT license that can be found in the
...
+        match self {
+            RangeBound::Infinite => None,
+            RangeBound::Inclusive(v) | RangeBound::Exclusive(v) => Some(v),
+        }
+    }
+
+    /// Convert this pgx [`RangeBound`] into the equivalent Postgres [`pg_sys::RangeBound`].
+    ///
+    /// Note that the `lower` property is always set to false as a [`RangeBound`] doesn't know the
+    /// end on which it's placed.
```
workingjubilee, line 46: The other question is if we should be exposing an `into_pg` that allows constructing a `RangeBound` that doesn't actually know whether it is the `lower` range or not yet says it is not, instead of requiring people to use a similar function on the larger `Range`. I don't know if it's common to just manipulate `RangeBound`s separately through Postgres functions, though.

pgx/src/datum/range.rs
```diff
@@ -9,56 +9,281 @@ Use of this source code is governed by the MIT license that can be found in the
...
+    /// Consumes `self` and returns the contained [`RangeBound`]s.
+    ///
+    /// # Panics
+    ///
+    /// This function panics if this [`Range`] represents an "empty" range.
+    #[inline]
+    #[track_caller]
+    pub fn unwrap(self) -> (RangeBound<T>, RangeBound<T>) {
+        self.inner.unwrap()
+    }
```
workingjubilee, line 248: This feels like a dubious function to expose since `into_inner` and `take` already exist, unlike `replace` where we need to adjust for the fact that `Option::replace` isn't right for that use.

workingjubilee (commented):
Sorry it took me a bit to get through things on top of the squirreling on other topics. I probably need to glance over it again still.

README.md
eeeebbbbrrrr: yes

README.md
eeeebbbbrrrr: yes.  All these were there before you're just noticing them b/c intellij forced me to reformat the table.

pgx/src/datum/range.rs
eeeebbbbrrrr: all the bindgen generated structs already `#[derive(Default])`, which of course will set `infinite` to `false`.  Not sure it's worth special-casing this one thing.

README.md
workingjubilee: I guess the other part of my question is: should it be?

pgx/src/datum/range.rs
eeeebbbbrrrr: we can make that function private.  It's used in the `IntoDatum` impl.

pgx/src/datum/range.rs
eeeebbbbrrrr: I argued with myself over this one.  Happy to remove it.

pgx/src/datum/range.rs
eeeebbbbrrrr: I think I did this to appease the compiler.  I'll double-check.

pgx/src/datum/range.rs
workingjubilee: ...wow you're right, that sure is... an impl of `Default` there.

workingjubilee: ...I guess we can uhhh, use... that... then...
**wow**.

README.md
eeeebbbbrrrr: When we get around to more complete APIs for these types we can lock it down. I suppose

pgx/src/datum/range.rs
eeeebbbbrrrr: removed.

pgx/src/datum/range.rs
eeeebbbbrrrr: yeah, It was kinda being used by the tests, but changing them to just `.clone()` its `RangeBound` reference is totally fine.  done.

eeeebbbbrrrr:
@workingjubilee I've addressed all your concerns and for fun, added a few `RangeBound::is_XXX()` functions, along with implementing Display for Range.

workingjubilee (approved):
Thank you! Yes, this looks good now.

PR #1051, opened 2023-02-21
Some cleanups around postgrestd and memctx handling

thomcc:


eeeebbbbrrrr:
@thomcc, can we also remove the `postgrestd` (and `plrust`) features in this PR?  They're both unnecessary now.  

I guess you can also just delete the `PgMemoryContexts::leak_trivial_alloc()` function.  We don't use it at all and it seems like not a thing we should present to users anyways.

workingjubilee (approved):
I'll take care of that.

PR #1052, opened 2023-02-21
a better Error type when conversion from Date to time::Date doesn't work

eeeebbbbrrrr:


pgx/src/datum/date.rs
```diff
@@ -98,24 +98,54 @@ impl Date {
...
-    fn from(date: time::Date) -> Self {
-        Date::from_pg_epoch_days(date.to_julian_day() - POSTGRES_EPOCH_JDATE)
+mod with_time_crate {
+    use crate::{Date, POSTGRES_EPOCH_JDATE};
+    use core::fmt::{Display, Formatter};
+    use std::error::Error;
+
+    #[derive(Debug, PartialEq, Clone)]
+    #[non_exhaustive]
+    pub struct TryFromDateError(pub Date);
```
thomcc, line 108: Do we want to export this in a way users can access it?

thomcc (approved):
It looks mostly okay to me, but the error type probably should be exported unless we have a reason not to.

pgx/src/datum/date.rs
eeeebbbbrrrr: I don’t see that they’ll need to reference it by name.

eeeebbbbrrrr: It’s under this time-crate feature flag anyways. And honestly, I hope that’ll go away soon

eeeebbbbrrrr: alrighty.  done

PR #1055, opened 2023-02-22
Cleanup `PgTrigger` such that its public API can now be safe.

eeeebbbbrrrr:
This pushes more responsibility on to the user constructing a `PgTrigger`, but that responsibility was already there anyways.

We also add the concept of an `Old` trigger tuple, to go along with the existing `New`.  `Current` has been removed entirely -- it's not a common procedural language idiom and was too confusing next to both `New` and `Old` exist.

The `PgTriggerSafe` struct is completely deleted -- it's no longer necessary.

`PgTrigger` now has a lifetime and borrows from its source `pg_sys::FunctionCallInfoBaseData`.
    
This ensures that a `PgTrigger` instance, which essentially borrows pointers from an "fcinfo" doesn't outlive the "fcinfo" from which it was created.  We also clean up the members of `PgTrigger` to just the bare minimum we need.
    
Another benefit here is that the `PgTrigger::relation(&self)` function can now directly use the `RelationData` pointer that's already been opened for us by Postgres.
    
Adding a lifetime to `PgTrigger` caused a bunch of touch points in our tests and examples, so I went ahead and fixed another issue at the same time.

`#[pg_trigger]` functions now must return `Result<Option<PgHeapTuple>, E>`.  It is desired that a trigger function be able to "return NULL" to indicate that the operation that caused it to fire should **not** be performed.  Prior to this change there was no way to do that.  A test has been added for this case as well.

This all came out of https://github.com/tcdi/plrust/issues/209


pgx/src/trigger_support/trigger_tuple.rs
```diff
@@ -1,14 +1,13 @@
...
-    /// a different one or skip the operation.
+    /// Represents the new database row for INSERT/UPDATE operations in row-level triggers.
     New,
+
+    /// Represents the old database row for UPDATE/DELETE operations in row-level triggers.
+    Old,
+
+    /// Represents the old database row for UPDATE/DELETE operations or the new database row for
+    /// INSERT operations in row-level triggers.
+    Current,
```
workingjubilee, line 12: Hm... I don't think I understand the difference between `TriggerTuple::Current` and either of the other two variants, now. It seems there's only Current. It's not intuitive how to use the other two? If these are outputs instead of inputs, then if we are able to intuit which approach is appropriate, we shouldn't need `TriggerTuple::Current` to tell us that, we should simply _do it_.

pgx/src/trigger_support/pg_trigger.rs
```diff
@@ -247,27 +269,4 @@ impl PgTrigger {
     pub fn fcinfo(&self) -> &pg_sys::FunctionCallInfo {
         self.fcinfo.borrow()
     }
-
-    /// Eagerly evaluate the data in this `PgTrigger` and build a safely accessible structure
-    /// which mimics the data provided to a PL/pgSQL trigger.
-    pub unsafe fn to_safe(&self) -> Result<PgTriggerSafe, PgTriggerError> {
```
workingjubilee, line 253: It's not immediately obvious: do we lose the "lazy evaluation" quality here? This seems fine otherwise.

pgx/src/trigger_support/pg_trigger.rs
```diff
@@ -159,51 +193,39 @@ impl PgTrigger {
...
-    /// The caller should already have at least AccessShareLock on the relation ID, else there are nasty race conditions.
-    ///
-    /// As such, this function is unsafe as we cannot guarantee that this requirement is true.
-    pub unsafe fn relation(&self) -> Result<crate::PgRelation, PgTriggerError> {
-        let relation = PgRelation::open(self.relation_data.rd_id);
-        Ok(relation)
+    pub fn relation(&self) -> Result<crate::PgRelation, PgTriggerError> {
+        // SAFETY:  the creator of this PgTrigger has asserted that the provided "fcinfo" struct
+        // contains proper trigger data, and that would include the fact that the relation already
+        // has at least a share lock created by Postgres
```
workingjubilee, line 205: Have we checked the fcinfo struct for this at this point? Can we? It's one thing to rely on Postgres getting the layout correct, but I don't like relying on Postgres not null-initializing a field.

pgx/src/heap_tuple.rs
```diff
@@ -88,31 +90,68 @@ impl<'a> PgHeapTuple<'a, AllocatedByPostgres> {
...
+                }
+            }
         };
 
-        if tuple.is_null() {
-            return None;
-        }
+        // at this point we should not be trying to return a NULL trigger tuple.
+        // We should have done that above as an early return
+        assert!(!tuple.is_null());
```
workingjubilee, line 143: What logic path do we have that could possibly have reached this point?

pgx/src/trigger_support/trigger_tuple.rs
workingjubilee: Hm, so I guess `New` and `Current` were intended in lieu of `New` and `Old` once upon a time?

pgx/src/trigger_support/pg_trigger.rs
eeeebbbbrrrr: Since `PgTriggerSafe` is gone users will only use `PgTrigger` and that struct *is* lazy.  It does next to nothing during its `from_fcinfo()` function, which is how it's created.

pgx/src/trigger_support/trigger_tuple.rs
eeeebbbbrrrr: Yeah, and I can't tell ya why.

pgx/src/trigger_support/pg_trigger.rs
eeeebbbbrrrr: I don't know that we can.  Especially the invariant of "table should have an AccessShareLock".

I'll take a look at the `from_fcinfo()` function and see if there's a little bit more validation we can do.

pgx/src/heap_tuple.rs
eeeebbbbrrrr: This was just a guard against the `trigger_data.tg_trigtuple/tg_newtuple` fields being NULL when we should have the proper logic in place to ensure that won't be possible, based on how the trigger function is being called from Postgres.

I'll expand the comment.

pgx/src/trigger_support/pg_trigger.rs
workingjubilee: Is it possible for someone to wrap this struct in an `Arc`, hang it in a `static` like `static SHENANIGANS: Arc<Mutex<Arc<PgTrigger>` and then later access this struct and attempt to call this function on it when we don't have a lock?

eeeebbbbrrrr: That's the least of the problems:

```rust
pub struct PgTrigger {
    trigger: pg_sys::Trigger,
    trigger_data: PgBox<pgx_pg_sys::TriggerData>,
    relation_data: pg_sys::RelationData,
    #[allow(dead_code)]
}
```

We can maybe add a lifetime to `PgTrigger` and tie it to the `fcinfo` used to create it?  That would prevent such things, yeah?

workingjubilee: Hm. I don't know if a lifetime is enough. The appropriate interface may be to accept a closure and run that with "safe access to the unsafe accessors", without yielding that type to the outside world. I'm not sure, I haven't seen many triggers so I don't know what they look like.

eeeebbbbrrrr:
```rust
#[pg_trigger]
fn trigger_example(
    trigger: &pgx::PgTrigger,
) -> Result<PgHeapTuple<'_, impl WhoAllocated>, TriggerError> {
    let old = trigger.current().ok_or(TriggerError::NullOld)?;

    let mut current = old.into_owned();
    let col_name = "title";

    if current.get_by_name(col_name)? == Some("Fox") {
        current.set_by_name(col_name, "Bear")?;
    }

    Ok(current)
}
```

That's an example from our, um, examples.

eeeebbbbrrrr: Postgres is going to give us the underlying `fcinfo` struct properly setup.

If the concern is if a user can then hold onto this `trigger: &pgx::Trigger` reference for longer than this function... I'm all ears on how to fix that.  I guess by virtue of the fact it is a reference already means it has a lifetime associated with it.

workingjubilee: Conditionalizing it on `&mut pg_sys::FunctionCallInfo` might be enough since that's exclusive access, however, if a programmer gets access to that `&mut pg_sys::FunctionCallInfo`, they can run `mem::take` on the FunctionCallInfo. That would be invalid since it's not `Default`, but...

```rust
mem::replace(&mut fcinfo, ptr::null_mut())
```
is not.

It may need `Pin` if they can touch it, assuming we want to be able to rely on it being valid after its use. The problem with it currently is that the current unsafe constructor (correctly) consumes the fcinfo.

workingjubilee: Well that `&PgTrigger` is fine.

workingjubilee: Okay, if `&mut PgTrigger` or owned PgTrigger is only accessible from inside the "function wrapper" and not the "user function" then we're cool.

eeeebbbbrrrr: I believe that https://github.com/tcdi/pgx/commit/5fc5a19f78e617adc494ea42aedc8ad13bdb6db4 addresses this nicely.  make sure to read its commit message  ;)

pgx/src/trigger_support/trigger_tuple.rs
eeeebbbbrrrr: `Current` is dead to us:  https://github.com/tcdi/pgx/commit/b68a29f85db91c60253fc1029b49fd0d302b2cfc

pgx/src/heap_tuple.rs
eeeebbbbrrrr: See if this is any better:  https://github.com/tcdi/pgx/commit/0f840d23cc9db2f1334711cf6e8a89f84b9e9df6

eeeebbbbrrrr:
Thanks for the review @workingjubilee, plus the extra offline discussions.  I think as of [0f840d2](https://github.com/tcdi/pgx/pull/1055/commits/0f840d23cc9db2f1334711cf6e8a89f84b9e9df6) everything is addressed to the best of my abilities.

workingjubilee (approved):
This looks good!

PR #1059, opened 2023-02-28
Bump tempfile dep to 3.4.0, and several other deps.

thomcc:
This is mostly to bump tempfile, which removes the dep on `remove_dir_all`, which has a (low severity) security vuln in the version our lockfile reports.

Edit: And then I bumped pretty much everything else that was not at the latest semver-major version (or rather, that wasn't compatible with it). I don't think any of these were public deps.


thomcc:
> E: The repository 'https://packages.microsoft.com/ubuntu/20.04/prod focal Release' is no longer signed.
> Error: Process completed with exit code 100.

That's a new error...

cargo-pgx/src/command/install.rs
```diff
@@ -400,6 +400,9 @@ pub(crate) fn get_version(manifest_path: impl AsRef<Path>) -> eyre::Result<Strin
                     .wrap_err("Couldn't parse manifest")?;
 
                 let version = manifest.package.ok_or(eyre!("no `[package]` section found"))?.version;
+                let Inheritable::Set(version) = version else {
```
thomcc, line 403: I should make sure the manifest is parsed with https://docs.rs/cargo_toml/latest/cargo_toml/struct.Manifest.html#method.complete_from_path, so that this case can't happen.

eeeebbbbrrrr (approved):

PR #1060, opened 2023-03-01
Add safe quoting functions

### LINKED ISSUE #561: Safe wrappers for SQL string interpolation functions
Provide safe implementations of postgres quoting functions in support of dynamic SQL / query building

Function List [Docs](https://www.postgresql.org/docs/14/functions-string.html)
```sql
parse_ident ( qualified_identifier text [, strict_mode boolean DEFAULT true ] ) → text[]
-- parse_ident('"SomeSchema".someTable') → {SomeSchema,sometable}

quote_ident ( text ) → text
-- quote_ident('Foo bar') → "Foo bar"

quote_literal ( text ) → text
-- quote_literal(E'O\'Reilly') → 'O''Reilly'

quote_literal ( anyelement ) → text
-- quote_literal(42.5) → '42.5'

format ( formatstr text [, formatarg "any" [, ...] ] ) → text
-- format('Hello %s, %1$s', 'World') → Hello World, World
```

The implementation for those functions is [here](https://github.com/postgres/postgres/blob/27b77ecf9f4d5be211900eda54d8155ada50d696/src/backend/utils/adt/quote.c)

----

From [discord comment](https://discord.com/channels/561648697805504526/749454413407584258/973621714837708841) it is currently possible to wrap these functions in user code using:

#### Re-expose via SQL
```rust
use pgx::*;

#[pg_extern]
fn quote_ident(ident: String) -> Option<String> {
    let quoted_ident = unsafe {
        direct_function_call::<String>(pg_sys::quote_ident, vec![ident.into_datum()]).unwrap()
    };
    Spi::get_one::<String>(&format!("SELECT {quoted_ident} FROM some_table"))
}
```

#### Expose to rust
```rust
use pgx::*;

#[pg_extern]
fn quote_ident(ident: String) -> String {
    unsafe {
        direct_function_call::<String>(pg_sys::quote_ident, vec![ident.into_datum()]).unwrap()
    }
}
```

JelteF:
When building SQL that should be executed using SPI quoting identifiers
and literals is useful. Postgres provides C functions for this purpose,
but these are unsafe to call. This adds safe wrappers around those
functions.

Fixes #561

The only open question I have is:
Should this be part of the `spi` module, or should a new one be created? These
functions are not part of `spi` related files in the Postgres source code, but
issue #561 was tagged as `spi`. So, for now I put them there.

### DIFF
--- a/pgx-tests/src/tests/spi_tests.rs
+++ b/pgx-tests/src/tests/spi_tests.rs
@@ -478,4 +478,34 @@ mod tests {
         // cidr is binary coercible to inet
         Spi::get_one::<pgx::Inet>("select '10.0.0.1/32'::cidr")
     }
+
+    #[pg_test]
+    fn test_quote_identifier() {
+        assert_eq!("unquoted", spi::quote_identifier("unquoted"));
+        assert_eq!(r#""actually-quoted""#, spi::quote_identifier("actually-quoted"));
+        assert_eq!(r#""quoted-string""#, spi::quote_identifier(String::from("quoted-string")));
+    }
+
+    #[pg_test]
+    fn test_quote_qualified_identifier() {
+        assert_eq!(
+            r#"unquoted."actually-quoted""#,
+            spi::quote_qualified_identifier("unquoted", "actually-quoted")
+        );
+        assert_eq!(
+            r#""actually-quoted".unquoted"#,
+            spi::quote_qualified_identifier("actually-quoted", "unquoted")
+        );
+        assert_eq!(
+            r#""actually-quoted1"."actually-quoted2""#,
+            spi::quote_qualified_identifier("actually-quoted1", "actually-quoted2")
+        );
+    }
+
+    #[pg_test]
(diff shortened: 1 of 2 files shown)


JelteF:
I expanded the doc comments a bit. Postgres source code comments were also quite minimal. One thing I realised though: Postgres its version of `quote_qualified_identifier` also allows `NULL` as the first argument. In that case it has the same behaviour as `quote_identifier`. To translate that literally to rust would really have `quote_qualified_identifier` take a `Option<StringLike>`. But that would mean you'd have to normally wrap the argument in a `Some(...)`, which seems a bit unergonomic to me for the normal usecase. So I suggest we don't do that and make people call `quoted_identifier` if they don't have a qualifier.

eeeebbbbrrrr:
> which seems a bit unergonomic to me for the normal usecase.

Yeah, I agree.

This looks good and I'm merging now.  Thanks for the PR and follow-up commit, @JelteF!

PR #1065, opened 2023-03-07
Include contrib/ in pgx init builds

ewencp:
Specifically, this will make contrib extensions available so testing when they are needed can be done through the normal workflow.

### DIFF
--- a/cargo-pgx/README.md
+++ b/cargo-pgx/README.md
@@ -101,7 +101,7 @@ $ cargo pgx init
 
 `cargo pgx init` is required to be run once to properly configure the `pgx` development environment.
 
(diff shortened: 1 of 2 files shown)


ewencp:
This could be hidden behind a flag to `cargo pgx init`, but I don't see much value in that since most extensions would need to be able to operate with others also available (and even loaded).

eeeebbbbrrrr:
I'm gonna let CI chew on this just to make sure it all still works, but I don't really see a problem with this.

The question of "how do I install contrib package X?" has come up a few times over the years, so yeah, I think this just makes sense.

Would you mind updating the `cargo-pgx/README.md` to indicate that this is now what `cargo pgx init` does?  If you'll do that, and CI passes, we'll get this merged and in the next point release.

ewencp:
Sounds good, note added. Thanks!

eeeebbbbrrrr:
I've installed quite a few of the contrib packages over the years, but I can't say I've installed every-single-one.  I had a though of "are there extra system dependencies this will need?", but CI passed on both linux and mac so I think that was just my anxiety acting up for no good reason.

Merging.  Thanks for the PR and welcome to the contributor club!

PR #1066, opened 2023-03-07
Add support for specifying GUC flags

JelteF:
When creating GUCs they can be passed many flags to change their
behaviour. This PR adds support for most of those flags to `pgx`.

The following flags were not included for the following reasons:
1. `LIST_INPUT` & `LIST_QUOTE`: these require more thought, because the
   GUC becomes a list. Also `LIST_QUOTE` is not even allowed to be used
   by extensions, because of problems with `pg_dump`.
2. `GUC_NOT_IN_SAMPLE`: No extension GUC is listed in
   `postgresql.conf.sample` anyway.
3. `GUC_UNIT_MEMORY`, `GUC_UNIT_TIME` & `GUC_UNIT`: These are masks, not
   flags. So these should not be passed to the `define_xxx_guc` functions.

In passing adds some basic docs for all the items in the `guc` module.

### DIFF
--- a/pgx-macros/src/lib.rs
+++ b/pgx-macros/src/lib.rs
@@ -839,6 +839,7 @@ fn impl_postgres_type(ast: DeriveInput) -> syn::Result<proc_macro2::TokenStream>
     Ok(stream)
 }
 
+/// Derives the `GucEnum` trait, so that normal Rust enums can be used as a GUC.
 #[proc_macro_derive(PostgresGucEnum, attributes(hidden))]
 pub fn postgres_guc_enum(input: TokenStream) -> TokenStream {
     let ast = parse_macro_input!(input as syn::DeriveInput);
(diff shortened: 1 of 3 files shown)


JelteF:
To be clear, this is a breaking API change since it adds an extra argument to already existing function definitions. But it's one that's easy to fix for anyone running into it. And the flags are common enough to use that it seems worth the breakage to me given this crate is not 1.0 yet.

eeeebbbbrrrr:
This is nice, thanks!

I'm debating if we should hold this for a pgx 0.8.0 release or just go ahead and include it in the next 0.7.x point release (which I think will happen in a day or two).

It is a breaking changes, and we're not necessarily afraid to make breaking changes since, as you pointed out, we're not 1.0 yet, but GUCs are pretty popular with pgx extensions.  Maybe that's a reason to just do it now.

JelteF:
Pushed a new version with slightly improved docs.

Sidenote: you probably want to enable this Github setting on the repo: https://github.blog/changelog/2022-08-23-new-options-for-controlling-the-default-commit-message-when-merging-a-pull-request/
That way you get much more useful commit messages by default when merging a PR.

eeeebbbbrrrr (approved):

PR #1071, opened 2023-03-11
Make ErrorReportWithLevel::detail_with_backtrace an Option<String>

### LINKED ISSUE #1062: Empty log DETAIL: line
This is a pretty minor one. Whenever I use `pgx::log!("something")`, the log contains my LOG: message, followed by a line with a DETAIL: message which is empty. There is nothing following the DETAIL:.

I can turn this off using `set log_error_verbosity = 'TERSE';`, though that might cause me to miss seeing something useful.

It would be best to suppress the DETAIL: line if there is no detail to report.

I'm using 0.7.2

ewencp:
Without this, the ErrorReport always has a detail field, even if it's an empty string. This is included in the format string, so it affects the output even if there isn't originally a detail string. Fixes #1062.

### DIFF
--- a/pgx-pg-sys/src/submodules/panic.rs
+++ b/pgx-pg-sys/src/submodules/panic.rs
@@ -184,16 +184,16 @@ impl ErrorReportWithLevel {
     }
 
     /// Get the detail line with backtrace. If backtrace is not available, it will just return the detail.
-    pub fn detail_with_backtrace(&self) -> String {
+    pub fn detail_with_backtrace(&self) -> Option<String> {
         match (self.detail(), self.backtrace()) {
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr:
I've updated this PR, but please note that we prefer PRs against the `develop` branch.

ewencp:
Oops, got it right with first PR, screwed it up with this one. Sorry about that!

I get the desire to have `master` be equivalent to current release, but it might be worth considering changing the default branch in the repo's settings so PRs get the right target by default. It has the downside that it'll be the branch showed by default when people arrive at the repo, but the README.md is probably the main thing people look at and it changes relatively infrequently.

I'll probably have at least a few more PRs as I'm noticing gaps in what's exposed currently so I'm somewhat invested in making the dev process as simple as possible.

PR #1075, opened 2023-03-14
Avoid some slow HashMap usage where possible

thomcc:
Because I was a little bit concerned about HashMap usage after the most recent postgrestd PR, I checked if we had some in `pgx`. Neither of these were concerning uses of `HashMap` from that perspective, but they... would be expected to have bad performance. They were both both maps which seem to use densely packed small integers as keys, so they can be replaced by vec/array.

This does that, by using `EnumMap<K, Option<V>>` in one place, and just a vector and offset in the other.

My one concern is if `SpiHeapTupleData`'s "map" needs to allow for being expanded after construction (specifically, if `TupleDescData::natts` can increase while the `SpiHeapTupleData` is alive). This doesn't seem to be the case (at least, we don't really handle it properly if it is), but it would mean my change to `set_by_ordinal` is probably wrong.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -611,6 +611,26 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7fcaabb2fef8c910e7f4c7ce9f67a1283a1715879a7c230ca9d6d1ae31f16d91"
 
 [[package]]
+name = "enum-map"
+version = "2.5.0"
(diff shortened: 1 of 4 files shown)


eeeebbbbrrrr (approved):

eeeebbbbrrrr:
>  if SpiHeapTupleData's "map" needs to allow for being expanded after construction

It does not.  I think these changes are fine, assuming CI agrees.

PR #1077, opened 2023-03-15
pgrx-pg-sys: include more headers with hooks

skyzh:
close https://github.com/tcdi/pgx/issues/1073

<del>`libpq/auth.h` is not included because it requires `ssl.h`. It might be confusing if people (like me) is configuring pgx-managed Postgres with openssl but not build pgx-pg-sys with openssl. So we still cannot use `ClientAuthentication_hook`.</del> <del>Added now. Users just need to ensure openssl includes are always in path when building pgx-pg-sys.</del> Doesn't work, so we won't add `ClientAuthentication_hook` for now.

I only built pg15 locally and let's see if the rest builds on CI 🤣 It builds.


eeeebbbbrrrr:
Hi @skyzh.  I wish I had merged this prior to the rename to "pgrx", but I missed it.  Would you mind making this PR against the `develop` branch and targeting the PR there too?  I'd love to get this included in the next pgrx release.

skyzh:
Sure, I can fix this later today 🤪

skyzh:
The PR should be ready for review now :)

eeeebbbbrrrr:
Thanks.  CI failed, but looks like some kind of weird transient networking thing.  Re-running the test jobs now.

Issue #1079, opened 2023-03-22
Changing the API for sound multi-threading


JelteF:
From [DESIGN-DECISIONS.md][design], this is the current situation:

> As it relates to any Postgres thing (calling a function, allocated memory, anything at all from pgx::*::*), there is none! Postgres itself is purposely not thread safe and neither is pgx.
>
>It's difficult (right now) to ask the Rust compiler to ensure code is not using pgx from a thread. We have been discussing some ideas. 

[design]: https://github.com/tcdi/pgx/blob/master/DESIGN-DECISIONS.md#thread-safety

The trick suggested here seems pretty nice: https://users.rust-lang.org/t/limit-function-call-only-in-current-thread/58935

It exposes all thread-unsafe functions as methods on a struct, of which an instance can only be created in a single thread.

thomcc:
We don't use the approach you suggest because it would be a pretty invasive and difficult to make change. When I maintained `imgui-rs` that's basically how we did it, and the ergonomics are, honestly, pretty bad.

workingjubilee:
Yeah I can't think of a way to implement that kind of construct which wouldn't make even "build this on one thread and use it from that thread" hard. Given our dynamic checks _do_ manage pretty well, we probably shouldn't torch the crate's ergonomics for the single-thread case to satisfy the multi-threading case.

JelteF:
I didn't realize there were dynamic checks. Those sound fine to me. Thanks for updating the docs.

thomcc:
Yeah, see https://github.com/pgcentralfoundation/pgrx/blob/develop/pgrx-pg-sys/src/submodules/thread_check.rs

PR #1081, opened 2023-03-23
Discuss multithreading ergonomics

workingjubilee:
This takes a stronger opinion on multithreading, per #1079, and the PR in question was merged, so document thusly.


thomcc (approved):

thomcc:
Still should probably have @eeeebbbbrrrr sign off but this looks good to me.

workingjubilee:
I think the solution that would be an improvement, thinking about it, would be if we had some mechanism that _sent_ all interactions with Postgres to a single thread. That way the ergonomics tradeoff is that at least you can always do it correctly. But that would be, yes, to put it mildly, an invasive change.

workingjubilee:
Right. I don't think my idea there is necessarily something we should drop everything and implement, just "this is something that _could_ pull its weight", as opposed to "this is something that _will not_ pull its weight".

eeeebbbbrrrr (approved):

PR #1083, opened 2023-03-27
Make `Array::as_slice` unsafe

thomcc:
I happened to stumble across this, and it seems kind of bad. Making it unsafe would at least stop it from being usable in `plrust`, although IDK if that's currently possible anyway.


thomcc:
Note: this is explained in https://github.com/tcdi/pgx/issues/627

PR #1084, opened 2023-03-27
Use a slice instead of a vec in direct_function_call and related functions

thomcc:
This is breaking change to this API, so perhaps it should be held off to the next release.

This always struck me as pointlessly heavyweight to allocate this on each call, so now it just passes a slice.


workingjubilee:
Maybe you could use `impl Borrow<[Option<pg_sys::Datum>]>` instead to avoid a breaking change!

( ...uh, probably don't actually do that. )

workingjubilee:
Yeah, I'm being silly.

PR #1094, opened 2023-04-05
Enforce UTF-8 correctness

workingjubilee:


yrashk:
May be we should add the fact that it may panic (if it is not valid UTF-8) to the functions' documentation?

eeeebbbbrrrr:
> May be we should add the fact that it may panic (if it is not valid UTF-8) to the functions' documentation?

Panic?  What panic?  It’s undefined behavior!  The kind our parents don’t like!

We need to at least document this.  I’m not sure pgx, being a generic framework, should take a stance here. And if we do it shouldn’t be only because of this. There’s many other places we take random char *’s from Postgres too. 

But this is definitely a subtle edge case and one of those hidden assumptions pgx-based extensions might miss.

eeeebbbbrrrr:
Also I’m tired and didn’t realize this was a PR (I thought it was an issue). Duh. 

I’m not sure this is the right answer either. It’s properly implemented but is it the right thing?

there’s a concern for plrust with this and we’d have some additional options over there like refusing to load at all or be more fine grained and refuse to create specific functions that take/return Strings. 

I think my concern here is still generally that pgx makes assumptions in other places it shouldn’t. Might be worth tracking them all down and seeing what it looks like.

eeeebbbbrrrr:
Maybe we can ask the database if it’s utf8 and keep the _unchecked() version in that case. 

One concern is performance. Rust’s string-fu is **strong** and I’d expect pgx extensions to take advantage of that. plrust functions too. 

Perf ought not trump safety but we already trust Postgres, so perhaps a middle-ground is a good way forward?

workingjubilee:
Some notes for tomorrow's discussion:
- Setting character encoding is done at CREATE DATABASE time.
- PostgreSQL exposes `convert_to` and `convert_from` which can reencode.
- The `SQL_ASCII` encoding does not imply any encoding.

workingjubilee:
That could result in mysteriously low performance without the DBA ever realizing why, however.

eeeebbbbrrrr:
> That could result in mysteriously low performance without the DBA ever realizing why, however.

Yeah, but at least if it the code was using a Cow (or our own wrapper type around it) it'd be easier to know that "oh, there's some transcoding happening, probably".  

I wouldn't want us to replace our `impl From/IntoDatum for String/&str` with this -- it'd need to be an opt-in thing that'd at least be some kind of escape hatch for the developer.

workingjubilee:
Doing that for a `Cow` type seems overly magical, doing it for a specific nominal type seems fine.

workingjubilee:
My thoughts:

- In the case of PGX used in a DB where someone controls both the DB encoding and the extensions, we want the option to compile it without such checks. No optimized heuristic is going to be as good as **not** checking, but this requires total DBA control.
- In the case of PGX used in a legacy DB where the DB encoding is in practice trivial (UTF-8, or "UTF-8, but mostly because it's ASCII"), which is the case for the `SQL_ASCII` encoding, we want to check minimally and inform the user if the encoding is not what they thought it was.
- In the case of PGX used in a legacy DB where the DB encoding is _complicated_ and definitely not UTF-8, we want the ability to support the user in doing transcoding of some sort, most likely by pawning the job off on ICU or Postgres.

To me, this points to panic-by-default and then working out the optimizations we want in followups.

eeeebbbbrrrr:
> panic-by-default

This means significant overhead in the common, happy path of the database actually being UTF8.  This would likely kill ZDB indexing performance, for example.

Do you see room to compromise on this and only panic if the database isn't UTF8?

workingjubilee:
I think the UTF-8 check is actually not going to be that slow. Not as slow as transcoding, for sure. :thinking:

But sure, we can memoize a check for if the database uses a known guaranteed UTF-8 or trivial encoding by default.

eeeebbbbrrrr:
How you wanna proceed?  I’m happy to take over this PR and finish it. I’m also happy for you to do it. Or we can merge this and do a follow up on Monday.

eeeebbbbrrrr (approved):

eeeebbbbrrrr:
Ship it

workingjubilee:
Added the requested docos to this PR though, since that's immediately relevant.

PR #1095, opened 2023-04-07
Memoize DB encoding for faster validation

workingjubilee:
This performs a check for whether the database is UTF-8, and then uses that information to reduce text validation. An initial check is performed at database load time. If the result discovers a completely incompatible, not-even-ASCII encoding, then the extension will bail. This shouldn't happen as they aren't supported server-side but maybe Postgres will implement support in the future?

### DIFF
--- a/pgx/src/lib.rs
+++ b/pgx/src/lib.rs
@@ -28,6 +28,7 @@ extern crate bitflags;
 extern crate alloc;
 extern crate core;
 
+use once_cell::sync::Lazy;
 // expose our various derive macros
 pub use pgx_macros;
 pub use pgx_macros::*;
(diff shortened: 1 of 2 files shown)


pgx/src/lib.rs
```diff
@@ -307,6 +308,38 @@ macro_rules! pg_sql_graph_magic {
     };
 }
 
+pub(crate) static UTF8DATABASE: Lazy<Utf8Compat> = Lazy::new(|| {
+    let encoding_int = unsafe { pgx_pg_sys::GetDatabaseEncoding() };
+    match encoding_int as core::ffi::c_uint {
+        pg_sys::pg_enc_PG_UTF8 => Utf8Compat::Yes,
+        // The 0 encoding. It... may be UTF-8
+        pg_sys::pg_enc_PG_SQL_ASCII => Utf8Compat::Maybe,
```
workingjubilee, line 316: This (and the logic that plays off it) optimistically assumes that PGX _may_ be active in a UTF-8-conformant database, and that if this is false, normal UTF-8 validation will sniff this out pretty quickly. This implies a performance penalty, however.

pgx/src/lib.rs
```diff
@@ -319,4 +352,12 @@ macro_rules! pg_sql_graph_magic {
 #[allow(unused)]
 pub fn initialize() {
     pg_sys::panic::register_pg_guard_panic_hook();
+    match Lazy::force(&UTF8DATABASE) {
+        Utf8Compat::Maybe => notice!("database encoding is not validated as UTF-8"),
+        Utf8Compat::Ascii => {
+            warning!("database encoding is not UTF-8, extension will accept ASCII only")
+        }
```
workingjubilee, line 359: These various encodings include a few which may share a few codepoint encodings with UTF-8, but the decision to only trust ASCII data from them is because that check is simpler ("high bit set? yes? okay!") and the shared codepoints are more incidental than by design.

pgx/src/lib.rs
```diff
@@ -319,4 +352,12 @@ macro_rules! pg_sql_graph_magic {
 #[allow(unused)]
 pub fn initialize() {
     pg_sys::panic::register_pg_guard_panic_hook();
+    match Lazy::force(&UTF8DATABASE) {
+        Utf8Compat::Maybe => notice!("database encoding is not validated as UTF-8"),
```
workingjubilee, line 356: There are encodings which can seem like UTF-8 for some number of bytes, and SQL_ASCII implies that we won't necessarily detect them before we complete some number of transactions. That's why the choice to emit a notice about this up-front.

pgx/src/lib.rs
```diff
@@ -319,4 +352,12 @@ macro_rules! pg_sql_graph_magic {
...
+        Utf8Compat::Ascii => {
+            warning!("database encoding is not UTF-8, extension will accept ASCII only")
+        }
+        Utf8Compat::No => unreachable!("unsupported database encoding encountered"),
```
workingjubilee, line 360: Hm. Maybe I should just remove this variant and bail up-front, when creating the enum. I'll come back to this on Monday.

workingjubilee:
Technically, this PR also could do some work on the `IntoDatum` side, but honestly, I'm not sure I want to take the stance that PGX is in charge of telling you what data to *not* shove into your database? I think PGX may have justifiable reason to restrict its concerns to making sure the data pulled into Rust meets Rust's type requirements.

eeeebbbbrrrr (changes requested):
@workingjubilee I'm not sure pgx itself should emit notices or warnings when an extension is loaded.  Extensions can be loaded in various ways and I don't think it's nice of us to spam the log and, especially in the case of `warning!` with a default postgresql.conf, the client.

We're not panicking here anymore, so I think it's fine to just get the Lazy force'd and not say anything.

eeeebbbbrrrr (approved):
LGTM!  Ship it!

PR #1096, opened 2023-04-07
Add support for explicitly specifying SECURITY INVOKER/DEFINER functions

### LINKED ISSUE #1064: Missing possible important feature "SECURITY DEFINER" in #[pg_extern]
Why do we need it? If you have roles, and let's say you have an anonymous user to whom users will be auto-assigned until they login, and then switch to their role, the anon user may need to call the login function, which accesses private schemas, e.g. auth. Users with a read and write. While it is possible to write this function in PG UDF, it is safer to write it in rust to hide code from some DB admin.  

We can add the Security Definer by hand to the generated function but it will be good to include it in the source. Unless, I missed the process

JohnHVancouver:
fix #1064

Before by default functions were security invoker only. This allows specifying security invoker/definer

e.g. pg_extern(security_definer)


JohnHVancouver:
Specifying functions like this work now:

```
#[pg_extern(security_definer)]
fn hello_security_definer() -> &'static str {
    "Hello, security_definer"
}

#[pg_extern(security_invoker)]
fn hello_security_invoker() -> &'static str {
    "Hello, security_invoker"
}
```

```
postgres=# select proname, prosecdef from pg_proc where proname like 'hello_security%';
        proname         | prosecdef 
------------------------+-----------
 hello_security_definer | t
 hello_security_invoker | f
(2 rows)
```

JohnHVancouver:
Need to base off of develop

PR #1097, opened 2023-04-07
Add support for explicitly specifying SECURITY INVOKER/DEFINER functions

### LINKED ISSUE #1064: Missing possible important feature "SECURITY DEFINER" in #[pg_extern]

JohnHVancouver:
fix #1064

Before by default functions were security invoker only. This allows specifying security invoker/definer

e.g. pg_extern(security_definer)


JohnHVancouver:
Allows to explicitly specify invoker/definer. It'll error if both are specified but that doesn't seem too bad (and consistent).

```
postgres=# select proname, prosecdef from pg_proc where proname like 'hello_security%';
        proname         | prosecdef 
------------------------+-----------
 hello_security_definer | t
 hello_security_invoker | f
(2 rows)

postgres=# select hello_security_definer(), hello_security_invoker();
 hello_security_definer  | hello_security_invoker  
-------------------------+-------------------------
 Hello, security_definer | Hello, security_invoker
(1 row)
```

eeeebbbbrrrr:
Thanks @JohnHVancouver!  Would you mind adding a test for this as well?  There's the `pgx-tests` crate.  I bet one of the existing modules would be good enough here.  I suppose the test could ensure that a function with these attributes has the proper values in `pgproc`.

JohnHVancouver:
Added test cases

eeeebbbbrrrr (approved):

workingjubilee (approved):
Thank you!

Issue #1106, opened 2023-04-17
This Project Is Getting a New Name:  `pgrx`


eeeebbbbrrrr:
## Goodbye `pgx`, Hello `pgrx`

For various reasons it is necessary for us to rename `pgx`.  

The core team, through much deliberation and consternation, have decided on **`pgrx`**.

This is short for "**P**ost**G**res **R**ust E**x**tension", which isn't far off from whatever `pgx` might have meant (I'm not sure there was an official definition!).

`pgrx` is close enough to `pgx` that a global, case-preserving search-n-replace on your code should work just fine.  That's exactly what we did to rename pgx, along with renaming directories and a few files that contained "pgx" in the filename.

We realize this is a huge inconvenience, but having the courage to change the things we can goes a long way.  The underlying reasons are entirely outside of the core team's control.

## When?

We intend the name change to be official in what will be `pgrx v0.7.4`.  

This release will be **exactly the same** as `pgx v0.7.4`, only with everything renamed.  There will be no other code changes.

We intend to rename to `pgrx v0.7.4` today, April 17, 2023.

This should allow existing pgx users to migrate their code to the new name without fear of other changes sneaking into their code.

## What About `crates.io`?

The currently released `pgx-0.7.4` crates will continue to live on crates.io for all eternity, however they will no longer be updated.

## What About `cargo-pgx`?

It will be called `cargo-pgrx` and will default to a (now renamed) `$PGRX_HOME` directory of `~/.pgrx/` instead of `~/.pgx/`.

This means you'll need to `cargo uninstall cargo-pgx && cargo install cargo-pgrx --locked && cargo pgrx init`.

## Everything's Been Renamed

When released there will be no references "pgx" in `pgrx`'s source code.  That is, except for "pgxs", which is a Postgres `Makefile` thing and unrelated to `pg(r)x` itself.

PR #1107, opened 2023-04-17
Rename to pgrx

eeeebbbbrrrr:
This does the needful behind issue #1106


workingjubilee (approved):
We discussed the article question, which was the only serious question I had.

eeeebbbbrrrr:
> We discussed the article question, which was the only serious question I had.

yeah, I think it's fine for the files in `./articles/` to get the rename treatment too, since they're under source control.

Published versions on corporate blog can be looked at later, or even never.

.github/docker/Dockerfile.alpine:3.16
```diff
@@ -1,11 +1,11 @@
...
+# Example of how to build this container (ran from pgrx checkout directory):
+#   docker build --build-arg PG_MAJOR_VER=14 -f .github/docker/Dockerfile.alpine\:3.16 -t pgrx-alpine .
 #
 # Example of how to run this container with tests after building the image:
-#   docker run pgx-alpine cargo test --no-default-features --features pg14 --locked
+#   docker run pgrx-alpine cargo test --no-default-features --features pg14 --locked
 #
 # Note that "PG_MAJOR_VER" build arg in the build step must match the
-# "--features pgXX" in the run step
+# "--features pgrxX" in the run step
```
syvb, line 8: This (and all the other instances of `pgXX`) shouldn't have been changed

eeeebbbbrrrr: nice catch.  thanks.  we're still reviewing this too.

workingjubilee:
Yeah, that seems fine, and it definitely seems lower priority now that I realize the tcdioss website won't automatically update.

Maybe we should open an issue to track following up on that and making a final decision. But "we discussed it and made a decision that will suffice for now" was the main thing I wanted.

CONTRIBUTING.md
```diff
@@ -98,10 +98,10 @@ Do this *in order*:
...
-You agree that all code you submit in pull requests to https://github.com/tcdi/pgx/pulls
+You agree that all code you submit in pull requests to https://github.com/tcdi/pgrx/pulls
 is offered according to the MIT License, thus may be freely relicensed and sublicensed,
 and that you are satisfied with the existing copyright notice as of opening your PR.
 
 [crates.io]: https://crates.io
-[pgx@crates.io]: https://crates.io/crates/pgxa
-[pgx@docs.rs]: https://docs.rs/pgx/latest/pgx
\ No newline at end of file
+[pgrx@crates.io]: https://crates.io/crates/pgrxa
```
eeeebbbbrrrr, line 106: this link was wrong in the orignal too.  we'll fix it in post.

eeeebbbbrrrr:
Alrighty, this is going to master right now as a squash merge, then merged into develop, then the repo here on GitHub will be renamed.  Then published to crates.io from the master branch.

PR #1109, opened 2023-04-17
Fix broken links to zombodb/pgrx

syvb:
Some links to zombodb/pgx got replaced with zombodb/pgrx, which GitHub doesn't redirect here. This fixes them by updating those links to tcdi/pgrx.


eeeebbbbrrrr:
You're a pal, @syvb!

Issue #1111, opened 2023-04-18
`build.rs` requires $PGRX_HOME/config.toml when it doesn't need to


eeeebbbbrrrr:
`build.rs` supports reading the variables `pg_config` provides from the environment, and in this case it never needs to execute `pg_config`, which also means that in this case it doesn't need to know where `pg_config` is on disk.  Which means in this case we don't require `$PGRX_HOME/config.toml`.

Unfortunately, `build.rs` blindly tries to open `$PGRX_HOME/config.toml` no matter what.

A downstream effect of this is that an installed PL/Rust extension requires that file during `CREATE FUNCTION` and it shouldn't.

eeeebbbbrrrr:
https://github.com/tcdi/pgrx/pull/1112 is the fix for this

PR #1112, opened 2023-04-18
Cleanup `build.rs` and friends issue #1111

eeeebbbbrrrr:
This PR "cleans up" `build.rs` so that it now tracks, via `cargo:rerun-if-env-changed`, every environment variable it uses.  The justification here is that if `build.rs` uses an environment variable, then `build.rs` will need to re-run if that envar changes.  It could be this is overly aggressive in that `cargo` probably does this for us for things like `TARGET` or `CARGO_MANIFEST_DIR` or `OUT_DIR`, but it seems prudent to not have to then justify why we track some envars and not others.

It also doesn't try to open `$PGRX_HOME/config.toml` in the case where the `pg_config` data is wholly described as environment variables.

And it also removes from `Pgrx::home()` the code that would just blindly try to create `$PGRX_HOME` if it doesn't exist.  It instead moves that to the Err case when it's called via `cargo pgrx init`.


workingjubilee:
I'm throwing the Will It Blend tests at this just to make sure.

cargo-pgrx/src/command/init.rs
```diff
@@ -118,9 +119,22 @@ impl CommandExecute for Init {
...
+#[tracing::instrument(skip_all)]
 pub(crate) fn init_pgrx(pgrx: &Pgrx, init: &Init) -> eyre::Result<()> {
-    let dir = Pgrx::home()?;
+    let pgrx_home = match Pgrx::home() {
+        Ok(path) => path,
+        Err(e) => match e {
+            PgrxHomeError::NoHomeDirectory => return Err(e.into()),
+            PgrxHomeError::IoError(e) => return Err(e.into()),
+            PgrxHomeError::MissingPgrxHome(path) => {
+                // $PGX_HOME doesn't exist, but that's okay as `cargo pgrx init` is the right time
```
workingjubilee, line 130:
```suggestion
                // $PGRX_HOME doesn't exist, but that's okay as `cargo pgrx init` is the right time
```

pgrx-pg-sys/build.rs
```diff
@@ -88,24 +88,22 @@ impl bindgen::callbacks::ParseCallbacks for PgrxOverrides {
...
     let compile_cshim =
-        std::env::var("CARGO_FEATURE_CSHIM").unwrap_or_else(|_| "0".to_string()) == "1";
+        env_tracked("CARGO_FEATURE_CSHIM").unwrap_or_else(|| "0".to_string()) == "1";
 
-    let is_for_release = std::env::var("PGRX_PG_SYS_GENERATE_BINDINGS_FOR_RELEASE")
-        .unwrap_or("0".to_string())
-        == "1";
-    println!("cargo:rerun-if-env-changed=PGRX_PG_SYS_GENERATE_BINDINGS_FOR_RELEASE");
+    let is_for_release =
+        env_tracked("PGRX_PG_SYS_GENERATE_BINDINGS_FOR_RELEASE").unwrap_or("0".to_string()) == "1";
```
workingjubilee, line 106: Trivial, I know, and preexisting, but it bugs me these aren't consistent.
```suggestion
        env_tracked("PGRX_PG_SYS_GENERATE_BINDINGS_FOR_RELEASE").unwrap_or_else(|| "0".to_string()) == "1";
```

pgrx-pg-sys/build.rs
```diff
@@ -116,20 +114,15 @@ fn main() -> eyre::Result<()> {
...
-    println!("cargo:rerun-if-changed=cshim");
-    emit_missing_rerun_if_env_changed();
+    emit_rerun_if_changed();
 
-    let pg_configs: Vec<(u16, PgConfig)> = if std::env::var(
+    let pg_configs: Vec<(u16, PgConfig)> = if env_tracked(
         "PGRX_PG_SYS_GENERATE_BINDINGS_FOR_RELEASE",
     )
     .unwrap_or("false".into())
         == "1"
```
workingjubilee, line 123: can't this be more elegantly stated as
```suggestion
    let pg_configs: Vec<(u16, PgConfig)> = if env_tracked(
        "PGRX_PG_SYS_GENERATE_BINDINGS_FOR_RELEASE",
    ).as_ref() == Some("1")
```

pgrx-pg-sys/build.rs
```diff
@@ -252,7 +255,7 @@ fn generate_bindings(
         .wrap_err_with(|| format!("failed to rewrite items for pg{}", major_version))?;
     let oids = format_builtin_oid_impl(oids);
 
-    let dest_dirs = if std::env::var("PGRX_PG_SYS_GENERATE_BINDINGS_FOR_RELEASE")
+    let dest_dirs = if env_tracked("PGRX_PG_SYS_GENERATE_BINDINGS_FOR_RELEASE")
         .unwrap_or("false".into())
         == "1"
```
workingjubilee, line 260:
```suggestion
    let dest_dirs = if env_tracked("PGRX_PG_SYS_GENERATE_BINDINGS_FOR_RELEASE").as_ref() == Some("1")
```

pgrx-pg-sys/build.rs
```diff
@@ -760,7 +763,7 @@ fn env_tracked(s: &str) -> Option<String> {
 }
 
 fn target_env_tracked(s: &str) -> Option<String> {
-    let target = std::env::var("TARGET").unwrap();
+    let target = env_tracked("TARGET").unwrap();
```
workingjubilee, line 766: I'm surprised that we need to add `env_tracked` to this, I thought Cargo was already tracking this? Are we sure we won't confuse Cargo on this point?

pgrx-pg-sys/build.rs
```diff
@@ -311,8 +314,8 @@ struct BuildPaths {
 impl BuildPaths {
     fn from_env() -> Self {
         // Cargo guarantees these are provided, so unwrap is fine.
-        let manifest_dir = std::env::var_os("CARGO_MANIFEST_DIR").map(PathBuf::from).unwrap();
-        let out_dir = std::env::var_os("OUT_DIR").map(PathBuf::from).unwrap();
+        let manifest_dir = env_tracked("CARGO_MANIFEST_DIR").map(PathBuf::from).unwrap();
+        let out_dir = env_tracked("OUT_DIR").map(PathBuf::from).unwrap();
```
workingjubilee, line 318: Same concerns re: "should we really be adding this to the env tracking"?

workingjubilee (commented):
Largely looks good. Some mild nits. Somewhat concerned about duplicating the tracking command confusing cargo.

pgrx-pg-sys/build.rs
workingjubilee: Also looking again this should actually be another case of `.as_ref() == Some("1")`

eeeebbbbrrrr:
so... I blindly did this for **every** environment variable.  My thought process was that if build.rs uses an envar, then it'll need to be rerun when that envar changes.

workingjubilee:
Whee, errant mouseclicks.

workingjubilee:
Hmm. Given the big build test suite passed on a previous version of this PR, we should be okay? I feel like my main concern is really just making sure the commit message or a comment or whatever includes "yeah this was a more... interesting decision" so that if we find an unusual bug later related to the build, shuffling through the `git blame` eventually turns up something.

eeeebbbbrrrr:
> Hmm. Given the big build test suite passed on a previous version of this PR, we should be okay? I feel like my main concern is really just making sure the commit message or a comment or whatever includes "yeah this was a more... interesting decision" so that if we find an unusual bug later related to the build, shuffling through the `git blame` eventually turns up something.

I updated the PR description.  If that works for you, it'll become the commit message -- once CI passes the additional cleanups I just pushed.

workingjubilee (approved):
Yep, looks good to me!

thomcc:
> It could be this is overly aggressive in that cargo probably does this for us for things like `TARGET` or `CARGO_MANIFEST_DIR` or `OUT_DIR`, but it seems prudent to not have to then justify why we track some envars and not others.

There used to be issues where doing this for the ones cargo manages would cause spurious rebuilds. I think these have been fixed, but it's possibly worth checking.

If it helps, https://github.com/rust-lang/cc-rs/blob/main/src/lib.rs#L3132-L3139 is what I've used in the past (although the main reason cc-rs tries to avoid it is not that bug, but because emitting *any* `rerun-if-env-changed` vars can impact if rebuilds happen).

eeeebbbbrrrr:
> If it helps, https://github.com/rust-lang/cc-rs/blob/main/src/lib.rs#L3132-L3139 is what I've used in the past (although 

It's funny, I had this same thought yesterday afternoon while walking the dog.

> the main reason cc-rs tries to avoid it is not that bug, but because emitting any rerun-if-env-changed vars can impact if rebuilds happen).

Well, we use `rerun-if-env-changed` to cause rebuilds, so like, /shrug?

I feel like this should be fine.  If it causes rebuilds when it shouldn't then I'd argue that's a `cargo` bug and we should report it there and then work around it like you do in cc-rs.  I'm gonna go ahead and merge and we can see what happens during our daily development.

eeeebbbbrrrr:
> I feel like this should be fine. If it causes rebuilds when it shouldn't then I'd argue that's a cargo bug and we should report it there and then work around it like you do in cc-rs. I'm gonna go ahead and merge and we can see what happens during our daily development.

As usual, @thomcc was correct!  :)

I've put up PR #1127 to address the constant rebuilds that started happening after this PR was merged.

I didn't narrow it down specifically because I got bored, but it's at least caused by one of the envars we access that begins with "CARGO":  `CARGO_CFG_TARGET_OS`, `CARGO_FEATURE_PGxx`, or `CARGO_CFG_PLRUSTC`.

I still say this is some kind of cargo bug.  Crate authors shouldn't have to do this.

PR #1115, opened 2023-04-19
Fixup iterator impls for Array

workingjubilee:
There were a bunch of small problems in these.

This is unrelated to, and thus can be landed in parallel with, the zero-copy Array work. No point in gunking up an already-large PR with additional work.

### DIFF
--- a/pgrx/src/datum/array.rs
+++ b/pgrx/src/datum/array.rs
@@ -438,11 +438,6 @@ impl<'a, T: FromDatum> Iterator for ArrayIntoIterator<'a, T> {
         // TODO: This code is dangerously under-exercised in the test suite.
         self.array.nelems - self.curr
     }
-
-    fn nth(&mut self, n: usize) -> Option<Self::Item> {
-        // TODO: This code is dangerously under-exercised in the test suite.
-        self.array.get(self.curr + n)
-    }
 }
 
 impl<'a, T: FromDatum> FromDatum for VariadicArray<'a, T> {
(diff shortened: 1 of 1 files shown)


pgrx/src/datum/array.rs
```diff
@@ -424,18 +416,24 @@ impl<'a, T: FromDatum> Iterator for ArrayIntoIterator<'a, T> {
...
-        self.array.nelems
+    fn count(self) -> usize {
+        // TODO: This code is dangerously under-exercised in the test suite.
+        self.array.nelems - self.curr
     }
 
     fn nth(&mut self, n: usize) -> Option<Self::Item> {
-        self.array.get(n)
+        // TODO: This code is dangerously under-exercised in the test suite.
+        self.array.get(self.curr + n)
```
thomcc, line 436: This is still wrong, nth needs to advance the iterator.

workingjubilee: I both realized that and also realized I should just drop this because I fiddled more with my array walking code and realized there's no point in overloading `nth`.

workingjubilee: Fixed (like a cat).

workingjubilee:
In the further direction of "these PRs actually relate to each other more than I thought", the other one has a conflicting (necessary) edit with the "reduce unsafe" commit, so I am going to drop that one.

eeeebbbbrrrr (approved):

PR #1123, opened 2023-04-25
fix bug with `impl<'a> FromDatum for &'a CStr`

eeeebbbbrrrr:
It needs to know how to copy itself into a different `MemoryContext`.

Also add a `impl SqlTranslatable for CString` and relax the lifetime bounds on the impl for `&CStr`


eeeebbbbrrrr:
This fixes a UAF when going through SPI:

```rust
let cstr = Spi::get_one::<&CStr>("SELECT 'hello'::cstring")?.unwrap();
// `cstr` now points to freed memory.  :(
```

PR #1125, opened 2023-04-26
fix (zero-copy)-`Array` so that when working with fixed-size, pass-by-value…

eeeebbbbrrrr:
Datums we still treat them as Datums instead of assuming those N bytes in memory represent the type.  This isn't true for arrays of enums and might not be true for other types as well.


eeeebbbbrrrr:
This is going to prompt a 0.8.2 release soon as we're sure this passes tests on aarch64 -- a thing we unfortunately need to manually determine.

BradyBonnette (approved):
Looks good, also runs tests successfully in a one-off aarch64 Ubuntu environment using pg13 .

PR #1126, opened 2023-04-27
Update version to 0.8.2

eeeebbbbrrrr:
pgrx v0.8.2 is a small bugfix release.

It fixes a bug decoding arrays of enums, and perhaps other pass-by-value datums, that could lead to reading past the end of the backing array structure along with fixing a regression regarding compilation speeds.

When upgrading, please make sure to run `cargo install cargo-pgrx --locked`.

## Changes

- PR #1125 - the `Array<T>` type can now properly decode T's that are enums annotated with `#[derive(PostgresEnum)]` and should also work properly with other pass-by-value datum types that require going through `FromDatum::from_datum()`.  This is a fairly important bugfix as with v0.8.0/1 it's possible for `Array<T>` to read past the end of the array.

- PR #1127 - due to no good deed going unpunished, and some over-zealousness in PR #1111, consecutive "cargo build" commands such as `cargo pgx run/test/install` would cause the `pgrx-pg-sys/build.rs` file to run again, causing a full rebuild of (at least) that crate.  This was an unintended side-effect and #1127 fixes it.

- PR #1128 - remove a few unused dependencies

**Full Changelog**: https://github.com/tcdi/pgrx/compare/v0.8.1...v0.8.2

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -305,7 +305,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-pgrx"
-version = "0.8.1"
+version = "0.8.2"
 dependencies = [
  "atty",
  "cargo_metadata",
@@ -1477,7 +1477,7 @@ dependencies = [
 
 [[package]]
 name = "pgrx"
-version = "0.8.1"
+version = "0.8.2"
 dependencies = [
  "atomic-traits",
  "bitflags",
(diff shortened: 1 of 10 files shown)


BradyBonnette (approved):
Looks great to me 👍

PR #1127, opened 2023-04-27
Hardcode a list of environment variables that, when accessed, we should **not** `rerun-if-env-changed` with.

eeeebbbbrrrr:
Some quick investigation seems to indicate that doing so for at least one of the keys we access that begins with "CARGO" causes `build.rs` to be re-built and run on every build.


BradyBonnette (approved):
Looks like it improved build times in both x86_64 and aarch64 environments!

PR #1135, opened 2023-05-11
include generic xlog related api

kemingy:
Hi, thanks for the great tool. I'm working on a postgres extension that requires indexing. Those APIs are necessary to manipulate the pages.


eeeebbbbrrrr:
Thanks for the PR.  I'll merge soon as CI passes, but it'll be a few weeks before we cut another release.

PR #1139, opened 2023-05-16
Date time overhaul

eeeebbbbrrrr:
Working on overhauling pgrx's `Date`, `Time`, `TimeWithTimeZone`, `Timestamp`, `TimestampWithTimeZone`, and `Interval` types.

We now delegate most everything back into Postgres so we're no longer responsible for re-implementing date/time math which is impossible to get correct.

The various types can now actually be constructed, and the ones `with time zone` support construction at a specific timezone and also being shifted to a specific timezone.

I've also ripped out support for the `time` crate -- the code around that is too much highly-specific knowledge between both Postgres and `time` in order to confidently maintain going forward.  Users that use that will need to make their own wrappers to and implement `time` support themselves.

There's still a little more work to do, including:

- [x] Better error handling -- need to invent a custom error type instead of the generic `PgSqlErrorCodes`
- [x] More `impl From` conversions between the types
- [x] `impl Add/Sub`
- ~~`impl Index` if possible so that `let hour = timestamp[DateTimeParts::Hour];` ~~ not possible
- [x] Drastically simplify the `Serialize` implementations and also properly implement the corresponding `Deserialize` trait
- [x] doc comments
- [x] additional unit tests
- [x] top-level constructor functions like `now()`, `current_timestamp()`, etc.  There's a handful of these in `timestamp.c`
- [x] misc. functions from https://www.postgresql.org/docs/15/functions-datetime.html like `age` and `date_trunc`
- [x] an example that shows off some of the simple usage


pgrx/src/datum/time.rs
```diff
@@ -51,70 +107,102 @@ impl IntoDatum for Time {
...
+        .execute()
+    }
+
+    /// Construct a new [`Time`] from its constituent parts.
+    ///
+    /// Elides the overhead of trapping errors for out-of-bounds parts
+    ///
+    /// # Panics
+    ///
+    /// This function panics if any part is out-of-bounds
```
thomcc, line 144: `unchecked` seems a bit of an odd name then.

thomcc (approved):
This looks really good.

Note that I wouldn't be able to catch issues like you calling the wrong postgres function (I tried to cross-reference the docs for a while, well, I gave up).

I think we might want to think about whether or not `pub` is right for the wrapper types. I trust your judgement here, although given that they already have conversion functions, I'm unsure it's needed.

I also have some nits about naming (really just whether or not `unchecked` is right, although it probably is consistent with how we use it elsewhere in pgrx.

eeeebbbbrrrr:
Okay, I got rid of the `pub` member for all the date/time types, added `fn into_inner()` for each of them, and added a couple of missing `impl From` traits too.

I have thought quite a bit for days about a better name for `new_unchecked()` and can't come up with one.  Can't come up with a worse name either!  So we'll just live with it.  If we come up with something between now and the next release, great.  If not, we can always depreciate.

Issue #1141, opened 2023-05-22
`pgrx::FATAL` and `pgrx::PANIC` do not terminate the process immediately


usamoi:
The macros `pgrx::FATAL` and `pgrx::PANIC` do not terminate the postgres process, but raise a panic that could be caught by `catch_unwind`.

We should immediately delegate FATAL and PANIC to Postgres.

eeeebbbbrrrr:
I'll have a PR up for this in a few minutes.

eeeebbbbrrrr:
fixed in PR #1142.  Thanks for finding this @usamoi!  It'll be included in the next release.

PR #1142, opened 2023-05-23
fix issue #1141:  immediately report FATAL and PANIC

eeeebbbbrrrr:
Errors reported at the FATAL or PANIC levels should immediately be reported to Postgres since they ultimately abort the transaction.

ERRORs continue to be converted into panics, and everything else is also immediately reported to Postgres but since they only emit messages, there's no change to flow control.

As a drive-by, I also added a few functions to the "bad_ideas" example to interactively test this change.  It's not really feasible for the test suite to somehow test that a test causes the backend, or the entire cluster, to crash, which is what PANIC and FATAL do.


thomcc (approved):
For these postgres will exit, so it's fine.

eeeebbbbrrrr:
Alrighty, thanks @thomcc.  Gonna merge it.

PR #1147, opened 2023-05-25
Add Array slice support

eeeebbbbrrrr:
Add support for getting a slice from an `Array<T>` where `T` is a one of these primitives:  i8, i16, i32, i64, f32, f64.  It's possible that we could support other fixed-size types as well, but this is a good place to start.

We also implement `IntoDatum` for `Array<T>`, add some tests, and expand the `arrays` example a little bit.

### DIFF
--- a/pgrx-examples/arrays/src/lib.rs
+++ b/pgrx-examples/arrays/src/lib.rs
@@ -100,8 +100,74 @@ fn return_vec_of_customtype() -> Vec<SomeStruct> {
     vec![SomeStruct {}]
 }
 
+#[pg_schema]
+mod vectors {
+    use pgrx::prelude::*;
+
+    extension_sql!(
+        r#"
+            do $$ begin raise warning 'creating sample data in the table ''vectors.data''.'; end; $$;
+            create table vectors.data as select vectors.random_vector(768) v from generate_series(1, 100000);
+    "#,
+        name = "vectors_data",
+        requires = [random_vector]
+    );
+
+    #[pg_extern]
+    fn sum_vector_array(input: Array<f32>) -> f32 {
+        input.iter_deny_null().map(|v| v as f32).sum()
+    }
+
+    #[pg_extern]
+    fn sum_vector_vec(input: Vec<f32>) -> f32 {
+        input.into_iter().map(|v| v as f32).sum()
+    }
+
+    #[pg_extern]
+    fn sum_vector_slice(input: Array<f32>) -> Result<f32, ArrayError> {
+        Ok(input.as_slice()?.iter().map(|v| *v as f32).sum())
+    }
+
+    #[pg_extern]
+    pub fn sum_vector_simd(values: Array<f32>) -> Result<f32, ArrayError> {
+        // this comes directly from https://stackoverflow.com/questions/23100534/how-to-sum-the-values-in-an-array-slice-or-vec-in-rust/67191480#67191480
+        // and https://www.reddit.com/r/rust/comments/13q56bp/comment/jlgq2t6/?utm_source=share&utm_medium=web2x&context=3
+        //
+        // the expectation here is that the compiler will autovectorize this code using SIMD
+        use std::convert::TryInto;
+
+        const LANES: usize = 16;
+
+        let values = values.as_slice()?;
+        let chunks = values.chunks_exact(LANES);
+        let remainder = chunks.remainder();
+
+        let sum = chunks.fold([0.0f32; LANES], |mut acc, chunk| {
+            let chunk: [f32; LANES] = chunk.try_into().unwrap();
+            for i in 0..LANES {
+                acc[i] += chunk[i];
+            }
+            acc
+        });
+
+        let remainder: f32 = remainder.iter().copied().sum();
+
+        let mut reduced = 0.0f32;
+        for i in 0..LANES {
+            reduced += sum[i];
+        }
+        Ok(reduced + remainder)
+    }
+
+    #[pg_extern]
+    fn random_vector(len: i32) -> Vec<f32> {
+        (0..len).map(|_| rand::random()).collect()
+    }
+}
+
 #[cfg(test)]
 pub mod pg_test {
+    use pgrx::pg_schema;
 
     pub fn setup(_options: Vec<&str>) {
         // perform one-off initialization when the pg_test framework starts
--- a/pgrx-examples/arrays/src/lib.rs
+++ b/pgrx-examples/arrays/src/lib.rs
@@ -124,12 +124,12 @@ mod vectors {
     }
 
     #[pg_extern]
-    fn sum_vector_slice(input: Array<f32>) -> Result<f32, ArrayError> {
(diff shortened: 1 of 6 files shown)


eeeebbbbrrrr:
Using the updated `pgrx-examples/arrays` example which now creates a table of float4[] data, we can see the performance differences between various Array usages.  Slices are great for performance and then lend themselves to coaxing the compiler to do autovectorization if you're smart enough to know how to do that (I am not and leaned on stack overflow (https://stackoverflow.com/questions/23100534/how-to-sum-the-values-in-an-array-slice-or-vec-in-rust/67191480#67191480).

Pretty cool tho.

```sql
[v15.3][660234] arrays=# explain analyze select vectors.sum_vector_array(v) from vectors.data;
                                                 QUERY PLAN                                                 
------------------------------------------------------------------------------------------------------------
 Seq Scan on data  (cost=0.00..1719.90 rows=86632 width=4) (actual time=0.096..568.560 rows=100000 loops=1)
 Planning Time: 0.094 ms
 Execution Time: 571.133 ms
(3 rows)

Time: 571.870 ms
[v15.3][660234] arrays=# explain analyze select vectors.sum_vector_vec(v) from vectors.data;
                                                 QUERY PLAN                                                  
-------------------------------------------------------------------------------------------------------------
 Seq Scan on data  (cost=0.00..1887.00 rows=100000 width=4) (actual time=0.039..694.131 rows=100000 loops=1)
 Planning Time: 0.074 ms
 Execution Time: 696.723 ms
(3 rows)

Time: 697.345 ms
[v15.3][660234] arrays=# explain analyze select vectors.sum_vector_slice(v) from vectors.data;
                                                 QUERY PLAN                                                  
-------------------------------------------------------------------------------------------------------------
 Seq Scan on data  (cost=0.00..1887.00 rows=100000 width=4) (actual time=0.033..420.596 rows=100000 loops=1)
 Planning Time: 0.035 ms
 Execution Time: 423.167 ms
(3 rows)

Time: 423.709 ms
[v15.3][660234] arrays=# explain analyze select vectors.sum_vector_simd(v) from vectors.data;
                                                 QUERY PLAN                                                  
-------------------------------------------------------------------------------------------------------------
 Seq Scan on data  (cost=0.00..1887.00 rows=100000 width=4) (actual time=0.028..373.343 rows=100000 loops=1)
 Planning Time: 0.037 ms
 Execution Time: 375.922 ms
(3 rows)

Time: 376.443 ms
```

pgrx-examples/arrays/src/lib.rs
```diff
@@ -100,8 +100,74 @@ fn return_vec_of_customtype() -> Vec<SomeStruct> {
...
+
+    #[pg_extern]
+    pub fn sum_vector_simd(values: Array<f32>) -> Result<f32, ArrayError> {
+        // this comes directly from https://stackoverflow.com/questions/23100534/how-to-sum-the-values-in-an-array-slice-or-vec-in-rust/67191480#67191480
+        // and https://www.reddit.com/r/rust/comments/13q56bp/comment/jlgq2t6/?utm_source=share&utm_medium=web2x&context=3
+        //
+        // the expectation here is that the compiler will autovectorize this code using SIMD
+        use std::convert::TryInto;
+
+        const LANES: usize = 16;
```
thomcc, line 139: This is definitely overkill (and will probably pessimize) but we don't really care here.

pgrx/src/datum/array.rs
```diff
@@ -282,6 +289,120 @@ impl<'a, T: FromDatum> Array<'a, T> {
...
+
+impl<'a> Array<'a, f64> {
+    /// Returns a slice of `f64`s which comprise this [`Array`].
+    ///
+    /// # Errors
+    ///
+    /// Returns a [`ArrayError::ArrayContainsNulls`] error if this [`Array`] contains one or more
+    /// SQL "NULL" values.  In this case, you'd likely want to fallback to using [`Array::iter()`].
+    pub fn as_slice(&self) -> Result<&[f64], ArrayError> {
+        if self.contains_nulls() {
```
thomcc, line 306: At some point it'd be nice to find a way to optimize `contains_nulls`, but for now this is great.

thomcc (approved):
Looks great!


I kind of wonder about having an `as_slice_and_nulls() -> (&[f64], &BitSlice)` or something. The null values would be unspecified but definitely initialized (since they come from disk).

That said, for now this is fine.

pgrx-examples/arrays/src/lib.rs
eeeebbbbrrrr: I took that directly from SO.  And like a good programmer I just assume it works. 

If you’re up for improving it for clarity and maybe even perf, I’d love that. My brain isn’t yet trained to work with these concepts.

pgrx/src/datum/array.rs
eeeebbbbrrrr: I looked at it and don’t see how. BitSlice doesn’t have a fast “at least one zero/one” short-circuit path.

eeeebbbbrrrr:
> Looks great!
> 
> I kind of wonder about having an `as_slice_and_nulls() -> (&[f64], &BitSlice)` or something. The null values would be unspecified but definitely initialized (since they come from disk).
> 
> That said, for now this is fine.

At that point the caller would end up implementing most of the logic we do in order to iterate the thing. Not sure it’s worth it. 

Could generally be useful to expose an iterator of the bitset, I suppose. Thoughts?

thomcc:
Re: as_slice_and_nulls, Yeah, it was mostly a half-thought-out idea. I haven't looked at it closely to see if it has legs.

pgrx-examples/arrays/src/lib.rs
thomcc: I'll look at it when I have less on my plate/when I feel like futzing around with some optimizations.

thomcc (approved):
Yep, still looks great.

BradyBonnette (approved):

pgrx-examples/arrays/src/lib.rs
jorgecarleitao: This idea is heavily used and tested in [arrow2](https://github.com/jorgecarleitao/arrow2/blob/main/src/compute/aggregate/sum.rs#L25) and is being used by [Polars](https://www.pola.rs/) and others for improved performance on aggregations.

We have a [repo](https://github.com/DataEngineeringLabs/simd-benches) with different implementations and corresponding benches.

Great work!

thomcc: The overkill was just 16 rather than 8 or 4, especially given that if the fold fails to be fully inlined there will be a lot of stack copies.

jorgecarleitao: (sorry, my comment was not meant to argue on that particular point!)

The background of 16 is that for a 32 bit width (e.g. f32 in SO), we have a total of 16*32=512 bits) so that the compiler can issue instructions from the [AVX-512]( https://en.wikipedia.org/wiki/AVX-512) set.

In arrow2 we have small trait system to generalize this across different types of bit widths so that the product is always 512 bits.

PR #1148, opened 2023-05-26
Few small array changes

thomcc:
- Optimizations (just adding `#[inline]`, `size_hint`, and `ExactSizeIterator` impls)
- Fix (I think?) incorrect lifetime bound in the internal `as_slice` (public API has the right bound, though).

### DIFF
--- a/pgrx/src/datum/array.rs
+++ b/pgrx/src/datum/array.rs
@@ -377,7 +377,7 @@ impl<'a> Array<'a, i8> {
 }
 
 #[inline(always)]
-fn as_slice<'a, T: Sized + FromDatum>(array: &Array<'a, T>) -> Result<&'_ [T], ArraySliceError> {
+fn as_slice<'a, T: Sized + FromDatum>(array: &'a Array<'_, T>) -> Result<&'a [T], ArraySliceError> {
     if array.contains_nulls() {
         return Err(ArraySliceError::ContainsNulls);
     }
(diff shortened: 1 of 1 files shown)


pgrx/src/datum/array.rs
```diff
@@ -648,8 +660,16 @@ impl<'a, T: FromDatum> Iterator for ArrayTypedIterator<'a, T> {
             element
         }
     }
+
+    #[inline]
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let len = self.array.raw.len().saturating_sub(self.curr);
```
thomcc, line 666: Using saturating_sub is probably slightly paranoid here, our `next()` impls generally check based relative comparison to the length (not strict equality), so it seems more honest.

pgrx/src/datum/array.rs
```diff
@@ -722,23 +752,15 @@ impl<'a, T: FromDatum> Iterator for ArrayIntoIterator<'a, T> {
         Some(element)
     }
 
+    #[inline]
     fn size_hint(&self) -> (usize, Option<usize>) {
-        // If asking for size, it's not clear if they want "actual size"
```
thomcc, line 726: It's asking for actual size, e.g. returning `Some(None)` counts as returning an element.

eeeebbbbrrrr (approved):

eeeebbbbrrrr (approved):

PR #1149, opened 2023-05-26
don't free the backing Array Datum pointer on drop, even if it was detoasted

eeeebbbbrrrr:


thomcc (approved):
Yeah, this is unfortunate, but could be worse.

PR #1150, opened 2023-05-26
Wait for postmaster.pid during `cargo pgrx test`

eeeebbbbrrrr:
Teach the pgrx test framework to wait for up to ~10s if the `postmaster.pid` file exists.

I've seen some instances where it's still there when we try to start the postmaster again for another test and waiting a bit longer allows it to get properly cleaned up.


thomcc (approved):
Seems fine. Odd that it fixes the issue without entering the wait loop but, so it goes.

PR #1151, opened 2023-05-26
upgrade all dependencies (except for `syn`) to their latest versions

eeeebbbbrrrr:
Resolve depreciation warning in `build.rs`.

Note that `cargo-pgrx` did get its `syn` updated to the latest since it also uses `prettyplease`.  This is fine for it, but much more complex for the rest of the code, so `syn` stays at 1.0.109.


thomcc (approved):
Looks great

PR #1154, opened 2023-05-30
Add --configure-flag option for cargo pgrx init

ewencp:
This allows passing through flags to the `configure` script when building postgres. For example, using
`'--configure-flag=--with-ssl=openssl'` enables openssl support, including contrib extensions like pgcrypto.

### DIFF
--- a/cargo-pgrx/README.md
+++ b/cargo-pgrx/README.md
@@ -118,7 +118,7 @@ However, if the "path to pg_config" is the literal string `download`, then `pgrx
 
 When the various `--pgXX` options are specified, these are the **only** versions of Postgres that `pgrx` will manage for you.
 
-You'll also want to make sure you have the "postgresql-server-dev" package installed for each version you want to manage yourself.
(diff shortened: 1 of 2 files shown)


ewencp:
I previously implemented #1065 to include contrib extensions. However, not everything gets included because some require additional configs to be enabled -- pgcrypto is specifically relevant to my use case so I used that as an example. Using this to pass `--with-ssl`, `--with-includes` and `--with-libraries` gets the extension installed so it can be usable. Since it's just passing things along to `configure`, it also works for core postgres functionality that's toggleable at build time.

eeeebbbbrrrr (changes requested):
If you will, please update the `cargo-pgrx/README.md` with this new flag.  Then I think this will be ready to merge.

ewencp:
Updated. That section is getting a bit unwieldy for adding docs to, it might need to be broken up into subsections for different use cases at some point to make it more manageable (and perhaps not all flags need to be documented there as long as they are documented clearly enough in the `--help` output).

eeeebbbbrrrr (approved):

eeeebbbbrrrr:
Thanks, @ewencp.

It's to have at least something in pgrx where the complaint is "too many docs"!  If we had the resources, I'd kick off a mdbook project for pgrx.

PR #1157, opened 2023-06-02
Fix version check

thomcc:
I was going to add a version check to plrustc, so I started by copying the last time I wrote a version check. Turns out, last time I wrote it, I got it wrong, so this fixes that.

### DIFF
--- a/pgrx-pg-sys/build.rs
+++ b/pgrx-pg-sys/build.rs
@@ -31,7 +31,7 @@ fn is_nightly() -> bool {
         return false;
     }
     let rustc = env_tracked("RUSTC").map(PathBuf::from).unwrap_or_else(|| "rustc".into());
-    let output = match std::process::Command::new(rustc).arg("--verbose").output() {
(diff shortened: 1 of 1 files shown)


pgrx-pg-sys/build.rs
```diff
@@ -31,7 +31,7 @@ fn is_nightly() -> bool {
         return false;
     }
     let rustc = env_tracked("RUSTC").map(PathBuf::from).unwrap_or_else(|| "rustc".into());
-    let output = match std::process::Command::new(rustc).arg("--verbose").output() {
+    let output = match std::process::Command::new(rustc).arg("--version").output() {
```
BradyBonnette, line 34: 🤔

eeeebbbbrrrr: that's a hilarious bug.

```
$ rustc --verbose
error: no input filename given

$  rustc --version
rustc 1.69.0 (84c898d65 2023-04-16)
```

Both of those pass the conditional at the end of the function.  🤦

eeeebbbbrrrr (approved):

PR #1159, opened 2023-06-05
`PgVarlena` now knows how to be used by `#[derive(PostgresEq/Ord/Hash)]`

eeeebbbbrrrr:
The `PgVarlena` type was lacking some trait impls for doing proper comparisions of the `T` it wraps.  Additionally, the macros that implement the PostgresEq/Ord/Hash operators didn't know about `PgVarlena` and were emitting incorrect Rust wrappers.

Add a new example in `pgrx-examples/operators/src/pgverlena.rs` with a unit test.

As a drive-by, `impl std::fmt::Write for StringInfo` (in `stringinfo.rs`).

### DIFF
--- a/pgrx-examples/operators/src/pgvarlena.rs
+++ b/pgrx-examples/operators/src/pgvarlena.rs
@@ -49,6 +49,7 @@ pub struct PgVarlenaThing {
 }
 
 impl PgVarlenaThing {
+    #[allow(dead_code)]
     pub fn new(a: u64, b: u64, c: i32, d: [u8; 5]) -> Self {
         Self { a, b, c, d }
     }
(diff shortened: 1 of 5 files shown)


pgrx-examples/operators/src/pgvarlena.rs
```diff
@@ -0,0 +1,135 @@
...
+        f.write_str(&self.c.to_string())?;
+        f.write_char(';')?;
+        f.write_str(&format!("{:?}", self.d))
+    }
+}
+
+// implement function to parse string representation of this type and a function to convert it
+// to a string
+impl PgVarlenaInOutFuncs for PgVarlenaThing {
+    fn input(input: &CStr) -> PgVarlena<Self>
```
thomcc, line 72: This could also be cleaned up, but it's a bit more involved and I'm not sure how much we care.

thomcc (approved):
There's a few nitpicks about the example but they don't really matter so it's your call.

I don't yet fully understand the stuff the rest of the code is doing, so I'm mostly just rubberstamping that part.

PR #1162, opened 2023-06-06
Simplify and fix some correctness issues with fixed-size byval arrays

thomcc:
- Simplifies the code a good amount, mostly reling on postgres's documentation about byval types to be correct (and for the optimizer see through `u32 -> Datum -> u32` and such).
- Inlines some datum functions which are required for the optimizer to actually see through that in all cases.
- Debug asserts about alignment more consistently, but it shouldn't ever be hit on x86_64 or aarch64.
- if target_pointer_width == "32" some of the slice impls were busted, so they've been disabled on 32 bit. (Note that 32 bit support is quite experimental, and there are probably more things we do wrong there)

### DIFF
--- a/pgrx/src/datum/array.rs
+++ b/pgrx/src/datum/array.rs
@@ -132,25 +132,13 @@ impl<'a, T: FromDatum> Array<'a, T> {
         // which implementation is correct for the type of element in this Array.
         let slide_impl: ChaChaSlideImpl<T> = match elem_layout.pass {
             PassBy::Value => match elem_layout.size {
-                // the layout size is fixed and it exactly matches the size of `T`
-                //
-                // Note that this doesn't guarantee that the elements are actually `T`s, only
-                // that they'll fit into it.  It's the caller's responsibility to make sure
-                Size::Fixed(size)
-                    if (size as usize) == std::mem::size_of::<T>()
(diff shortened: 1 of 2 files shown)


pgrx/src/datum/array.rs
```diff
@@ -418,132 +408,43 @@ mod casper {
...
-            2
-        }
+    /// Safety: Equivalent to a (potentially) aligned read of `ptr`, which
+    /// should be `Copy` (ideally...).
+    #[track_caller]
+    #[inline(always)]
+    pub(super) unsafe fn byval_read<T: Copy>(ptr: *const u8) -> T {
+        let ptr = ptr.cast::<T>();
+        debug_assert!(is_aligned(ptr), "not aligned to {}: {ptr:p}", std::mem::align_of::<T>());
+        ptr.cast::<T>().read()
```
eeeebbbbrrrr, line 423: any need to do this cast twice?  Second one seems superfluous

eeeebbbbrrrr (changes requested):

eeeebbbbrrrr:
This is v nice, @thomcc.  Great cleanup and simplification and correctness.  Much appreciated!  Feel free to merge when you're ready.

PR #1164, opened 2023-06-08
Properly Support `RETURNS TABLE (T)` (returning a table with 1 field)

eeeebbbbrrrr:
Postgres treats a SRF that returns a 1-field table as if it is a `RETURNS SETOF T`, so we need to do that too.

This adds a standalone function in the `srf` example, and a few tests that ensure this also works when returning it wrapped in `Result` or `Option` (or both!)

### DIFF
--- a/pgrx-examples/srf/src/lib.rs
+++ b/pgrx-examples/srf/src/lib.rs
@@ -40,6 +40,11 @@ fn result_table<'a>() -> Result<
     Ok(Some(TableIterator::new(vec![(Some(1), Some(2))])))
 }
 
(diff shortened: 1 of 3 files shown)


thomcc (approved):
Seems fine, this code continues to be gnarly as hell, but that's how it is I suppose.

eeeebbbbrrrr:
> Seems fine, this code continues to be gnarly as hell, but that's how it is I suppose.

Yeah.  It's pretty much at the point where some considered design needs to happen before we start overloading it with more functionality.

Issue #1087, opened 2023-03-31
Error "couldn't parse postmaster" when using Postgres.app on MacOS 


clowder:
Running `cargo pgx install` fails with the error "couldn't parse postmaster" when building against a Postgres version managed by Postgres.app. 

rustc 1.68.2
pgx 0.7.4

```bash
$> cargo pgx install
       Using PgConfig("pg15") and `pg_config` from /Applications/Postgres.app/Contents/Versions/latest/bin/pg_config
    Building extension with features pg15
     Running command "/opt/homebrew/Cellar/rust/1.68.1/bin/cargo" "build" "--features" "pg15" "--no-default-features" "--message-format=json-render-diagnostics"
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
  Installing extension
     Copying control file to /Applications/Postgres.app/Contents/Versions/15/share/postgresql/extension/foo.control
     Copying shared library to /Applications/Postgres.app/Contents/Versions/15/lib/postgresql/foo.so
 Discovering SQL entities
Error: 
   0: couldn't parse postmaster
   1: Unsupported file format

Location:
   /Users/clowder/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo-pgx-0.7.4/src/command/schema.rs:506

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ SPANTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

   0: cargo_pgx::command::schema::create_stub with postmaster_path=/Applications/Postgres.app/Contents/Versions/15/bin/postmaster postmaster_stub_dir=/Users/clowder/.pgx/postmaster_stubs/Applications/Postgres.app/Contents/Versions/15/bin
      at /Users/clowder/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo-pgx-0.7.4/src/command/schema.rs:467
   1: cargo_pgx::command::schema::generate_schema with pg_version=15.2 profile=Dev test=false path=/Applications/Postgres.app/Contents/Versions/15/share/postgresql/extension/foo--0.0.0.sql features=["pg15"]
      at /Users/clowder/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo-pgx-0.7.4/src/command/schema.rs:172
   2: cargo_pgx::command::install::install_extension with pg_version=15.2 profile=Dev test=false features=["pg15"]
[...]

PR #1166, opened 2023-06-09
Support building against macOS universal binaries

clowder:
FIXES: #1087

A "universal binary" is a fat binary which contains multiple builds for different architectures. When Postgres is packaged in this format it's necessary to slice out the build for the current architecture before trying to parse it. The program terminates early if the universal binary doesn't support the current architecture.

The list of supported architectures in `slice_arch32` is limited by [`object`][object-arch], which supports fewer than [Rust][rust-arch].

The universal binary test fixture was simply chosen because of it's small size relative to other binaries in Postgres.

[^1]: https://github.com/gimli-rs/object/issues/454#issuecomment-1200668989

[rust-arch]: http://doc.rust-lang.org/1.68.2/std/env/consts/constant.ARCH.html
[object-arch]: https://docs.rs/object/latest/object/read/macho/trait.FatArch.html#method.architecture


eeeebbbbrrrr:
Hey, this is cool.  Thanks a bunch.  I'm sure @thomcc will be excited to review this in detail.

> The universal binary test fixture was simply chosen because of it's small size relative to other binaries in Postgres.

Any chance we can rename this to anything other than `pg_config`?  I just know that name will cause confusion down the road since pgrx uses the real pg_config everywhere.

clowder:
@eeeebbbbrrrr Thanks for the quick reply!

> Any chance we can rename this to anything other than `pg_config`? I just know that name will cause confusion down the road since pgrx uses the real pg_config everywhere.

Happy to do this. Just to clarify, it's literally a universal binary version of the pg_config utility. With that in mind, would reshuffling the parts be enough? For example `cargo-pgrx/tests/fixtures/pg_config.macos-universal`. Or should we just give it some opaque name? For example `cargo-pgrx/tests/fixtures/macos-universal`.

eeeebbbbrrrr:
Yeah, i'd prefer it be called something like `test-macos-universal-binary` or similar.  I suppose we don't care what the binary actually is.

Also, I can't tell in the PR, but please make sure it doesn't accidentally have the execute bit (+x) set.

clowder:
> Yeah, i'd prefer it be called something like `test-macos-universal-binary` or similar. I suppose we don't care what the binary actually is.

Updated. I opted for `macos-universal-binary` because `test` was already in the path.

> Also, I can't tell in the PR, but please make sure it doesn't accidentally have the execute bit (+x) set.

It did. This has been fixed now.

eeeebbbbrrrr:
Nice, thanks!  I just realized I never approved you for CI, so now that's happening.  I'd still like @thomcc to give this a look and approve.  He's more experienced with this kind of thing than I am.

clowder:
> Yeah, glad I turned on CI. It doesn't like something here: https://github.com/tcdi/pgrx/actions/runs/5269612633/jobs/9528132717#step:10:160

Sorry about that, I missed something renaming that fixture. Sorted now.

cargo-pgrx/src/command/schema.rs
```diff
@@ -556,3 +558,82 @@ fn create_stub(
...
+                    panic!("Failed to slice architecture '{arch}' from universal binary.")
+                }
+            }
+        }
+        _ => object::File::parse(&*data),
+    }
+}
+
+fn slice_arch32<'a>(data: &'a [u8], arch: &str) -> Option<&'a [u8]> {
+    let target = match arch {
```
thomcc, line 581: Most of these almost certainly don't work, but I suppose it's harmless.

thomcc (approved):
Sorry for the delay, this looks fine to me.

PR #1167, opened 2023-06-09
Underspecify rather than overspecify the serde/serde_json/serde_derive versions we require

thomcc:
Allows making some issues with plrust dependency handling much less of a problem in practice than they might otherwise be.


eeeebbbbrrrr (approved):
Thanks @thomcc

PR #1169, opened 2023-06-15
Postgres 16beta1 Support

eeeebbbbrrrr:
This lands support for Postgres 16beta1.  It does it in such a way that it won't be a gigantic pain in the ass to soon add support for subsequent betas and release candidates.

It's understood by the codebase as just "pg16", so we won't be able to support, say, beta1, beta2, and rc1 concurrently, but that's okay.  That sounds terrible anyhow.

There were a few internal pgrx changes to fix compatibility and get tests passing, but nothing major.  

Note that the `GucSetting`s for strings have changed from `GucSetting<Option<&'static str>>` to `GucSetting<Option<&'static CStr>>` because the latter is what they really are.

I note to whoever might merge this PR... I think it would be better to do a regular merge here, rather than a squash merge.  There's only 3 commits (so far), and I think I'd prefer to see those reflected as such in the git history.


eeeebbbbrrrr:
Need to figure out CI support for 16beta1.

pgrx-pg-config/src/lib.rs
```diff
@@ -588,7 +649,36 @@ impl Pgrx {
...
+}
+
+pub fn is_supported_major_version(v: u16) -> bool {
+    for pgver in SUPPORTED_VERSIONS() {
+        if v == pgver.major {
+            return true;
+        }
+    }
+
+    false
```
workingjubilee, line 680:
```suggestion
SUPPORTED_VERSIONS().into_iter().any(|pgver| pgver.major == v)
```

cargo-pgrx/src/command/version.rs
```diff
@@ -12,16 +13,17 @@ pub(crate) fn pgrx_default(supported_major_versions: &[u16]) -> eyre::Result<Pgr
...
+    use std::collections::BTreeMap;
     use url::Url;
 
     use crate::command::build_agent_for_url;
 
     pub(super) struct PostgreSQLVersionRss;
 
     impl PostgreSQLVersionRss {
-        pub(super) fn new(supported_major_versions: &[u16]) -> eyre::Result<Vec<PgVersion>> {
+        pub(super) fn new(supported_verions: &[PgVersion]) -> eyre::Result<Vec<PgVersion>> {
```
workingjubilee, line 26:
```suggestion
        pub(super) fn new(supported_versions: &[PgVersion]) -> eyre::Result<Vec<PgVersion>> {
```

cargo-pgrx/src/command/version.rs
```diff
@@ -35,7 +37,10 @@ mod rss {
                 Err(e) => return Err(e.into()),
             };
 
-            let mut versions = Vec::new();
+            let mut versions: BTreeMap<u16, PgVersion> = BTreeMap::from_iter(
+                supported_verions.iter().map(|pgver| (pgver.major, pgver.clone())),
```
workingjubilee, line 41:
```suggestion
                supported_versions.iter().map(|pgver| (pgver.major, pgver.clone())),
```

pgrx-pg-config/src/lib.rs
```diff
@@ -270,7 +326,12 @@ impl PgConfig {
 
     pub fn postmaster_path(&self) -> eyre::Result<PathBuf> {
         let mut path = self.bin_dir()?;
-        path.push("postmaster");
+
+        if self.major_version()? > 15 {
+            path.push("postgres")
+        } else {
+            path.push("postmaster");
+        }
```
workingjubilee, line 334: I am curious if checking for postgres unconditionally would break anything, but perhaps it is best to satisfy that curiosity later.

pgrx-pg-sys/src/lib.rs
```diff
@@ -360,30 +383,39 @@ mod all_versions {
...
-            // means it'll have this header before it
-            *(pointer
-                .cast::<::std::os::raw::c_char>()
-                .sub(std::mem::size_of::<*mut ::std::os::raw::c_void>())
-                .cast())
-        };
-
-        assert!(MemoryContextIsValid(context));
+        // Postgres versions <16 don't export the "GetMemoryChunkContext" function.  It's a "static inline"
+        // function in `memutils.h`, so we port it to Rust right here
```
workingjubilee, line 387: Exciting.

pgrx-pg-config/src/lib.rs
```diff
@@ -588,7 +649,36 @@ impl Pgrx {
...
+        PgVersion::new(11, PgMinorVersion::Latest, None),
+        PgVersion::new(12, PgMinorVersion::Latest, None),
+        PgVersion::new(13, PgMinorVersion::Latest, None),
+        PgVersion::new(14, PgMinorVersion::Latest, None),
+        PgVersion::new(15, PgMinorVersion::Latest, None),
+        PgVersion::new(
+            16,
+            PgMinorVersion::Beta(1),
+            Some(
+                Url::parse(
```
workingjubilee, line 664: I wonder if this can be made const fn and it simply hasn't had anyone try yet.

pgrx-pg-sys/src/lib.rs
```diff
@@ -22,9 +22,11 @@ Use of this source code is governed by the MIT license that can be found in the
 #[cfg(
     any(
         // no features at all will cause problems
-        not(any(feature = "pg11", feature = "pg12", feature = "pg13", feature = "pg14", feature = "pg15")),
+        not(any(feature = "pg11", feature = "pg12", feature = "pg13", feature = "pg14", feature = "pg15", feature = "pg16")),
   ))]
-std::compile_error!("exactly one one feature must be provided (pg11, pg12, pg13, pg14, pg15)");
+std::compile_error!(
+    "exactly one one feature must be provided (pg11, pg12, pg13, pg14, pg15, pg16)"
```
workingjubilee, line 28:
```suggestion
    "exactly one feature must be provided (pg11, pg12, pg13, pg14, pg15, pg16)"
```

pgrx/src/datum/range.rs
```diff
@@ -385,8 +385,22 @@ where
...
+                {
+                    pg_sys::make_range(
+                        typecache,
+                        &mut lower_bound,
+                        &mut upper_bound,
+                        is_empty,
+                        std::ptr::null_mut(),
+                    )
+                }
+            };
```
workingjubilee, line 403: These `cfg` can rest atop a `let`, reducing the number of nested brackets.

workingjubilee:
```suggestion
            #[cfg(not(feature = "pg16"))]
            let range_type =
                pg_sys::make_range(typecache, &mut lower_bound, &mut upper_bound, is_empty);
            #[cfg(feature = "pg16")]
            let range_type = pg_sys::make_range(
                typecache,
                &mut lower_bound,
                &mut upper_bound,
                is_empty,
                std::ptr::null_mut(),
            );
```

pgrx/src/guc.rs
```diff
@@ -114,72 +114,78 @@ where
...
+            } else {
+                Some(CStr::from_ptr(self.value.get() as *const _))
+            }
+        }
     }
 
-    unsafe fn as_ptr(&self) -> *mut *mut std::os::raw::c_char {
-        self.char_p.as_ptr()
+    fn as_ptr(&self) -> *mut *mut std::os::raw::c_char {
+        self.value.as_ptr() as *mut std::os::raw::c_char as *mut _
```
workingjubilee, line 188: is there a reason `as *mut *mut std::os::raw::c_char` doesn't work?
```suggestion
        self.value.as_ptr() as *mut *mut std::os::raw::c_char
```

pgrx/src/tupdesc.rs
```diff
@@ -174,13 +175,11 @@ impl<'a> PgTupleDesc<'a> {
     ```
     */
     pub fn for_composite_type(name: &str) -> Option<PgTupleDesc<'a>> {
-        unsafe {
-            let mut typoid = pg_sys::Oid::INVALID;
-            let mut typmod = 0;
-            pg_sys::parseTypeString(name.as_pg_cstr(), &mut typoid, &mut typmod, true);
+        let typoid = PgTryBuilder::new(|| Some(regtypein(name)))
+            .catch_when(PgSqlErrorCode::ERRCODE_UNDEFINED_OBJECT, |_| None)
+            .execute()?;
```
workingjubilee, line 180: Huh, what's the difference?

workingjubilee (commented):
I need to examine the GetMemoryChunkContext stuff more closely, but I had a few nits and a question that are independent.

pgrx/src/tupdesc.rs
workingjubilee: Ah I see, previously we were saying `missing_ok = true` and letting an `InvalidOid` potentially pass through, which would become a None. But `parseTypeString` in new Postgres versions now has an `ErrorSaveContext*`, and we don't have an API design that supports using that instead for error handling. So you chose to switch to regtypein, which does that business for us, and we do throw-catch based errors instead, and this gives us something consistent across all our supported Postgres versions with effectively equivalent functionality. We might want to add ErrorSaveContext support, sounds kinda Rusty to do, but... later.

pgrx-pg-config/src/lib.rs
```diff
@@ -193,16 +221,32 @@ impl PgConfig {
...
+        if version.len() == 1 {
+            // it's hopefully a "beta" or "rc" release
+            let first = &version[0];
+
+            if first.contains("beta") {
+                beta = true;
+                version = first.split("beta").collect();
+            } else if first.contains("rc") {
+                rc = true;
+                version = first.split("beta").collect();
```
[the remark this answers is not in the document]
eeeebbbbrrrr: Good catch, thanks!

pgrx/src/tupdesc.rs
eeeebbbbrrrr: Exactly.  

Also, in looking at it in detail, I think that `parseTypeString` was probably the wrong thing to be using in the first place, but it just happened to work out okay.  Using the `regtypein` function I think would actually be the preferred way of doing this.

eeeebbbbrrrr:
> I need to examine the GetMemoryChunkContext stuff more closely, but I had a few nits and a question that are independent.

I've addressed everything that you mentioned here.  All good catches, thanks a bunch!

As far as the MemoryContextChunk business, it's actually simple for pg16... it's the other versions where we had to port their static inline function to Rust (which I did last year, so that code has been around for awhile).  For reference, here's where it came from:  https://github.com/postgres/postgres/blob/4be308edee4495fe72075397efa0f6e0e0f9693b/src/include/utils/memutils.h#L114

workingjubilee (approved):
This PR includes breaking changes, but aside from that, seems good to me.

PR #1177, opened 2023-06-20
Fix bug with converting a `pg_sys::Datum` into a `pgrx::Date`

eeeebbbbrrrr:
The cast to `isize` was wrong.  It needs to be a cast to `pg_sys::DateADT` which is an `i32`.


thomcc (approved):
Looks good to me

thomcc:
For anybody keeping track this is a large positive number that fits in an isize, but needs to come through as a negative i32.

PR #1180, opened 2023-06-29
Fix Arrays with leading nulls

eeeebbbbrrrr:
This test fails with some form of:

```
compression method lz4 not supported
Postgres location: toast_compression.c:187
```

A classic off-by-one error caused this:
If element 0 was null, and 1 was non-null, data started with 1,
but iteration would advance before unpacking.
This allowed moving to uninit data, or just skipping elements.

### DIFF
--- a/pgrx-tests/src/tests/array_tests.rs
+++ b/pgrx-tests/src/tests/array_tests.rs
@@ -484,7 +484,7 @@ mod tests {
     }
 
     #[pg_test]
-    fn test_text_array() -> Result<(), Box<dyn std::error::Error>> {
+    fn test_text_array_as_vec_string() -> Result<(), Box<dyn std::error::Error>> {
         let a = Spi::get_one::<Array<String>>(
             "SELECT ARRAY[NULL, NULL, NULL, NULL, 'the fifth element']::text[]",
         )?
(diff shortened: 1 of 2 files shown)


pgrx-tests/src/tests/array_tests.rs
```diff
@@ -482,4 +482,16 @@ mod tests {
...
+
+    #[pg_test]
+    fn test_text_array() -> Result<(), Box<dyn std::error::Error>> {
+        let a = Spi::get_one::<Array<String>>(
+            "SELECT ARRAY[NULL, NULL, NULL, NULL, 'the fifth element']::text[]",
+        )?
+        .expect("spi result was NULL")
+        .into_iter()
+        .collect::<Vec<_>>();
+        assert_eq!(a, vec![None, None, None, None, Some(String::from("the fifth element"))]);
```
BradyBonnette, line 494: If this passes on all distros in every PG configuration, does that mean it is..

... multi-pass?

eeeebbbbrrrr: when it does, YES!  hahaha

eeeebbbbrrrr:
A different failure, but probably the same?

```
attempt to subtract with overflow
Postgres location: varlena.rs:236
```

workingjubilee:
No idea why it is specifically failing to link and run the versioned so tests.

workingjubilee (approved):
I assert that the remaining test failures are unrelated.

Issue #1179, opened 2023-06-24
Supporting PostgreSQL forks


workingjubilee:
Some users wish to use PGRX with "almost Postgres" databases: forks that are 90% conformant to how Postgres operates and its internal sources, but that deviate in some regard. These can hypothetically be supported, but will require external assistance to understand the deviations of the fork, as we are, broadly speaking, specialists in Postgres, not Greenplum and so on. It would be especially helpful to have assistance in running and troubleshooting CI for such a fork.

These databases likely will also have to be defined separately from the normal version features, e.g. not `pg12` but `greenplum7`. Or perhaps two features, with both `greenplum7` and `pg12`? Either way.

Example issues:
- https://github.com/postgresml/postgresml/issues/780

workingjubilee:
Specific problems that need to be addressed to enable compiling some PGRX-based extensions against Greenplum 7, which is based on PostgreSQL 12 and thus would normally use this header as the basis of our generated bindings: https://github.com/tcdi/pgrx/blob/master/pgrx-pg-sys/include/pg12.h
- [ ] `heap_copytuple` not present in the included headers
- [ ] `heap_form_tuple` not present in the included headers
- [ ] `MemoryContextDelete` not present in the included headers
- [ ] New field in `BackgroundWorker`: `bgw_start_rule`
- [ ] `ProcessInterrupts` now expects two arguments, a `*const c_char` and `c_int`?

The relevant code in PGRX needs to either be changed to use different functions when whatever appropriate `cfg` is decided on is enabled, a new set of headers need to be defined for that database version to compile against, or whatever else.

cyccbxhl:
Hello,

I am currently working on adding support for Greenplum 7 to pgrx.  I've introduced a new feature flag called "gp7". At this moment,  it can compile within an existing GP project by using the following commands:
```
export PGRX_PG_CONFIG_PATH=${Where-your-GP-pg_config}
cargo build --features gp7 --no-default-features
```
May I contribute my code to your project? If so, please let me know what else needs to be done. As far as I'm aware, the remaining tasks include integrating tests for GP7, determining whether there is a need to interface with various cargo-pgrx commands (such as cargo pgrx init/run), as well as some documentation work.

lij55:
> Hello,
> 
> I am currently working on adding support for Greenplum 7 to pgrx. I've introduced a new feature flag called "gp7". At this moment, it can compile within an existing GP project by using the following commands:
> 
> ```
[...]

PR #1184, opened 2023-07-03
varlena: support bitflag in bigendian format.

Sasasu:
Postgresql has an endian-dependent format for the bitflag in varlena. currently pgrx only support smallendian. adding bigendian support.

Greenplum uses the bigendian format for all platforms.

----

I will double check if there is some copy&past error tomorrow. :>

### DIFF
--- a/pgrx/src/varlena.rs
+++ b/pgrx/src/varlena.rs
@@ -281,7 +281,6 @@ pub fn rust_byte_slice_to_bytea(slice: &[u8]) -> PgBox<pg_sys::bytea> {
     }
 }
 
-#[cfg(target_endian = "little")]
 mod varlena_littleendian {
     use super::*;
 
@@ -375,7 +374,6 @@ mod varlena_littleendian {
     }
 }
 
-#[cfg(target_endian = "big")]
 mod varlena_bigendian {
     use super::*;
 
--- a/pgrx/src/varlena.rs
+++ b/pgrx/src/varlena.rs
@@ -12,17 +12,10 @@ Use of this source code is governed by the MIT license that can be found in the 
 use crate::{pg_sys, PgBox};
 use core::{slice, str};
 
-/// # Safety
-///
-/// The caller asserts the specified `ptr` really is a non-null, palloc'd [`pg_sys::varlena`] pointer
-#[inline(always)]
-pub unsafe fn set_varsize_4b(ptr: *mut pg_sys::varlena, len: i32) {
-    // #define SET_VARSIZE_4B(PTR,len) \
-    // 	(((varattrib_4b *) (PTR))->va_4byte.va_header = (((uint32) (len)) << 2))
-
-    // SAFETY:  A varlena can be safely cast to a varattrib_4b
-    (*ptr.cast::<pg_sys::varattrib_4b>()).va_4byte.as_mut().va_header = (len as u32) << 2;
-}
+#[cfg(target_endian = "big")]
+pub use varlena_bigendian::*;
+#[cfg(target_endian = "little")]
+pub use varlena_littleendian::*;
 
 /// # Safety
 ///
@@ -31,18 +24,6 @@ pub unsafe fn set_varsize_4b(ptr: *mut pg_sys::varlena, len: i32) {
 pub unsafe fn set_varsize(ptr: *mut pg_sys::varlena, len: i32) {
     // #define SET_VARSIZE(PTR, len)				SET_VARSIZE_4B(PTR, len)
     set_varsize_4b(ptr, len)
-}
-
-/// # Safety
-///
-/// The caller asserts the specified `ptr` really is a non-null, palloc'd [`pg_sys::varlena`] pointer
-#[inline(always)]
-pub unsafe fn set_varsize_1b(ptr: *mut pg_sys::varlena, len: i32) {
-    // #define SET_VARSIZE_1B(PTR,len) \
-    // 	(((varattrib_1b *) (PTR))->va_header = (((uint8) (len)) << 1) | 0x01)
-
-    // SAFETY:  A varlena can be safely cast to a varattrib_1b
-    (*ptr.cast::<pg_sys::varattrib_1b>()).va_header = ((len as u8) << 1) | 0x01
 }
 
 /// # Safety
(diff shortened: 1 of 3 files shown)


thomcc:
> Greenplum uses the bigendian format for all platforms.

This means `target_endian = "little"` is wrong, no?

eeeebbbbrrrr:
The pgrx project has no facilities for testing on big endian architectures.  I haven't yet validated that the new big endian impls of these macros are correct/identical-to-postgres-impls, but even if they are, we can't accept this simply because we can't test it.

Is this part of a larger effort by Greenplum to offer pgrx support?  If so, I'd like an opportunity to speak to someone in the company.  My email is e_ridge @ tcdi.com

pgrx/src/varlena.rs
```diff
@@ -12,17 +12,10 @@ Use of this source code is governed by the MIT license that can be found in the
...
-    // #define SET_VARSIZE_4B(PTR,len) \
-    // 	(((varattrib_4b *) (PTR))->va_4byte.va_header = (((uint32) (len)) << 2))
-
-    // SAFETY:  A varlena can be safely cast to a varattrib_4b
-    (*ptr.cast::<pg_sys::varattrib_4b>()).va_4byte.as_mut().va_header = (len as u32) << 2;
-}
+#[cfg(target_endian = "big")]
+pub use varlena_bigendian::*;
+#[cfg(target_endian = "little")]
+pub use varlena_littleendian::*;
```
workingjubilee, line 18: Given what you said
```suggestion
#[cfg(any(feature = "greenplum7", target_endian = "big"))]
pub use varlena_bigendian::*;
#[cfg(not(any(feature = "greenplum7", target_endian = "big")))]
pub use varlena_littleendian::*;
```

pgrx/src/varlena.rs
```diff
@@ -375,3 +280,192 @@ pub fn rust_byte_slice_to_bytea(slice: &[u8]) -> PgBox<pg_sys::bytea> {
         ))
     }
 }
+
+#[cfg(target_endian = "little")]
```
workingjubilee, line 284: There really isn't a benefit to putting the cfg here.

```suggestion
```

pgrx/src/varlena.rs
```diff
@@ -375,3 +280,192 @@ pub fn rust_byte_slice_to_bytea(slice: &[u8]) -> PgBox<pg_sys::bytea> {
...
+    /// #define VARATT_IS_1B_E(PTR) ((((varattrib_1b *) (PTR))->va_header) == 0x01)
+    /// ```
+    #[inline]
+    pub unsafe fn varatt_is_1b_e(ptr: *const pg_sys::varlena) -> bool {
+        let va1b = ptr as *const pg_sys::varattrib_1b;
+        (*va1b).va_header == 0x01
+    }
+}
+
+#[cfg(target_endian = "big")]
```
workingjubilee, line 378: There really isn't a benefit to putting the cfg here.

```suggestion
```

workingjubilee (changes requested):
This, of course, highlights a deeper problem, which is that we need to actually sit down and agree on the feature for enabling Greenplum support.

pgrx/src/varlena.rs
eeeebbbbrrrr: Just to be clear, @workingjubilee's suggestion here is just for demonstration purposes.  We are not in any position to invent a `greenplum7` feature yet.  Such a thing will require a significant amount of planning, design, and understanding on our part, along with the ability to test it... and test it, apparently, on disparate CPU architectures.

Sasasu: My goal is to add greenplum support, but this PR is not related to greenplum. and this PR did not add the `greenplum7` feature. If this PR gets merged, I will add gp7 support based on this PR.

Sasasu:
> > Greenplum uses the bigendian format for all platforms.
> 
> This means `target_endian = "little"` is wrong, no?

this PR is not related to greenplum. for pgrx `target_endian = "little"` is correct.
but greenplum remove the littleendian code, and always use bigendian bitflag on all platforms for an unknown reason.

this bitflag does not use to calculate, so it is possible to use bigendian bitflag on littleendian platforms.

this PR is adding the support for bitflag in bigendian format. not related to greenplum.

workingjubilee:
I understand that, but we strongly prefer to be able to test changes for validation of their correctness, **especially** with the varlena-reading tests. It is very hard for us to validate changes are correct without a big-endian system to test on, and unfortunately it isn't possible to use Rust's byte-swapping facilities such as [`from_be_bytes`](https://doc.rust-lang.org/std/primitive.u32.html#method.from_be_bytes) and [`to_le_bytes`](https://doc.rust-lang.org/std/primitive.u32.html#method.to_le_bytes) to reimplement these macros, to my understanding of the bit-manipulations. If it were, I would unhesitatingly recommend using an "endian-oblivious" implementation.

Most of all, I don't want to approve this and then find out some other part of our code is busted as hell on big-endian systems. Better for it to fail in some blatantly obvious way... I have many grievances with Postgres but it is surprisingly good at recovering from rampant data corruption and erroring when it enters invalid states. I'd rather someone who tried to use PGRX on a big-endian system encounter an early, spectacular failure, than get far enough for them to induce a more subtle data corruption that Postgres does not detect.

Sasasu:
> Most of all, I don't want to approve this and then find out some other part of our code is busted as hell on big-endian systems. 

this is Ok, there is no modern CPU support big-endian anymore. I agree with you.
I will close this PR after I verified on my machine. there still have a problem in `get_nullable_datum` or `varsize_any_exhdr`.

---

and, just an ask,  Can you accept moving most of functions to cshim? This will introduce some function call overhead, which will slightly reduce performance. this ABI problem is already handled in the C side (by static inline and macro functions)

---

for Greenplum support, there are much more works to do. not only the things mentioned in issue 1179.
I will complete my usage first.

eeeebbbbrrrr:
> this is Ok, there is no modern CPU support big-endian anymore. I agree with you.
I will close this PR after I verified on my machine. there still have a problem in get_nullable_datum or varsize_any_exhdr.

If you're talking about https://github.com/tcdi/pgrx/pull/1184/commits/e3464a1f8c4d5c4926aa5dbf717606079d552c5a, please make a separate PR for that -- it does look like it confused length in bytes with number of elements.

> and, just an ask, Can you accept moving most of functions to cshim? This will introduce some function call overhead, which will slightly reduce performance. this ABI problem is already handled in the C side (by static inline and macro functions)

Most the functions being discussed in PR were in the cshim at one point.  The primary problem with the cshim is the ability to easily cross-compile it.  Some downstream pgrx users enjoy cross compiling from x86_64 to aarch64.  In fact, that's a key feature of PL/Rust.

Internally, we've discussed inventing some very lightweight "C to Rust" transformer that could programmatically port these types of C macros.  Of course, this a) doesn't exist yet, and b) wouldn't cover everything, but I think it's a better direction.  We definitely agree that hand-porting these macros is error-prone.

Generally speaking, any of the Postgres #define macros or static inline functions that aren't central to general pgrx operation -- things that would fine if unavailable when the `cshim` feature is disabled -- are fine candidates for the cshim.  Core support for the varlena type is not one of those things, however.

> for Greenplum support, there are much more works to do. not only the things mentioned in issue 1179.
I will complete my usage first.

We'd like to support Greenplum.  

It's important to recognize that pgrx is developed with an understanding of the official Postgres sources/internals.  Supporting deviations from that will require a lot of effort.

As I mentioned earlier, I'd love an opportunity to talk to someone at Greelplum/vmware about this.  In order for the pgrx core team to agree to support multiple, different products we're going to need some coordination first.  I don't want to encourage you or your team to start sending us Greenplum PRs without that coordination.

radarwave:
@eeeebbbbrrrr Thanks for your comments, it would be great to talk with you, and I believe we discussed Zombodb&Greenplum integration before. As a start, maybe we can discuss this in Greenplum open-source Slack channel with both the engineering and product teams. If needed, we can coordinate a time for the meeting.

Slack: greenplum.slack.com
Channel: #extensions

Sasasu:
> If you're talking about https://github.com/tcdi/pgrx/commit/e3464a1f8c4d5c4926aa5dbf717606079d552c5a, please make a separate PR for that
https://github.com/tcdi/pgrx/pull/1190

> Core support for the varlena type is not one of those things, however.
Ok, let us wait for https://github.com/rust-lang/rust-bindgen/pull/2369

>  I'd love an opportunity to talk to someone at Greelplum/vmware about this.
From my side, this is a 20% time project. there are some people who want to use rust in his project and find pgrx is incompatible with Greenplum. and I can write Rust, and I am helping with his project. I don't know the plan at the company level.

and pgrx is working with Greenplum on my fork. I think we need more tests and start internal usage first (not the company plan). I will close this PR one week after (or you can close this PR at any time).

workingjubilee:
> and pgrx is working with Greenplum on my fork

Is https://github.com/Sasasu/pgrx/commit/9f83dbb16fe49020b05cb735d832324da150a147 the fork you mean?

Sasasu:
> Is https://github.com/Sasasu/pgrx/commit/9f83dbb16fe49020b05cb735d832324da150a147 the fork you mean?

yes, this commit breaks Postgresql support. but add Greenplum support (with cfg = pg12).

eeeebbbbrrrr:
My [comment over there](https://github.com/pgcentralfoundation/pgrx/issues/1179#issuecomment-2088740201) applies here too.

PR #1185, opened 2023-07-03
Type roundtrip tests

eeeebbbbrrrr:
A new set of tests to simply check that we can "roundtrip" a Datum.  That is, we return the same value as we were given in a `#[pg_extern]` function argument.

Essentially this tests that the type's `FromDatum` and `IntoDatum` impls match, and in the case of arrays, tests that the iterator impl for `Array<T>` works for its individual datums.

In doing so, this uncovered a bug with converting a `pg_sys::Point` (and `pg_sys::BOX`) into a Datum, so those are fixed in here as well.

The addition of `#[derive(Copy)]` to the "Complex" type in pgrx-tests is just to make it work well with the roundtrip tests here.

### DIFF
--- a/pgrx-tests/src/tests/struct_type_tests.rs
+++ b/pgrx-tests/src/tests/struct_type_tests.rs
@@ -16,7 +16,7 @@ use pgrx::AllocatedByRust;
 
 use crate::get_named_capture;
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Copy)]
 #[repr(C)]
 pub struct Complex {
     x: f64,
--- a/pgrx-tests/src/tests/mod.rs
+++ b/pgrx-tests/src/tests/mod.rs
@@ -42,6 +42,7 @@ mod pgrx_module_qualification;
 mod postgres_type_tests;
 mod range_tests;
 mod result_tests;
+mod roundtrip_tests;
 mod schema_tests;
 mod shmem_tests;
 mod spi_tests;
--- a/pgrx-tests/src/tests/roundtrip_tests.rs
+++ b/pgrx-tests/src/tests/roundtrip_tests.rs
@@ -48,20 +48,21 @@ mod tests {
 
     #[allow(unused_imports)]
     use crate as pgrx_tests;
+    use crate::tests::struct_type_tests::Complex;
 
     use super::RandomData;
 
     macro_rules! roundtrip {
         ($fname:ident, $tname:ident, $rtype:ty, $expected:expr) => {
-            #[pg_extern]
+            #[pg_extern(requires = [ "create_complex_type" ])] // the "Complex" type comes from another crate, and we need its schema fully created before it can be used here
             fn $fname(i: $rtype) -> $rtype {
                 i
             }
 
             #[pg_test]
             fn $tname() -> Result<(), Box<dyn Error>> {
-                let value: $rtype = $expected.into();
-                let expected: $rtype = value.clone();
+                let value: $rtype = $expected;
+                let expected: $rtype = Clone::clone(&value);
                 let result: $rtype = Spi::get_one_with_args(
                     &format!("SELECT {}($1)", stringify!(tests.$fname)),
                     vec![(PgOid::from(<$rtype>::type_oid()), value.into_datum())],
@@ -132,6 +133,7 @@ mod tests {
     );
 
     roundtrip!(rt_random_data, test_rt_random_data, RandomData, RandomData::random());
+    roundtrip!(rt_complex, test_rt_complex, PgBox<Complex>, Complex::random());
 
     // -----------
     // arrays of the above
@@ -408,6 +410,20 @@ mod tests {
             Some(RandomData::random()),
             None,
             Some(RandomData::random()),
+            None
+        ]
+    );
+
+    roundtrip!(
+        rt_array_complex,
+        test_rt_array_complex,
+        Vec<Option<PgBox<Complex>>>,
+        vec![
+            None,
+            Some(Complex::random()),
+            Some(Complex::random()),
+            None,
+            Some(Complex::random()),
             None
         ]
     );
--- a/pgrx-tests/src/tests/roundtrip_tests.rs
+++ b/pgrx-tests/src/tests/roundtrip_tests.rs
@@ -1,14 +1,55 @@
+use rand::distributions::{Alphanumeric, Standard};
+use rand::Rng;
+
+use pgrx::Date;
+
+#[derive(pgrx::PostgresType, Clone, PartialEq, Eq, Debug, serde::Serialize, serde::Deserialize)]
+pub struct RandomData {
+    i: u64,
+    s: String,
+    a: Vec<Date>,
+}
+
+impl RandomData {
+    fn random() -> Self {
+        RandomData {
+            i: rand::random(),
+            s: rand::thread_rng()
+                .sample_iter(Alphanumeric)
+                .take(rand::thread_rng().gen_range(0..=1000))
+                .map(char::from)
+                .collect(),
+            a: rand::thread_rng()
+                .sample_iter(Standard)
+                .take(rand::thread_rng().gen_range(0..=1000))
+                .map(|_: u32| {
+                    Date::new(
+                        rand::thread_rng().gen_range(1..=3000),
+                        rand::thread_rng().gen_range(1..=12),
+                        rand::thread_rng().gen_range(1..=28),
+                    )
+                    .unwrap()
+                })
+                .collect(),
+        }
+    }
+}
+
 #[cfg(any(test, feature = "pg_test"))]
 #[pgrx::pg_schema]
 mod tests {
-    #[allow(unused_imports)]
-    use crate as pgrx_tests;
-    use pgrx::pg_sys::BuiltinOid;
-    use pgrx::prelude::*;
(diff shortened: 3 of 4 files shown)


thomcc (approved):
This is not only a great improvement, it's also will make slotting in some property testing (where we randomly generate the values which get round-tripped) very straightforward to add.

I'll admit I'm mostly rubber-stamping the entity-graph change tho, I'm not 100% sure what that change is trying to account for, and it looks like (especially after this) it has pretty good test coverage anyway.

workingjubilee:
I'll take a look at the array problem in a bit.

eeeebbbbrrrr:
> I'll admit I'm mostly rubber-stamping the entity-graph change tho, I'm not 100% sure what that change is trying to account for, and it looks like (especially after this) it has pretty good test coverage anyway.

Apparently the syntax tree that results from a `macro_rules!` expansion is a little different (maybe in relation to the #[pg_extern] procmacro too), at least in the "return value" location, and it was a case that we didn't account for, and one where what made sense was to recurse.  So I moved that gigantic match statement into its own function, and added the `syn::Type::Group` case to recurse into it.

I spent quite some time staring at `cargo expand` output and the macros aren't generating any code that's different than you'd write by hand, but something in how the resulting tree gets passed into pg_extern is different.

Thanks for coming to my TED talk.

pgrx-sql-entity-graph/src/pg_extern/returning.rs
```diff
@@ -53,297 +53,278 @@ impl Returning {
...
+            syn::Type::Paren(ref mut type_paren) => match &mut *type_paren.elem {
+                syn::Type::Macro(ref mut type_macro) => Self::parse_type_macro(type_macro),
+                other => {
+                    return Err(syn::Error::new(
+                        other.span(),
+                        &format!("Got unknown return type (type_paren): {type_paren:?}"),
+                    ))
                 }
+            },
+            syn::Type::Group(tg) => return Self::match_type(&tg.elem),
```
eeeebbbbrrrr, line 306: This is really the only new code here.  All the surrounding code is unchanged -- it was only pulled out into a standalone function, so there's one or two less indentations which is causing the code to reformat and now everything looks different.

eeeebbbbrrrr:
I've added the other type variations I wanted to.  One is a type derived by `#[derive(PostgresType)]` and the other is a fixed-size type wrapped in a `PgBox`.  

The PgBox case required implementing some convenience traits to make it easier to unit test, which is probably good anyways.

workingjubilee:
One of the asserts should go away if you rebase this on develop.

eeeebbbbrrrr:
I've rebased and (force) pushed this PR.  Lets see what CI does with it now.

workingjubilee:
Can we split this into three PRs?

1. land the type-roundtrip tests for everything but Point and BOX
2. land the test reporting improvements
3. a larger-scale investigation into type-roundtrip for Point and BOX

The goal is to verify that 1 and 2 _aren't_ the problem.

pgrx/src/pgbox.rs
```diff
@@ -368,15 +369,64 @@ impl<T, AllocatedBy: WhoAllocated> PgBox<T, AllocatedBy> {
...
+where
+    T: Clone,
+{
+    /// Copies the wrapped `T` into [`PgMemoryContexts::CurrentMemoryContext`].
+    fn clone(&self) -> Self {
+        unsafe {
+            let copy = PgMemoryContexts::CurrentMemoryContext
+                .copy_ptr_into(self.as_ptr(), std::mem::size_of::<T>());
+
+            PgBox::from_pg(copy)
```
workingjubilee, line 382: VERY uncertain about this.

pgrx/src/pgbox.rs
```diff
@@ -368,15 +369,64 @@ impl<T, AllocatedBy: WhoAllocated> PgBox<T, AllocatedBy> {
...
+}
+
+impl<T, AllocatedBy: WhoAllocated> AsRef<T> for PgBox<T, AllocatedBy> {
+    fn as_ref(&self) -> &T {
+        match self.ptr.as_ref() {
+            Some(ptr) => unsafe { ptr.as_ref() },
+            None => panic!("Attempt to dereference null pointer during `AsRef::as_ref()` of PgBox"),
+        }
+    }
+}
```
workingjubilee, line 422: feeling nervous about the soundness of this one, even with the null check, since it's easy to create a PgBox from a potentially-dubious pointer.

eeeebbbbrrrr:
> Can we split this into three PRs?

Maybe that's a good idea.  I'll come back to this on Monday with fresh eyes.

pgrx-tests/src/tests/struct_type_tests.rs
```diff
@@ -16,13 +16,31 @@ use pgrx::AllocatedByRust;
...
+impl Complex {
+    pub fn random() -> PgBox<Complex> {
+        unsafe {
+            let mut c = PgBox::<Complex>::alloc0();
+            c.x = rand::random();
+            c.y = rand::random();
+            c.into_pg_boxed()
+        }
+    }
+}
```
workingjubilee, line 42: Mmm I don't know if I like this.

pgrx/src/pgbox.rs
workingjubilee: To elaborate: pgrx has made wrong assumptions in the past that T is all that is interesting about T, i.e. that the type is "flat" and not e.g. managing memory held elsewhere. This does something that could extend the lifespan of T but fail to extend the life of the memory that T is managing.

pgrx/src/pgbox.rs
eeeebbbbrrrr: I can not impl `AsRef` and instead add a `unsafe fn as_ref(self) -> &T` function to `PgBox` that does this?  Maybe have it return `Option<&T>`?

I did this only to make that roundtrip test macro a little easier to write for PgBox'd things.

pgrx-tests/src/tests/struct_type_tests.rs
eeeebbbbrrrr: It's just generating some random test data.  If you got a better idea, I'm all ears.  Not sure how much any of has has to "like" it?

pgrx/src/pgbox.rs
eeeebbbbrrrr: I can get rid of this entirely.  Like the AsRef impl, this was just to make the roundtrip tests easier to write.

Maybe what we can do is `impl Clone for PgBox<pg_sys::Point>` and do that the "RightWay" rather than a blanket impl that, as you point out, probably ain't gonna work for every situation.

Thoughts?

workingjubilee: This needs to not only do this shallow copy of the memory that PgBox handles, it needs to do a "deep clone", so that if T is, say, `Vec<Box<[i32]>>`, it clones all the boxed slices too.

workingjubilee: Sure! This actually can work fine if it restricts to T being a Copy type.

pgrx-tests/src/tests/struct_type_tests.rs
workingjubilee: `Box<Complex>` should be fine if it's just for testing? We can just leak allocations, it's not a big deal.

pgrx/src/pgbox.rs
workingjubilee: I had mostly decided there was probably something cursed about a new impl to set this weird cascade of failures off, and started looking for something dubious, but I think this may be less of a problem than I thought.

pgrx-tests/src/tests/struct_type_tests.rs
eeeebbbbrrrr: `Box<Complex>` isn't a thing that we can FromDatum/IntoDatum.  That's why this does `PgBox<Complex>`.  I thought your complaint was about the `rand` usage!

pgrx/src/pgbox.rs
eeeebbbbrrrr: > Sure! This actually can work fine as-is if it restricts to T being a Copy type.

Yeah, I think maybe that was my intention anyways -- but obviously missed the mark.

pgrx-tests/src/tests/struct_type_tests.rs
workingjubilee: Oh, you're right.

pgrx/src/pgbox.rs
workingjubilee: I've decided this is no more dubious than PgBox already was.

eeeebbbbrrrr:
Okay, I've got this PR dialed into just the "roundtrip" macro test stuff now.

The other changes were split out into PRs #1203 and #1204 and merged into `develop`

pgrx-tests/src/tests/struct_type_tests.rs
```diff
@@ -16,7 +16,7 @@ use pgrx::AllocatedByRust;
 
 use crate::get_named_capture;
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Copy)]
```
eeeebbbbrrrr, line 19: Makes it easier to work with in the new roundtrip macro tests added in this PR.

workingjubilee (approved):
This seems to have worked out the problems!

PR #1186, opened 2023-07-04
Try to smartly propagate fs errors

### LINKED ISSUE #1183: Avoid using `Path::exists`
Its use should probably be considered more discouraged than it already is, as it is prone to [reporting non-existence in cases of permission errors](https://doc.rust-lang.org/std/path/struct.Path.html#method.exists) which can make already-confusing build errors even harder to debug. This will improve diagnostics during builds.

Found while glancing at https://github.com/timescale/timescaledb-toolkit/pull/762

workingjubilee:
Path::exists coerces filesystem errors to "false".
When used in restricted, networked, or other
"interesting" filesystems, like many build environs,
this can lose important error data instead of passing it on.
Try to improve error messaging via propagating errors.

This fixes #1183.

### DIFF
--- a/cargo-pgrx/src/command/package.rs
+++ b/cargo-pgrx/src/command/package.rs
@@ -110,7 +110,9 @@ pub(crate) fn package_extension(
     is_test: bool,
     features: &clap_cargo::Features,
 ) -> eyre::Result<()> {
-    if !out_dir.exists() {
+    let out_dir_exists =
+        out_dir.try_exists().wrap_err("failed to access out-dir while packaging extension")?;
+    if !out_dir_exists {
         std::fs::create_dir_all(&out_dir)?;
     }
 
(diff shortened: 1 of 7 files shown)


cargo-pgrx/src/command/package.rs
```diff
@@ -110,7 +110,9 @@ pub(crate) fn package_extension(
     is_test: bool,
     features: &clap_cargo::Features,
 ) -> eyre::Result<()> {
-    if !out_dir.exists() {
+    let out_dir_exists =
+        out_dir.try_exists().wrap_err("failed to access out-dir while packaging extension")?;
```
eeeebbbbrrrr, line 114: can this error contain the value of `out_dir`?

eeeebbbbrrrr (approved):
LGTM with one little ask.

cargo-pgrx/src/command/package.rs
workingjubilee: absolutely!

PR #1189, opened 2023-07-05
Include security labels header

### LINKED ISSUE #1187: Add support for Security Labels ?
Hi !

Thanks for this framework !  it is really helpful. 

Do you have plans to support Security Labels ? Although it is a very rare feature, it can be really powerful way to describe an ad-hoc security policy for an extension

At least functions like `register_label_provider` and `GetSecurityLabel` would quite usefull for me.

https://doxygen.postgresql.org/seclabel_8c.html
https://www.postgresql.org/docs/current/sql-security-label.html

Maybe it's already available but I couldn't find how in the docs...

daamien:
Closes #1187

### DIFF
--- a/pgrx-pg-sys/include/pg11.h
+++ b/pgrx-pg-sys/include/pg11.h
@@ -48,6 +48,7 @@ Use of this source code is governed by the MIT license that can be found in the 
 #include "commands/event_trigger.h"
 #include "commands/explain.h"
 #include "commands/proclang.h"
+#include "commands/seclabel.h"
 #include "commands/tablespace.h"
 #include "commands/tablecmds.h"
 #include "commands/trigger.h"
--- a/pgrx-pg-sys/include/pg12.h
+++ b/pgrx-pg-sys/include/pg12.h
@@ -50,6 +50,7 @@ Use of this source code is governed by the MIT license that can be found in the 
 #include "commands/event_trigger.h"
 #include "commands/explain.h"
 #include "commands/proclang.h"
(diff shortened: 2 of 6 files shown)


workingjubilee:
Out of curiosity, what sort of uses do you intend to put this to? All the functions in these headers will remain `unsafe` to use.

eeeebbbbrrrr:
It's missing a similar change to `pg16.h`.  I wonder if this was originally created from `master` even tho the target here is `develop`?

daamien:
Hi !

@workingjubilee I'm working on an extension called PostgreSQL Anonymizer where I need to register several Security Label Providers which allow users to define a making policy for a given column like this:

```sql
SECURITY LABEL FOR anon ON COLUMN player.name
IS 'MASKED WITH FUNCTION anon.fake_last_name()';
```

The security label provider is created during by the _PG_init function

https://gitlab.com/dalibo/postgresql_anonymizer/-/blob/master/anon.c#L420

Then the masking rules are read using `GetSecurityLabel`.

As we're considering moving toward Rust, I'm trying to build a prototype with `pgrx`

@eeeebbbbrrrr I'm not sure what happened, but I just rebased my PR from tcdi:develop.

eeeebbbbrrrr (approved):

eeeebbbbrrrr:
Thanks @daamien.  Welcome to the contributors club!  This will be in our next release, which I hope will be very soon.

daamien:
thanks a lot !

PR #1191, opened 2023-07-05
Array-walking is aligned

workingjubilee:
The alignment-handling code has been correct for some time.

Remove the questioning assert, extract the align-up pattern into a function on `layout::Align`, and precompute the fixed-size offset to use for fixed-size elements in arrays.

### DIFF
--- a/pgrx/src/datum/array.rs
+++ b/pgrx/src/datum/array.rs
@@ -154,11 +154,11 @@ impl<'a, T: FromDatum> Array<'a, T> {
                 // determined at runtime based on the length of the string
                 Size::CStr => Box::new(casper::PassByCStr),
 
-                // Array elements are fixed sizes yet Postgres wants us to pass around by reference
-                Size::Fixed(size) => Box::new(casper::PassByFixed {
-                    align: elem_layout.align.as_usize(),
-                    size: size as usize,
-                }),
+                // Array elements are fixed sizes yet the data is "pass-by-reference"
+                // Most commonly, this is because of elements larger than a Datum.
+                Size::Fixed(size) => {
+                    Box::new(casper::PassByFixed::from_size_and_align(size, elem_layout.align))
+                }
             },
         };
 
@@ -384,6 +384,7 @@ fn as_slice<'a, T: Sized + FromDatum>(array: &'a Array<'_, T>) -> Result<&'a [T]
 }
 
 mod casper {
+    use crate::layout::Align;
     use crate::{pg_sys, varlena, Array, FromDatum};
 
     // it's a pop-culture reference (https://en.wikipedia.org/wiki/Cha_Cha_Slide) not some fancy crypto thing you nerd
@@ -495,9 +496,19 @@ mod casper {
     }
 
     pub(super) struct PassByFixed {
-        pub(super) align: usize,
-        pub(super) size: usize,
+        pub(super) padded_size: usize,
     }
+
+    impl PassByFixed {
+        pub(super) fn from_size_and_align(size: u16, align: Align) -> Self {
+            let align = align.as_usize();
+            let size = size as usize;
+            let align_mask = size & (align - 1);
+            let align_offset = if align_mask != 0 { align - align_mask } else { 0 };
+            PassByFixed { padded_size: size + align_offset }
+        }
+    }
+
     impl<T: FromDatum> ChaChaSlide<T> for PassByFixed {
         #[inline]
         unsafe fn bring_it_back_now(&self, array: &Array<T>, ptr: *const u8) -> Option<T> {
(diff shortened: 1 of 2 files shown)


workingjubilee:
This addresses the assert failure in https://github.com/tcdi/pgrx/pull/1185

eeeebbbbrrrr (approved):

eeeebbbbrrrr:
This looks fine.  I guess the problem here is that we were continuing to assert something that's actually not true, or otherwise properly handled by the follow-up changes after zero-copy-arrays first landed?

workingjubilee:
Correct. I am not even sure if it was still unhandled by the end of the first PR, but by the time of the `Box<dyn ChaChaSlide>` refactor, it was definitely not.

workingjubilee:
I need to add an assert elsewhere.

workingjubilee:
Added appropriate debug asserts for "YES this pointer is inbounds of the array!"

workingjubilee:
Oh lol, this debug assert is wrong.

pgrx/src/datum/array.rs
```diff
@@ -628,6 +623,11 @@ impl<'a, T: FromDatum> Iterator for ArrayIterator<'a, T> {
         let Some(is_null) = array.null_slice.get(*curr) else { return None };
         *curr += 1;
 
+        // Make sure a previous iter didn't push out of bounds!
+        #[cfg(debug_assertions)]
+        if !is_null {
+            assert!(*ptr < array.raw.end_ptr())
+        };
```
workingjubilee, line 630: This would have caught https://github.com/tcdi/pgrx/pull/1180

pgrx/src/datum/array.rs
```diff
@@ -495,9 +487,19 @@ mod casper {
     }
 
     pub(super) struct PassByFixed {
-        pub(super) align: usize,
-        pub(super) size: usize,
+        pub(super) padded_size: usize,
+    }
+
+    impl PassByFixed {
+        pub(super) fn from_size_and_align(size: u16, align: Align) -> Self {
```
thomcc, line 494: Chatted on discord about this impl, had a few tweaks. Also please `#[inline]` this function so iterators in external crates can see how the size was computed (admittedly after devirtualizing).

pgrx/src/datum/array.rs
```diff
@@ -628,6 +623,11 @@ impl<'a, T: FromDatum> Iterator for ArrayIterator<'a, T> {
         let Some(is_null) = array.null_slice.get(*curr) else { return None };
         *curr += 1;
 
+        // Make sure a previous iter didn't push out of bounds!
+        #[cfg(debug_assertions)]
+        if !is_null {
```
thomcc, line 628: Nit: I'd prefer `if cfg!(debug_assertions) && !is_null` instead of the `#[cfg(...)]`, to avoid annoyances from conditional compilation (future changes missing a rename or having unused var warnings -- altho they'd be caught quickly).

Or even `if !is_null { debug_assert!(...); }`

workingjubilee: clearly it should be
```suggestion
        // If we might read, make sure a previous iter didn't push out of bounds!
        debug_assert!(is_null || *ptr < array.raw.end_ptr());
```

pgrx/src/datum/array.rs
```diff
@@ -628,6 +606,10 @@ impl<'a, T: FromDatum> Iterator for ArrayIterator<'a, T> {
         let Some(is_null) = array.null_slice.get(*curr) else { return None };
         *curr += 1;
 
+        // Make sure a previous iter didn't push out of bounds!
+        if !is_null {
+            debug_assert!(*ptr < array.raw.end_ptr());
```
thomcc, line 611: Note that the comparison is wrong if `ptr` wrapped around the address space and other such edge cases. Might not be worth giving a shit about, since handling that correctly is annoying.

Actually, I have some code in a non-work project that cares about that, so I tweaked it for pgrx (hopefully not breaking it in the process) which gives <https://gist.github.com/thomcc/b2730e3b1a56c4d8cdebf29814ea2351>. It can probably be simplified for our use though (we perhaps don't need the `size` param), and has not been tested after I tweaked it for pgrx.

pgrx/src/datum/array.rs
```diff
@@ -450,7 +440,7 @@ mod casper {
 
     /// Array elements are [`pg_sys::varlena`] types, which are pass-by-reference
     pub(super) struct PassByVarlena {
-        pub(super) align: usize,
+        pub(super) align: Align,
```
workingjubilee, line 443: Carrying a full `usize` is unnecessary here and in any case makes the result that much slightly-less typesafe.

pgrx/src/datum/array.rs
thomcc: If you go that route, we have a couple other places where we have similar debug assertions comparing against `end_ptr`, might be worth fixing them too.

pgrx/src/datum/array.rs
```diff
@@ -585,6 +561,8 @@ impl<'a, T: FromDatum> Iterator for ArrayTypedIterator<'a, T> {
         if *curr >= array.raw.len() {
             None
         } else {
+            // Make sure a previous iter didn't push out of bounds!
+            debug_assert!(*ptr < array.raw.end_ptr());
```
thomcc, line 565: This might be the "couple of other places" I am thinking of. I knew I had seen it >_>

pgrx/src/datum/array.rs
workingjubilee: Can we land that in a separate PR? I probably want it but we should discuss it as its own beast because that raises questions about more of our codebase and this patch has already scope-creeped a bit.

thomcc: Sure.

thomcc (approved):

PR #1196, opened 2023-07-06
Fixes macos-11 tests

BradyBonnette:
For some reason, `brew` suddenly stopped working on macos-11 GHA instances. The only place we were using brew was to install `sccache`, so these changes remove the use of brew to install sccache and instead install sccache manually. These changes also run all tests on macos-11 instead of selective tests.

NOTE: This PR is targeting `tcdi:master` since they're just CI changes.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -450,22 +450,28 @@ jobs:
     - name: Install cargo-pgrx
       run: cargo install --path cargo-pgrx/ --debug --force
 
-    # - name: Run 'cargo pgrx init'
-    #   run: |
-    #     set -x
-    #     cargo pgrx init --pg$PG_VER $(which pg_config)
+    - name: Print sccache stats
+      run: sccache --show-stats
 
-    # - name: Run base-level tests
-    #   run: |
-    #     set -x
-    #     cargo test \
-    #       --features "pg$PG_VER" --no-default-features \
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -397,9 +400,11 @@ jobs:
         echo ""
 
         echo "----- Install sccache -----"
-        brew update
-        brew install sccache
-        mkdir -p /Users/runner/Library/Caches/Mozilla.sccache
+        curl -L https://github.com/mozilla/sccache/releases/download/v0.5.4/sccache-v0.5.4-x86_64-apple-darwin.tar.gz | tar xz
+        mv -f sccache-v0.5.4-x86_64-apple-darwin/sccache /usr/local/bin
```
thomcc, line 404: I would like it if we could avoid this becoming stale. I don't see an easy fix for this though, so @BradyBonnette, do you mind subscribing to the github notifications for new sccache releases and keeping this up to date?

(I would volunteer but rust project stuff overwhelms my github notification box, and I'd definitely fail to notice a release)

BradyBonnette: There might be something we could do internally. Ill take this conversation into Discord because there are considerations here.

BradyBonnette:
This needs to go into `tcdi:develop`, and for some reason I cannot change the base branch. So I have to create a new PR.

PR #1197, opened 2023-07-07
Fixes macos-11 tests

BradyBonnette:
For some reason, `brew` suddenly stopped working on macos-11 GHA instances. The only place we were using brew was to install `sccache`, so these changes remove the use of brew to install sccache and instead install sccache manually. These changes also run all tests on macos-11 instead of selective tests.


BradyBonnette:
See https://github.com/tcdi/pgrx/pull/1196 for any previous conversations

PR #1198, opened 2023-07-08
Pgcentralfoundation updates

eeeebbbbrrrr:
Updates around copyright and license headers now that pgrx is owned by PgCentralFoundation, Inc.

I've made an executive decision here to prefix the copyright headers with `//LICENSE` or `#LICENSE` or `--LICENSE` depending on the file type.

The reason for this that tooling around this sort of thing (updating copyright headers) is absolutely terrible.  I'm not knocking the tool we kinda use (https://github.com/chasinglogic/licensure) but it has no ability to update existing headers and it seems to have some bugs regarding whitespace and line wrapping.  So... having something we can programmatically yank out of the sources every year to update the damn year will make this a little easier.  

I should give `licensure` some PRs.

There's no code changes here.


ChronicallyJD (approved):
I didn't review every file of course but through the dozen, it looks good!

PR #1203, opened 2023-07-10
Testing help

eeeebbbbrrrr:
Just some cleanups for testing, mostly around `framework.rs`

### DIFF
--- a/pgrx-sql-entity-graph/src/pg_extern/returning.rs
+++ b/pgrx-sql-entity-graph/src/pg_extern/returning.rs
@@ -22,7 +22,7 @@ use std::convert::TryFrom;
 use syn::parse::{Parse, ParseStream};
 use syn::punctuated::Punctuated;
 use syn::spanned::Spanned;
-use syn::{GenericArgument, PathArguments, Token, Type};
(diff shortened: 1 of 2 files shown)


pgrx-sql-entity-graph/src/pg_extern/returning.rs
```diff
@@ -54,297 +54,278 @@ impl Returning {
...
+            syn::Type::Paren(ref mut type_paren) => match &mut *type_paren.elem {
+                syn::Type::Macro(ref mut type_macro) => Self::parse_type_macro(type_macro),
+                other => {
+                    return Err(syn::Error::new(
+                        other.span(),
+                        &format!("Got unknown return type (type_paren): {type_paren:?}"),
+                    ))
                 }
+            },
+            syn::Type::Group(tg) => return Self::match_type(&tg.elem),
```
eeeebbbbrrrr, line 307: This is the actual code change here -- supporting this variant.

All the surrounding code was moved into a new function called `match_type()` and we recurse here.  So the big diff is really just whitespace due to less indentation

workingjubilee (commented):
This looks good, formatting noise aside.

PR #1206, opened 2023-07-12
 Add a workaround for the pg16/homebrew/icu4c situation 

thomcc:
Seems to work, testing a little more.

### DIFF
--- a/cargo-pgrx/src/command/init.rs
+++ b/cargo-pgrx/src/command/init.rs
@@ -21,7 +21,7 @@ use std::collections::HashMap;
 use std::fs::File;
 use std::io::{Read, Write};
 use std::path::PathBuf;
-use std::process::Stdio;
+use std::process::{Command, Stdio};
 
 use std::sync::{Arc, Mutex};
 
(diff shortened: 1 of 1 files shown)


cargo-pgrx/src/command/init.rs
```diff
@@ -275,6 +275,77 @@ fn untar(bytes: &[u8], pgrxdir: &PathBuf, pg_config: &PgConfig) -> eyre::Result<
...
+        let msg = "\
+            Homebrew seems to be in use, but the `icu4c` does not appear to be \
+            installed. Without this, the build of PostgreSQL 16 may fail. You \
+            may be able to fix this by installing it using:\n\
+            \n\
+            $ brew install icu4c\n\
+            \n\
+            before retrying. Alternatively, if you are certain you do not need it \
+            you may disable ICU support during the build by invoking `cargo pgrx init` \
+            with `--configure-flags=--without-icu`, as follows:\n\
```
eeeebbbbrrrr, line 325: ... but the `icu4c` package is not installed.

cargo-pgrx/src/command/init.rs
```diff
@@ -298,7 +369,13 @@ fn configure_postgres(pg_config: &PgConfig, pgdir: &PathBuf, init: &Init) -> eyr
     for var in PROCESS_ENV_DENYLIST {
         command.env_remove(var);
     }
-
+    // Work around the fact that pg16 requires icu to be installed and available
+    // via pkg-config, but doesn't add it to the pkg-config path. (Does a decent
+    // number of checks to try to avoid causing trouble when things were
+    // actually fine)
+    if cfg!(target_os = "macos") && pg_config.major_version().unwrap_or(0) >= 16 {
```
eeeebbbbrrrr, line 376: can we make this `== 16`?  Just thinking that assuming what 17 will do might hide a problem from future us.  I'd probably rather 17 fail too when we first add support for it.

eeeebbbbrrrr (changes requested):

eeeebbbbrrrr:
Not in this PR, but I suppose we need a CI job that uses `cargo-pgrx` to install PG16 on MacOS.  

@BradyBonnette, would that be difficult?

cargo-pgrx/src/command/init.rs
thomcc: I think it just controls collations, if their extension doesn't care about that ISTM like it's reasonable to disable ICU.

cargo-pgrx/src/command/init.rs
thomcc: Done.

eeeebbbbrrrr (approved):

eeeebbbbrrrr:
LGTM.  Merging.

Issue #1209, opened 2023-07-13
SpiTupleTable is unsound


david-monroe:
While investigating memory reclamation of SpiTupleTable, I realized that its lifetime is not bound by the lifetime of the surrounding SPI connection. Therefore, SPI_finish frees the memory referenced by the SpiTupleTable without restricting the lifetime of the SpiTupleTable.

```sql
CREATE OR REPLACE FUNCTION three()
    RETURNS TEXT LANGUAGE plrust
AS
$$
    // create the cursor we actually care about
    let mut res = Spi::connect(|c| c.open_cursor("select 'hello' from generate_series(1, 10000)", None).fetch(10000).unwrap());

    // here we just perform some allocations to make sure that the previous cursor gets invalidated
    for i in 0..1000 {
        Spi::connect(|c| c.open_cursor("select 1", None).fetch(1).unwrap());
    }

    // later elements are probably more likely to point to deallocated memory
    for i in 0..1000 {
        res.next();
    }

    // segfault
    Ok(res.next().unwrap().get::<String>(1)?)
$$;

select three();
```

```
2023-07-13 17:37:11.654 UTC [51] LOG:  server process (PID 58) was terminated by signal 11: Segmentation fault
2023-07-13 17:37:11.654 UTC [51] DETAIL:  Failed process was running: select three()
2023-07-13 17:37:11.654 UTC [51] LOG:  terminating any other active server processes
2023-07-13 17:37:11.655 UTC [51] LOG:  all server processes terminated; reinitializing
```

---

While this example uses cursors, the same is possibly true for plain queries as well.

eeeebbbbrrrr:
Thanks.  We'll look into this immediately.

eeeebbbbrrrr:
https://github.com/pgcentralfoundation/pgrx/pull/1210 fixes this such that the above code won't compile anymore, which I think is the correct answer here.

My pgrx-based interpretation of what this code should like like is:

```rust
#[pg_extern]
fn issue1209() -> Result<Option<String>, Box<dyn std::error::Error>> {
    let res = Spi::connect(|c| {
        let mut cursor = c.open_cursor("SELECT 'hello' FROM generate_series(1, 10000)", None);
        let table = cursor.fetch(10000)?;
        table.into_iter().map(|row| row.get(1)).collect::<Result<Vec<_>, _>>()
    })?;

    Ok(res.first().cloned().flatten())
}
```

The function body ought to be paste-able into a plrust function.

[...]

PR #1210, opened 2023-07-13
Fix issue #1209

eeeebbbbrrrr:
Cleanup lifetimes such that a `Spi::connect()` block can't return things that should be borrowed from the lifetime of that SPI connection, which includes cursors (`SpiCursor`).

This causes the test from #1209 to not compile, which is the desired result, with:

```
error[E0515]: cannot return value referencing temporary value
   --> pgrx-tests/src/tests/spi_tests.rs:516:13
    |
516 |               c.open_cursor("select 'hello' from generate_series(1, 10000)", None)
    |               ^-------------------------------------------------------------------
    |               |
    |  _____________temporary value created here
    | |
517 | |                 .fetch(10000)
518 | |                 .unwrap()
    | |_________________________^ returns a value referencing data owned by the current function
    |
    = help: use `.collect()` to allocate the iterator

error[E0515]: cannot return value referencing function parameter `c`
   --> pgrx-tests/src/tests/spi_tests.rs:516:13
    |
516 |               c.open_cursor("select 'hello' from generate_series(1, 10000)", None)
    |               ^-------------------------------------------------------------------
    |               |
    |  _____________`c` is borrowed here
    | |
517 | |                 .fetch(10000)
518 | |                 .unwrap()
    | |_________________________^ returns a value referencing data owned by the current function
    |
    = help: use `.collect()` to allocate the iterator

error[E0515]: cannot return value referencing temporary value
   --> pgrx-tests/src/tests/spi_tests.rs:523:30
    |
523 |             Spi::connect(|c| c.open_cursor("select 1", None).fetch(1).unwrap());
    |                              -------------------------------^^^^^^^^^^^^^^^^^^
    |                              |
    |                              returns a value referencing data owned by the current function
    |                              temporary value created here
    |
    = help: use `.collect()` to allocate the iterator

error[E0515]: cannot return value referencing function parameter `c`
   --> pgrx-tests/src/tests/spi_tests.rs:523:30
    |
523 |             Spi::connect(|c| c.open_cursor("select 1", None).fetch(1).unwrap());
    |                              -------------------------------^^^^^^^^^^^^^^^^^^
    |                              |
    |                              returns a value referencing data owned by the current function
    |                              `c` is borrowed here
    |
    = help: use `.collect()` to allocate the iterator
```

I am not sure how to write a test with pgrx' test suite to ensure something **doesn't** compile.  That'll be easy to do over in PL/Rust, but I feel like we should have some kind of regression test around this here.  Thoughts @thomcc or @workingjubilee?

### DIFF
--- a/pgrx-tests/src/tests/spi_tests.rs
+++ b/pgrx-tests/src/tests/spi_tests.rs
@@ -179,7 +179,7 @@ mod tests {
     }
 
     #[pg_test]
-    fn test_inserting_null() -> Result<(), Box<dyn Error>> {
+    fn test_inserting_null() -> Result<(), pgrx::spi::Error> {
         Spi::connect(|mut client| {
             client.update("CREATE TABLE tests.null_test (id uuid)", None, None).map(|_| ())
         })?;
--- a/pgrx-tests/src/tests/spi_tests.rs
+++ b/pgrx-tests/src/tests/spi_tests.rs
@@ -509,6 +509,19 @@ mod tests {
         assert_eq!("'quoted-string'", spi::quote_literal(String::from("quoted-string")));
     }
 
+    #[pg_test]
+    fn can_return_borrowed_str() -> Result<(), Box<dyn std::error::Error>> {
+        let res = Spi::connect(|c| {
+            let mut cursor = c.open_cursor("SELECT 'hello' FROM generate_series(1, 10000)", None);
+            let table = cursor.fetch(10000)?;
+            table.into_iter().map(|row| row.get::<&str>(1)).collect::<Result<Vec<_>, _>>()
+        })?;
+
+        let value = res.first().cloned().flatten().map(|s| s.to_string());
+        assert_eq!(Some("hello".to_string()), value);
+        Ok(())
+    }
+
     // TODO:  The point of this test is to **not** compile.  How to write a test for that?
     // #[pg_test]
     // fn issue1209() -> Result<Option<String>, Box<dyn std::error::Error>> {
--- a/pgrx-tests/src/tests/spi_tests.rs
+++ b/pgrx-tests/src/tests/spi_tests.rs
@@ -13,13 +13,14 @@ mod tests {
     #[allow(unused_imports)]
     use crate as pgrx_tests;
     use pgrx::IntoDatum;
+    use std::error::Error;
 
     use pgrx::prelude::*;
     use pgrx::spi;
 
     #[pg_test(error = "syntax error at or near \"THIS\"")]
     fn test_spi_failure() -> Result<(), spi::Error> {
-        Spi::connect(|client| client.select("THIS IS NOT A VALID QUERY", None, None)).map(|_| ())
+        Spi::connect(|client| client.select("THIS IS NOT A VALID QUERY", None, None).map(|_| ()))
     }
 
     #[pg_test]
@@ -178,9 +179,9 @@ mod tests {
     }
 
     #[pg_test]
-    fn test_inserting_null() -> Result<(), pgrx::spi::Error> {
+    fn test_inserting_null() -> Result<(), Box<dyn Error>> {
         Spi::connect(|mut client| {
-            client.update("CREATE TABLE tests.null_test (id uuid)", None, None)
+            client.update("CREATE TABLE tests.null_test (id uuid)", None, None).map(|_| ())
         })?;
         assert_eq!(
             Spi::get_one_with_args::<i32>(
@@ -369,7 +370,7 @@ mod tests {
         let err = Spi::connect(|client| {
             let prepared =
                 client.prepare("SELECT $1", Some(vec![PgOid::BuiltIn(PgBuiltInOids::INT4OID)]))?;
-            client.select(&prepared, None, None)
+            client.select(&prepared, None, None).map(|_| ())
         })
         .unwrap_err();
 
@@ -404,7 +405,7 @@ mod tests {
     #[pg_test(error = "CREATE TABLE is not allowed in a non-volatile function")]
     fn test_readwrite_in_readonly() -> Result<(), spi::Error> {
         // This is supposed to run in read-only
-        Spi::connect(|client| client.select("CREATE TABLE a ()", None, None)).map(|_| ())
+        Spi::connect(|client| client.select("CREATE TABLE a ()", None, None).map(|_| ()))
     }
 
     #[pg_test]
@@ -510,7 +511,7 @@ mod tests {
     }
 
     #[pg_test]
-    fn can_return_borrowed_str() -> Result<(), Box<dyn std::error::Error>> {
+    fn can_return_borrowed_str() -> Result<(), Box<dyn Error>> {
         let res = Spi::connect(|c| {
             let mut cursor = c.open_cursor("SELECT 'hello' FROM generate_series(1, 10000)", None);
             let table = cursor.fetch(10000)?;
@@ -524,7 +525,7 @@ mod tests {
 
     // TODO:  The point of this test is to **not** compile.  How to write a test for that?
     // #[pg_test]
-    // fn issue1209() -> Result<Option<String>, Box<dyn std::error::Error>> {
+    // fn issue1209() -> Result<Option<String>, Box<dyn Error>> {
     //     // create the cursor we actually care about
     //     let mut res = Spi::connect(|c| {
     //         c.open_cursor("select 'hello' from generate_series(1, 10000)", None)
@@ -544,5 +545,12 @@ mod tests {
     //
     //     // segfault
     //     Ok(res.next().unwrap().get::<String>(1)?)
+    // }
+    //
+    // #[pg_extern]
+    // fn issue1209_prepared_stmt(q: &str) -> Result<Option<String>, Box<dyn std::error::Error>> {
+    //     let prepared = { Spi::connect(|c| c.prepare(q, None))? };
+    //
+    //     Ok(Spi::connect(|c| (&prepared).execute(&c, Some(1), None)?.first().get(1))?)
     // }
 }
--- a/pgrx-tests/src/tests/spi_tests.rs
+++ b/pgrx-tests/src/tests/spi_tests.rs
@@ -508,4 +508,28 @@ mod tests {
         assert_eq!("'quoted-with-''quotes'''", spi::quote_literal("quoted-with-'quotes'"));
         assert_eq!("'quoted-string'", spi::quote_literal(String::from("quoted-string")));
     }
+
+    // TODO:  The point of this test is to **not** compile.  How to write a test for that?
+    // #[pg_test]
+    // fn issue1209() -> Result<Option<String>, Box<dyn std::error::Error>> {
+    //     // create the cursor we actually care about
+    //     let mut res = Spi::connect(|c| {
+    //         c.open_cursor("select 'hello' from generate_series(1, 10000)", None)
+    //             .fetch(10000)
+    //             .unwrap()
+    //     });
+    //
+    //     // here we just perform some allocations to make sure that the previous cursor gets invalidated
+    //     for _ in 0..1000 {
+    //         Spi::connect(|c| c.open_cursor("select 1", None).fetch(1).unwrap());
+    //     }
+    //
+    //     // later elements are probably more likely to point to deallocated memory
+    //     for _ in 0..1000 {
+    //         res.next();
+    //     }
+    //
+    //     // segfault
+    //     Ok(res.next().unwrap().get::<String>(1)?)
+    // }
 }
--- a/pgrx-tests/src/tests/spi_tests.rs
+++ b/pgrx-tests/src/tests/spi_tests.rs
@@ -522,37 +522,4 @@ mod tests {
         assert_eq!(Some("hello".to_string()), value);
         Ok(())
     }
-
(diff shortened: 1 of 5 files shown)


workingjubilee:
TIME FOR COMPILETEST

workingjubilee:
@eeeebbbbrrrr The library of choice to test what you want to test is https://docs.rs/ui_test/latest/ui_test/ and it enables what are called "UI tests" by rustc's testing infrastructure. I said "compiletest" because it is a revised version of https://github.com/rust-lang/rust/tree/1b3e68692592d71938df8e7fd8e53fbe5e7ef58c/src/tools/compiletest

david-monroe:
Unfortunately I don't think it is that simple.

1. SpiHeapTupleData is also just a pointer to allocated data that is freed by SPI_finish.

2. There is `impl<'a> FromDatum for &'a str` which I think allows you to get a `&'static str` pointing into temporary data via

    ```rust
        pub fn get<T: IntoDatum + FromDatum>(&self, ordinal: usize) -> Result<Option<T>> {
    ```

You should do a thorough analysis of lifetimes in this crate.

workingjubilee:
We have indeed been doing that. I have been working on various possible solutions for actually guaranteeing type safety over FFI. Unfortunately it's very hard to get someone who both:
- is very familiar with the "objective lifetimes" of Postgres
- is very familiar with how Rust's lifetime system works

And it's hard to simply fuse two people together, so while Eric knows the former and I know the latter, we can't necessarily do a Vulcan mind meld and even if we did, there's a lot of code to review and fix.

As you might have noticed, approximately **everything** implements FromDatum and it is a bound for **everything** thus it is used **everywhere** so "just adding a lifetime parameter" probably is close to "rewrite the entire 34000 lines of pgrx, pgrx-macros, and pgrx-sql-entity-graph".

eeeebbbbrrrr:
> Unfortunately I don't think it is that simple.
> 
> 1. SpiHeapTupleData is also just a pointer to allocated data that is freed by SPI_finish.

There's a follow-up commit here that addresses this.

eeeebbbbrrrr:
I want to spend a little more time analyzing this, but I think this PR is now catching the things.

It looks like there were two problems here:

- not enough of the Spi types and functions have the proper (any?) lifetime annotations to ensure their instances or return values all associated with the surrounding `Spi::connect()` lifetime.

- `SpiHeapTupleDataEntry::value()` is allocating the "FromDatum" version of the backing Datum in the wrong Postgres `MemoryContext`

pgrx/src/spi.rs
```diff
@@ -952,22 +951,26 @@ impl<'a> Query for PreparedStatement<'a> {
...
     ///
     /// # Panics
     ///
     /// This function will panic if the supplied `query` string contained a NULL byte
-    pub fn prepare(&self, query: &str, args: Option<Vec<PgOid>>) -> Result<PreparedStatement> {
+    pub fn prepare(
+        &self,
+        query: &str,
+        args: Option<Vec<PgOid>>,
+    ) -> Result<PreparedStatement<'conn>> {
```
workingjubilee, line 973: If `SpiClient<'conn>`, then that implies SpiClient is bound by the `'conn` lifetime... so this should correctly relate the `&self` and `PreparedStatement<'conn>` lifetimes. But then there's `&str`, which implies another lifetime.

I think normal function variance should take care of how it only needs to outlive `'conn`, but it will also be easy to add another lifetime parameter later if it doesn't.

pgrx/src/spi.rs
```diff
@@ -876,9 +875,9 @@ impl<'a> PreparedStatement<'a> {
     }
 }
 
-impl<'a: 'b, 'b> Query for &'b PreparedStatement<'a> {
+impl<'conn> Query<'conn> for &PreparedStatement<'conn> {
```
workingjubilee, line 878: Hmmm, previously we said `'a` outlives `'b`. This may fall prey to the wrong lifetime being inferred, but an easy fix if that turns out to be the case.

pgrx/src/spi.rs
```diff
@@ -409,25 +409,26 @@ impl<'a> Query for &'a str {
 }
 
 #[derive(Debug)]
-pub struct SpiTupleTable {
+pub struct SpiTupleTable<'conn> {
     #[allow(dead_code)]
     status_code: SpiOkCodes,
-    table: Option<*mut pg_sys::SPITupleTable>,
+    table: Option<&'conn mut pg_sys::SPITupleTable>,
```
workingjubilee, line 415: Do we later use this to access data allocated past the SPITubleTable, or does it alias with anything else? I notice it doesn't get a lot of action (directly) in our test infra, so I'm wondering if the new `&'conn mut` will break existing user code.

pgrx/src/spi.rs
```diff
@@ -324,9 +324,9 @@ fn prepare_datum(datum: Option<pg_sys::Datum>) -> (pg_sys::Datum, std::os::raw::
     }
 }
 
-impl<'a> Query for &'a str {
+impl<'conn> Query<'conn> for &str {
```
workingjubilee, line 327: Hmm. How long *does* the `&str` need to live (or not-live), exactly?

workingjubilee (commented):
Non-blocking, I think.

workingjubilee:
> As you might have noticed, approximately **everything** implements FromDatum and it is a bound for **everything** thus it is used **everywhere** so "just adding a lifetime parameter" probably is close to "rewrite the entire 34000 lines of pgrx, pgrx-macros, and pgrx-sql-entity-graph".

Which, to be clear, _is_ what is going to happen, effectively.

> not enough of the Spi types and functions have the proper (any?) lifetime annotations to ensure their instances or return values all associated with the surrounding `Spi::connect()` lifetime.

I have some questions about those.

pgrx/src/spi.rs
eeeebbbbrrrr: The `&str` (ie, `Self`) only needs to live as long as it takes for us to copy it into a `CString` to send off to Postgres as the query string.  Same us true for the `impl Query for &String` impl.  

I wonder if we should have instead just done:  `impl<'conn, Q: AsRef<str>> Query<'conn> for Q` to better express what we're implementing for.

tl;dr is that we don't care how long the `&str` lives b/c we copy it.

pgrx/src/spi.rs
eeeebbbbrrrr: Being an implementation detail, I'm unclear how it would break existing user code?

If that thing were to leak out to the public API tho, at least now it has a proper lifetime and might affect compilation, which we'd want, yeah?

workingjubilee: There's a few details where internal changes can actually leak out due to Rust unfortunately inferring certain information that is otherwise fairly laborious to annotate on _every single type ever_, and I can never off-handedly remember exactly which things get affected by this. I think `*mut T` to `&'a mut T` is usually nonbreaking, however, it's `&'a T` to `&'a mut T` that reliably gotchas people.

This is sometimes referred to as the "perfect derive problem" or "semver hazards": https://github.com/rust-lang/lang-team/blob/master/design-meeting-minutes/2022-04-12-implied-bounds-and-perfect-derive.md

pgrx/src/spi.rs
workingjubilee: Ah okay.

pgrx/src/spi.rs
eeeebbbbrrrr: The only way to create a `PreparedStatement` is through `SpiClient::prepare()` which can only happen within `Spi::connect()` and now as of 0e5b5ae48b54bcd9f856c836c6e90c00a11740c9 all the lifetime connections should be sorted such that it's also no possible to return a `PreparedStatement` from `Spi::connect(|c| ...)`.

That said, I added these bounds back (with a better name) because yes, unlike the `&str` case, the connection needs to "outlive" the prepared statement.  Per the Postgres docs (https://www.postgresql.org/docs/current/spi-spi-prepare.html), a prepared statement only lives as long as you're under `SPI_connect()`.

pgrx/src/spi.rs
eeeebbbbrrrr: `.prepare()` isn't also borrowing from `&self`, so I think this is okay.  IOW, it doesn't matter how long `&self` lives after we get the resulting `PreparedStatement<'conn>`.  What matter is only that the prepared statement can't somehow escape the 'conn.  KAAAAHN!!!!

eeeebbbbrrrr:
> I want to spend a little more time analyzing this, but I think this PR is now catching the things.
> 
> It looks like there were two problems here:
> 
> * not enough of the Spi types and functions have the proper (any?) lifetime annotations to ensure their instances or return values all associated with the surrounding `Spi::connect()` lifetime.
> * `SpiHeapTupleDataEntry::value()` is allocating the "FromDatum" version of the backing Datum in the wrong Postgres `MemoryContext`

To amend the first bullet point there...  The `SpiClient` that `Spi::connect()` gives the closure needs to be an anonymous lifetime, not one dictated by the caller, which is what I first did in [67aad55](https://github.com/pgcentralfoundation/pgrx/pull/1210/commits/67aad5543a39b101ec9b4749129d9fefebe5d927).

This stuff is incredibly complex.

Our intention here was always that any Spi-related things created within the scope of `Spi::connect()` cannot escape that scope.  The only thing that can are returned Datums, which we have special support for to allocate in the parent memory context.  That's essentially point two above in that we were missing that in one place.

I want to think on this a bit more before we merge, but I'm feeling pretty good about it now.

pgrx-tests/src/tests/spi_tests.rs
```diff
@@ -178,9 +179,9 @@ mod tests {
     }
 
     #[pg_test]
-    fn test_inserting_null() -> Result<(), pgrx::spi::Error> {
+    fn test_inserting_null() -> Result<(), Box<dyn Error>> {
```
eeeebbbbrrrr, line 182: I don’t think this was necessary

pgrx/src/spi.rs
eeeebbbbrrrr: Jubilee, is there an action item here?  I’m not really sure.

pgrx/src/spi.rs
```diff
@@ -1292,7 +1266,7 @@ impl SpiHeapTupleData {
     pub fn get<T: IntoDatum + FromDatum>(&self, ordinal: usize) -> Result<Option<T>> {
         self.get_datum_by_ordinal(ordinal).map(|entry| entry.value())?
     }
-
+T
```
eeeebbbbrrrr: WT…T?

pgrx-tests/src/tests/spi_tests.rs
eeeebbbbrrrr: fixed

pgrx/src/spi.rs
eeeebbbbrrrr: fixed

pgrx/src/spi.rs
workingjubilee: No, I don't think so, I'm just noting it as something to keep an eye on and to try to write more tests that exercise the more niche points of `pgrx::spi`.

workingjubilee: Well, there was the question about aliasing as usual, and also about range of access, given that I'm not immediately sure if `pg_sys::SPITubleTable` is like `ArrayType` or not.

thomcc: Range of access doesn't really matter. Stacked borrowed forbids it, but treed-borrows allows `&mut Header` with trailing data, and I've been told that even though TB is likely to be replaced, that problem will certainly remain fixed in the replacement.

IOW, this only matters for running in miri (under default settings), which isn't a concern.

pgrx-tests/src/tests/spi_tests.rs
```diff
@@ -508,4 +509,50 @@ mod tests {
...
+    // }
+    //
+    // #[pg_extern]
+    // fn issue1209_prepared_stmt(q: &str) -> Result<Option<String>, Box<dyn std::error::Error>> {
+    //     use pgrx::spi::Query;
+    //
+    //     let prepared = { Spi::connect(|c| c.prepare(q, None))? };
+    //
+    //     Ok(Spi::connect(|c| prepared.execute(&c, Some(1), None)?.first().get(1))?)
+    // }
```
workingjubilee, line 557: We should handle the `ui_test` things in a later PR, it's a small amount of setup to establish the first ones and best not to block this on that.

pgrx/src/spi.rs
```diff
@@ -324,16 +316,16 @@ fn prepare_datum(datum: Option<pg_sys::Datum>) -> (pg_sys::Datum, std::os::raw::
...
-    type Result = Result<SpiTupleTable>;
+    type Result = Result<SpiTupleTable<'conn>>;
 
     /// # Panics
     ///
     /// This function will panic if somehow the specified query contains a null byte.
     fn execute(
         self,
-        _client: &SpiClient,
+        _client: &SpiClient<'conn>,
```
workingjubilee, line 328: This is making me realize that I don't understand why Query is its own trait instead of being some kind of extension trait on SpiClient, if not **to** specifically help with this kind of unsound pattern, because it breaks the core lifetime relationship and is implemented the same way in all(?) cases.

However, I am not going to think about that too hard right now as... well, it's a preexisting oddity.

workingjubilee (approved):
I believe this is an improvement and should resolve the immediate problem.

PR #1212, opened 2023-07-14
[donotland] Looking into the versioned-so thing

thomcc:
This just turns on verbose output for the linker, since we aren't getting error messages from CI now.


workingjubilee:
Resolved in #1214.

PR #1214, opened 2023-07-14
Stop SpiClient soundness from regressing

### LINKED ISSUE #1211: Add support for "compiletest" tests
Something that came out of PR #1210 is that there's some code patterns we'd like to ensure don't compile, which would be a way to help us enforce some borrowing rules/requirements around (at least) Spi.

That code is:

```rust
#[pg_test]
fn issue1209() -> Result<Option<String>, Box<dyn Error>> {
    // create the cursor we actually care about
    let mut res = Spi::connect(|c| {
        c.open_cursor("select 'hello' from generate_series(1, 10000)", None)
            .fetch(10000)
            .unwrap()
    });

    // here we just perform some allocations to make sure that the previous cursor gets invalidated
    for _ in 0..1000 {
        Spi::connect(|c| c.open_cursor("select 1", None).fetch(1).unwrap());
    }

    // later elements are probably more likely to point to deallocated memory
    for _ in 0..1000 {
        res.next();
    }

    // segfault
    Ok(res.next().unwrap().get::<String>(1)?)
}

#[pg_extern]
fn issue1209_prepared_stmt(q: &str) -> Result<Option<String>, Box<dyn std::error::Error>> {
    use pgrx::spi::Query;

    let prepared = { Spi::connect(|c| c.prepare(q, None))? };

    Ok(Spi::connect(|c| prepared.execute(&c, Some(1), None)?.first().get(1))?)
}
```

@workingjubilee suggested we use the `ui_test` crate (https://github.com/pgcentralfoundation/pgrx/pull/1210#issuecomment-1634730902).

workingjubilee:
...by adding a regression test with trybuild!
Adds ui tests using trybuild's test runner.

Fixes #1211.

### DIFF
--- a/pgrx-tests/tests/ui/escaping-spiclient-1209-cursor.rs
+++ b/pgrx-tests/tests/ui/escaping-spiclient-1209-cursor.rs
@@ -1,0 +1,27 @@
+use pgrx::prelude::*;
+use std::error::Error;
+
+#[pg_test]
+fn issue1209() -> Result<Option<String>, Box<dyn Error>> {
+    // create the cursor we actually care about
+    let mut res = Spi::connect(|c| {
+        c.open_cursor("select 'hello' from generate_series(1, 10000)", None)
+            .fetch(10000)
+            .unwrap()
+    });
+
+    // here we just perform some allocations to make sure that the previous cursor gets invalidated
+    for _ in 0..1000 {
+        Spi::connect(|c| c.open_cursor("select 1", None).fetch(1).unwrap());
+    }
+
+    // later elements are probably more likely to point to deallocated memory
+    for _ in 0..1000 {
+        res.next();
+    }
+
+    // segfault
+    Ok(res.next().unwrap().get::<String>(1)?)
+}
+
+fn main() {}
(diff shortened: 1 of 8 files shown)


pgrx-tests/tests/ui/escaping-spiclient-1209-cursor.rs
```diff
@@ -0,0 +1,27 @@
...
+    // later elements are probably more likely to point to deallocated memory
+    for _ in 0..1000 {
+        res.next();
+    }
+
+    // segfault
+    Ok(res.next().unwrap().get::<String>(1)?)
+}
+
+fn main() {}
```
workingjubilee, line 27: I don't like the frivolous mains but whatever.

workingjubilee:
I got CI functional but this now needs cleanup in order to be ready to land again, it is not acceptable in its current state.

workingjubilee:
I hate that this works.

pgrx-tests/tests/ui/escaping-spiclient-1209-cursor.rs
thomcc: might be able to avoid with `#![crate_type = "lib"]`? not sure.

thomcc (approved):
Seems fine. Worried about some of the big constants WRT #1218 but they shouldn't be *too* big...

pgrx-tests/tests/ui/escaping-spiclient-1209-cursor.rs
workingjubilee: Tried, unfortunately didn't work.

PR #1215, opened 2023-07-14
Fix SPI datum allocation memory leaks

eeeebbbbrrrr:
The intent of SPI has been that it use whatever `CurrentMemoryContext` is prior to calling `Spi::connect()` to allocate Datums.  This allows the `Spi::connect()` closure to return type instances that might be backed by Postgres-allocated memory (`&str` and `&[u8]`, specifically).

While operating in Spi-land, we're inside a short-lived "SPI Proc" memory context -- we enter it upon `SPI_connect()` and it's free'd at `SPI_finish()`.  Unfortunately, an incorrect assumption was made that this memory context's parent is `CurrentMemoryContext` at the time `SPI_connect()` is called.  This is not the case.

Postgres attaches this "SPI Proc" context to `TopTransactionContext` instead.  As such, pgrx' use of `PgMemoryContexts::CurrentMemoryContext.parent()` to find the context for datum allocation is the wrong one.

This PR adjusts the Spi wrapper internals such that the `SpiClient` holds onto a `PgMemoryContext` and that it is set to `CurrentMemoryContext` **before** we call `SPI_connect().`

This resolves the first complaint from https://github.com/pgcentralfoundation/pgrx/issues/1209#issuecomment-1635614847 that Spi is leaky.

### DIFF
--- a/pgrx/src/spi.rs
+++ b/pgrx/src/spi.rs
@@ -232,8 +232,26 @@ impl Spi {
     }
 }
 
+#[derive(Debug)]
+struct DatumMemoryContext(PgMemoryContexts);
+impl Clone for DatumMemoryContext {
+    #[inline]
+    fn clone(&self) -> Self {
+        Self(PgMemoryContexts::For(self.0.value()))
+    }
+}
+
+impl DatumMemoryContext {
+    #[inline]
+    fn to_inner(&self) -> PgMemoryContexts {
+        self.clone().0
+    }
+}
+
 // TODO: should `'conn` be invariant?
+#[derive(Debug, Clone)]
 pub struct SpiClient<'conn> {
+    mcxt: DatumMemoryContext,
     __marker: PhantomData<&'conn SpiConnection>,
 }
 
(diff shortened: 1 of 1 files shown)


pgrx/src/spi.rs
```diff
@@ -232,8 +232,26 @@ impl Spi {
...
+}
+
+impl DatumMemoryContext {
+    #[inline]
+    fn to_inner(&self) -> PgMemoryContexts {
+        self.clone().0
+    }
+}
+
 // TODO: should `'conn` be invariant?
```
workingjubilee, line 251: an excellent question, January Eric.

david-monroe:
If this allocates into the current function context, then it still disables the hack mentioned in https://docs.rs/pgrx/latest/pgrx/spi/struct.SpiCursor.html#important-notes-about-memory-usage.

pgrx/src/spi.rs
workingjubilee: The answer is **extremely** yes.

eeeebbbbrrrr:
I don't believe this is going to make it, and that's fine!

PR #1216, opened 2023-07-17
`AnyNumeric` is no longer backed by Postgres-allocated memory

eeeebbbbrrrr:
Until now, `AnyNumeric` (and by extension, `Numeric`) has been backed by a Postgres-allocated pointer to `pg_sys::NumericData`.  This pointer would have been allocated by the `CurrentMemoryContext` and could become freed if that memory context were to be freed prior to the Rust `AnyNumeric` wrapper going out of scope.

`pg_sys::NumericData` is actually a "varlena" in disguise.  This property allows us to copy the backing bytes into Rust-owned memory, which is then properly managed for the life of the `AnyNumeric` wrapper.

Conversion from an `AnyNumeric` back to the generic `Datum` is simply a matter of copying that data into Postgres-allocated memory or projecting as one via a simple cast.

### DIFF
--- a/pgrx/src/datum/numeric.rs
+++ b/pgrx/src/datum/numeric.rs
@@ -14,7 +14,7 @@ use std::iter::Sum;
 
 use crate::numeric_support::convert::{from_primitive_helper, FromPrimitiveFunc};
 pub use crate::numeric_support::error::Error;
-use crate::{direct_function_call, pg_sys, varsize, PgMemoryContexts};
+use crate::{direct_function_call, pg_sys};
 
 /// A wrapper around the Postgres SQL `NUMERIC(P, S)` type.  Its `Precision` and `Scale` values
 /// are known at compile-time to assist with scale conversions and general type safety.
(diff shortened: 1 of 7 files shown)


eeeebbbbrrrr:
This came out of the investigations I've been doing to cleanup SPI (PR for that in the next day or so).

This appears to be the only type we support that looks like it's rust-owned data, but is actually backed by a Postgres-allocated pointer.

All of the other `FromDatum`-supporting types (which are not references or have their own lifetimes) appear to be `Copy`  or otherwise rust-owned.  I'll do another look at everything tomorrow.

pgrx/src/datum/numeric.rs
```diff
@@ -40,30 +40,12 @@ pub struct Numeric<const P: u32, const S: u32>(pub(crate) AnyNumeric);
...
-    fn drop(&mut self) {
-        if self.need_pfree {
-            unsafe {
-                pg_sys::pfree(self.inner.cast());
-            }
-        }
-    }
+    // we represent a NUMERIC as opaque bytes -- we never inspect these ourselves, only a pointer
+    // to them (cast as a [`Datum`]) are passed to Postgres
+    pub(crate) inner: Vec<u8>,
```
thomcc, line 48: This never grows, so `Box<[u8]>` would be slightly more efficient

thomcc (approved):
Looks good to me, one minor nit.

pgrx/src/datum/numeric.rs
eeeebbbbrrrr: How does one make a `[u8]` from arbitrary bytes?

pgrx/src/datum/numeric_support/datum.rs
```diff
@@ -22,42 +22,48 @@ impl FromDatum for AnyNumeric {
...
-        } else {
-            memory_context.switch_to(|_| {
-                // copy the Datum into this MemoryContext and then create the AnyNumeric over that
-                let copy = pg_sys::pg_detoast_datum_copy(datum.cast_mut_ptr());
-                Some(AnyNumeric { inner: copy.cast(), need_pfree: true })
-            })
+            // copy us into a Vec<u8>
+            let size = varsize_any(numeric);
+            let mut bytes = vec![0; size];
+            std::ptr::copy_nonoverlapping(numeric.cast::<u8>(), bytes.as_mut_ptr(), size);
```
eeeebbbbrrrr, line 45: @thomcc how would this allocate a `Box<[u8]>` instead?

pgrx/src/datum/numeric.rs
thomcc: you can use `.into()` to make a `Vec<u8>` into a `Box<[u8]>`

eeeebbbbrrrr: hmm.  Okay.  Down here (https://github.com/pgcentralfoundation/pgrx/pull/1216#discussion_r1265746694), is there a more direct way to do that?

pgrx/src/datum/numeric_support/datum.rs
thomcc: Just convert to `Box<[u8]>` after -- It should be zero-cost in this case.

thomcc (approved):

Issue #1217, opened 2023-07-18
Detecting memory errors in testing


workingjubilee:
For certain classes of errors, a test may return the correct result, without detecting a problem actual programs "in the large" may encounter. For example, writing and reading to freed memory may appear to work -- perhaps nobody has started to use that memory for anything else. Ensuring our tests fail in such cases is highly desirable. Thankfully, a large amount of research has gone into detecting such errors with minimal changes to the program under test.

1. "Dynamic" instrumentation tools, of which the main option for memory checking is Valgrind.
    - These are slow, often 10-50x slower than normal (unusable on production-sized workloads, but perhaps viable for tests), but provide high-quality detection and don't require rebuilding tested programs (although they can benefit from it, and Postgres supports such).
    - One major downside is they often require x86_64 Linux, and anything else is partially supported at best.
2. "Static" instrumentation tools, such as Address Sanitizer (ASAN), Memory Sanitizer (MSAN), and so on. These are like Valgrind but an order of magnitude faster and more portable. The main tradeoff is each individual one does not necessarily detect as many errors, and the program *must* be compiled with sanitizer support enabled.
    - For cases involving dynamic code loading (so, Postgres extensions), all involved objects must be linked against a single dynamic library for a sanitizer, not a different version. This is likely but needs research to make sure, for instance, the same LLVM version can be used to build Postgres and our Rust code, amongst other possible complications.
    - Without additional annotations, it is often not possible to detect many errors in programs which use pools like Postgres memory contexts. This applies to Valgrind as well, but Postgres already has annotations for Valgrind, not for sanitizers.
    - A final hurdle: sanitizers cannot necessarily handle `fork`, which Postgres uses to spawn processes. It's unclear if this will cause issues. Valgrind handles this if told to trace forked children.
3. Compiling Postgres with additional options when testing to aid in detecting such issues.
    - This is not limited to assertions, but there are flags for memory context checking, wiping freed memory, and so on. There are a lot of these, and we should start setting several additional ones in the cargo-pgrx-built postgres.
4. Using a hardened system allocator. [Scudo](https://llvm.org/docs/ScudoHardenedAllocator.html) is currently the most well-supported of these, but [the electric fence malloc](https://linux.die.net/man/3/efence) is the same basic idea. These replace malloc/free/etc with a version which catches some kinds of misuse.
[...]

PR #1218, opened 2023-07-18
Initial valgrind support

thomcc:
Part of #1217. To use you `--valgrind` to `cargo pgrx init` (this does not prevent use in non-valgrind cases), and then set `USE_VALGRIND=1 cargo test`. This also covers the "Compiling Postgres with additional options when testing to aid in detecting such issues." part of #1217.

`pg_ctl` hates being run under valgrind on my VM for some reason, so this just covers test and build. Still looking into this.


thomcc:
Something in this still sometimes crashes valgrind...

thomcc:
This seems to work, and uncovered no new issues (still running on some ). Usage guide:

1. Install valgrind, headers, libs (on Fedora `sudo dnf install valgrind valgrind-devel valgrind-tools-devel` is enough).

2. `cargo pgrx init --valgrind`. The only major downside to using this as your primary pgrx installation is that its slow, you can still run without valgrind. 

3. Set `USE_VALGRIND` in the environment when running tests, for example `USE_VALGRIND=1 cargo test`. `valgrind` must be on the path for this to work. This takes about 15 minutes on a very beefy cloud server, so manage your timing expectations accordingly.

thomcc:
We're green on all versions 11-16. I'll figure out the linking issue, it seems to be similar to what I was looking into in #1212, unfortunately. It doesn't occur locally, though.

cargo-pgrx/src/command/init.rs
```diff
@@ -70,6 +70,13 @@ pub(crate) struct Init {
     base_testing_port: Option<u16>,
     #[clap(long, help = "Additional flags to pass to the configure script")]
     configure_flag: Vec<String>,
+    /// Compile PostgreSQL with the necessary flags to detect a good amount of
+    /// memory errors when run under Valgrind.
+    ///
+    /// Building PostgreSQL with these flags requires that Valgrind be
+    /// installed, but the resulting build is usable without valgrind.
+    #[clap(long)]
+    valgrind: bool,
```
workingjubilee, line 79: I don't understand why these two clauses are simultaneously true ?_?

I suppose it's not critical that I do, but

MEMORY_CHECKING.md
```diff
@@ -0,0 +1,21 @@
...
+
+For some background see the writeup in <https://github.com/pgcentralfoundation/pgrx/issues/1217>.
+
+## Running with Memory Checking
+
+### Valgrind
+
+1. Install valgrind, headers, libs (on Fedora `sudo dnf install valgrind valgrind-devel valgrind-tools-devel` is enough).
+
+2. Run `cargo pgrx init --valgrind`. The only major downside to using this as your primary pgrx installation is that its slow, but you can still run without valgrind.
```
workingjubilee, line 11: I don't understand why these two clauses are simultaneously true ?_?

I suppose it's not critical that I do, but

pgrx-tests/src/framework.rs
```diff
@@ -756,3 +786,7 @@ fn find_on_path(program: &str) -> Option<PathBuf> {
     let paths = std::env::var_os("PATH")?;
     std::env::split_paths(&paths).map(|p| p.join(program)).find(|abs| abs.exists())
 }
+
+fn use_valgrind() -> bool {
+    std::env::var_os("USE_VALGRIND").is_some()
```
workingjubilee, line 791: Is it intentional here that `USE_VALGRIND=0` would work?

cargo-pgrx/src/command/init.rs
thomcc: Valgrind must be installed because postgres does something like `#include <valgrind.h>` to bring some macros into scope. Those macros expand to asm statements that are essentially just spicy nops when valgrind is not connected.

pgrx-tests/src/framework.rs
thomcc: Somewhat, in that I intentionally don't care about that case.

cargo-pgrx/src/command/init.rs
workingjubilee: Oh, I meant "why is it still slow"?

thomcc: I edited. But it's probably slow due to the -Og which might actually be unnecessary given that we aren't debugging Postgres, just the extensions.

pgrx-tests/src/framework.rs
workingjubilee: Can this be `false` on `""`, then? It's easier to do `USE_VALGRIND="" cargo test` than `unset USE_VALGRIND` if necessary.

cargo-pgrx/src/command/init.rs
workingjubilee: Oh okay!

pgrx-tests/src/framework.rs
thomcc: Sure.

workingjubilee:
```suggestion
fn use_valgrind() -> bool {
    std::env::var_os("USE_VALGRIND").is_some_and(|s| s.len > 0)
```

thomcc: `s.len()` you mean (what you wrote shouldn't compile, I think). I'd probably implement it as `std::env::var_os("USE_VALGRIND").unwrap_or_default() != ""` though.

workingjubilee: durf.

workingjubilee: gonna make me RFC property syntax I swear

workingjubilee: Anyway either way you want to write it is fine~

thomcc: I'm not picky about it.

cargo-pgrx/src/command/init.rs
thomcc: I documented this, and removed -Og

workingjubilee (approved):
This seems to work great, aside from having to uninstall half of my local setup to properly install valgrind (because that also, properly speaking, should involve syncing my packages) (because I use Arch, btw)!

PR #1219, opened 2023-07-18
Modularize pgrx::spi

workingjubilee:


pgrx/src/spi/tuple.rs
```diff
@@ -0,0 +1,529 @@
...
+        self
+    }
+
+    /// Restore the state of iteration back to before the start.
+    ///
+    /// This is useful to iterate the table multiple times
+    pub fn rewind(mut self) -> Self {
+        self.current = -1;
+        self
+    }
```
workingjubilee, line 40:
```suggestion
impl<'conn> SpiTupleTable<'conn> {
    /// `SpiTupleTable`s are positioned before the start, for iteration purposes.
    ///
    /// This method moves the position to the first row.  If there are no rows, this
    /// method will silently return.
    pub fn first(mut self) -> Self {
        self.current = 0;
        self
    }

    /// Restore the state of iteration back to before the start.
    ///
    /// This is useful to iterate the table multiple times
    pub fn rewind(mut self) -> Self {
        self.current = -1;
        self
    }
```

These functions should probably instead use `&mut self`. They could still return `&mut Self` if we really like that style of API.

pgrx/src/spi/tuple.rs
```diff
@@ -0,0 +1,529 @@
...
+
+    #[inline]
+    fn count(self) -> usize
+    where
+        Self: Sized,
+    {
+        self.size
+    }
+}
```
workingjubilee, line 308: I am not confident that it's appropriate for this type to implement Iterator itself.

pgrx/src/spi/tuple.rs
```diff
@@ -0,0 +1,529 @@
...
+        if ordinal < 1 || ordinal > self.columns() {
+            Err(Error::SpiError(SpiErrorCodes::NoAttribute))
+        } else {
+            Ok(())
+        }
+    }
+}
+
+impl<'conn> SpiHeapTupleDataEntry<'conn> {
+    pub fn value<T: IntoDatum + FromDatum>(&self) -> SpiResult<Option<T>> {
```
workingjubilee, line 487: Part of the problem really does go down all the way here, where it becomes possible to extract types with only FromDatum as the effective boundary on them.

pgrx/src/spi/cursor.rs
```diff
@@ -0,0 +1,118 @@
...
+///     // ... more code ...
+///     Ok(())
+///     // <--- second SpiTupleTable gets freed by Spi::connect at this point
+/// })
+/// # }
+/// ```
+pub struct SpiCursor<'client> {
+    pub(crate) ptr: NonNull<pg_sys::PortalData>,
+    pub(crate) __marker: PhantomData<&'client SpiClient<'client>>,
+}
```
workingjubilee, line 72: It's not entirely obvious to me why this is the lifetime marker instead of e.g. only allowing access to SpiCursor via borrowing it from an SpiClient.

workingjubilee: In general, `pgrx::spi` has a problem where the historical implementations were either slightly clumsy, which is to be expected, or too clever by half. This is a bad combination, as it became "idiomatic", as it were, to follow the existing codebase style and also construct more peculiar implementations that rely on either unnecessary or over-clever idioms, instead of trying to use a much more direct and "boring" approach. We'd still have to fix things either way, however, so it's not worth dwelling on.

pgrx/src/spi/tuple.rs
eeeebbbbrrrr: L306 seems wrong, at least. 

Why do you not think it's appropriate?  It's fairly natural to work with the results of a query as a forward-only stream of rows.

pgrx/src/spi/tuple.rs
eeeebbbbrrrr: sure.

pgrx/src/spi.rs
```diff
@@ -182,6 +188,8 @@ pub enum Error {
     NoTupleTable,
 }
 
+pub type Error = SpiError;
```
eeeebbbbrrrr, line 191: 👍

eeeebbbbrrrr (approved):

pgrx/src/spi/tuple.rs
workingjubilee: Because it allows rewinding, so it's not very "forward-only".

eeeebbbbrrrr: > Because it allows rewinding, so it's not very "forward-only".

It only allows rewinding because it can.  I agree that's definitely not an inherent property of "a forward-only stream of rows".

workingjubilee: It's usually better to implement IntoIterator if it normally allows going backwards like that.

PR #1220, opened 2023-07-18
Remove `SpiConnection`

eeeebbbbrrrr:
This removes the `SpiConnection` struct and cleans up how `SpiClient` is used.

Also, the various `'conn` lifetimes have been renamed to `'client`.

I think there's more magic that can happen as a follow-up to this.

### DIFF
--- a/pgrx/src/spi.rs
+++ b/pgrx/src/spi.rs
@@ -240,13 +240,13 @@ impl Spi {
     }
 
     pub fn get_one<A: FromDatum + IntoDatum>(query: &str) -> Result<Option<A>> {
-        Spi::connect(|mut client| client.update(query, Some(1), None)?.first().get_one())
+        Spi::connect(|client| client.update(query, Some(1), None)?.first().get_one())
     }
 
     pub fn get_two<A: FromDatum + IntoDatum, B: FromDatum + IntoDatum>(
         query: &str,
     ) -> Result<(Option<A>, Option<B>)> {
-        Spi::connect(|mut client| client.update(query, Some(1), None)?.first().get_two::<A, B>())
+        Spi::connect(|client| client.update(query, Some(1), None)?.first().get_two::<A, B>())
     }
 
     pub fn get_three<
@@ -256,25 +256,21 @@ impl Spi {
     >(
         query: &str,
     ) -> Result<(Option<A>, Option<B>, Option<C>)> {
-        Spi::connect(|mut client| {
-            client.update(query, Some(1), None)?.first().get_three::<A, B, C>()
-        })
+        Spi::connect(|client| client.update(query, Some(1), None)?.first().get_three::<A, B, C>())
     }
 
     pub fn get_one_with_args<A: FromDatum + IntoDatum>(
         query: &str,
         args: Vec<(PgOid, Option<pg_sys::Datum>)>,
     ) -> Result<Option<A>> {
-        Spi::connect(|mut client| client.update(query, Some(1), Some(args))?.first().get_one())
+        Spi::connect(|client| client.update(query, Some(1), Some(args))?.first().get_one())
     }
 
     pub fn get_two_with_args<A: FromDatum + IntoDatum, B: FromDatum + IntoDatum>(
         query: &str,
         args: Vec<(PgOid, Option<pg_sys::Datum>)>,
     ) -> Result<(Option<A>, Option<B>)> {
-        Spi::connect(|mut client| {
-            client.update(query, Some(1), Some(args))?.first().get_two::<A, B>()
-        })
+        Spi::connect(|client| client.update(query, Some(1), Some(args))?.first().get_two::<A, B>())
     }
 
     pub fn get_three_with_args<
@@ -285,7 +281,7 @@ impl Spi {
         query: &str,
         args: Vec<(PgOid, Option<pg_sys::Datum>)>,
     ) -> Result<(Option<A>, Option<B>, Option<C>)> {
-        Spi::connect(|mut client| {
+        Spi::connect(|client| {
             client.update(query, Some(1), Some(args))?.first().get_three::<A, B, C>()
         })
     }
@@ -308,7 +304,7 @@ impl Spi {
         query: &str,
         args: Option<Vec<(PgOid, Option<pg_sys::Datum>)>>,
     ) -> std::result::Result<(), Error> {
-        Spi::connect(|mut client| client.update(query, None, args).map(|_| ()))
+        Spi::connect(|client| client.update(query, None, args).map(|_| ()))
     }
 
     /// explain a query, returning its result in json form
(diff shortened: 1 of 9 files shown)


pgrx/src/spi/client.rs
```diff
@@ -5,22 +5,26 @@ use std::ptr::NonNull;
...
     ///
     /// # Panics
     ///
     /// This function will panic if the supplied `query` string contained a NULL byte
-    pub fn prepare(
-        &self,
-        query: &str,
-        args: Option<Vec<PgOid>>,
-    ) -> SpiResult<PreparedStatement<'conn>> {
+    pub fn prepare(&self, query: &str, args: Option<Vec<PgOid>>) -> SpiResult<PreparedStatement> {
```
workingjubilee, line 27: I'm surprised this is valid?

pgrx/src/spi/client.rs
```diff
@@ -119,44 +128,20 @@ impl<'conn> SpiClient<'conn> {
...
-    /// when SpiConnection is dropped, we make sure to disconnect from SPI
+impl Drop for SpiClient {
     fn drop(&mut self) {
         // best efforts to disconnect from SPI
         // SPI_finish() would only complain if we hadn't previously called SPI_connect() and
-        // SpiConnection should prevent that from happening (assuming users don't go unsafe{})
+        // SpiClient will prevent that from happening (assuming users don't go unsafe{})
         Spi::check_status(unsafe { pg_sys::SPI_finish() }).ok();
     }
 }
```
workingjubilee, line 147: So the problem here is that now that it's on SpiClient, people can just run `mem::forget(client)` safely.

pgrx/src/spi/client.rs
eeeebbbbrrrr: How so?

Do you mean it might need to be 
```rust
fn prepare<'client>(&'client self, ...) -> SpiResult<PreparedStatement<'client>> {}
```

pgrx/src/spi/client.rs
eeeebbbbrrrr: The user only sees a `&mut client`.  Wouldn't `mem::forget()` on a reference be a noop?

pgrx/src/spi/client.rs
```diff
@@ -5,22 +5,26 @@ use std::ptr::NonNull;
 use crate::pg_sys::{self, PgOid};
 use crate::spi::{PreparedStatement, Query, Spi, SpiCursor, SpiError, SpiResult, SpiTupleTable};
 
-// TODO: should `'conn` be invariant?
-pub struct SpiClient<'conn> {
-    __marker: PhantomData<&'conn SpiConnection>,
-}
+pub struct SpiClient;
```
workingjubilee, line 8: This allows creating an SpiClient out of nothing and then forcing `SPI_finish` prematurely, simply by writing
```rust
let x = SpiClient;
mem::drop(x);
```

This could then cause problems where memory is now freed by Postgres but the Rust lifetime may not have ended. This approach will not work.

eeeebbbbrrrr: oh yeah, that shouldn't have been `pub`.  `pub(super)` is what I wanted!

workingjubilee: `Spi::connect` depends on exposing access to `SpiClient` in its current API.

pgrx/src/spi/client.rs
workingjubilee: I expected Rust to ask for `SpiResult<PreparedStatement<'_>> at least.

pgrx/src/spi/client.rs
eeeebbbbrrrr: yeah yeah yeah.  fix incoming....

eeeebbbbrrrr: smash that refresh button!

pgrx/src/spi/client.rs
eeeebbbbrrrr: even better:

```
error[E0382]: borrow of moved value: `client`
   --> pgrx-tests/src/tests/spi_tests.rs:228:13
    |
226 |         Spi::connect(|client| {
    |                       ------ move occurs because `client` has type `&mut SpiClient`, which does not implement the `Copy` trait
227 |             std::mem::forget(client);
    |                              ------ value moved here
228 |             client.update("CREATE TABLE tests.cursor_table (id int)", None, None)?;
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ value borrowed here after move

```

pgrx/src/spi.rs
```diff
@@ -366,7 +361,7 @@ impl Spi {
     /// This function will panic if for some reason it's unable to "connect" to Postgres' SPI
     /// system.  At the time of this writing, that's actually impossible as the underlying function
     /// ([`pg_sys::SPI_connect()`]) **always** returns a successful response.
-    pub fn connect<R, F: FnOnce(SpiClient<'_>) -> R>(f: F) -> R {
+    pub fn connect<R, F: FnOnce(&mut SpiClient) -> R>(f: F) -> R {
```
workingjubilee, line 364: I'm worried about contravariance, though if I remember correctly the lifetime parameter was already contravariant.

eeeebbbbrrrr: I had a thought that this would now get us closer to how scoped threads work -- like we were discussing on discord earlier today.  

I haven't looked at this yet tho.

This formulation at least clearly indicates that SpiThings are borrowed from the mutable SpiClient instance, rather than needing to live as long as its lifetime.  At least that's how I see it in my head.

pgrx/src/spi/client.rs
workingjubilee: I didn't notice the change to `&mut SpiClient` from `SpiClient<'_>` when I said that. This seems like it could be an improvement.

workingjubilee:
Whyfor the rename to `'client`, btw? I figure SpiClient is still modeling a connection but if you view it differently that's interesting to know.

eeeebbbbrrrr:
> Whyfor the rename to `'client`, btw? I figure SpiClient is still modeling a connection but if you view it differently that's interesting to know.

My thought was as things are laid out in this PR, it’s the `&mut SpiClient` that’s in control of the lifetime. So I thought “the client provides these things” or "these things should borrow from a client". 

Conceptually yeah, it’s a “connection”. I have no particular preference. I'm just trying to keep things straight in my head as we sort though this stuff.

workingjubilee:
I mulled it over a bit and I'm not entirely sure this is the right direction because while the cleanup is overall desirable, the signature change could subtly break something and there's some sinister things we can probably do with signatures that look more like our current instance.

eeeebbbbrrrr:
I don't believe this is going to make it, and that's fine!

PR #1225, opened 2023-07-24
Changes GHA workflows to use new upgraded runners

BradyBonnette:
Also did a one-off to make sure that nightly/WIB works with the new setup


eeeebbbbrrrr (approved):
LGTM.  We'll need to cherry-pick back into `develop` and resolve the merge conflicts -- we'll basically want the changes from both sides except for the "midway cleanup" part.

PR #1227, opened 2023-07-25
Modularize the interior of pgrx-pg-sys

workingjubilee:
The pgrx-pg-sys crate has been gradually doing better, but the primary module was still a mess of bizarre configuration code strewn about. There's a certain inevitability of that, in some ways, but we can recover a lot of implicit information by organizing it, with fairly little effort. Now it's easy to tell the difference between macro ports and cshim functions at a glance.


thomcc (approved):
Seems fine.

PR #1229, opened 2023-07-25
Add support for handling SIGINT and SIGCHLD from bgworker

JelteF:
Both of these are pretty useful signals to be able to handle from a
background worker.
- `SIGINT` allows doing something special when `pg_cancel_backend` is
  called on the pid of your background worker
- `SIGCHLD` is useful when your background worker spawns other processes
  and wants to get notified when they crash/finish.

### DIFF
--- a/pgrx/src/bgworkers.rs
+++ b/pgrx/src/bgworkers.rs
@@ -107,7 +107,10 @@ impl BackgroundWorker {
         .expect("'extra' is not valid UTF8")
     }
 
-    /// Have we received a SIGUP?
+    /// Have we received a SIGHUP since previously calling this function? This resets the
+    /// internal boolean that tracks if SIGHUP was received. So when calling this twice in
+    /// in a row, the second time will always return false.
+    #[must_use = "you aren't getting the same bool back if you call this twice"]
     pub fn sighup_received() -> bool {
         unsafe {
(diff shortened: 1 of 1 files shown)


pgrx/src/bgworkers.rs
```diff
@@ -121,6 +125,24 @@ impl BackgroundWorker {
...
+    pub fn sigint_received() -> bool {
+        unsafe {
+            assert!(!pg_sys::MyBgworkerEntry.is_null(), "BackgroundWorker associated functions can only be called from a registered background worker");
+        }
+        // toggle the bool to false, returning whatever it was
+        GOT_SIGINT.swap(false, Ordering::SeqCst)
+    }
+
+    /// Have we received a SIGCHLD?
+    pub fn sigchld_received() -> bool {
```
workingjubilee, line 138: Please add documentation to these functions that indicates they mutate the worker's state, so you **better** actually use this boolean instead of just calling it for funsies. Perhaps also with something like `#[must_use = "you aren't getting the same bool back if you call this twice"]`?

You may update the other functions already like these two for consistency, or allow that to happen in another PR, but I do not want to merge another one of these without such.

JelteF: done

workingjubilee (approved):
Excellent, thank you so much!

PR #1235, opened 2023-07-26
Ignores UI tests for MUSL environments

BradyBonnette:
The new UI tests introduced in `develop` seem to be incompatible with musl environments such as Alpine. The current working theory is that there could be a bug in `TryBuild` that ignores/replaces certain rustc flags. In particular, it seems to be disregarding the `"-Ctarget-feature=-crt-static"` flags that are necessary for pgrx to build and be used in Alpine.

Also includes a couple minor updates, such as bumping the postgresql-alpine Docker version.

### DIFF
--- a/.github/docker/Dockerfile.alpine:3.18
+++ b/.github/docker/Dockerfile.alpine:3.18
@@ -1,5 +1,5 @@
 # Example of how to build this container (ran from pgrx checkout directory):
-#   docker build --build-arg PG_MAJOR_VER=14 -f .github/docker/Dockerfile.alpine\:3.16 -t pgrx-alpine .
+#   docker build --build-arg PG_MAJOR_VER=14 -f .github/docker/Dockerfile.alpine\:3.18 -t pgrx-alpine .
 #
 # Example of how to run this container with tests after building the image:
 #   docker run pgrx-alpine cargo test --no-default-features --features pg14 --locked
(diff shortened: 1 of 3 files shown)


BradyBonnette:
Note that I have the "Will It Blend" tests enabled for now, just to be sure they work. Before this is merged into `develop`, Ill change them back to be ran as nightly again.

thomcc:
I think it's more that cargo doesn't respect it as a rustflag when cross compiling, and the workarounds (`.cargo/config.toml`/`CARGO_TARGET_*_RUSTFLAGS`) aren't respected/relevant. But we haven't looked into it enough for me to feel good filing a bug upstream about it.

BradyBonnette:
That's right. Locally I tried changing both `.cargo/config.toml` to include:
```toml
[target.'cfg(target_env="musl")']
rustflags = ["-Ctarget-feature=-crt-static"]
```
and that did not work. I also tried running the tests with something such as: `CARGO_TARGET_X86_64_UNKNOWN_LINUX_MUSL_RUSTFLAGS="-Ctarget-feature=-crt-static" cargo test ui --no-default-features --features "pg14" -- --exact` and that also did not work.

workingjubilee (approved):
Yeah, this is fine.

workingjubilee:
Also to clarify for anyone who comes back and rereads this:

The main reason waiving these tests on a musl target is fine is because they don't add anything to rerun in different envs if they cause trouble for us in CI. The tests are fundamentally tests to guarantee attempts to invoke functions in certain ways fail, because we designed our API to prohibit it. We know the Rust toolchain isn't going to have so much platform-specific variance that these work on some targets and not others.

In a certain sense, if the tests themselves fail to compile at all on a specific target environment, uhm... mission accomplished???

We should reenable them someday when musl distros work as smoothly for Rust as glibc distros do, which is a process mostly happening upstream in Rust's compiler/std/libc bindings.

PR #1237, opened 2023-07-28
Add postmaster related api

JelteF:
For an extension I'm writing I want to connect back to Postgres. So I
need to know what port and address Postgres is running on. The
`postmaster.h` header contains that information, so this starts
including that header in `pgrx_pg_sys`.


JelteF:
Oh sorry, I realized I opened this PR against master...

eeeebbbbrrrr:
Guess what?  We’re making “develop” the default branch and are just getting rid of all this confusion. We get caught by it too!

Issue #1222, opened 2023-07-19
Improve `cargo pgrx init` error message when bzip2 is not installed


thomcc:
If you don't have `bzip2` installed `cargo pgrx init` seems to fail with this relatively cryptic error:

```
$ cargo pgrx init
   Discovered Postgres v11.20, v12.15, v13.11, v14.8, v15.3, v16beta2
  Downloading Postgres v16beta2 from https://ftp.postgresql.org/pub/source/v16beta2/postgresql-16beta2.tar.bz2
  Downloading Postgres v12.15 from https://ftp.postgresql.org/pub/source/v12.15/postgresql-12.15.tar.bz2
  Downloading Postgres v15.3 from https://ftp.postgresql.org/pub/source/v15.3/postgresql-15.3.tar.bz2
  Downloading Postgres v11.20 from https://ftp.postgresql.org/pub/source/v11.20/postgresql-11.20.tar.bz2
  Downloading Postgres v14.8 from https://ftp.postgresql.org/pub/source/v14.8/postgresql-14.8.tar.bz2
  Downloading Postgres v13.11 from https://ftp.postgresql.org/pub/source/v13.11/postgresql-13.11.tar.bz2
     Removing /home/fedora/.pgrx/11.20
    Untarring Postgres v11.20 to /home/fedora/.pgrx/11.20
      Removing /home/fedora/.pgrx/14.8
    Untarring Postgres v14.8 to /home/fedora/.pgrx/14.8
     Removing /home/fedora/.pgrx/12.15
    Untarring Postgres v12.15 to /home/fedora/.pgrx/12.15
     Removing /home/fedora/.pgrx/13.11
    Untarring Postgres v13.11 to /home/fedora/.pgrx/13.11
     Removing /home/fedora/.pgrx/15.3
    Untarring Postgres v15.3 to /home/fedora/.pgrx/15.3
     Removing /home/fedora/.pgrx/16beta2
    Untarring Postgres v16beta2 to /home/fedora/.pgrx/16beta2
Error:
   0: Broken pipe (os error 32)

Location:
   cargo-pgrx/src/command/init.rs:274

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ SPANTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

   0: cargo_pgrx::command::init::download_postgres with pg_version=11.20
      at cargo-pgrx/src/command/init.rs:210
[...]

PR #1238, opened 2023-07-28
Addresses cargo-pgrx error reporting

BradyBonnette:
There are some situations where `cargo pgrx` reports an error but important details are missing, as it relates to the spawning of child process. One example of this is when `cargo pgrx` downloads and compiles PostgreSQL from source, as demonstrated by https://github.com/pgcentralfoundation/pgrx/issues/1222

The changes here seek to make any error output due to a spawned child process be reported to the end user instead of being silently swallowed and disregarded.

### DIFF
--- a/cargo-pgrx/src/command/init.rs
+++ b/cargo-pgrx/src/command/init.rs
@@ -271,14 +271,35 @@ fn untar(bytes: &[u8], pgrxdir: &PathBuf, pg_config: &PgConfig) -> eyre::Result<
         .wrap_err("failed to spawn `tar`")?;
 
     let stdin = child.stdin.as_mut().expect("failed to get `tar`'s stdin");
-    stdin.write_all(bytes)?;
-    stdin.flush()?;
+    let mut error = None;
+
+    // In order to capture and report any output from the child's stderr (in the event of an error),
+    // we need to wait until `.wait_with_output()` is called. Bailing early here via `?` or `.expect()`
+    // means only the immediate error is reported, such as "Broken Pipe". The same applies for the subsequent
+    // `.flush()` call immediately after.
+    match stdin.write_all(bytes) {
+        Ok(_) => (),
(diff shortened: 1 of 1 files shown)


thomcc (approved):
LGTM

BradyBonnette:
@thomcc I might go through some of the other places where `Command` is used and see if I can verify that those report similar to this.

cargo-pgrx/src/command/init.rs
```diff
@@ -271,14 +271,35 @@ fn untar(bytes: &[u8], pgrxdir: &PathBuf, pg_config: &PgConfig) -> eyre::Result<
...
+    // `.flush()` call immediately after.
+    match stdin.write_all(bytes) {
+        Ok(_) => (),
+        Err(e) => error = Some(e),
+    }
+
+    match stdin.flush() {
+        Ok(_) => (),
+        Err(e) => error = Some(e),
+    }
```
workingjubilee, line 288: Is it possible for both of these to error? We'd hypothetically be dumping one error on the floor in favor of another, again.

workingjubilee (commented):
Yeah it _might_ be nice to factorize this out a bit if we have a bunch of `Command::spawn` sites that look like this. Not strictly necessary, tho'.

cargo-pgrx/src/command/init.rs
thomcc: I would be very surprised if flush did anything on this type. But we should probably either take the first error or use chaining, if we care.

BradyBonnette: I thought about this as I was writing it, but I wasn't entirely sure how to handle it without making it "shift right" or making a plate of spaghetti out of it.

workingjubilee: You will see me tend to incrementally construct and then immediately pattern-match on tuples a lot in my code for this reason.

PR #1239, opened 2023-07-28
 Add a env flag that can be set to skip `#[pg_test]`-generated tests.

thomcc:
Requested by @SaltyDavidselph, who wants to run only `#[test]` functions, but not `#[pg_test]` ones.


workingjubilee (approved):
Wonder if we will want to instead do something like `PGRX_TEST="blah"` at some point, but seems fine.

Issue #1076, opened 2023-03-15
Custom types - getting "ERROR: type 'x' already exists" in multiple cases


nguiard:
Hi,
Thanks for this awesome library.

I'm fooling around with custom types.

- This works:
```rust
use pgx::prelude::*;
use serde::{Serialize, Deserialize};

pgx::pg_module_magic!();


#[derive(Debug, Serialize, Deserialize, PostgresType)]
pub struct A;

#[pg_extern]
fn myfunc() -> A {
    A
}
```

- However, this does not anymore, when you wrap the return in a `Result`
```rust
use pgx::prelude::*;
use pgx::spi;
use serde::{Serialize, Deserialize};

pgx::pg_module_magic!();


#[derive(Debug, Serialize, Deserialize, PostgresType)]
pub struct A;

#[pg_extern]
fn myfunc() -> Result<A, spi::Error> {
    Ok(A)
}
```
It compiles fine of course, but when you try to `create extension`, you get `ERROR:  type "a" already exists`. (also, it's not dependent on `spi`, changing `spi::Error` to `std::error::Error` produces the same result).

It seems that this is tied to ordering problems in the generated sql file. Because this also fails with the same error:
```rust
use pgx::prelude::*;
use pgx::spi;
use serde::{Serialize, Deserialize};

pgx::pg_module_magic!();


#[derive(Debug, Serialize, Deserialize, PostgresType)]
pub struct B;

#[derive(Debug, Serialize, Deserialize, PostgresType)]
pub struct A;

#[pg_extern]
fn myfunc(_x: B) -> Result<A, spi::Error> {
    Ok(A)
}
```

But, rename `struct A` to `struct Z`, and it now works!
```rust
use pgx::prelude::*;
use pgx::spi;
use serde::{Serialize, Deserialize};

pgx::pg_module_magic!();


#[derive(Debug, Serialize, Deserialize, PostgresType)]
pub struct B;

#[derive(Debug, Serialize, Deserialize, PostgresType)]
pub struct Z;

#[pg_extern]
fn myfunc(_x: B) -> Result<Z, spi::Error> {
    Ok(Z)
}
```

marcusllorente-msft:
Hi, I am encountering the same error as well. @nguiard I don't see why renaming the struct from A to Z would work, and I can't seem to make it work on my end when replicating the change from `struct A` to `struct Z`. Any insight?

eeeebbbbrrrr:
[...]

PR #1240, opened 2023-07-31
Cleanup the error when cargo-pgrx version doesn't match Cargo.toml

eeeebbbbrrrr:
As was discussed in the comments in another issue
(https://github.com/pgcentralfoundation/pgrx/issues/1076#issuecomment-1656213886), the error message when `cargo-pgrx`'s version doesn't match the pgrx crate dependency version is confusing.

This cleans it up and also tells you which Cargo.toml file it is.

### DIFF
--- a/cargo-pgrx/src/metadata.rs
+++ b/cargo-pgrx/src/metadata.rs
@@ -65,11 +65,15 @@ pub fn validate(
             .collect::<Vec<_>>()
             .join(", ");
 
-        return Err(eyre!("The installed `cargo-pgrx` v{cargo_pgrx_version} \
+        return Err(eyre!(
+            "The installed `cargo-pgrx` v{cargo_pgrx_version} \
         is not compatible with the {mismatches} {} in `{}`. `cargo-pgrx` \
-        and pgrx dependency versions must be identical.", 
-            if many == 1 { "dependency" } else { "dependencies"}, 
-            path.map(|p| p.as_ref().display().to_string()).unwrap_or_else(|| "./Cargo.toml".to_string()).yellow()));
+        and pgrx dependency versions must be identical.",
+            if many == 1 { "dependency" } else { "dependencies" },
+            path.map(|p| p.as_ref().display().to_string())
+                .unwrap_or_else(|| "./Cargo.toml".to_string())
+                .yellow()
+        ));
     }
 
     Ok(())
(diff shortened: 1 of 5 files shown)


cargo-pgrx/src/metadata.rs
```diff
@@ -58,5 +57,18 @@ pub fn validate(metadata: &Metadata) -> eyre::Result<()> {
...
 
+    if !mismatches.is_empty() {
+        let many = mismatches.len();
+        let mismatches = mismatches
+            .into_iter()
+            .map(|(p, v)| format!("`{p} = {v}`"))
+            .collect::<Vec<_>>()
+            .join(", ");
+
+        return Err(eyre!("The installed `cargo-pgrx` v{cargo_pgrx_version} is not compatible with the {mismatches} {} in `{}`.  `cargo-pgrx` and pgrx dependency versions must be identical.",
```
thomcc: this line is getting very long, so it might be nice to split it up.

thomcc (approved):
Small nit, but otherwise looks fine.

cargo-pgrx/src/metadata.rs
workingjubilee:
```suggestion
        return Err(eyre!("The installed `cargo-pgrx` v{cargo_pgrx_version} \
        is not compatible with the {mismatches} {} in `{}`. `cargo-pgrx` \
        and pgrx dependency versions must be identical.", 
```

workingjubilee (approved):
Looks great, though apparently my suggestion was not perfectly formatted!

eeeebbbbrrrr:
> Looks great, though apparently my suggestion was not perfectly formatted!

it is now.

cargo-pgrx/src/metadata.rs
```diff
@@ -58,5 +57,24 @@ pub fn validate(metadata: &Metadata) -> eyre::Result<()> {
...
+        let many = mismatches.len();
+        let mismatches = mismatches
+            .into_iter()
+            .map(|(p, v)| format!("`{p} = {v}`"))
+            .collect::<Vec<_>>()
+            .join(", ");
+
+        return Err(eyre!(
+            "The installed `cargo-pgrx` v{cargo_pgrx_version} \
+        is not compatible with the {mismatches} {} in `{}`. `cargo-pgrx` \
```
thomcc:
```suggestion
            is not compatible with the {mismatches} {} in `{}`. `cargo-pgrx` \
```

cargo-pgrx/src/metadata.rs
```diff
@@ -58,5 +57,24 @@ pub fn validate(metadata: &Metadata) -> eyre::Result<()> {
...
+        and pgrx dependency versions must be identical.",
```
thomcc:
```suggestion
            and pgrx dependency versions must be identical.",
```

thomcc (approved):

PR #1241, opened 2023-07-31
Fix issue #1076: Properly handle dependency graph of `Result<T, _>`

eeeebbbbrrrr:
When a `#[pg_extern]` function returns `Result<T, E>`, the entity graph was using the "type_id" of that whole type instead of just `T`, which is what would be in the graph.

This detects that type from the first argument of the return type and uses it in the code generation.

This closes issue #1076.

### DIFF
--- a/pgrx-sql-entity-graph/src/used_type.rs
+++ b/pgrx-sql-entity-graph/src/used_type.rs
@@ -31,6 +31,7 @@ use super::metadata::FunctionMetadataTypeEntity;
 pub struct UsedType {
     pub original_ty: syn::Type,
     pub resolved_ty: syn::Type,
+    pub resolved_ty_inner: Option<syn::Type>,
     /// Set via `composite_type!()`
     pub composite_type: Option<CompositeTypeMacro>,
     /// Set via `VariadicArray` or `variadic!()`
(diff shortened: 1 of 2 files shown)


pgrx-sql-entity-graph/src/used_type.rs
```diff
@@ -283,12 +284,43 @@ impl UsedType {
...
+        // if the Type is like `Result<T, E>`, this finds the `T`
+        let mut resolved_ty_inner: Option<syn::Type> = None;
+        if result {
+            if let syn::Type::Path(tp) = &resolved_ty {
+                if let Some(first_segment) =
+                    tp.path.segments.first().map(|segment| Some(segment)).unwrap_or(None)
+                {
+                    if let syn::PathArguments::AngleBracketed(ab) = &first_segment.arguments {
+                        if let Some(first_arg) =
+                            ab.args.first().map(|arg| Some(arg)).unwrap_or(None)
```
thomcc, line 296: Hmm, I guess this just assumes that in `Foo<T, U, V, W, ...>` we care about `T` and that's it. It's not ideal but that is true for Result, Option, etc, so I suppose this is fine for now and a strict improvement.

eeeebbbbrrrr: Correct. And it only does this for Results. 

Impossible to know type_ids for a Result when you allow the E to be anything.

thomcc: Ah, yeah, sorry, I missed this was only for Result.

thomcc (approved):

PR #1242, opened 2023-08-01
Add operator and cache related api

VoVAllen:
Thanks for the pgrx project! It makes our iteration on the vector extension much easier and faster

Also thanks to @usamoi, who is the major developer of pgvecto.rs.


thomcc (approved):

workingjubilee (approved):
Nice, thank you!

PR #1243, opened 2023-08-01
Remove references to master branch

thomcc:
This probably shouldn't be merged before actually removing the branch.

The one exception to this is `finalize-release.sh`, which probably doesn't need to exist anymore after this? https://github.com/pgcentralfoundation/pgrx/blob/develop/finalize-release.sh#L13-L18

@BradyBonnette does the change to the CI look right?


eeeebbbbrrrr:
> The one exception to this is finalize-release.sh, which probably doesn't need to exist anymore after this

~~And `prepare-release.sh` probably needs some changes, which I'll do when I do the next release here in a bit.~~

It looks like it's fine as-is, despite the default branch change.

eeeebbbbrrrr (approved):

eeeebbbbrrrr:
@thomcc why don't you go ahead and remove the `finalize-release.sh` script in this PR.  Then I'll merge this after the next release.

thomcc:
Done. I had also forgotten to save the changes to CONTRIBUTING.md, which had a mention of the master branch.

.github/workflows/will-it-blend-develop.yml
```diff
@@ -22,25 +22,7 @@ jobs:
...
-            echo "Running via schedule, so using branch develop"
-            echo "NIGHTLY_BUILD_REF=develop" >> $GITHUB_ENV
-          else
-            echo "Not running via schedule, so using branch $GITHUB_HEAD_REF"
-            echo "NIGHTLY_BUILD_REF=$GITHUB_HEAD_REF" >> $GITHUB_ENV
-          fi
-
       - uses: actions/checkout@v3
-        with:
-          ref: ${{ env.NIGHTLY_BUILD_REF }}
```
BradyBonnette, line 43: Sorry, somehow I overlooked my @ notification. My apologies, @thomcc 

I think removing this will be OK and should work once develop is set as the default branch. If things fall apart of course I can go in and correct some things. We will also need to be sure that the manual run of WIB will work after these changes go through.

BradyBonnette (approved):
Looks good to me. If Will It Blend misbehaves I can always go in and address it, I am not overly concerned about it.

.github/workflows/will-it-blend-develop.yml
workingjubilee: It already is, these are just simplifications of CI since we now don't need manual overrides and they... _should_ be no-ops?

BradyBonnette: Yeah, it's best to remove the unnecessary things as to avoid confusion. I hated that I originally had to add in the logic above to make it work, and am delighted that it now gets to be ripped out.

PR #1249, opened 2023-08-03
Only enable line tables for profile.dev

thomcc:
This reduces the size of target after running tests from ~8GB to ~5GB, and makes them complete a good amount faster (generating debug info is slow). It only really makes a difference if you're going to use a debugger, in which case perhaps you should just turn that flag off.

The size and speed saving make me think it should be done generally, rather than only for CI.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -431,8 +431,11 @@ jobs:
         # ls -lath `$(which pg_config) --pkglibdir` `$(which pg_config) --sharedir`/extension
         echo ""
 
-        echo "----- Output Cargo version -----"
-        cargo --version
+        echo "----- Stable Rust ----"
+        rustup update stable
+        rustup default stable
+        rustc -vV
+        cargo -vV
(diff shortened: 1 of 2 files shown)


eeeebbbbrrrr (approved):

eeeebbbbrrrr:
This failed on MacOS with an error that seems related, but I don't quite follow what it wants:

```
error: failed to parse manifest at `/Users/runner/work/pgrx/pgrx/Cargo.toml`

Caused by:
  expected a boolean or an integer
  in `profile.dev.debug`
Error: Process completed with exit code 101.
```

Do we have different versions of rust between linux and mac in CI or does mac rust not support this value for `profile.dev.debug`?

workingjubilee:
`macos-11` doesn't include the latest Rust version, apparently. Fixed in 2d0b4b700fde16e31a70b4041d215389b290ab44

workingjubilee:
Landing this for my sanity (re: not having my HD fill up instantly).

PR #1250, opened 2023-08-04
Support for the user specified version of the init command

### LINKED ISSUE #149: Feature request: cargo pgx init --pg12=https://ftp.postgresql.org/pub/source/v12.6/postgresql-12.6.tar.bz2
It would be very helpful for Nix-based environments (or for reproducible builds in general) to specify an exact version of postgres to use during the `cargo pgx init` phase.

I have tried using a nix-shell with `cargo pgx init --pg12=${postgresql_12}/bin/pg_config`, however, since the nix store paths are not writable, this doesn't help much (e.g. pgx run, pgx install and pgx package will not work). I have also tried copying the nix postgresql_12 to a local writable directory, then using that directory during pgx init, however, the pg_config of that writable copy will still produce absolute nix store path outputs (nix has a very opinionated method to guarantee reproducible builds). Using `pgx init --pg12=download` will not produce reproducible builds in a clean CI environment.

Also see comment https://github.com/zombodb/pgx/issues/105#issuecomment-828832451 by @Hoverbear.

Please consider allowing `init` to accept a specific download url (or a part of the url), or alternatively, to accept a full version number.

Examples:

```bash
cargo pgx init --pg12=https://ftp.postgresql.org/pub/source/v12.6/postgresql-12.6.tar.bz2
cargo pgx init --pg12=12.6
```

ongchi:
In current version, user pass path to `pg_config` tool or `download` to use local installed or download and build from source from postgres latest release tarball.

This pr extends init option to support custom version:

```sh
cargo pgrx init --pg12 12.10
```

Passing `latest` as option will also download and install the latest version from source release, which is identical to `download`, while the `download` option still remains for backward compatibility.

Closes: #149

### DIFF
--- a/cargo-pgrx/src/command/init.rs
+++ b/cargo-pgrx/src/command/init.rs
@@ -13,7 +13,7 @@ use crate::CommandExecute;
 use eyre::{eyre, WrapErr};
 use owo_colors::OwoColorize;
 use pgrx_pg_config::{
-    get_c_locale_flags, prefix_path, PgConfig, PgConfigSelector, Pgrx, PgrxHomeError,
+    get_c_locale_flags, prefix_path, PgConfig, PgConfigSelector, PgVersion, Pgrx, PgrxHomeError,
 };
 use rayon::prelude::*;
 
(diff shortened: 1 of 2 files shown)


eeeebbbbrrrr:
I haven’t looked at the PR yet but I’m not keen on this idea. 

We’ve already had a PR for this and I didn’t find the argument very persuasive.  EDIT: sorry I guess it was the issue linked above. 

If pgrx is to support specific major versions then it needs a lot more consideration than just this. 

I still think the correct way to handle this workflow is to install that version locally yourself and use cargo pgrx init with a different PGRX_HOME. 

In day-to-day development of a pgrx extension there should be no differences between point releases of a major version. So the only reason to install a specific one would be to evaluate an extension against an older Postgres that probably had a bug. I don’t see this as under pgrx’ perview.

ongchi:
Thank you for your feedback and I understand your decision. I'll close this PR.

PR #1254, opened 2023-08-10
update for pg16beta3 support

eeeebbbbrrrr:


BradyBonnette:
@eeeebbbbrrrr 

So something I am noticing with the CI output -- it's still using beta2, which leads me to believe the Postgres team has not  updated their APT packages yet.

E.g. https://github.com/pgcentralfoundation/pgrx/actions/runs/5822617826/job/15787637585?pr=1254#step:5:108

The reason we were using apt packages is to cut down on the build times. Should we wait for beta3 to be available in their official apt repository before considering this good, change CI to use `cargo pgrx` to build/install, or just leave it as-is?

eeeebbbbrrrr:
I saw that too.  I don't think we should wait.  Tests pass locally on pg16beta3 from the official website.  They'll get their apt repo updated soon enough.

BradyBonnette (approved):

PR #1256, opened 2023-08-10
spi: simplify (optimize?) Datum preparation

vrmiguel:
Hey there!

This PR aims to simplify the argtypes/datums/nulls preparation code, also reducing code duplication by abstracting it into a function

I imagine that this code also runs faster since it does not allocate intermediate Vecs

I know Rust is capable of reusing allocations in some instances of `vec.into_iter().map(f).collect()` but only in scenarios where the transformation keeps the same type size and alignment ([source](https://doc.rust-lang.org/src/alloc/vec/in_place_collect.rs.html), [Godbolt](https://godbolt.org/z/7ovfn85zW))), which is not the case in here

### DIFF
--- a/pgrx/src/spi/query.rs
+++ b/pgrx/src/spi/query.rs
@@ -3,6 +3,9 @@ use std::marker::PhantomData;
 use std::ops::Deref;
 use std::ptr::NonNull;
 
+use libc::c_char;
+use pg_sys::{Datum, Oid};
+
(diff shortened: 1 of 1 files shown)


workingjubilee:
Thom is in the middle of reworking adjacent code in this API so I'm tagging him in to make sure this doesn't merge conflict.

thomcc (approved):
This is fine.

thomcc:
Merging so I can rebase my stuff.

PR #1258, opened 2023-08-14
Add proptest support

workingjubilee:
Our proptest support requires `PgTestRunner`, which shims `TestRunner` in what will be increasingly custom ways, in order to support running inside `#[pg_test]`. 

For most of this PR I focused on extensively documenting expected patterns so that it's easier to understand how to write new ones while minimally referring to the main library's documentation, which is a bit... dense. I had some additional work on adding value trees for complex-to-construct types that I will be following up with in a later PR.

What I thought would also be an "easy sample test" demonstrated we shouldn't have a `From<pg_sys::DateADT>` (an alias for i32) impl for Date. Thus this also breaks public API by removing `From<i32> for Date`, downgrading it to a `TryFrom` that respects the boundaries found by the test's shrinking to simplified case. `Date::saturating_from_raw` is added for accepting the inner type and saturating it to `-infinity` or `infinity` if it is out-of-range for a Date. This is necessary in order to make the "sample" test pass.


pgrx-tests/src/tests/proptests.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// Hypothesis: We can pass random dates directly into Postgres functions and get them back.
+    #[pg_test]
+    pub fn date_spi_roundtrip() {
+        // 2. Constructing the Postgres-adapted test runner
+        let mut proptest = PgTestRunner::default();
+        // 3. A strategy for creating and refining values
+        let strat = prop::num::i32::ANY.prop_map(|i| {
+            Date::try_from(i).unwrap_or(unsafe {
+                Date::from_pg_epoch_days(i32::clamp(i, -2451545, 2147483494 - 2451546))
+            })
```
workingjubilee: This should be extracted into a function on the Date type.

pgrx-tests/src/proptest.rs
```diff
@@ -0,0 +1,48 @@
...
+        test: impl Fn(S::Value) -> TestCaseResult,
+    ) -> Result<(), TestError<<S as Strategy>::Value>> {
+        self.0.run(strategy, |value| {
+            PgTryBuilder::new(AssertUnwindSafe(|| test(value)))
+                .catch_others(|err| match err {
+                    CaughtError::PostgresError(err)
+                    | CaughtError::ErrorReport(err)
+                    | CaughtError::RustPanic { ereport: err, .. } => {
+                        Err(TestCaseError::Fail(err.message().to_owned().into()))
+                    }
```
workingjubilee, line 43: This should be extracted into a function on CaughtError.

pgrx/src/datum/date.rs
```diff
@@ -64,6 +47,32 @@ impl From<TimestampWithTimeZone> for Date {
...
+/// - Some values such as `i32::MIN` and `i32::MAX` have special meanings as infinities
+impl TryFrom<pg_sys::DateADT> for Date {
+    type Error = DateTimeConversionError;
+    #[inline]
+    fn try_from(value: pg_sys::DateADT) -> Result<Self, Self::Error> {
+        if (i32::MIN + 1..-2451545).contains(&value) {
+            Err(DateTimeConversionError::FieldOverflow) // but it's an underflow...
+                                                        // well, underflow isn't right either
+        } else if ((2147483494 - 2451545)..i32::MAX).contains(&value) {
+            Err(DateTimeConversionError::FieldOverflow)
```
workingjubilee: These need proper variants.

pgrx-tests/src/proptest.rs
```diff
@@ -0,0 +1,48 @@
...
+
+impl PgTestRunner {
+    pub fn run<S: Strategy>(
+        &mut self,
+        strategy: &S,
+        test: impl Fn(S::Value) -> TestCaseResult,
+    ) -> Result<(), TestError<<S as Strategy>::Value>> {
+        self.0.run(strategy, |value| {
+            PgTryBuilder::new(AssertUnwindSafe(|| test(value)))
+                .catch_others(|err| match err {
```
workingjubilee, line 38: This experience makes it clear to me that `catch_others` is kind of a weird name, and also that we could probably use a simpler sugared constructor for "please just run one damn closure and catch it".

pgrx-tests/src/tests/proptests.rs
workingjubilee: This is important for the correctness of the changes from the test, so I can't skimp on it before landing this.

pgrx/src/datum/date.rs
workingjubilee: This is important for the correctness of the changes from the test, so I can't skimp on it before landing this.

pgrx-tests/src/proptest.rs
workingjubilee: That can be deferred into a later PR.

pgrx-tests/src/proptest.rs
workingjubilee: That can be deferred into a later PR.

workingjubilee:
Yep, Mac issue seems resolved.

pgrx/src/datum/date.rs
eeeebbbbrrrr: Where do those magic numbers come from?  Are they Date bounds?  Is there a const in pg_sys for them?

eeeebbbbrrrr (approved):
This looks good to me. I am super jazzed to have this foundation in place. Plus it already found issues. Sooo good!

pgrx/src/datum/date.rs
workingjubilee: `DATE_END_JULIAN` and `POSTGRES_EPOCH_JDATE`, so yeah.

workingjubilee: Added `DatetimeConversionError::OutOfRange` to address this.

thomcc (approved):
Looks great!

PR #1261, opened 2023-08-16
Assume commutation when deriving PostgresEq

workingjubilee:
This fixes https://github.com/pksunkara/pgx_ulid/pull/16 without requiring additional work on the part of the programmer. However, there is a design decision here: do we require explicit annotation before assuming commutative equality or not?

For this version of this branch, I chose the route of applying stronger trait bounds to demand that the equality implementation adheres to the Rust requirements for a "total" equality. However, as the SQL we emit is used deep inside Postgres for various forms of algebraic reasoning, including, as I understand it, index maintenance, I can't promise that we don't wind up running afoul of Postgres using this data in a way that will incur UB if the implementation is actively malicious, as can be done with implementations of the following kind:

```rust
impl PartialEq for AlwaysRandom {
    fn eq(&self, other: &Self) -> bool {
        rand::random()
    }
}

impl Eq for AlwaysRandom { }
```

Note, however, this was always true, as we emitted negator annotations, which means even this was probably already broken:

```rust
impl PartialEq for AlwaysWrong {
    fn eq(&self, other: &Self) -> bool {
        false
    }
}

impl Eq for AlwaysWrong {}
```


workingjubilee:
Oh, we also emit the PARALLEL SAFE and IMMUTABLE annotations, too!

thomcc (approved):
This makes sense to me.

osdiab:
Hello, I wanted to check if something is blocking a release from being pushed including this change - https://github.com/pksunkara/pgx_ulid/pull/16 is waiting on this to be deployed, and until then joins don't work predictably with `pgx_ulid`. Sorry for the urgency but it's blocking progress at my company so I'm really looking forward for this to be deployed. Thanks!

osdiab:
tagging @eeeebbbbrrrr on this since it seems like you maintain this library—thanks for the great contributions.

workingjubilee:
I am capable of cutting releases, and @eeeebbbbrrrr is still on vacation.

osdiab:
Great—is there anything the community can do to help unblock a release?

workingjubilee:
At the current moment, the only delay is that I am trying to reproduce https://github.com/pgcentralfoundation/pgrx/pull/1263 to determine whether it should go into the release.

workingjubilee:
Do you only depend on ulid, or are there other pgrx-based extensions in use?

osdiab:
Just `ulid`, yeah it's just that joins of depth greater than 1 fail every time so basically either we wait this out or we stop using `ulid`—but since a fix is already here i'd rather the former!

PR #1262, opened 2023-08-17
Demand Ord for PostgresOrd

workingjubilee:
Previously, the cmp invocation could be misinterpreted as Iterator.
Code exploiting this bug is unlikely, but it could hit other errors,
so this way the error is much nicer.

Also makes the previous trybuild test for PostgresEq more realistic.

### DIFF
--- a/pgrx-tests/tests/ui/total_ord_for_postgres_ord.rs
+++ b/pgrx-tests/tests/ui/total_ord_for_postgres_ord.rs
@@ -1,0 +1,45 @@
+use core::cmp::Ordering;
+use pgrx::prelude::*;
+use serde::{Deserialize, Serialize};
+
+#[derive(Copy, Clone, Serialize, Deserialize, PartialEq, PartialOrd, PostgresType, PostgresOrd)]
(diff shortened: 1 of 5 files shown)


pgrx-tests/tests/ui/total_ord_for_postgres_ord.rs
```diff
@@ -0,0 +1,45 @@
...
+
+pub struct BrokenIter {}
+
+impl Iterator for BrokenIter {
+    type Item = i32;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        todo!()
+    }
+}
```
workingjubilee: This entire repro could be greatly simplified by simply implementing `fn cmp`.

thomcc:
I think it could probably happen without any traits, just with an inherent `fn cmp(&self, o: &Self)`, no?

thomcc (approved):

workingjubilee:
The LHS didn't even need to be &self, so this could have been a mutating comparison hypothetically!

Issue #1257, opened 2023-08-10
Investigate if `fncollation` provides enough info for text encoding checks


workingjubilee:
FunctionCallInfoBaseData has this field and it contains orderings like "C", "POSIX", language-specific orderings, etc. It seems like this might provide info about what sort of text input that Postgres is expecting. I can't look at it right now but it might be useful to inspect later.

workingjubilee:
Examined closer:

It is not _very_ useful for this purpose. It only provides a substitute collation if the database's current collation is not suitable for the query. However, a given build of Postgres has enough firepower in text conversion utilities to transcode to and from UTF-8.

PR #1266, opened 2023-08-20
WIP: Initial work on a dynamic function call interface.  

eeeebbbbrrrr:
I was bored while on vacation, so I thought I'd start some code around being able to dynamically call any function that can be called from SQL.  

The implementation details are that this resolves the function by its name and argument types and calls it directly.  There's no Spi overhead here.  There is some allocation overhead and catalog access overhead, but that's insignificant when compared to everything Spi does.  Any function in `pg_catalog.pg_proc` can be called.

As the TODOs indicate, there's still more work to do.

We can bikeshed the API and names and such, but the general thought process is:

```sql
CREATE FUNCTION some_function(int4, text, bool[]) 
RETURNS text 
STRICT 
LANGUAGE sql AS $$ 
   SELECT '<whatever>' 
$$;
```

```rust
let result = pgrx::fcall::<&str>(
   "some_function",   // can be schema-qualified
   &[&Some(1i32), &Some("foo"), &Some(vec![true, true, false])]
);
assert_eq!(result, "<whatever>");
```

I'm happy for input now while it's still WIP, rather than later when it's mostly complete as I envision it.  We need this capability for PL/Rust so we need to make sure we get the user-facing API correct, at least.

Some things I don't think we should support in the first iteration of this:

- Postgres v11 (it EOL's in November -- not worth the extra code)
- Calling set-returning-functions (`RETURNS SET OF t` and `RETUNS TABLE (...)`)
- Functions with OUT parameters
- Automatically performing type casting/coersion for function arguments

Some things I need to figure out:

- Calling a function by Oid needs a collation Oid for any function that has TEXT arguments -- where does this come from?
- I think `IntoDatum::into_datum(self)` should actually be `IntoDatum::as_datum(&self)` (or whatever).  The point is that converting a Rust value into a Datum doesn't need to consume the Rust value.  Not sure how this interacts with what @thomcc is working on

### DIFF
--- a/pgrx/src/fcall.rs
+++ b/pgrx/src/fcall.rs
@@ -1,0 +1,143 @@
+//LICENSE Portions Copyright 2019-2021 ZomboDB, LLC.
+//LICENSE
+//LICENSE Portions Copyright 2021-2023 Technology Concepts & Design, Inc.
+//LICENSE
+//LICENSE Portions Copyright 2023-2023 PgCentral Foundation, Inc. <contact@pgcentral.org>
+//LICENSE
+//LICENSE All rights reserved.
+//LICENSE
+//LICENSE Use of this source code is governed by the MIT license that can be found in the LICENSE file.
+use crate::{
+    direct_function_call, list::PgList, pg_sys, pg_sys::AsPgCStr, Array, FromDatum, IntoDatum,
+};
+
+pub trait FCallArg {
+    fn as_datum(&self) -> Option<pg_sys::Datum>;
+    fn type_oid(&self) -> pg_sys::Oid;
+}
+
+impl<T: IntoDatum + Clone> FCallArg for Option<T> {
+    fn as_datum(&self) -> Option<pg_sys::Datum> {
+        // TODO:  would prefer not to need `Clone`, but that requires changes to `IntoDatum`
+        self.as_ref().map(|v| Clone::clone(v).into_datum()).flatten()
+    }
+
+    fn type_oid(&self) -> pg_sys::Oid {
+        T::type_oid()
+    }
+}
+
+// TODO:  This should return `Result<Option<T>, FCallError>`
+//        We'll want to trap the common errors like ERRCODE_UNDEFINED_FUNCTION and ERRCODE_AMBIGUOUS_FUNCTION
+//        along with making some of our own error types
+pub fn fcall<T: FromDatum>(fname: &str, args: &[&dyn FCallArg]) -> Option<T> {
+    let mut arg_types = Vec::with_capacity(args.len());
+    let mut arg_datums = Vec::with_capacity(args.len());
+    for (oid, datum) in args.iter().map(|a| (a.type_oid(), a.as_datum())) {
+        arg_types.push(oid);
+        arg_datums.push(datum);
+    }
+    unsafe {
+        // let Postgres parse the function name -- it could be schema-qualified and Postgres knows
+        // the parsing rules better than we do
+        let ident_parts = direct_function_call::<Array<&str>>(
+            pg_sys::parse_ident,
+            &[fname.into_datum(), true.into_datum()],
+        );
+
+        // convert those into a PgList of each part as a `pg_sys::String`
+        let func_oid = {
+            // TODO:  `PgList` requires the "cshim" feature, which is **not** turned on for PL/Rust
+            //         I think we'll need to fully port Postgres' "List" type to Rust...
+            let mut parts_list = PgList::new();
+            ident_parts
+                .unwrap()
+                .iter_deny_null()
+                .map(|part| pg_sys::makeString(part.as_pg_cstr()))
+                .for_each(|part| parts_list.push(part));
+
+            // ask Postgres to find the function.  This will look for the possibly-qualified named
+            // function following the normal SEARCH_PATH rules, ensuring its argument type Oids
+            // exactly match the ones from the user's input arguments.  It does not evaluate the
+            // return type, so we'll have to do that later
+            let func_oid = pg_sys::LookupFuncName(
+                parts_list.as_ptr(),
+                args.len().try_into().unwrap(),
+                arg_types.as_ptr(),
+                false,
+            );
+
+            // free the individual `pg_sys::String` parts we allocated above
+            parts_list.iter_ptr().for_each(|s| {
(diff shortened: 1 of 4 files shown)


workingjubilee:
I don't think we should expose a (safe) non-consuming method of generating Datums unless we go through the entire PGRX API surface and make it so every way to implicitly do a reverse conversion from a Datum is `unsafe`. It's dubious enough as-is, but the current way makes double-frees at least less-likely.

pgrx/src/fcall.rs
```diff
@@ -0,0 +1,143 @@
...
+//LICENSE Portions Copyright 2023-2023 PgCentral Foundation, Inc. <contact@pgcentral.org>
+//LICENSE
+//LICENSE All rights reserved.
+//LICENSE
+//LICENSE Use of this source code is governed by the MIT license that can be found in the LICENSE file.
+use crate::{
+    direct_function_call, list::PgList, pg_sys, pg_sys::AsPgCStr, Array, FromDatum, IntoDatum,
+};
+
+pub trait FCallArg {
```
workingjubilee, line 14: Incorrect implementations of this could lead to UB, even for something as simple as just lying about the OID, thus it must be

```suggestion
pub unsafe trait FCallArg {
```

pgrx/src/fcall.rs
```diff
@@ -0,0 +1,143 @@
...
+    fn as_datum(&self) -> Option<pg_sys::Datum>;
+    fn type_oid(&self) -> pg_sys::Oid;
+}
+
+impl<T: IntoDatum + Clone> FCallArg for Option<T> {
```
workingjubilee, line 19:
```suggestion
unsafe impl<T: IntoDatum + Clone> FCallArg for Option<T> {
```

pgrx/src/fcall.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    fn type_oid(&self) -> pg_sys::Oid {
+        T::type_oid()
+    }
+}
+
+// TODO:  This should return `Result<Option<T>, FCallError>`
+//        We'll want to trap the common errors like ERRCODE_UNDEFINED_FUNCTION and ERRCODE_AMBIGUOUS_FUNCTION
+//        along with making some of our own error types
+pub fn fcall<T: FromDatum>(fname: &str, args: &[&dyn FCallArg]) -> Option<T> {
```
workingjubilee, line 33: We recently had an interesting report that returning Results leads to worse UX in some cases than panicking, but I am not sure exactly what the context was. Not actionable yet.

pgrx/src/fcall.rs
```diff
@@ -0,0 +1,143 @@
...
+        //  - IN/OUT/TABLE arg types: I think we should only support IN in v1 of this
+        //  - Make sure the return type `T` is compatible with the resolved function's return type
+        //  - ??
+        //
+
+        //
+        // The following code is Postgres-version specific.  Right now, it's compatible with v12+
+        // v11 will need a different implementation.
+        //
+        // NB:  Which I don't want to do since it EOLs in 3 months
```
workingjubilee, line 99: I support your laziness.

pgrx/src/fcall.rs
```diff
@@ -0,0 +1,143 @@
...
+        // for `args.len()` arguments
+        let fcinfo = pg_sys::palloc0(
+            std::mem::size_of::<pg_sys::FunctionCallInfoBaseData>()
+                + std::mem::size_of::<pg_sys::NullableDatum>() * args.len(),
+        ) as *mut pg_sys::FunctionCallInfoBaseData;
+
+        // initialize it
+        let fcinfo_ref = fcinfo.as_mut().unwrap();
+        fcinfo_ref.flinfo = &mut flinfo;
+        fcinfo_ref.fncollation = pg_sys::InvalidOid; // TODO:  We need to get a collation from somewhere?
```
workingjubilee, line 116: Ah, I have an excuse to follow up on this tomorrow:
- https://github.com/pgcentralfoundation/pgrx/issues/1257

pgrx/src/fcall.rs
workingjubilee: However:
```suggestion
// TODO:  This should return `Result<Option<T>, FCallError>`
//        We'll want to trap the common errors like ERRCODE_UNDEFINED_FUNCTION and ERRCODE_AMBIGUOUS_FUNCTION
//        along with making some of our own error types
#[cfg(not(feature = "pg11"))]
pub fn fcall<T: FromDatum>(fname: &str, args: &[&dyn FCallArg]) -> Option<T> {
```

eeeebbbbrrrr:
> I don't think we should expose a non-consuming method of generating Datums unless we go through the entire PGRX API surface and make it so every way to implicitly do a reverse conversion from a Datum is `unsafe`. It's dubious enough as-is, but the current way makes double-frees at least less-likely.

Yeah. We need some definitions around what it means to convert a Rust value into a Datum. 

For wholly owned values I think it’s obvious — possibly palloc the equivalent Postgres type. But for borrowed values like Array, I am not sure… reuse the already-palloc’d stuff via consumption or allocate a copy?

pgrx/src/fcall.rs
eeeebbbbrrrr: > We recently had an interesting report that returning Results leads to worse UX in some cases than panicking, but I am not sure exactly what the context was. Not actionable yet.

I saw that in passing. If it’s not reporting nicely that’s our fault, not specifically the fault of Result.

pgrx/src/fcall.rs
eeeebbbbrrrr: If IntoDatum didn’t consume then we wouldn’t need this trait at all.

workingjubilee:
I suspect that we will wind up in cases where we want to do either (since one is a clone and one is a borrow, basically) and make sure we provide separate implementations for each.

pgrx/src/fcall.rs
workingjubilee: The answer to that issue's question is no. But also the answer to "what do we do here?" seems to be that, in most cases, the "direct function call" frameworks inside Postgres either explicitly pass a collation or assume InvalidOid is an acceptable input collation. There is a TODO to use `DEFAULT_COLLATION_OID` in a few such places, however.

PR #1271, opened 2023-08-24
Better error messages when `pg_config` isn't found.

eeeebbbbrrrr:
We do a little better job letting the user know why `pg_config` couldn't be found.
Screenshot 2023-08-24 at 10 43 34 AM

### DIFF
--- a/pgrx-pg-config/src/lib.rs
+++ b/pgrx-pg-config/src/lib.rs
@@ -417,7 +417,18 @@ impl PgConfig {
                 Ok(output) => Ok(String::from_utf8(output.stdout).unwrap().trim().to_string()),
                 Err(e) => match e.kind() {
                     ErrorKind::NotFound => Err(e).wrap_err_with(|| {
-                        format!("Unable to find `{}` on the system $PATH", "pg_config".yellow())
+                        let pg_config_str = pg_config.display().to_string();
+
+                        if pg_config_str == "pg_config" {
+                            format!("Unable to find `{}` on the system $PATH", "pg_config".yellow())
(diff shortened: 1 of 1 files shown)


pgrx-pg-config/src/lib.rs
```diff
@@ -417,7 +417,18 @@ impl PgConfig {
                 Ok(output) => Ok(String::from_utf8(output.stdout).unwrap().trim().to_string()),
                 Err(e) => match e.kind() {
                     ErrorKind::NotFound => Err(e).wrap_err_with(|| {
-                        format!("Unable to find `{}` on the system $PATH", "pg_config".yellow())
+                        let pg_config_str = pg_config.display().to_string();
+
+                        if pg_config_str == "pg_config" {
+                            format!("Unable to find `{}` on the system $PATH", "pg_config".yellow())
+                        } else if pg_config_str.contains('~') {
+                            format!("The specified pg_config binary, `{}`, does not exist.  It contains a `~`, indicating your shell didn't properly expand your home directory", pg_config_str.yellow())
```
workingjubilee: This should only happen initially in the path, so we should use `starts_with` instead of `contains`, and soften the wording slightly.

eeeebbbbrrrr: - "... indicating computers were a mistake"?
- "... indicating the shell failed to expand it"?
- "The specified pg_config binary, `{}`, does not exist.  The shell didn't expand the `~`"?  <-- this

workingjubilee: Either seems good!

workingjubilee (approved):
Great!

PR #1273, opened 2023-08-24
Fix GetMemoryChunkContext port

workingjubilee:
- The name was needlessly backwards-ish.
- `MemoryContextIsValid` performed a potentially-UB read.


workingjubilee:
Ah, Thom informs me the read wasn't actually that much of a problem: although the alignment question is still slightly undecided, it is much more close to decided than other opsem questions, and it is probably going to come down as not concerned with this. The documentation still needed to be clearer, however, I think.

PR #1274, opened 2023-08-30
Memoize git hash and extension metadata

levkk:
Copy extension update files 100x faster.

### DIFF
--- a/cargo-pgrx/src/command/install.rs
+++ b/cargo-pgrx/src/command/install.rs
@@ -25,7 +25,7 @@ use std::sync::{Arc, Mutex};
 
 /// Type used for memoizing expensive to calculate values.
 /// Arc<Mutex> is needed to get around compiler safety checks.
-type MemoizeKeyValue = Arc<Mutex<HashMap<String, String>>>;
+type MemoizeKeyValue = Arc<Mutex<HashMap<PathBuf, String>>>;
 
 /// Install the extension from the current crate to the Postgres specified by whatever `pg_config` is currently on your $PATH
 #[derive(clap::Args, Debug)]
@@ -404,7 +404,7 @@ pub(crate) fn find_library_file(
 static CARGO_VERSION: Lazy<MemoizeKeyValue> = Lazy::new(|| Arc::new(Mutex::new(HashMap::new())));
 
 pub(crate) fn get_version(manifest_path: impl AsRef<Path>) -> eyre::Result<String> {
-    let path_string = manifest_path.as_ref().display().to_string();
+    let path_string = manifest_path.as_ref().to_owned();
 
     if let Some(version) = CARGO_VERSION.lock().unwrap().get(&path_string) {
         return Ok(version.clone());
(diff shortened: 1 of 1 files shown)


thomcc (approved):
We could probably use RwLock if there's ever contention here, but there shouldn't be for now. This basically looks fine to me. Thanks.

cargo-pgrx/src/command/install.rs
```diff
@@ -13,12 +13,19 @@ use crate::profile::CargoProfile;
...
+use std::collections::HashMap;
 use std::fs;
 use std::io::BufReader;
 use std::path::{Path, PathBuf};
 use std::process::Stdio;
+use std::sync::{Arc, Mutex};
+
+/// Type used for memoizing expensive to calculate values.
+/// Arc<Mutex> is needed to get around compiler safety checks.
+type MemoizeKeyValue = Arc<Mutex<HashMap<String, String>>>;
```
workingjubilee: I don't think it matters in practice, and this version is more amenable to us later threadpooling the IO if it gets truly dreadful, but we could alternatively use https://doc.rust-lang.org/stable/std/macro.thread_local.html to defeat the borrow-checker's concerns.

levkk: I'm down to try, I've never used thread_locals before.

cargo-pgrx/src/command/install.rs
```diff
@@ -395,8 +401,16 @@ pub(crate) fn find_library_file(
     Ok(library_file_path)
 }
 
+static CARGO_VERSION: Lazy<MemoizeKeyValue> = Lazy::new(|| Arc::new(Mutex::new(HashMap::new())));
+
 pub(crate) fn get_version(manifest_path: impl AsRef<Path>) -> eyre::Result<String> {
-    match get_property(&manifest_path, "default_version")? {
+    let path_string = manifest_path.as_ref().display().to_string();
```
thomcc: Not really the best way to do this but we do it everywhere in this code so it's fine.

levkk: Should I use `PathBuf`?

cargo-pgrx/src/command/install.rs
thomcc: It's probably better to take a lock for the duration of a hashmap read than to have a thread do IO we've already done on another thread.

workingjubilee: Yeah, I think this is fine here, I was only noting the alternate implementation choice because of my pedagogic compulsion.

cargo-pgrx/src/command/install.rs
thomcc: ~~I think its fine. I wouldn't have approved it if I thought it needed to change. We do this all over cargo-pgrx anyway.~~

Well, you did it anyway while I wrote this, so it's hard to complain about it (if nothing else it's clearer as PathBuf).

workingjubilee: That's great, thank you!

workingjubilee (approved):

PR #1275, opened 2023-09-01
spi: Add support for specifying mutating mode to PreparedStatement

XeniaLu:
* Add support for specifying `mutating` flag to PreparedStatement.
 * Add `prepare_mut` function that execute the mutating query.

### DIFF
--- a/pgrx/src/spi/query.rs
+++ b/pgrx/src/spi/query.rs
@@ -148,6 +148,7 @@ impl<'conn> Query<'conn> for &str {
 pub struct PreparedStatement<'conn> {
     pub(super) plan: NonNull<pg_sys::_SPI_plan>,
     pub(super) __marker: PhantomData<&'conn ()>,
+    pub(super) mutating: bool,
 }
 
 /// Static lifetime-bound prepared statement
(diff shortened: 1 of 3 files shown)


eeeebbbbrrrr:
Thanks for the PR.  We're doing some overhauls to Spi and I'm not sure how this kind of change fits into those changes.  I'll leave this for @thomcc as he's the one leading that work.

I see the need for this change so we'll make sure we get the functionality covered somehow.

pgrx/src/spi/query.rs
```diff
@@ -228,6 +233,9 @@ impl<'conn: 'stmt, 'stmt> Query<'conn> for &'stmt PreparedStatement<'conn> {
         limit: Option<libc::c_long>,
         arguments: Self::Arguments,
     ) -> SpiResult<SpiTupleTable<'conn>> {
+        if self.mutating {
+            Spi::mark_mutable();
+        }
```
workingjubilee: My understanding is Postgres will do this for us, and we can instead use our knowledge to actually skip some FFI overhead on our side, so we should, instead of this, make our bool passed to Postgres be the expr `self.mutating || Spi::is_xact_still_immutable()` on line 260.

pgrx/src/spi/query.rs
```diff
@@ -258,6 +266,9 @@ impl<'conn: 'stmt, 'stmt> Query<'conn> for &'stmt PreparedStatement<'conn> {
     }
 
     fn open_cursor(self, _client: &SpiClient<'conn>, args: Self::Arguments) -> SpiCursor<'conn> {
+        if self.mutating {
+            Spi::mark_mutable();
+        }
```
workingjubilee: Same comment as the other case.

workingjubilee (commented):
Hello, sorry about the extremely long turnaround time on review, I wound up taking point on following up on Thom's work.

workingjubilee (approved):
Thanks!

PR #1277, opened 2023-09-02
Introduce pure-Rust replacement for PgList

workingjubilee:
I'm allowing the original PgList type to continue to exist for code that might still be relying on PgList for a bit longer and because this way we can have a deprecation period for the original type.

The high level overview:
- `*mut pg_sys::List` is used like a peculiar cousin of a `Vec<pg_sys::ListCell>` since Postgres 13. Before that it was a true linked list. This is wrapped by `enum pgrx::list::List<T>` which is really a funny-looking `Option<ListHead<T>>`.
- ...`ListHead<T>` is itself a funny-looking `NonNull<pg_sys::List>`.
- These allow obtaining `{&,&mut} [ListCell<T>]`, a slice of strongly-typed `pg_sys::ListCell`s.
- `unsafe trait Enlist` describes a union variant of `pg_sys::ListCell`.
- All ways to poke inside a `ListCell<T>` are bounded by `T: Enlist`.
- All ways to get a `List<T>` require `unsafe` assertions about the List's validity for various things... unless it's a null pointer, oddly enough. That is, `List::Nil` is publicly constructible quite on purpose, and otherwise you need to construct a `ListHead<T>` first (only via an `unsafe fn`), or you can't build the other variant of `List<T>`.

~~This is currently bounded by Postgres version~~ Fixed that.

In order to obtain some coherent ergonomics for `pg_sys::NodeTag`, it is now a Rust enum. This is kosher because **all node tag values can change when you add even one variant**, so it is not quite as unstable as a *true* Rust enum, but it *is* incoherent if you aren't built directly against the EXACT same version of Postgres. They are NOT stable from version to version, except, fortunately, for the zeroed variant (`NodeTag::T_Invalid`).


workingjubilee:
I still need to port the tests currently using PgList over, so that'll be fun. For what it's worth, I basically half am putting this up so I can see if ZomboDB can be persuaded to build with it, since ZomboDB has more uses of PgList than our entire test suite.

workingjubilee:
There's a bunch of other API that should be stamped out on this for Postgres parity, suffice to say that the main holdup was that `list::Drain` provides the remaining basis for any of the most interesting API in Postgres.

pgrx/src/lib.rs
```diff
@@ -97,7 +96,7 @@ pub use htup::*;
 pub use inoutfuncs::*;
 pub use itemptr::*;
 #[cfg(feature = "cshim")]
-pub use list::*;
+pub use list::old_list::*;
```
eeeebbbbrrrr, line 99: I’m reading top-down so apologies if it’s settled elsewhere… old_list should be behind the cshim feature.

pgrx/src/list.rs
eeeebbbbrrrr, line 1: Woah. Okay. This is gonna require some coffee. Goodness

eeeebbbbrrrr (commented):
@workingjubilee This is incredible. You now probably understand Postgres’ List impl better than the -hackers do!

I need to spend more time with this and I will after the long weekend.

pgrx/src/list.rs
```diff
@@ -12,257 +12,617 @@
...
+
+        #[inline]
+        pub fn capacity(&self) -> usize {
+            match self {
+                List::Nil => 0,
+                List::Cons(head) => head.capacity(),
+            }
+        }
+
+        pub fn into_nullable(self) -> *mut pg_sys::List {
```
eeeebbbbrrrr: I don't understand "nullable" here.  At the end of the day, this is returning the underlying list pointer.

I was looking for the function to "get the raw List pointer so it can be passed off to Postgres" and this looks like function, but the name doesn't make sense to me.

workingjubilee: It's a nullable i.e. maybe-null pointer.

In the rest of pgrx, we often rely on Postgres returning non-null pointers, and this one never has such a guarantee.

eeeebbbbrrrr: pgrx has historically used `into_pg` for this and `from_pg` for `downcast_from_nullable`.  I guess that's what I was looking for.

I'm not arguing or anything.  This function name and also `downcast_from_nullable` just aren't what I was expecting.

workingjubilee: I could use `downcast_from_pg` but it's a "true" downcast in the sense that it's fallible.

pgrx/src/list.rs
```diff
@@ -12,257 +12,617 @@
...
+            // interpret the cell as a pointer, assuming pointers are Pod
+            let cell_addr = unsafe { self.cell.ptr_value };
+            f.debug_tuple("ListCell").field(&cell_addr).finish()
         }
     }
 
-    pub unsafe fn from_pg(list: *mut pg_sys::List) -> Self {
-        PgList { list, allocated_by_pg: true, _marker: PhantomData }
+    // Note this well: the size of a `ListCell<T>`'s type doesn't matter.
+    // Thus it isn't acceptable to implement Enlist for types larger than `pg_sys::ListCell`.
```
workingjubilee:
```suggestion
    // Note: the size of `ListCell<T>`'s generic `T` doesn't matter,
    // thus it isn't acceptable to implement Enlist for a `T` larger than `pg_sys::ListCell`.
```

pgrx/src/list.rs
workingjubilee: In practice, pgrx has used inconsistent names in a few places, e.g. https://github.com/pgcentralfoundation/pgrx/blob/047b1d1fc9e9c4007c871e226fa81e294f8bf5e6/pgrx/src/datum/array.rs#L161

and in my head it adopts an opposite polarity as e.g. [`as_non_null_ptr`](https://doc.rust-lang.org/std/ptr/struct.NonNull.html#method.as_non_null_ptr).

To me, the `from_pg` and `into_pg` names imply strongly the type is actually *from Postgres* or *going into Postgres*, but in actuality we have people composing usage of pgrx with *other active C code that also binds against Postgres*. This makes me uneasy with the idea that this other miscellaneous C code, which may also deal in Postgres types, should be treated as "inside the trust boundary" in the same way that Postgres itself *necessarily* is, and uneasy with names that may undermine someone's caution.

This, all together (alongside my gradually crumbling sanity as I tried to understand Postgres) is why I prioritized my personal taste when initially opening the PR. But, despite all those factors in tension, I don't necessarily feel married to their current names, especially if they feel non-intuitive. The boring conventional `into_ptr` and `downcast_ptr` would be fine.

pgrx/src/list.rs
```diff
@@ -12,257 +12,617 @@
...
+                unsafe { grow_list(list, size + list.length as usize) };
+            };
+            self
         }
     }
 
-    #[inline]
-    pub fn iter_ptr(&self) -> impl Iterator<Item = *mut T> + '_ {
-        PgListIteratorPtr { list: &self, pos: 0 }
+    unsafe fn grow_list(list: &mut pg_sys::List, target: usize) {
```
eeeebbbbrrrr: This doesn't cover the "debug" branches that the Postgres version has (DEBUG_LIST_MEMORY_USAGE) but I think that's okay.

pgrx/src/list.rs
```diff
@@ -12,257 +12,617 @@
...
+                panic!("Context free list?");
+            }
+            let buf = pg_sys::MemoryContextAlloc(context, alloc_size);
+            if buf == ptr::null_mut() {
+                panic!("List allocation failure");
+            }
+            ptr::copy_nonoverlapping(list.elements, buf.cast(), list.length as _);
+            // If the old buffer is pointers, we would like everyone dereferencing them to segfault,
+            // if OIDs, Postgres will surface errors quickly on InvalidOid, etc.
+            ptr::write_bytes(list.elements, 0, list.length as _);
```
eeeebbbbrrrr: However, Postgres only does this when `CLOBBER_FREED_MEMORY` is defined.  How faithful to their version do we want to be?  I suspect they don't want it in --release builds and maybe we don't either?  

For reference: https://github.com/postgres/postgres/blob/a899d07b24a6e519790041eeea5d96cc534f384e/src/backend/nodes/list.c#L196

pgrx/src/list/old_list.rs
```diff
@@ -0,0 +1,254 @@
...
+    allocated_by_pg: bool,
+    _marker: PhantomData<T>,
+}
+impl<T> Default for PgList<T> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<T> PgList<T> {
```
eeeebbbbrrrr, line 15: The person who made this should be ashamed.

pgrx/src/list.rs
eeeebbbbrrrr: I don't want us to get bogged down in naming, especially as you point out, we're pretty inconsistent.  So I'll let you decide what you think is right and would be most expected by a random Rust programmer.  I didn't find these names obvious, but my opinions are biased.

eeeebbbbrrrr (approved):
Jubilee, this looks great.  Super great.

We do need some tests around this, but they can come in a separate PR if you prefer.

pgrx/src/list.rs
workingjubilee: I would prefer to have one consistent approach throughout, basically. I think a short memset is unlikely to be a performance concern.

workingjubilee: basically the reason to always do it is so we don't have to bind the valgrind headers, and we may still surface some bugs this way, because they use `VALGRIND_MAKE_MEM_NOACCESS`. I don't believe we have a good way to do that while keeping everything "pure Rust". Maybe I'm wrong, @thomcc?

workingjubilee:
Contemplating moving this into a `node` module, however there are things that would go into `node` that we would definitely want to split out, probably, like `param`, so maybe not. Yeah, already talked myself out of it: the `List` type is used for not-immediately-relevant-to-nodes reasons, like tuples. The organization of the code in Postgres is purely historical.

pgrx/src/list/old_list.rs
```diff
@@ -0,0 +1,254 @@
+use crate::{is_a, pg_sys, void_mut_ptr};
+use std::marker::PhantomData;
+
+pub struct PgList<T> {
+    list: *mut pg_sys::List,
+    allocated_by_pg: bool,
+    _marker: PhantomData<T>,
+}
+impl<T> Default for PgList<T> {
+    fn default() -> Self {
```
thomcc, line 10: nit: inline

pgrx/src/list.rs
thomcc: Don't clobber memory on release builds, probably.

workingjubilee: I corrected the clobber pattern and made it a `#[cfg(debug_assertions)]` thing!

workingjubilee:
PgList was... untested. I couldn't find tests that were easy to simply update to the new version without reworking unrelated API, or that actually tested the powers of PgList. As my goal was to improve on the state of PgList, and because this is blocking #1279 now, I'm going to merge this as-is!

Caveat emptor!

workingjubilee:
Seriously, GitHub? Your large diff threshold is below 500 lines?

workingjubilee:
The break here is the NodeTag change, not anything else.

PR #1279, opened 2023-09-05
Dynamic function call

eeeebbbbrrrr:
(This replaces #1266 because I don't know how to use git.)

I was bored while on vacation, so I thought I'd start some code around being able to dynamically call any function that can be called from SQL.  

The implementation details are that this resolves the function by its name and argument types and calls it directly.  There's no Spi overhead here.  There is some allocation overhead and catalog access overhead, but that's insignificant when compared to everything Spi does.  (most) Any function in `pg_catalog.pg_proc` can be called.

We can bikeshed the API and names and such, but the general thought process is:

```sql
CREATE FUNCTION some_function(int4, text, bool[]) 
RETURNS text 
STRICT 
LANGUAGE sql AS $$ 
   SELECT '<whatever>' 
$$;
```

```rust
let result = pgrx::fcall::<&str>(
   "some_function",   // can be schema-qualified
   &[&Arg::Value(1i32), &Arg::Value("foo"), &Arg::Value(vec![true, true, false])]
);
assert_eq!(result, "<whatever>");
```

I'm happy for input now while it's still WIP, rather than later when it's mostly complete as I envision it.  We need this capability for PL/Rust so we need to make sure we get the user-facing API correct, at least.

Some things I don't think we should support in the first iteration of this:

- Postgres v11 (it EOL's in November -- not worth the extra code)
- Calling set-returning-functions (`RETURNS SET OF t` and `RETUNS TABLE (...)`)
- Functions with OUT parameters
- Automatically performing type casting/coersion for function arguments
- Named arguments

Some things I need to figure out:

- ~~Calling a function by Oid needs a collation Oid for any function that has TEXT arguments -- where does this come from?~~
- ~~I think `IntoDatum::into_datum(self)` should actually be `IntoDatum::as_datum(&self)` (or whatever).  The point is that converting a Rust value into a Datum doesn't need to consume the Rust value.  Not sure how this interacts with what @thomcc is working on~~ Doing a Clone is fine for now.

### DIFF
--- a/pgrx/src/fcall.rs
+++ b/pgrx/src/fcall.rs
@@ -105,7 +105,8 @@ pub fn fcall<T: FromDatum + IntoDatum>(fname: &str, args: &[&dyn FCallArg]) -> R
     // if the function is STRICT and at least one of our argument values is `None` (ie, NULL)...
     // we must return `None` now and not call the function.  Passing a NULL argument to a STRICT
     // function will likely crash Postgres
-    if pg_proc.proisstrict() && arg_datums.iter().any(|d| d.is_none()) {
+    let isstrict = pg_proc.proisstrict();
+    if isstrict && arg_datums.iter().any(|d| d.is_none()) {
         return Ok(None);
     }
 
@@ -140,7 +141,7 @@ pub fn fcall<T: FromDatum + IntoDatum>(fname: &str, args: &[&dyn FCallArg]) -> R
         // setup the argument array
         let args_slice = fcinfo_ref.args.as_mut_slice(args.len());
         for (i, datum) in arg_datums.into_iter().enumerate() {
-            assert!(datum.is_some()); // no NULL datums here, please
+            assert!(!isstrict || (isstrict && datum.is_some())); // no NULL datums if this function is STRICT
 
             let arg = &mut args_slice[i];
             (arg.value, arg.isnull) =
--- a/pgrx/src/fcall.rs
+++ b/pgrx/src/fcall.rs
@@ -141,7 +141,14 @@ pub type Result<T> = std::result::Result<T, FCallError>;
 /// and will immediately abort the transaction.  To catch such errors yourself, use [`fcall`] with
 /// [`PgTryBuilder`].
 ///
-/// # Example
+/// # Limitations
+///
+/// Currently, [`fcall`] does not support:
+///     - Functions that `RETURNS SET OF $type` or `RETURNS TABLE(...)`
+///     - Functions with `IN_OUT` or `OUT` arguments
+///     - Postgres 11
+///
+/// # Examples
 ///
 /// ## Calling a UDF
 ///
--- a/pgrx/src/fcall.rs
+++ b/pgrx/src/fcall.rs
@@ -11,12 +11,12 @@ use crate::{
     direct_function_call, list::PgList, pg_sys, pg_sys::AsPgCStr, Array, FromDatum, IntoDatum,
 };
 
-pub trait FCallArg {
+pub unsafe trait FCallArg {
     fn as_datum(&self) -> Option<pg_sys::Datum>;
     fn type_oid(&self) -> pg_sys::Oid;
 }
 
-impl<T: IntoDatum + Clone> FCallArg for Option<T> {
+unsafe impl<T: IntoDatum + Clone> FCallArg for Option<T> {
     fn as_datum(&self) -> Option<pg_sys::Datum> {
         // TODO:  would prefer not to need `Clone`, but that requires changes to `IntoDatum`
         self.as_ref().map(|v| Clone::clone(v).into_datum()).flatten()
--- a/pgrx/src/fcall.rs
+++ b/pgrx/src/fcall.rs
@@ -7,6 +7,12 @@
 //LICENSE All rights reserved.
 //LICENSE
 //LICENSE Use of this source code is governed by the MIT license that can be found in the LICENSE file.
+
+use pgrx_pg_sys::errcodes::PgSqlErrorCode;
+use pgrx_pg_sys::PgTryBuilder;
+use std::panic::AssertUnwindSafe;
+
+use crate::pg_catalog::pg_proc::{PgProc, ProArgMode, ProKind};
 use crate::{
     direct_function_call, list::PgList, pg_sys, pg_sys::AsPgCStr, Array, FromDatum, IntoDatum,
 };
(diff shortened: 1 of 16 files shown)


pgrx/src/pg_catalog/pg_proc.rs
```diff
@@ -0,0 +1,316 @@
...
+
+impl From<i8> for ProVolatile {
+    fn from(value: i8) -> Self {
+        match value as u8 {
+            b'i' => ProVolatile::Immutable,
+            b's' => ProVolatile::Stable,
+            b'v' => ProVolatile::Volatile,
+
+            // there's just no ability to move forward if given a value that we don't know about
+            _ => panic!("unrecognized `ProVolatile`: `{}`", value),
```
thomcc: nit:
```suggestion
            v => panic!("unrecognized `ProVolatile`: `{:?}`", v as char),
```
since these use ASCII it will probably be easier to debug if it's formatted as a character than a number.

pgrx/src/pg_catalog/pg_proc.rs
```diff
@@ -0,0 +1,316 @@
...
+impl From<i8> for ProKind {
+    fn from(value: i8) -> Self {
+        match value as u8 {
+            b'f' => ProKind::Function,
+            b'p' => ProKind::Procedure,
+            b'a' => ProKind::Aggregate,
+            b'w' => ProKind::Window,
+
+            // there's just no ability to move forward if given a value that we don't know about
+            _ => panic!("unrecognized `ProKind`: `{}`", value),
```
thomcc:
```suggestion
            v => panic!("unrecognized `ProKind`: `{:?}`", v as char),
```

pgrx/src/pg_catalog/pg_proc.rs
```diff
@@ -0,0 +1,316 @@
...
+            b't' => ProArgMode::Table,
+
+            // there's just no ability to move forward if given a value that we don't know about
+            _ => panic!("unrecognized `ProArgMode`: `{}`", value),
+        }
+    }
+}
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub enum ProKind {
```
thomcc, line 47: Should this be `#[non_exhaustive]`?

pgrx/src/pg_catalog/pg_proc.rs
```diff
@@ -0,0 +1,316 @@
...
+use crate::{pg_sys, FromDatum, IntoDatum};
+use std::ptr::NonNull;
+
+/// Provides a safe wrapper around a Postgres "SysCache" entry from `pg_catalog.pg_proc`.
+pub struct PgProc {
+    inner: NonNull<pg_sys::HeapTupleData>,
+}
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub enum ProArgMode {
```
thomcc, line 22: Should this be `#[non_exhaustive]`?

pgrx/src/pg_catalog/pg_proc.rs
```diff
@@ -0,0 +1,316 @@
...
+                pg_proc_oid.into_datum().unwrap(),
+            );
+            let inner = NonNull::new(entry)?;
+            Some(PgProc { inner })
+        }
+    }
+
+    /// Owner of the function
+    pub fn proowner(&self) -> pg_sys::Oid {
+        // SAFETY:  `proowner` has a NOT NULL constraint
```
thomcc: Are these actually safety constraints? If so, `get_attr` should be an `unsafe fn`. Otherwise, you should remove `SAFETY:` from the comment (here and below)

pgrx/src/fcall.rs
```diff
@@ -0,0 +1,218 @@
...
+use pgrx_pg_sys::errcodes::PgSqlErrorCode;
+use pgrx_pg_sys::PgTryBuilder;
+use std::panic::AssertUnwindSafe;
+
+use crate::pg_catalog::pg_proc::{PgProc, ProArgMode, ProKind};
+use crate::{
+    direct_function_call, list::PgList, pg_sys, pg_sys::AsPgCStr, Array, FromDatum, IntoDatum,
+};
+
+pub unsafe trait FCallArg {
```
thomcc: Hmmmmmmmmmm...

pgrx/src/pg_catalog/pg_proc.rs
```diff
@@ -0,0 +1,316 @@
...
+    fn from(value: i8) -> Self {
+        match value as u8 {
+            b'i' => ProArgMode::In,
+            b'o' => ProArgMode::Out,
+            b'b' => ProArgMode::InOut,
+            b'v' => ProArgMode::Variadic,
+            b't' => ProArgMode::Table,
+
+            // there's just no ability to move forward if given a value that we don't know about
+            _ => panic!("unrecognized `ProArgMode`: `{}`", value),
```
thomcc:
```suggestion
            v => panic!("unrecognized `ProArgMode`: `{:?}`", v as char),
```

thomcc (commented):
rough pass, mostly skimming.

pgrx/src/fcall.rs
```diff
@@ -0,0 +1,218 @@
...
+        // for `args.len()` arguments
+        let fcinfo = pg_sys::palloc0(
+            std::mem::size_of::<pg_sys::FunctionCallInfoBaseData>()
+                + std::mem::size_of::<pg_sys::NullableDatum>() * args.len(),
+        ) as *mut pg_sys::FunctionCallInfoBaseData;
+
+        // initialize it
+        let fcinfo_ref = fcinfo.as_mut().unwrap();
+        fcinfo_ref.flinfo = &mut flinfo;
+        fcinfo_ref.fncollation = pg_sys::InvalidOid; // TODO:  We need to get a collation from somewhere?
```
thomcc: Is it ever unsafe to use the wrong one? If not, I think we should offer a version that uses InvalidOid (which is the default), and one that allows the user to provide it.

That seems to match how function calls are performed in the PG sources -- usually with InvalidOid, but sometimes the collation is provided externally (I haven't threaded through to figure out what logic the higher level callers use to pass the collation in, but it seems like it varies).

pgrx/src/fcall.rs
```diff
@@ -0,0 +1,218 @@
...
+    // return type, so we'll have to do that later
+    PgTryBuilder::new(AssertUnwindSafe(|| unsafe {
+        Ok(pg_sys::LookupFuncName(
+            parts_list.as_ptr(),
+            nargs.into(),
+            arg_types.as_ptr(),
+            false, // missing_ok is not ok
+        ))
+    }))
+    .catch_when(PgSqlErrorCode::ERRCODE_AMBIGUOUS_FUNCTION, |_| Err(FCallError::AmbiguousFunction))
```
thomcc: I wonder if it's better to catch these or allow them to propagate. I suppose it's more rusty to catch and return a result...

pgrx/src/pg_catalog/pg_proc.rs
eeeebbbbrrrr: Those notes were about the .unwrap() not panicking.  I'll change the wording.

pgrx/src/fcall.rs
eeeebbbbrrrr: ???????

pgrx/src/fcall.rs
eeeebbbbrrrr: > Is it ever unsafe to use the wrong one?

Using `InvalidOid` with a function that takes or returns a TEXT-type argument flat out fails with an error.  So we gotta provide some legitimate value somehow.  Jubilee had some ideas on Discord.

eeeebbbbrrrr:
I think this is done.

We can bikeshed the "fcall" names a bit if anyone wants.  I was just looking for something easy to type and wouldn't overpower the function name string argument when reading code that uses this.

If this were javascript I'd have named this `$` so there's that.

eeeebbbbrrrr:
This will ultimately need PR #1277.  So I'd actually like to get that merged first, then I'll do the work to update this.  The primary reason is not that the current PgList sucks (it does), but that it's hidden behind the `cshim` feature flag, and in order for `fcall` to work with PL/Rust it cannot be behind that feature.

pgrx/src/fcall.rs
```diff
@@ -0,0 +1,448 @@
...
+    if !matches!(pg_proc.prokind(), ProKind::Function) {
+        // It only makes sense to directly call regular functions.  Calling aggregate or window
+        // functions is nonsensical
+        return Err(FCallError::UnsupportedFunctionType);
+    } else if pg_proc.proargmodes().iter().any(|mode| *mode != ProArgMode::In) {
+        // Right now we only know how to support arguments with the IN mode.  Perhaps in the
+        // future we can support IN_OUT and TABLE return types
+        return Err(FCallError::UnsupportedArgumentModes);
+    } else if retoid == pg_sys::INTERNALOID
+        || pg_proc.proargtypes().iter().any(|oid| *oid == pg_sys::INTERNALOID)
```
thomcc: I literally came here to ask if we were doing this, and am super glad to see it's forbiddn!

pgrx/src/fcall.rs
```diff
@@ -0,0 +1,448 @@
...
+    {
+        // No idea what to do with the INTERNAL type.  Generally it's just a raw pointer but pgrx
```
thomcc: What to do is forbid it. It's entirely untyped. It's not entirely reasonable to use in a low level pgrx extension, but it should require unsafe code because you must know what you are doing.

pgrx/src/fcall.rs
```diff
@@ -0,0 +1,448 @@
...
+                datum.map(|d| (d, false)).unwrap_or_else(|| (pg_sys::Datum::from(0), true));
+        }
+
+        // call the function
+        // SAFETY: `flinfo` was create for us by `fmgr_info` above and `fn_addr` would have been properly set
+        let func = *(*fcinfo_ref.flinfo)
+            .fn_addr
+            .as_ref()
+            .expect("function initialization problem: fn_addr not set");
+        let result_datum = func(fcinfo);
```
thomcc: Just thought of this, but this is (potentially) an external call into a C postgres function, so if it errors it will longjmp, no? So I think we need a [`pg_guard_ffi_boundary`](https://github.com/pgcentralfoundation/pgrx/blob/develop/pgrx-pg-sys/src/submodules/ffi.rs#L92) around it, unless I'm mistaken

thomcc (changes requested):
(Oops, forgot to mark as changes requested -- this shouldn't land without the error-handling guard. Might be nice to add a test calling some function which raises an error, but checking that it isn't longjmping over rust frames might be really annoying)

pgrx/src/fcall.rs
eeeebbbbrrrr: I think it's more usable here to catch and convert to Result as the user may be implementing some kind of dynamic function calling system where they'd need to be able to recover from a function name that doesn't exist.

pgrx/src/fcall.rs
eeeebbbbrrrr: yeah, great catch!  Didn't even occur to me.  :(

eeeebbbbrrrr: Fixed up, with a test, in https://github.com/pgcentralfoundation/pgrx/pull/1279/commits/7616a6685c27f444afde64c6219b1ca2453c4a3d

workingjubilee:
The List bugs that were blocking you should be fixed by a simple rebase on develop.

eeeebbbbrrrr:
> The List bugs that were blocking you should be fixed by a simple rebase on develop.

yup, rebased this PR and CI passed and tests passed locally.  Excellent.

I just need an approval or two on this before we merge.

pgrx/src/fcall.rs
```diff
@@ -0,0 +1,461 @@
...
+///
+/// # Errors
+///
+/// - [`FCallError::NotDefaultArgument`] if the specified `argnum` does not have a `DEFAULT` clause
+/// - [`FCallError::DefaultNotConstantExpression`] if the `DEFAULT` clause is one we cannot evaluate
+fn create_default_value(pg_proc: &PgProc, argnum: usize) -> Result<Option<pg_sys::Datum>> {
+    let non_default_args_cnt = pg_proc.pronargs() - pg_proc.pronargdefaults();
+    if argnum < non_default_args_cnt {
+        return Err(FCallError::NotDefaultArgument(argnum));
+    }
```
workingjubilee: Is it the case that default arguments must only come last? What about functions with mixed or named notation?

pgrx/src/fcall.rs
```diff
@@ -0,0 +1,461 @@
...
+/// CREATE FUNCTION sum_array(a int4[]) RETURNS int4 LANGUAGE sql AS $$ SELECT sum(value) FROM (SELECT unnest($1) value) x $$;
+/// ```
+///
+/// ```rust,no_run
+/// use pgrx::fcall::{Arg, fcall};
+///
+/// let array = vec![1,2,3];
+/// let sum = fcall::<i32>("sum_array", &[&Arg::Value(array)]);
+/// assert_eq!(sum, Ok(Some(6)));
+/// ```
```
workingjubilee: What's the UX like for calling functions that normally used named notation? Is that even possible with this interface, or would we have to have another? Does Postgres internally canonize them to positional notation and you have to match the positional notation anyway, even if the argument fields get swapped?

eeeebbbbrrrr: Yeah, postgres rewrites them into positional args.  I didn't feel like it was worth supporting.  We could in the future with a different function that takes a Map instead of a slice for the argument set.

pgrx/src/fcall.rs
eeeebbbbrrrr: yes, default arguments must come last.

None of this supports named notation -- I should have listed that in the "things we don't support" up above.

pgrx/src/pg_catalog/mod.rs
workingjubilee, line 1: Hm.

workingjubilee (approved):
Looks good aiui with whatever naming stuff hashed out ~~and a comment about how functions using named notation may technically be "supported" but may be unstable (because they're secretly positional and there's not necessarily a promise on how exactly Postgres settles that positionality).~~

pgrx/src/pg_catalog/mod.rs
eeeebbbbrrrr: > because they're secretly positional and there's not necessarily a promise on how exactly Postgres settles that positionality

There's a function that figures this out.  I ran across it last week but don't remember the name.

Keep in mind, we're not trying to emulate SQL here.  If a user wants all the niceties of SQL they can use SPI.

workingjubilee: agreed, I just asked to make sure we had an explanation in-hand if a user finds our behavior surprising.

eeeebbbbrrrr:
Alrighty, @workingjubilee!  I believe this is done.

thomcc (approved):
My concerns are resolved, and it seems mostly fine.

pgrx/src/pg_catalog/mod.rs
eeeebbbbrrrr: Just a note to clarify that functions are not declared with named positional arguments, they're called that way.

workingjubilee: whoops!

PR #1285, opened 2023-09-11
when constructing a new ListCell, the 'next' node begins life as NULL

eeeebbbbrrrr:
This fixes a segfault when iterating a List.

CI helped to find this when testing #1279 on pg12.

There's still another bug but I haven't found it yet and therefore cannot explain what it is either!


workingjubilee (approved):
Ah right, it's not zero-initialized with MemoryContextAlloc.

PR #1286, opened 2023-09-11
Rename node_list.rs -> linked_list.rs

workingjubilee:
This name made sense at first, but it actually seems to be getting a bit confusing because of the use of "Nodes" elsewhere in Postgres.


eeeebbbbrrrr:
I was thinking this today actually.  Ship it!

PR #1289, opened 2023-09-11
Fix some List sizing bugs

workingjubilee:
Linked lists now correctly track their length.

There remains a bug inside the flat list resizing code. I padded the initial buffer more so that we can store a few more elements without doing that, but it still must be fixed. I will revisit and add more tests later.


workingjubilee:
THAT WASN'T A FIX YOU AIRHEAD.

workingjubilee:
I am a :sparkles:professional software developer:sparkles:.

PR #1294, opened 2023-09-13
Generate bindings for WAL recovery functions

feikesteenbergen:
While building a feature, I found out `GetXLogReplayRecPtr` was no longer available in PostgreSQL 15 and 16. Older versions did contain it just fine. In commit 70e81861 of the PostgreSQL repository, xlog.c was split: recovery related functions were moved into xlogrecovery.c

This commit ensures we also generate the bindings for this file

1: https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=70e81861fadd9112fa2d425c762e163910a4ee52

### DIFF
--- a/pgrx-pg-sys/include/pg15.h
+++ b/pgrx-pg-sys/include/pg15.h
@@ -30,6 +30,7 @@
 #include "access/sysattr.h"
 #include "access/table.h"
 #include "access/xact.h"
+#include "access/xlogrecovery.h"
 #include "catalog/dependency.h"
 #include "catalog/index.h"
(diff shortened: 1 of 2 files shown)


feikesteenbergen:
My guess is these bindings should also be regenerated, and that regeneration committed. I don't know the exact process and environment that should be doing that though.

eeeebbbbrrrr:
PG16 was just released and I'm going to publish a v0.10.1 release today (that's based off the `v0.10.0` tag and not develop) and I'll make sure this gets cherry-picked into that and that bindings are properly updated.

Issue #1293, opened 2023-09-12
Array of composite type inside composite type


alaneos777:
Hi everyone, I was trying to follow this [example](https://github.com/pgcentralfoundation/pgrx/blob/master/pgrx-examples/composite_type/src/lib.rs) and create a composite type that holds an array of another composite type. This is the whole code:

```rust
use pgrx::prelude::*;

pgrx::pg_module_magic!();

extension_sql!(
    r#"
    CREATE TYPE entity AS (id INT, lang TEXT);
    CREATE TYPE result AS (score DOUBLE PRECISION, entities entity[]);
    "#,
    name = "init_example",
    bootstrap
);

#[pg_extern]
fn create_result() -> pgrx::composite_type!("result") {
    let mut record = PgHeapTuple::new_composite_type("result").unwrap();
    record.set_by_name("score", 0.707).unwrap();
    let mut entity_records: Vec<pgrx::composite_type!("entity")> = Vec::new();
    for i in 0..10 {
        let mut entity_record = PgHeapTuple::new_composite_type("entity").unwrap();
        entity_record.set_by_name("id", i).unwrap();
        entity_record.set_by_name("lang", "en".to_string()).unwrap();
        entity_records.push(entity_record);
    }
    record.set_by_name("entities", entity_records).unwrap();
    return record;
}

/// This module is required by `cargo pgrx test` invocations.
/// It must be visible at the root of your extension crate.
#[cfg(test)]
pub mod pg_test {
    pub fn setup(_options: Vec<&str>) {
        // perform one-off initialization when the pg_test framework starts
    }

    pub fn postgresql_conf_options() -> Vec<&'static str> {
        // return any postgresql.conf settings that are required for your tests
        vec![]
    }
}
```

But when I want to call `select * from create_result();`, I got this:
```
example=# create extension example;
CREATE EXTENSION
example=# select * from create_result();
ERROR:  called `Result::unwrap()` on an `Err` value: IncompatibleTypes { rust_type: "alloc::vec::Vec<pgrx::heap_tuple::PgHeapTuple<pgrx::pgbox::AllocatedByRust>>", rust_oid: Oid(25463), datum_type: "record[]", datum_oid: Oid(2287) }
example=#
```

What do I need to change to achieve this scenario? Thanks in advance for your help.

eeeebbbbrrrr:
Thanks for the report.  We'll take a look at the bug.

eeeebbbbrrrr:
[...]

PR #1303, opened 2023-09-18
Fix issue #1293

eeeebbbbrrrr:
This allows for `PgHeapTuple.set_by_index()` to work correctly with arrays of composite types.

### DIFF
--- a/pgrx/src/datum/array.rs
+++ b/pgrx/src/datum/array.rs
@@ -741,6 +741,13 @@ impl<T: IntoDatum + FromDatum> IntoDatum for Array<'_, T> {
     fn type_oid() -> Oid {
         T::array_type_oid()
     }
+
+    fn composite_type_oid(&self) -> Option<Oid> {
+        // the composite type oid for a vec of composite types is the array type of the base composite type
+        self.get(0)
+            .map(|v| v.composite_type_oid().map(|oid| unsafe { pg_sys::get_array_type(oid) }))
+            .flatten()
+    }
 }
 
 impl<T: FromDatum> FromDatum for Vec<T> {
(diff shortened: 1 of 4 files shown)


workingjubilee (approved):

pgrx/src/heap_tuple.rs
```diff
@@ -352,17 +352,34 @@ impl<'a> PgHeapTuple<'a, AllocatedByRust> {
...
+                        if att.atttypid != array_of_composite_type_oid {
+                            return Err(TryFromDatumError::IncompatibleTypes {
+                                rust_type: std::any::type_name::<T>(),
+                                rust_oid: att.atttypid,
+                                datum_type: lookup_type_name(array_of_composite_type_oid),
+                                datum_oid: array_of_composite_type_oid,
+                            });
+                        }
+                    } else {
+                        // it's not an array type, so we'll do standard type compatability checks
```
workingjubilee:
```suggestion
                        // it's not an array type, so we'll do standard type compatibility checks
```

pgrx/src/datum/array.rs
```diff
@@ -741,6 +741,13 @@ impl<T: IntoDatum + FromDatum> IntoDatum for Array<'_, T> {
     fn type_oid() -> Oid {
         T::array_type_oid()
     }
+
+    fn composite_type_oid(&self) -> Option<Oid> {
+        // the composite type oid for a vec of composite types is the array type of the base composite type
+        self.get(0)
+            .map(|v| v.composite_type_oid().map(|oid| unsafe { pg_sys::get_array_type(oid) }))
+            .flatten()
```
workingjubilee, line 749: It looks like it should work but doesn't this needlessly whiff for an Array of size zero when it could instead be
```suggestion
        Some(unsafe { pg_sys::get_array_type(self.raw.oid()) })
```

eeeebbbbrrrr: hm.  Yeah, I suppose this would work for `Array`.  This was just a copy-paste from what I did for `Vec`, which in fact needs at least 1 value to figure this out.

PR #1304, opened 2023-09-18
Replace cshimmed fn

workingjubilee:
This returns to their rightful place some cshimmed fn that were accidentally dropped on the floor while I was trying to clean up pgrx-pg-sys to... make it harder to lose track of things... yeah! The best-laid plans of mice and men etc.

It then does something more devilish: by shadowing certain names, it allows us to outmode them entirely in the future.


eeeebbbbrrrr (approved):
LGTM!

PR #1305, opened 2023-09-19
[DO NOT MERGE] macOS CI: run tests on Postgres 15 and 16

workingjubilee:
Just running CI on the test itself to see what's what.


workingjubilee:
Okay, this validated what I wanted it to: the test is in fact a regression test.

PR #1307, opened 2023-09-19
v0.10: Backport "Add cshimmed fn back"

workingjubilee:
In case we do another patch release, since whoopsies.


eeeebbbbrrrr:
Thanks.  I think we will and I'll merge this now.

PR #1300, opened 2023-09-17
Fix install command for Postgres 16 on macOS

bayandin:
Since Postgres 16, the shared library extension on macOS is `.dylib` instead of `.so`.
Ref https://github.com/postgres/postgres/commit/b55f62abb2c2e07dfae99e19a2b3d7ca9e58dc1a

PR #1308, opened 2023-09-19
Bump to pgrx v0.10.2

workingjubilee:
Welcome to pgrx v0.10.2!

In pgrx v0.10.0, as part of a refactoring step in order to make refactoring easier in the future, without things like losing track of random functions, we... omitted a few functions! These depended on the cshim, but were not direct bindings, themselves. Oops. They're back now in `pgrx::pg_sys`.

## `cargo pgrx` works now for Postgres 16 on macOS

Installing extensions on macOS for Postgres 16 had problems due to recent changes to Postgres defaults. We have fixed this and expanded our test coverage for macOS to reduce the odds of this happening in the future, thanks to @bayandin in https://github.com/pgcentralfoundation/pgrx/pull/1300

## composites compose better

There were some problems with composite types and arrays, and specifically arrays of composite types in composite types. These are the sorts of things that cause edge cases in PgHeapTuple. They should now be appropriately supported, however, thanks to @eeeebbbbrrrr in https://github.com/pgcentralfoundation/pgrx/pull/1303

## `#[pg_schema]` will enable your bad ideas

The following should now work appropriately:
```rust
#[pgrx::pg_schema]
mod pg_catalog {
```

It does what you think it does. Since this can affect a global namespace in a way that makes it much easier to conflict with other extensions, it is best to avoid this. However, it is the sort of bad idea that extensions which "take over" the Postgres host extensively redefine the host's capabilities, like [ZomboDB](https://www.zombodb.com/) does, may have to resort to in order to accomplish their goal. Thanks to @eeeebbbbrrrr in https://github.com/pgcentralfoundation/pgrx/pull/1302

## Keep the bug reports coming

And thank _you_ for reporting your issues!


eeeebbbbrrrr (approved):
wonderful!

PR #1309, opened 2023-09-19
Enable proptest in more workflows

workingjubilee:


BradyBonnette:
So my browser completely failed to send my submission earlier, but this looks good.

PR #1311, opened 2023-09-19
Float Dockerfile's Fedora version

workingjubilee:
The Fedora 36 rpm isn't hosted by download.postgresql.org anymore and we likely will need to change the version more often, so let's start using the Fedora OS version provided as an argument. Also, bump to Fedora 38.

Necessary in order to get Will It Blend™ back to full functionality.


workingjubilee:
macOS failures seem to be transient weirdness from homebrew or GHA or something.

BradyBonnette:
> macOS failures seem to be transient weirdness from homebrew or GHA or something.

This is unsurprising, to say the least. :(

workingjubilee:
Looks like everything passed on the second go.

workingjubilee:
I am merging this so I don't get another notification about the workflow failing, now that it has been reenabled.

PR #1312, opened 2023-09-28
Organize bindgen arg sets in fn (nfc)

workingjubilee:
No functional changes.


thomcc (commented):
I'm not sure that this is an improvement honestly. Why is it a trait?

workingjubilee:
> I'm not sure that this is an improvement honestly. Why is it a trait?

for mostly-preserving the existing method call syntax, thassit. happy to make them free fn.

thomcc:
I'd prefer free functions, since then adding parameters and such doesn't require modifying as many places.

workingjubilee:
sounds good to me!

thomcc:
We talked about this on discord and I don't really care about this enough to object to it becoming a trait if that's a change @workingjubilee wants to make.

workingjubilee:
Free fn are fine. We discussed some options for improving things further and like... mostly, I'm okay with anything as long as it's somewhat more organized than "not at all". The main goal is to avoid this becoming impossible to read and sift through when more bindgen flags get added.

PR #1314, opened 2023-09-28
Consistently order generated impls

workingjubilee:
Using BTreeSet instead of HashSet allows ceasing to permeate `is_for_release` through bindgen, into parts which don't need to know about that detail. This removes a source of inconsistency in our build artifacts, and also avoids having to do an additional reallocate-and-sort step, by simply using a type that was always going to be sorted anyways.

Local testing shows no negative impacts on compile time, and a mild positive, but within noise variance... this is only a couple seconds on a compile that takes almost a minute, even on a powerful machine.


thomcc (approved):
Good call. It's odd how many places we explicitly check this (even after this we check it more than once) but that's an unrelated issue.

PR #1316, opened 2023-09-29
Make check builds ~25% faster by merging extern blocks

workingjubilee:
Merge all extern "C" blocks into one big one via bindgen. Then split them up into two blocks while rewriting items. This takes off ~20,000 lines of the 60,000+ line bindgen, which reduces the amount of time spent writing to disk, saving about 300 kilobytes in real bitstate.

This drops "clean check" builds on my machine from about ~43 seconds to ~32 seconds, so -25% or +33% faster, depending on how you count. Some of this may be due to less thrashing during macro expansion.


eeeebbbbrrrr:
Amazing!

thomcc (approved):

workingjubilee:
Doesn't seem to have made CI detectably faster on its own, but that's fine: I'm the one who needs the fast iteration cycles!

PR #1320, opened 2023-10-05
Include bindings for controlfile

feikesteenbergen:
> NOTE: I didn't include the updated generated bindings, as I'm still unsure as to what the proper invocation is, these still need to be generated.

The function get_controlfile is very useful for some monitoring experiments and solutions.

I'm working on a bgworker that starts even when in crash recovery. Part of the goal of this worker is to provide more insight into the progress of crash recovery:

- distance to consistency (in WAL bytes)
- distance since last checkpoint (in WAL bytes)

For that to work nicely, it would be great to have access to the `get_controlfile` function, which exposes this information in an easy digestable fashion.

### DIFF
--- a/pgrx-pg-sys/include/pg11.h
+++ b/pgrx-pg-sys/include/pg11.h
@@ -66,6 +66,7 @@
 #include "commands/user.h"
 #include "commands/vacuum.h"
 #include "common/config_info.h"
+#include "common/controldata_utils.h"
 #include "executor/executor.h"
 #include "executor/spi.h"
 #include "executor/tuptable.h"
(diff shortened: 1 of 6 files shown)


feikesteenbergen:
I'm not sure what I'm doing wrong, tests locally work, and this seems to fail all the time against postgres-16 only.

eeeebbbbrrrr:
This CI failure is our fault. It’s on my list to resolve today, actually. 

Thanks and hang tight!

eeeebbbbrrrr:
I fixed CI in the `develop` branch.  If you want to rebase this and push again, we should see things be all green here.

I'd kinda like for this PR to be green before we merge since it's related to bindings generation.

I plan on cutting a v0.10.x later this week, and I'm happy to cherry-pick this PR once we merge it into `develop`.

PR #1322, opened 2023-10-05
Add `cargo pgrx install --sudo`

eeeebbbbrrrr:
`cargo pgrx install --sudo` is a lightweight wrapper around `cargo pgrx package --release` that then uses `sudo cp $src $dest` to copy the extension file artifacts where `pg_config` thinks they should go on the system.

Typically, for package-manager installed Postgres installations, the extension target directories are owned by root and difficult for cargo-pgrx to deal with.  This gives the user a simple escape hatch for when they want to otherwise `cargo pgrx install`.

This idea came from @levkk over at PostgresML.


eeeebbbbrrrr:
If we approve of this, I'll also cherry-pick it into the `v0.10-develop` branch and go ahead and publish it.

eeeebbbbrrrr:
Screenshot 2023-10-05 at 10 06 16 AM

cargo-pgrx/src/command/sudo_install.rs
```diff
@@ -0,0 +1,92 @@
...
+        eprintln!("Using sudo to copy extension files from {}", outdir.display().cyan());
+        for src in output_files {
+            let src = src.canonicalize()?;
+            let dest = make_absolute(src.strip_prefix(&outdir)?).canonicalize()?;
+
+            // we're about to run `sudo` to copy some files, one at a time
+            let mut command = Command::new("sudo"); // NB:  If we ever support Windows...
+            command.arg("cp").arg(src).arg(dest);
+
+            println!("{} {:?}", "     Running".bold().green(), command);
```
levkk: Can make it prettier, e.g.

```rust
println!("{} sudo cp {} {}", "       Running".bold().green(), src.display(), dest.display());
```

eeeebbbbrrrr: okay, will do.

cargo-pgrx/src/command/sudo_install.rs
```diff
@@ -0,0 +1,92 @@
...
+            features: value.features,
+            verbose: value.verbose,
+        }
+    }
+}
+
+impl CommandExecute for SudoInstall {
+    fn execute(self) -> eyre::Result<()> {
+        fn make_absolute(p: &Path) -> PathBuf {
+            PathBuf::from_iter(vec![PathBuf::from("/").as_path(), p])
```
thomcc, line 59: Wow does this ever look wrong (it's not wrong for this use case ofc, but still)

cargo-pgrx/src/command/sudo_install.rs
```diff
@@ -0,0 +1,92 @@
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+use owo_colors::OwoColorize;
+
+use crate::command::package::Package;
+use crate::CommandExecute;
+
+/// Like `cargo pgrx install`, but uses `sudo` to copy the extension files
```
thomcc, line 10: I wonder if this would be better as `cargo pgrx install --sudo`. This is fine though.

I also slightly wonder if it's worth warning users who try to `sudo cargo pgrx install`, since it's both a footgun and a big security liability (there are definitely reasons to do it, so "warn" seems right rather than actually forbidding it). Neither here nor there though, just an idle thought.

cargo-pgrx/src/command/sudo_install.rs
```diff
@@ -0,0 +1,92 @@
...
+        for src in output_files {
+            let src = src.canonicalize()?;
+            let dest = make_absolute(src.strip_prefix(&outdir)?).canonicalize()?;
+
+            // we're about to run `sudo` to copy some files, one at a time
+            let mut command = Command::new("sudo"); // NB:  If we ever support Windows...
+            command.arg("cp").arg(src).arg(dest);
+
+            println!("{} {:?}", "     Running".bold().green(), command);
+            let output = command.output()?;
```
thomcc: Using `output()` here will cause stderr/stdout to be piped. I'd expect that to prevent `sudo` from prompting the user for their password, or does that somehow work?

cargo-pgrx/src/command/sudo_install.rs
levkk: Most people won't even be able to because Rust is not installed as the root user. If you try, it will yell at you for...a while and ask you to set all kinds of environment variables before it will allow itself to be installed as root. But yes, I agree, a simple `if whoami == root then exit 1` in this case would not be a terrible idea.

Thinking out loud, `cargo pgrx install` would, by default never really work on any Linux system because for most people this would resolve to the installation in `/usr` . It works great on Mac because brew installs into `/opt/homebrew` where the user is allowed read/write access.

cargo-pgrx/src/command/sudo_install.rs
levkk: Yeah why don't you just use `std::fs::canonicalize` together with `shellexpand` as needed?

eeeebbbbrrrr:
```rust
let src = src.canonicalize()?;
let dest = make_absolute(src.strip_prefix(&outdir)?).canonicalize()?;
```

The `src.strip_prefix(&outdir)` call here includes the leading `/`.  So without `make_absolute()` the path ends up being relative to the current working directory.  Ask me how I know.

cargo-pgrx/src/command/sudo_install.rs
eeeebbbbrrrr: I'll rejigger this so that it is `cargo pgrx install --sudo`.  The spelling of this isn't a hill I'm willing to die on.

If we want a `if whoami == root` check, then we can do that as another PR.  That feels orthogonal to this.

cargo-pgrx/src/command/sudo_install.rs
eeeebbbbrrrr: It worked for me on Linux.  See the screenshot up above, but you see the sudo password prompt, can enter your password, and continue.

I wonder if sudo does something tricky if it detects stdout/err is piped?  I suppose I can spawn the process and just wait for it, tho.  I'll do that too.

eeeebbbbrrrr:
Alrighty, I've cleaned things up in https://github.com/pgcentralfoundation/pgrx/pull/1322/commits/d9563c099be91f6c989d84a55de29a72cc998d39

~~CI is going to fail on pg16, but that's unrelated to this PR (and has been fixed/merged by #1326)~~

huh, it did not fail.  Even better!

eeeebbbbrrrr:
@thomcc if I can get an approval on this today, that'd be great.  I want to put this behind us and get a new pgrx release out tomorrow.

thomcc (approved):
Looks good!

eeeebbbbrrrr:
Thank you for updating the title/description, @workingjubilee.

PR #1326, opened 2023-10-10
Fix pg16 CI

eeeebbbbrrrr:
Taking what we learned from PL/Rust and pg16 and CI, this does essentially the same two things:

- removes all the old postgres/clang/llvm-related packages that github might have installed
- sets up clang14 only

The "problem" seems to be having multiple versions of clang installed at once.  There may be some way to say, "dear bindgen, only use this particular clang", but I think this is more expedient.  Also, it's probably best CI doesn't have a bunch of clutter we don't need.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -53,22 +53,21 @@ jobs:
         echo ""
 
         echo "----- Remove old postgres -----"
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr:
merging w/o review since it's "just CI", and we need it to work and someone already got confused by this in #1320

PR #1332, opened 2023-10-12
Add wrappers for postgres geometric types.

hamiltop:
For simple `Copy` types, just newtype them and add impls for FromDatum and ToDatum.

For more complex variable length types, create zero-copy and owned structs with impls for FromDatum and ToDatum as well as SqlTranslatable.

### DIFF
--- a/pgrx-tests/src/tests/geo_tests.rs
+++ b/pgrx-tests/src/tests/geo_tests.rs
@@ -13,24 +13,97 @@ mod tests {
     #[allow(unused_imports)]
     use crate as pgrx_tests;
 
+    use pgrx::datum::geo::*;
     use pgrx::prelude::*;
 
     #[pg_test]
-    fn test_point_into_datum() -> spi::Result<()> {
-        let p =
-            Spi::get_one::<pg_sys::Point>("SELECT '42, 99'::point")?.expect("SPI result was null");
-        assert_eq!(p.x, 42.0);
-        assert_eq!(p.y, 99.0);
+    fn test_point_datum() -> spi::Result<()> {
+        let p = Spi::get_one::<Point>("SELECT '42, 99'::point")?.expect("SPI result was null");
+        assert_eq!(p, Point { x: 42.0, y: 99.0 });
+        let p2 = Spi::get_one_with_args::<Point>(
+            "SELECT $1",
+            vec![(Point::type_oid().into(), p.into_datum())],
+        )?
+        .expect("SPI result was null");
+        assert_eq!(p, p2);
         Ok(())
     }
 
     #[pg_test]
-    fn test_box_into_datum() -> spi::Result<()> {
-        let b = Spi::get_one::<pg_sys::BOX>("SELECT '1,2,3,4'::box")?.expect("SPI result was null");
-        assert_eq!(b.high.x, 3.0);
-        assert_eq!(b.high.y, 4.0);
-        assert_eq!(b.low.x, 1.0);
-        assert_eq!(b.low.y, 2.0);
+    fn test_box_datum() -> spi::Result<()> {
+        let b = Spi::get_one::<Box>("SELECT '1,2,3,4'::box")?.expect("SPI result was null");
+        assert_eq!(b, Box { high: Point { x: 3.0, y: 4.0 }, low: Point { x: 1.0, y: 2.0 } });
+        let b2 = Spi::get_one_with_args::<Box>(
+            "SELECT $1",
+            vec![(Box::type_oid().into(), b.into_datum())],
+        )?
+        .expect("SPI result was null");
+        assert_eq!(b, b2);
+        Ok(())
+    }
+
+    #[pg_test]
+    fn test_lseg_datum() -> spi::Result<()> {
+        let l = Spi::get_one::<LineSegment>("SELECT '(1,2),(3,4)'::lseg")?
+            .expect("SPI result was null");
+        assert_eq!(l.p, [Point { x: 1.0, y: 2.0 }, Point { x: 3.0, y: 4.0 }]);
+        let l2 = Spi::get_one_with_args::<LineSegment>(
+            "SELECT $1",
+            vec![(LineSegment::type_oid().into(), l.into_datum())],
+        )?
+        .expect("SPI result was null");
+        assert_eq!(l.p, l2.p);
+        Ok(())
+    }
+
+    #[pg_test]
+    fn test_path_datum() -> spi::Result<()> {
+        // Closed path
+        let p = Spi::get_one::<Path>("SELECT '((1,2),(3,4))'::path")?.expect("SPI result was null");
+        assert_eq!(p.points(), [Point { x: 1.0, y: 2.0 }, Point { x: 3.0, y: 4.0 }]);
+        assert_eq!(p.closed(), true);
+        let p2 = Spi::get_one_with_args::<Path>(
+            "SELECT $1",
+            vec![(Path::type_oid().into(), p.clone().into_datum())],
+        )?
+        .expect("SPI result was null");
+        assert_eq!(p.points(), p2.points());
+        assert_eq!(p.closed(), p2.closed());
+
+        // Open path
+        let p = Spi::get_one::<Path>("SELECT '[(1,2),(3,4)]'::path")?.expect("SPI result was null");
+        assert_eq!(p.points(), [Point { x: 1.0, y: 2.0 }, Point { x: 3.0, y: 4.0 }]);
+        assert_eq!(p.closed(), false);
+        let p2 = Spi::get_one_with_args::<Path>(
+            "SELECT $1",
+            vec![(Path::type_oid().into(), p.clone().into_datum())],
+        )?
+        .expect("SPI result was null");
+        assert_eq!(p.points(), p2.points());
+        assert_eq!(p.closed(), p2.closed());
+
+        Ok(())
+    }
+
+    #[pg_test]
+    fn test_polygon_datum() -> spi::Result<()> {
+        let p = Spi::get_one::<Polygon>("SELECT '((1,2),(3,4),(0,5))'::polygon")?
+            .expect("SPI result was null");
+        assert_eq!(
+            p.points(),
+            [Point { x: 1.0, y: 2.0 }, Point { x: 3.0, y: 4.0 }, Point { x: 0.0, y: 5.0 }]
+        );
+        assert_eq!(
+            p.boundbox(),
+            Box { high: Point { x: 3.0, y: 5.0 }, low: Point { x: 0.0, y: 2.0 } }
+        );
+        let p2 = Spi::get_one_with_args::<Polygon>(
+            "SELECT $1",
+            vec![(Polygon::type_oid().into(), p.clone().into_datum())],
+        )?
+        .expect("SPI result was null");
+        assert_eq!(p.points(), p2.points());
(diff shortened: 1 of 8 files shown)


pgrx/src/datum/geo.rs
```diff
@@ -7,9 +7,30 @@
...
+use pgrx_pg_sys::pfree;
+use pgrx_sql_entity_graph::metadata::{
+    ArgumentError, Returns, ReturnsError, SqlMapping, SqlTranslatable,
+};
+
+use crate::{pg_sys, set_varsize_4b, FromDatum, IntoDatum, PgMemoryContexts};
+
+// Copy types
+
+pub type Box = pg_sys::BOX;
```
hamiltop, line 31: In [pgrx/src/datum/date.rs] the pattern is to do a transparent wrapping struct and add `From` impls. If that's the preferred pattern I can switch these over to that approach.

eeeebbbbrrrr: Give us a few days to look this over.  

I don’t know what the right answer is. I think we might prefer a newtype but then that’ll cause complications over in pl/rust land for types that are already exposed, like BOX. That’s not your responsibility of course, buts it’s something we need to take into consideration.

pgrx/src/datum/geo.rs
```diff
@@ -59,11 +150,11 @@ impl FromDatum for pg_sys::Point {
     }
 }
 
-impl IntoDatum for pg_sys::Point {
+impl IntoDatum for Point {
     fn into_datum(mut self) -> Option<pg_sys::Datum> {
         unsafe {
             let copy = PgMemoryContexts::CurrentMemoryContext
-                .copy_ptr_into(&mut self, std::mem::size_of::<pg_sys::Point>());
+                .copy_ptr_into(&mut self, std::mem::size_of::<Point>());
```
eeeebbbbrrrr: we tend to prefix things in `pg_sys` with `pg_sys::`.  I can't say we're 100% consistent with that throughout the codebase, but my preference would be to shift towards that direction, not away.

The reason is it makes it a lot easier when reading the code to know "this thing is defined by Postgres", which is almost always an important thing to know about a type when working inside pgrx.

So, if you will, please revert this particular change, and if there's anything else in the PR that elides `pg_sys::`, please add it.

pgrx/src/datum/geo.rs
eeeebbbbrrrr: I've been thinking about this a bit more.  I think it's fine to leave these as aliases.  In fact, it's probably preferred.

A newtype wrapper would imply that pgrx has some more knowledge about the type and, for example, enforces bounds or requires some invariants that we don't.

We basically had to do that for the date/time types b/c even tho the underlying type is something basic like a u32 (or whatever), not all u32s are valid dates or timestamps or whatnot.  I don't guess that's the case here.

It looks like Postgres does some validation on the string representation of these types, of course, but it doesn't care to check the actual values.

eeeebbbbrrrr (changes requested):
Thank you very much @hamiltop.  This is sincerely appreciated.

In addition to the little nit about prefixing things in pg_sys with `pg_sys::`, if you could update the table of types in the top-level README.md, that'd be great.

Our documentation is nowhere near where we'd like it to be, but that table is a great quick reference.

workingjubilee (changes requested):
Please split the "zero-copy" types into their own PR. We will need to evaluate them separately as they intersect with part of the MemCx project.

pgrx/src/datum/geo.rs
hamiltop: To be clear, this isn't quite "eliding `pg_sys::`". This is a change from `pg_sys::Point` to `geo::Point` (which happens to be an alias to `pg_sys::Point`). It's especially confusing here because they are both spelled and capitalized the same way, but other cases it's `Box` vs `pg_sys::BOX` and `LineSegment` vs `pg_sys::LSEG`.

Also, this particular line could also be rewritten as `std::mem::size_of::<Self>`.

I do like the more normal type names, especially in place of ALL_CAPS ones, but if `pg_sys::` is preferred I can make that change.

eeeebbbbrrrr: > This is a change from pg_sys::Point to geo::Point

hmm.  yeah, it is.

> Also, this particular line could also be rewritten as std::mem::size_of::<Self>.

Indeed.  Probably good to make that little change.

> but if pg_sys:: is preferred I can make that change.

No, this is correct.  You're right.

hamiltop: Updated.

hamiltop:
@workingjubilee Updated with removal of zero-copy types. Should I wait for `MemCx` to land before re-introducing the zero-copy types? Sounds like it's close.

workingjubilee:
Not that close, apparently. ^^ Sorry for the delays.

pgrx/src/datum/geo.rs
```diff
@@ -21,17 +42,17 @@ impl FromDatum for pg_sys::BOX {
...
     }
 }
 
-impl IntoDatum for pg_sys::BOX {
+impl IntoDatum for Box {
     fn into_datum(mut self) -> Option<pg_sys::Datum> {
         unsafe {
             let ptr = PgMemoryContexts::CurrentMemoryContext
-                .copy_ptr_into(&mut self, std::mem::size_of::<pg_sys::BOX>());
+                .copy_ptr_into(&mut self, std::mem::size_of::<Self>());
```
workingjubilee, line 55: While we're here, let's enforce this style:
```suggestion
                .copy_ptr_into(&mut self, mem::size_of::<Self>());
```

pgrx/src/datum/geo.rs
```diff
@@ -41,7 +62,77 @@ impl IntoDatum for pg_sys::BOX {
...
+            Some(line.read())
+        }
+    }
+}
+
+impl IntoDatum for Line {
+    fn into_datum(mut self) -> Option<pg_sys::Datum> {
+        unsafe {
+            let ptr = PgMemoryContexts::CurrentMemoryContext
+                .copy_ptr_into(&mut self, std::mem::size_of::<Self>());
```
workingjubilee, line 89:
```suggestion
                .copy_ptr_into(&mut self, mem::size_of::<Self>());
```

pgrx/src/datum/geo.rs
```diff
@@ -41,7 +62,77 @@ impl IntoDatum for pg_sys::BOX {
...
+            Some(lseg.read())
+        }
+    }
+}
+
+impl IntoDatum for LineSegment {
+    fn into_datum(mut self) -> Option<pg_sys::Datum> {
+        unsafe {
+            let ptr = PgMemoryContexts::CurrentMemoryContext
+                .copy_ptr_into(&mut self, std::mem::size_of::<Self>());
```
workingjubilee, line 123:
```suggestion
                .copy_ptr_into(&mut self, mem::size_of::<Self>());
```

pgrx/src/datum/geo.rs
```diff
@@ -72,3 +163,214 @@ impl IntoDatum for pg_sys::Point {
...
+            Some(circle.read())
+        }
+    }
+}
+
+impl IntoDatum for Circle {
+    fn into_datum(mut self) -> Option<pg_sys::Datum> {
+        unsafe {
+            let ptr = PgMemoryContexts::CurrentMemoryContext
+                .copy_ptr_into(&mut self, std::mem::size_of::<Self>());
```
workingjubilee, line 191:
```suggestion
                .copy_ptr_into(&mut self, mem::size_of::<Self>());
```

pgrx-tests/src/tests/geo_tests.rs
```diff
@@ -13,24 +13,97 @@ mod tests {
...
+            "SELECT $1",
+            vec![(Box::type_oid().into(), b.into_datum())],
+        )?
+        .expect("SPI result was null");
+        assert_eq!(b, b2);
+        Ok(())
+    }
+
+    #[pg_test]
+    fn test_lseg_datum() -> spi::Result<()> {
```
workingjubilee, line 46: These tests look good, but can we also get one or two that show the same types being passed around using the `fn_call` module? I've noticed sometimes there's odd discrepancies between various ways of moving data in/out using literals and SPI, and in/out using other interfaces. You don't have to do all of them, just whichever type seems most quirky.

pgrx/src/datum/geo.rs
workingjubilee: > We basically had to do that for the date/time types b/c even tho the underlying type is something basic like a u32 (or whatever), not all u32s are valid dates or timestamps or whatnot.

Yes, that is exactly the case. The types have implicit invariants that aren't going to be enforced by the "raw" `pg_sys` side, and lifting them into Rust allowed enforcing invariants. If **all pairs of points** are _valid BOXes_ (I'm talking only about Postgres's conception of validity), then it's fine to just use an alias.

pgrx/src/datum/geo.rs
```diff
@@ -41,7 +62,77 @@ impl IntoDatum for pg_sys::BOX {
...
+            Some(ptr.into())
+        }
+    }
+
+    fn type_oid() -> pg_sys::Oid {
+        pg_sys::LSEGOID
+    }
+}
+
+pub type Point = pg_sys::Point;
```
workingjubilee, line 133: All pairs of f64s are valid Points, right?

pgrx/src/datum/geo.rs
```diff
@@ -72,3 +163,214 @@ impl IntoDatum for pg_sys::Point {
         pg_sys::POINTOID
     }
 }
+
+pub type Circle = pg_sys::CIRCLE;
```
workingjubilee, line 167: Same "are there any implicit invariants?" question.

workingjubilee (changes requested):

workingjubilee:
A rough draft for `MemCx<'mcx>` landed in #1342.

eeeebbbbrrrr:
I know this has been out here for awhile now, but it actually looks complete?  @hamiltop / @workingjubilee, do y'all know of anything else that needs to happen?

I'm about to release v0.15.0 so this won't make it into that, but if this is otherwise good I can get it into a v0.15.1.

eeeebbbbrrrr:
Hi @hamiltop!  My apologies for allowing your work here to have been forgotten for so long.  

Since this was ready to merge, I asked Codex to get it updated and it did and force-pushed to your branch, since it also needed to rebase on current `develop`.  Below is a short list of what it did.

I'm merging this now.  Thanks, and again, please accept my apologies for losing track of this.

---

  Rebased this PR onto current `develop` and pushed the repaired branch.

  What changed:
  - Kept simple geometric types as `pg_sys` aliases under `pgrx::geo` rather than newtypes.
  - Added/updated `FromDatum` and `IntoDatum` support for `Box`, `Circle`, `Line`, `LineSegment`, `Point`, `Path`,
  and `Polygon`.
  - Added `SqlTranslatable` coverage for `LINE`/`LSEG` and owned `Path`/`Polygon`.
  - Added `#[pg_extern]` call convention support so the geo types work across function argument/return boundaries.
  - Updated the README type mapping to use `pgrx::geo::*`.
  - Added SPI roundtrip tests plus `fn_call` roundtrip tests for `Path` and `Polygon`.
  - Fixed a couple of implementation issues found during review: initialized `PATH.dummy` and added checked varlena
  size handling.

eeeebbbbrrrr (approved):

PR #1338, opened 2023-10-14
Use and pass jobserver instead of limiting cores directly

passcod:
Currently cargo-pgrx uses both rayon and some manual "max(1, cores / 3)" logic to use multiple cores without overwhelming the system.

This PR ditches these two mechanisms in favour of the jobserver concept, implemented by the jobslot crate. That is an implementation of the `make` jobserver protocol.

Basically a jobserver controls access to a limited pool of job slots: to do some compute, you grab a slot, blocking until one becomes available if necessary. A process in the jobserver context passes down the jobserver handle to subprocesses, who, if they're aware (like `make` is), all thus cooperate together.

In this PR, cargo-pgrx:
- passes the jobserver handle to all calls to `make`
- grabs a slot for all _compute-bound_ steps (ie the untar and configure steps)
- _doesn't_ grab a slot for _network-bound_ steps (ie the download step)
- starts a thread per `pg_config` resolution task

This naturally and granularly distributes tasks across all available cores, without restricting each `make` to a fixed amount of jobs. Notably, if eg building pg12 and pg13, and pg12 finishes completely, the `make` for pg13 may scale up to use the freed-up cores.

---

Before (4 cores, notice how it doesn't even start downloading/untarring pg16 until (off-screen) one of the compiles is fully done):
Screenshot from 2023-10-14 23-14-51

After (4 cores, notice how there's first 5 downloads and then it restricts itself to 4 compute tasks at a time, interleaving as necessary):
Screenshot from 2023-10-14 23-17-09

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -338,6 +338,7 @@ dependencies = [
  "eyre",
  "flate2",
  "fork",
+ "jobslot",
  "libloading 0.8.1",
  "nix",
  "num_cpus",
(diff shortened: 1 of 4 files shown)


workingjubilee (approved):
I only scanned this and am not very familiar with jobserver, but this looks like a good improvement.

eeeebbbbrrrr:
I know it’s WIP but looks like CI wants a “cargo fmt” and I’d want the readme updated before we merge. 

This is neat.

workingjubilee:
Conflicted with #1348

thomcc (approved):

passcod:
I'll fix up the conflicts on this and my other PR later today

workingjubilee:
No rush!

passcod:
Rebased

workingjubilee:
Thanks!

Issue #1341, opened 2023-10-17
Definite memory leak when calling dynamic function


ccleve:
In an attempt to diagnose #1330 I installed the new 0.11.0 and tried out the new dynamic function calls. I got a leak. Perhaps a fix for this one will also fix my issue.

Run this and watch memory usage climb to multiple gb:

```
use pgrx::fn_call::fn_call;
use pgrx::fn_call::Arg;
use pgrx::pg_extern;

#[pg_extern]
fn tester() {
    let bar = "bar";
    for _ in 0..10000000 {
        let result = fn_call::<String>("myfunc", &[&Arg::Value(bar)]);
    }
}

#[pg_extern]
fn myfunc(bar: &str) -> String {
    let mut out = "foo".to_owned();
    out.push_str(bar);
    out
}

```

Just run `select tester();` at the command line, and if you're on a Mac open up Activity Monitor / Memory tab and watch the memory usage grow for the postgres process.

eeeebbbbrrrr:
Thanks for the report. We’ll investigate

eeeebbbbrrrr:
I rewrote this to be even more aggressive:

```rust
#[pg_extern]
fn tester() {
    let bar = "bar";
    for _ in 0..10000000 {
        check_for_interrupts!();

        let s = fn_call::<String>("myfunc", &[&Arg::Value(bar)]);
        assert!(matches!(s, Ok(Some(_))))
    }
}

#[pg_extern]
fn myfunc(bar: &str) -> String {
    static CAPACITY: usize = 1024 * 1024 * 500; // 500 meg
    let mut s = String::with_capacity(CAPACITY);
    while s.len() < CAPACITY {
        s.push_str(bar)
    }
    s
}
```


What's happening here is that when `myfunc` is called and returns that 500 meg `String`, it's converted into a postgres-allocated (via `palloc`) `text` datum.  Then `fn_call` takes that and converts it back into an owned `String`.  We're going through Postgres' internal function calling interfaces here, and things are passed around as datums.

Anyways, that, what is now an intermediate palloc'd `text` datum, gets lost.  pgrx can't know to free it when it returns from `fn_call` because it doesn't really know where it came from.

As such, it's the user's responsibility to manage that memory, in an `unsafe` manner using (at least) a transient memory context, like so:

```rust
#[pg_extern]
fn tester() {
    let bar = "bar";
    for _ in 0..10000000 {
        check_for_interrupts!();

        unsafe {
[...]

PR #1342, opened 2023-10-18
Add `MemCx<'mcx>` for `List<'mcx, T>`

workingjubilee:
List requires reallocation in order to push to it with safe Rust, but it is defined as allocated via palloc, which has problems with lifetimes. In order to head off these problems, and begin work on related ones, introduce a type describing a MemoryContext with a lifetime, including scoped (therefore lifetime-bound) access to the current context.


pgrx/src/pg_sys.rs
```diff
@@ -14,11 +15,13 @@ pub use pgrx_pg_sys::*;
...
-        .get((index - 1) as _)
-        .expect("rt_fetch used out-of-bounds")
-        .cast()
+    mem::current_context(|cx| {
+        crate::list::List::<*mut core::ffi::c_void>::downcast_ptr_in_memcx(range_table, cx)
+            .expect("rt_fetch used on non-ptr List")
+            .get((index - 1) as _)
+            .expect("rt_fetch used out-of-bounds")
+            .cast()
+    })
```
workingjubilee: This diff suggests to me my approach is kinda incorrect, since I don't really "use" the context.

pgrx/src/mem.rs
```diff
@@ -0,0 +1,39 @@
...
+
+/// Acquire the current context and operate inside it.
+pub fn current_context<'curr, F, T>(f: F) -> T
+where
+    F: for<'clos> FnOnce(&'clos MemCx<'curr>) -> T,
+{
+    let memcx = unsafe { MemCx::from_ptr(pg_sys::CurrentMemoryContext) };
+    let ret = { f(&memcx) };
+    ret
+}
```
workingjubilee: Definitely not sure about soundness, but something similar to this is probably required, for ergonomics reasons.

pgrx/src/list/flat_list.rs
```diff
@@ -357,25 +361,25 @@ unsafe fn destroy_list(list: *mut pg_sys::List) {
...
-    list: List<T>,
+pub struct ListIter<'a, T> {
+    list: List<'a, T>,
     iter: RawCellIter<T>,
 }
 
 /// A list being drained.
 #[derive(Debug)]
-pub struct Drain<'a, T> {
+pub struct Drain<'a, 'cx, T> {
```
workingjubilee, line 372: Really need to move Drain itself "up".

Also: two lifetimes! spooky.

workingjubilee (commented):
UX thoughts: I probably didn't write the interfaces correctly. Might need redesign.

Wondered about maybe instead using something like `List<T, M>`, where `M` might be the thing infected with the lifetime? idk.

eeeebbbbrrrr:
Just a note to say I've skimmed over this.  I don't have any feedback yet other than it's exciting to see some code demonstrating the ideas.

pgrx-tests/src/tests/list_tests.rs
```diff
@@ -13,44 +13,44 @@ mod tests {
...
-        assert_eq!(50, list.len());
+            // Want to make sure the list length updates properly in the three major drain cases:
+            // from start of list, from inside the middle of the list, and from middle to tail.
+            let _ = list.drain(0..10);
+            assert_eq!(90, list.len());
+            let _ = list.drain(10..30);
+            assert_eq!(70, list.len());
+            let _ = list.drain(50..);
+            assert_eq!(50, list.len());
+        })
```
workingjubilee: Making sure Drain worked here was the most annoying thing to fix, since it wasn't intuitive until actually trialed, but was just an amount of lifetime poking so it was pretty okay, and our own implementation primitives are allowed to be messy to work with, I think: I don't expect anyone will hold on to a Drain for long.

pgrx/src/pg_sys.rs
workingjubilee: Yeah, this is probably going to be the most annoying part here. Maybe this would be better captured as another parameter to List, so non-allocating but owning uses don't have to worry about this.

pgrx/src/fn_call.rs
```diff
@@ -336,69 +337,79 @@ fn lookup_fn(fname: &str, args: &[&dyn FnCallArg]) -> Result<pg_sys::Oid> {
...
-        // SAFETY:  we've allocated a PgList with the proper String node elements representing its name
-        // and we've allocated Vec of argument type oids which can be represented as a pointer.
-        let mut fnoid =
-            pg_sys::LookupFuncName(parts_list.as_mut_ptr(), nargs as _, arg_types.as_ptr(), true);
-
-        if fnoid == pg_sys::InvalidOid {
-            // if that didn't find a function, maybe we've got some defaults in there, so do a lookup
-            // where Postgres will consider that
-            fnoid = pg_sys::LookupFuncName(
+    mem::current_context(|mcx| {
```
workingjubilee: The need to do this, and the problems with memory leaks in #1341, strongly suggests we should be accepting a MemCx argument to this function and also the public interface of `fn_call`.

thomcc (approved):
This is a good improvement.

workingjubilee:
Rebased and added _some_ (~300 words) of documentation on memory contexts. I think I will land this now, so that it's easier to build off of and mutate as-we-go.

PR #1346, opened 2023-10-20
Avoid colons in filenames

passcod:
Colons (`:`) are invalid in filenames on Windows, which makes it impossible (or at least unpleasant) to checkout the repo.

This PR uses underscores instead, but happy to switch to any other [allowed character](https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names).


workingjubilee:
Thank you!

workingjubilee:
Huh, that's odd, I didn't expect this to fail:
https://github.com/pgcentralfoundation/pgrx/actions/runs/6602309588/job/17934216699

passcod:
Oh, I missed one replace I think.

workingjubilee:
yup, picked up in https://github.com/pgcentralfoundation/pgrx/pull/1350

PR #1347, opened 2023-10-20
Untar directly in rust

passcod:
Good: doesn't require tar and bzip2 cli tools
Bad: more rust dependencies I guess

This PR also changes the untar process to be a little more resistant to unexpected crashes (or power failures), downloading and unpacking to a staging directory before moving the new sources in place. Happy to split that into its own PR if you want.


workingjubilee:
I am not immediately sure if we want to accept this (probably? I'd want to check with the others first), but I know I'd click the merge button on a making-the-process-more-resilient PR.

workingjubilee (approved):
Talked to @thomcc and he says this means we're compiling in and shipping libbzip2, but that seems fine. Preferable, even.

passcod:
Rebased

PR #1348, opened 2023-10-20
(cargo-pgrx init) Use toml serializer to write config.toml

passcod:
Writing the toml by formatting strings meant that on Windows, backslashes in paths weren't escaped, outputting invalid toml. This uses the `toml` crate and the existing `ConfigToml` struct to read/write the file intead.

---

Also in the readme it says

> If a new minor Postgres version is released in the future you can simply run `cargo pgrx init [args]` again, and your local version will be updated, preserving all existing databases and configuration.

That was a bit confusing to me, as e.g. running:

```console
$ cargo pgrx init --pg12 download --pg13 download
```

Followed by

```console
$ cargo pgrx init --pg13 download --pg14 download
```

Would "erase" the pg12 checkout from the config.toml file, even though the files remained there.

I took that as accidental and fixed it (ie, we read the config and modify it) in this PR, but of course can revert that bit if it was intentional.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -357,6 +357,7 @@ dependencies = [
  "serde_derive",
  "syn 2.0.38",
  "tempfile",
+ "toml",
  "tracing",
  "tracing-error",
  "tracing-subscriber",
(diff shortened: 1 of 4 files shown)


workingjubilee:
That seems to be correct!

PR #1350, opened 2023-10-23
Fixes nightly "cargo unlocked" tests

BradyBonnette:


BradyBonnette:
@workingjubilee See here for an example of all the "cargo unlocked" tests running: https://github.com/pgcentralfoundation/pgrx/actions/runs/6618439757

Once this is merged into develop, they should all run on the next scheduled time (which I think is ~ 3:00 am eastern).

workingjubilee:
Oh. Duh.

PR #1352, opened 2023-10-24
install --sudo destination path must exist to be canonicalized

levkk:
`fs::canonicalize` calls `realpath` which requires the dest path to exist. It doesn't when installing the extension for the first time (and a new version of the extension for subsequent times). So...best we can [1] do is trust that @eeeebbbbrrrr 's canonicalization works . :rocket: 

[1]: https://stackoverflow.com/questions/68231306/stdfscanonicalize-for-files-that-dont-exist


workingjubilee:
Oh, amusing. Thank you!

PR #1354, opened 2023-10-27
yeet IntoDatum for u32

workingjubilee:
impl from when `pg_sys::Oid` was not a newtype but an alias.

now unnecessary.


workingjubilee:
Alternatively, we could also implement `u{8,16,64}`, but I believe we should support those as integers instead of having u32 report `OIDOID`, and it might surface bugs due to unexpected "overflows" in places.

thomcc (approved):

PR #1355, opened 2023-10-27
Weaken bounds on Array impls

workingjubilee:
Weaken the bounds on various implementations for Array, to make things easier to refactor. In that vein, clean up a function from on IntoDatum that can be inside Array's IntoDatum impl.

This weakening is mostly irrelevant because you still can't construct such an Array.

### DIFF
--- a/pgrx/src/datum/array.rs
+++ b/pgrx/src/datum/array.rs
@@ -55,7 +55,7 @@ fn with_vec(elems: Array<String>) {
 }
 ```
 */
-pub struct Array<'a, T: FromDatum> {
+pub struct Array<'a, T> {
     null_slice: NullKind<'a>,
     slide_impl: ChaChaSlideImpl<T>,
     // Rust drops in FIFO order, drop this last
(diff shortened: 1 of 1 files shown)


pgrx/src/datum/into.rs
```diff
@@ -34,9 +34,6 @@ pub trait IntoDatum {
     fn composite_type_oid(&self) -> Option<Oid> {
         None
     }
-    fn array_type_oid() -> pg_sys::Oid {
-        unsafe { pg_sys::get_array_type(Self::type_oid()) }
-    }
```
workingjubilee: This is technically a breaking change.

workingjubilee:
Actually, I think I'll separate that out.

pgrx/src/datum/array.rs
```diff
@@ -275,6 +243,41 @@ impl<'a, T: FromDatum> Array<'a, T> {
...
+    /// and this can hypothetically cause a memory leak if so.
+    #[inline]
+    pub fn into_array_type(self) -> *const pg_sys::ArrayType {
+        // may be worth replacing this function when Toast<T> matures enough
+        // to be used as a public type with a fn(self) -> Toast<RawArray>
+
+        let Array { raw, .. } = self;
+        // Wrap the Toast<RawArray> to prevent it from deallocating itself
+        let mut raw = core::mem::ManuallyDrop::new(raw);
+        let ptr = raw.deref_mut().deref_mut() as *mut RawArray;
```
thomcc, line 259: you didn't add this, but lol.

thomcc (approved):

pgrx/src/datum/array.rs
workingjubilee: :see_no_evil:

PR #1357, opened 2023-10-28
Add `clippy -Awarnings` to CI and fix unsound fn

workingjubilee:
I have resolved to grumble less about clippy: it detected **serious soundness issues related to the implementation of `pg_aggregate`** after I ran it over the repo (ignoring all the warnings, of course). These functions are publicly reachable but `unsafe` to use, as they may accept and dereference e.g. null pointers.

Add clippy to CI, but allow all warnings, so it only monitors correctness lints. Mark the relevant unsound `pub fn` as `unsafe`, catching:
- `aggregate::Aggregate::in_memory_context`
- `htup::heap_tuple_get_datum`

### DIFF
--- a/pgrx-sql-entity-graph/src/aggregate/mod.rs
+++ b/pgrx-sql-entity-graph/src/aggregate/mod.rs
@@ -299,10 +299,12 @@ impl PgAggregate {
                 #[allow(non_snake_case, clippy::too_many_arguments)]
                 #pg_extern_attr
                 fn #fn_name(this: #type_state_without_self, #(#args_with_names),*, fcinfo: ::pgrx::pg_sys::FunctionCallInfo) -> #type_state_without_self {
-                    <#target_path as ::pgrx::aggregate::Aggregate>::in_memory_context(
-                        fcinfo,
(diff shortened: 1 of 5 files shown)


pgrx-sql-entity-graph/src/aggregate/mod.rs
```diff
@@ -299,10 +299,12 @@ impl PgAggregate {
...
-                    <#target_path as ::pgrx::aggregate::Aggregate>::in_memory_context(
-                        fcinfo,
-                        move |_context| <#target_path as ::pgrx::aggregate::Aggregate>::state(this, (#(#arg_names),*), fcinfo)
-                    )
+                    unsafe {
+                        <#target_path as ::pgrx::aggregate::Aggregate>::in_memory_context(
+                            fcinfo,
+                            move |_context| <#target_path as ::pgrx::aggregate::Aggregate>::state(this, (#(#arg_names),*), fcinfo)
+                        )
+                    }
```
workingjubilee, line 307: I honestly am not entirely sure if this is actually correct to assert, either, but I **think** this fn in practice only gets called by Postgres, making this one of the many undocumented `unsafe` blocks to audit later. ^^;

PR #1365, opened 2023-10-31
Remove Postgres 11 support

workingjubilee:
pgrx 0.12 will also require Postgres 12.


cargo-pgrx/README.md
```diff
@@ -146,6 +146,7 @@ Options:
                                                to have pgrx download/compile/install it [env: PG12_PG_CONFIG=]
       --pg13 <PG13>                            If installed locally, the path to PG13's `pgconfig` tool, or `download`
                                                to have pgrx download/compile/install it [env: PG13_PG_CONFIG=]
+  -v, --verbose...                             Enable info logs, -vv for debug, -vvv for trace
```
workingjubilee, line 149: I have no idea why this jumped up here but I am not fixing it right now.

cargo-pgrx/README.md
workingjubilee, line 1: This entire file has a bunch of README changes in which I actually ran the command and pasted in what I saw.

In many cases the output had diverged, significantly. It might be a good idea to think about a better way to keep these synced.

PR #1371, opened 2023-11-01
Use one 🦭 per crate

workingjubilee:
I discovered we had two Sealed traits inside pgrx while doing some rework, so I decided to unify them, as we don't actually need two. I also believe that the use of the pgrx::fn_call::seal::Sealed trait was incorrect: it was binding the type T, but with a blanket impl. This means downstream crates could impl FnCallArg for types other than Arg, as long as they had first implemented Clone (which they obviously can) and IntoDatum (which remains a public trait).

This didn't occur to me the first time I saw FnCallArg because I hadn't arrived at my current interpretation of what "This is only implemented for the Arg enum" was supposed to mean: This _must_ only be implemented for the Arg enum!

I have corrected the bounds according to what I believe was intended.

### DIFF
--- a/pgrx/src/fn_call.rs
+++ b/pgrx/src/fn_call.rs
@@ -14,19 +14,14 @@ use pgrx_pg_sys::PgTryBuilder;
 use std::panic::AssertUnwindSafe;
 
 use crate::pg_catalog::pg_proc::{PgProc, ProArgMode, ProKind};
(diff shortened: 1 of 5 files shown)


thomcc (approved):

workingjubilee:
( very hypothetically. )

PR #1374, opened 2023-11-03
Allow safe u32 -> Oid conversions

workingjubilee:
A while back Thom discovered this, but we didn't open the PR at the time.

Simply cast `u32 as i32` and pass it through SPI to do this by other means. While it is an incredibly bad idea to do so, unfortunately that means we cannot rely on the conversion being `unsafe`, as there is no way we would make SPI unsafe. That leaves no persuasive argument for not providing From.


workingjubilee:
wait, no it doesn't break. what was I thinking?

PR #1375, opened 2023-11-03
Update fcinfo fn toward actual usage

workingjubilee:
Functions that had no downstream users, some odd misdeclarations, and in one case a function we can expose finally as public because it's actually what we want people to be using, going forward (probably).

### DIFF
--- a/pgrx/src/fcinfo.rs
+++ b/pgrx/src/fcinfo.rs
@@ -231,24 +231,6 @@ pub unsafe fn pg_getarg_type(fcinfo: pg_sys::FunctionCallInfo, num: usize) -> pg
     pg_sys::get_fn_expr_argtype(fcinfo.as_ref().unwrap().flinfo, num as std::os::raw::c_int)
 }
 
-/// This is intended for Postgres functions that take an actual `cstring` argument, not for getting
-/// a varlena argument type as a CStr.
-///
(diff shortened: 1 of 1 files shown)


pgrx/src/fcinfo.rs
```diff
@@ -240,17 +240,11 @@ pub unsafe fn pg_getarg_type(fcinfo: pg_sys::FunctionCallInfo, num: usize) -> pg
...
 /// }
 ///```
-///
-/// # Safety
-///
-/// This function is unsafe for symmetry with the other related functions that deal with
-/// `PG_FUNCTION_INFO_V1` functions.  It has no specific safety invariants that must be met.
 #[inline]
-pub unsafe fn pg_return_void() -> pg_sys::Datum {
+pub fn pg_return_void() -> pg_sys::Datum {
```
workingjubilee, line 227: I should note, despite making this change, it's always better to be able to relax an unsafety than apply one, as it's (mostly) non-breaking to relax it.

thomcc (approved):
SGTM, I'm totally fine cutting things if it reduces the number of things we need to worry about while figuring out the soundness design changes.

eeeebbbbrrrr (approved):

eeeebbbbrrrr:
Go for it

PR #1378, opened 2023-11-06
Fix timezone test after daylight savings shift

thomcc:
This test fails for me locally because the offset is -18000 now. Maybe we should be more specific about the timezone we're asking about, but this seems fine too.


BlackHoleFox:
This fixed my issue in WSL2, many thanks.

Issue #1351, opened 2023-10-24
Could not start server; shmem too large


kczimm:
I'm having trouble starting the server when the size of shared memory is beyond a certain size (not particularly large). If I take the [shmem example](https://github.com/pgcentralfoundation/pgrx/tree/develop/pgrx-examples/shmem) and simply increase the size of the `Deque` to ~8MB, the server fails to start. 

```rust
static DEQUE: PgLwLock<heapless::Deque<Pgtest, 1_000_000>> = PgLwLock::new();
```

If I set the log level to `debug5`, the log says the following:

```
2023-10-24 16:14:10.031 UTC [3966800] DEBUG:  find_in_dynamic_libpath: trying "/home/ubuntu/.pgrx/15.4/pgrx-install/lib/postgresql/shmem"
2023-10-24 16:14:10.031 UTC [3966800] DEBUG:  find_in_dynamic_libpath: trying "/home/ubuntu/.pgrx/15.4/pgrx-install/lib/postgresql/shmem.so"
2023-10-24 16:14:10.031 UTC [3966800] DEBUG:  loaded library "shmem"
2023-10-24 16:14:10.031 UTC [3966800] DEBUG:  invoking IpcMemoryCreate(size=1129136128)
2023-10-24 16:14:10.031 UTC [3966800] DEBUG:  mmap(1130364928) with MAP_HUGETLB failed, huge pages disabled: Cannot allocate memory
2023-10-24 16:14:10.052 UTC [3966800] DEBUG:  dynamic shared memory system will support 224 segments
2023-10-24 16:14:10.052 UTC [3966800] DEBUG:  created dynamic shared memory control segment 2541292408 (8976 bytes)
```

## Note
I used the feature flag `pg15` simply because I already had my `postgresql.conf` setup for it. I've tried experimenting with `shared_buffers` size but like I said, this isn't even that large of shared memory. My `kernel.shmmax` and `kernel.shmall` settings are much larger than what I'm requesting also.

If anyone knows what is going on here, it would be much appreciated.

kczimm:
After further experimentation, I enabled `hugepages` with `sysctl` which made that debug print disappear from the log, but the server still fails to start.

kczimm:
https://kaiwantech.wordpress.com/2011/08/17/kmalloc-and-vmalloc-linux-kernel-memory-allocation-api-limits/

[...]

Issue #1223, opened 2023-07-21
PgLwLock breaks postgres reinitialization after backend crash


JelteF:
Normally when a backend crashes with a segfault, postmaster will kill all other backends and then reinitialize everything. But this breaks when using `PgLwLock`, because shared memory initialization of the extension is executed again and then PgLwLock complains that it cannot attach.
```
2023-07-21 16:08:56.754 CEST [538131] LOG:  background worker "Some background worker" (PID 538137) was terminated by signal 11: Segmentation fault
2023-07-21 16:08:56.755 CEST [538131] LOG:  all server processes terminated; reinitializing
2023-07-21 16:08:56.924 CEST [538131] FATAL:  Can't attach, lock is not in an empty state: PgLwLockInner
```

JelteF:
Easiest way to reproduce is to run the `shmem` example (which requires putting `shmem` in `shared_preload_libraries`). Then `kill -11 <random pg backend>`.

JelteF:
Backtrace in case its helpful:
```
2023-07-21 16:40:02.995 CEST [550346] LOG:  all server processes terminated; reinitializing
2023-07-21 16:40:03.125 CEST [550346] FATAL:  Can't attach, lock is not in an empty state: PgLwLockInner
2023-07-21 16:40:03.125 CEST [550346] DETAIL:
       0: pgrx_pg_sys::submodules::panic::register_pg_guard_panic_hook::{{closure}}::{{closure}}
                 at /home/jelte/opensource/pgrx/pgrx-pg-sys/src/submodules/panic.rs:311:39
       1: std::thread::local::LocalKey<T>::try_with
                 at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/std/src/thread/local.rs:270:16
       2: std::thread::local::LocalKey<T>::with
                 at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/std/src/thread/local.rs:246:9
       3: pgrx_pg_sys::submodules::panic::register_pg_guard_panic_hook::{{closure}}
                 at /home/jelte/opensource/pgrx/pgrx-pg-sys/src/submodules/panic.rs:308:9
       4: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
                 at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/alloc/src/boxed.rs:1999:9
[...]

PR #1379, opened 2023-11-07
Shared memory hash table

levkk:
Shared memory hash table. Implementation based (pretty much copied from) on `pg_stat_statements`. Subject to the same shared memory limitations as the rest of shared memory data structures, but still cool to have a native Postgres HashMap implementation in shared memory.


levkk:
```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x9)
    frame #0: 0x00000001a00f7b38 libsystem_platform.dylib`_platform_strcmp + 8
libsystem_platform.dylib`_platform_strcmp:
->  0x1a00f7b38 <+8>:  ldrb   w4, [x0], #0x1
    0x1a00f7b3c <+12>: ldrb   w5, [x1], #0x1
    0x1a00f7b40 <+16>: subs   x3, x4, x5
    0x1a00f7b44 <+20>: ccmp   w4, #0x0, #0x4, eq
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x9)
  * frame #0: 0x00000001a00f7b38 libsystem_platform.dylib`_platform_strcmp + 8
    frame #1: 0x000000019ffcd310 libsystem_c.dylib`hash_search + 116
    frame #2: 0x000000010191dfbc shmem.so`{closure#0} at pg13.rs:27674:1
    frame #3: 0x00000001019079f0 shmem.so`hash_search [inlined] pg_guard_ffi_boundary_impl<*mut core::ffi::c_void, pgrx_pg_sys::include::pg13::hash_search::{closure_env#0}> at ffi.rs:120:26
    frame #4: 0x0000000101907930 shmem.so`hash_search [inlined] pg_guard_ffi_boundary<*mut core::ffi::c_void, pgrx_pg_sys::include::pg13::hash_search::{closure_env#0}> at ffi.rs:94:14
    frame #5: 0x0000000101907930 shmem.so`hash_search at pg13.rs:27674:1
    frame #6: 0x00000001018ef51c shmem.so`insert at shmem_hash.rs:59:13
    frame #7: 0x00000001018a3d4c shmem.so`hash_table_insert at lib.rs:68:5
    frame #8: 0x00000001018a3dd4 shmem.so`hash_table_insert_wrapper_inner at lib.rs:67:1
    frame #9: 0x00000001018cc810 shmem.so`{closure#0} at lib.rs:67:1
    frame #10: 0x00000001018d6e0c shmem.so`do_call<shmem::hash_table_insert_wrapper::{closure_env#0}, ()> at panicking.rs:500:40
    frame #11: 0x00000001018d7b14 shmem.so`__rust_try + 32
    frame #12: 0x00000001018d6ab8 shmem.so`try<(), shmem::hash_table_insert_wrapper::{closure_env#0}> at panicking.rs:464:19
    frame #13: 0x00000001018aa644 shmem.so`catch_unwind<shmem::hash_table_insert_wrapper::{closure_env#0}, ()> at panic.rs:142:14
    frame #14: 0x00000001018bcdd0 shmem.so`run_guarded<shmem::hash_table_insert_wrapper::{closure_env#0}, ()> at panic.rs:408:11
    frame #15: 0x00000001018bff10 shmem.so`pgrx_extern_c_guard<shmem::hash_table_insert_wrapper::{closure_env#0}, ()> at panic.rs:385:11
    frame #16: 0x00000001018a3d84 shmem.so`hash_table_insert_wrapper at lib.rs:67:1
    frame #17: 0x0000000100d174b4 postgres`ExecInterpExpr(state=0x0000000145018970, econtext=<unavailable>, isnull=<unavailable>) at execExprInterp.c:704:8 [opt]
    frame #18: 0x0000000100d4a7d8 postgres`ExecResult [inlined] ExecEvalExprSwitchContext(state=0x0000000145018970, econtext=0x0000000145018670, isNull=0x000000016f275e4f) at executor.h:322:13 [opt]
    frame #19: 0x0000000100d4a7b0 postgres`ExecResult [inlined] ExecProject(projInfo=0x0000000145018968) at executor.h:356:9 [opt]
    frame #20: 0x0000000100d4a794 postgres`ExecResult(pstate=<unavailable>) at nodeResult.c:136:10 [opt]
    frame #21: 0x0000000100d1e3d0 postgres`standard_ExecutorRun [inlined] ExecProcNode(node=0x0000000145018558) at executor.h:248:9 [opt]
    frame #22: 0x0000000100d1e3b4 postgres`standard_ExecutorRun [inlined] ExecutePlan(estate=0x0000000145018320, planstate=0x0000000145018558, use_parallel_mode=<unavailable>, operation=CMD_SELECT, sendTuples=<unavailable>, numberTuples=0, direction=<unavailable>, dest=0x0000000135009158, execute_once=<unavailable>) at execMain.c:1632:10 [opt]
    frame #23: 0x0000000100d1e388 postgres`standard_ExecutorRun(queryDesc=0x000000014500eb20, direction=<unavailable>, count=0, execute_once=<unavailable>) at execMain.c:350:3 [opt]
    frame #24: 0x0000000100d1e2c0 postgres`ExecutorRun(queryDesc=<unavailable>, direction=<unavailable>, count=<unavailable>, execute_once=<unavailable>) at execMain.c:294:3 [opt] [artificial]
    frame #25: 0x0000000100e9847c postgres`PortalRunSelect(portal=0x000000014682d920, forward=<unavailable>, count=0, dest=0x0000000135009158) at pquery.c:921:4 [opt]
    frame #26: 0x0000000100e980b4 postgres`PortalRun(portal=0x000000014682d920, count=9223372036854775807, isTopLevel=<unavailable>, run_once=<unavailable>, dest=0x0000000135009158, altdest=0x0000000135009158, qc=0x000000016f2760e8) at pquery.c:765:18 [opt]
    frame #27: 0x0000000100e972e4 postgres`exec_simple_query(query_string="select hash_table_insert(10, 9);") at postgres.c:1238:10 [opt]
    frame #28: 0x0000000100e94808 postgres`PostgresMain(argc=<unavailable>, argv=<unavailable>, dbname=<unavailable>, username=<unavailable>) at postgres.c:0 [opt]
    frame #29: 0x0000000100e11500 postgres`BackendRun(port=0x0000000144f047e0) at postmaster.c:4557:2 [opt]
    frame #30: 0x0000000100e10c20 postgres`ServerLoop [inlined] BackendStartup(port=0x0000000144f047e0) at postmaster.c:4241:3 [opt]
    frame #31: 0x0000000100e10bfc postgres`ServerLoop at postmaster.c:1739:7 [opt]
    frame #32: 0x0000000100e0dff8 postgres`PostmasterMain(argc=<unavailable>, argv=<unavailable>) at postmaster.c:1412:11 [opt]
    frame #33: 0x0000000100d69b34 postgres`main(argc=8, argv=0x0000600001b30dc0) at main.c:210:3 [opt]
    frame #34: 0x000000019fd73f28 dyld`start + 2236
(lldb)
```

levkk:
@thomcc Hi Thom, I'm getting a segfault here and the backtrace (comment above) is showing that the linker picked up the wrong `hash_search` function. Instead of using the one in postgres, it's using the one in Mac OS's libc. Any ideas?

To repro:

1. `cargo pgrx run` the shmem example
2. `SELECT pg_backend_pid()`
3. `sudo lldb -p <backend pid>`
4. `SELECT hash_table_insert(10, 9)` _twice_. The first time, it'll be ok, the second you should get a segfault. Basically, calling the wrong `hash_search` function twice results in bad memory access.

Thank you!

pgrx/src/shmem_hash.rs
```diff
@@ -0,0 +1,220 @@
...
+            htab: PgLwLock::new(),
+            size,
+            phantom_key: std::marker::PhantomData,
+            phantom_value: std::marker::PhantomData,
+        }
+    }
+
+    /// Insert a key and value into the HashMap. If the key is already
+    /// present, it will be replaced. If the HashMap is full, return an error.
+    pub fn insert(&self, key: K, value: V) -> Result<(), Error> {
```
kczimm: If the map is full, the key and value should be returned.

workingjubilee:
```
  Error: bindgen failed for pg16

  Caused by:
     0: Unable to generate bindings for pg16
     1: clang diagnosed error: /usr/lib/llvm-14/lib/clang/14.0.0/include/emmintrin.h:2228:10: error: invalid conversion between vector type '__m128i' (vector of 2 'long long' values) and integer type 'int' of different size
        /usr/lib/llvm-14/lib/clang/14.0.0/include/emmintrin.h:2250:10: error: invalid conversion between vector type '__m128i' (vector of 2 'long long' values) and integer type 'int' of different size
        /usr/lib/llvm-14/lib/clang/14.0.0/include/emmintrin.h:2271:10: error: invalid conversion between vector type '__m128i' (vector of 2 'long long' values) and integer type 'int' of different size
        /usr/lib/llvm-14/lib/clang/14.0.0/include/emmintrin.h:2292:10: error: invalid conversion between vector type '__m128i' (vector of 2 'long long' values) and integer type 'int' of different size
        /usr/lib/llvm-14/lib/clang/14.0.0/include/emmintrin.h:2670:10: error: invalid conversion between vector type '__m128i' (vector of 2 'long long' values) and integer type 'int' of different size
        /usr/lib/llvm-14/lib/clang/14.0.0/include/emmintrin.h:2691:10: error: invalid conversion between vector type '__m128i' (vector of 2 'long long' values) and integer type 'int' of different size
        /usr/lib/llvm-14/lib/clang/14.0.0/include/emmintrin.h:2711:10: error: invalid conversion between vector type '__m128i' (vector of 2 'long long' values) and integer type 'int' of different size
        /usr/lib/llvm-14/lib/clang/14.0.0/include/emmintrin.h:2731:10: error: invalid conversion between vector type '__m128i' (vector of 2 'long long' values) and integer type 'int' of different size


  Location:
      pgrx-pg-sys/build.rs:733:10
```
not this shit again.

levkk:
Yeah I also lied about this being infinitely large...still hitting `out of memory` when allocating more than 512kb or so. Gonna keep debugging to find out why. Probably some shared memory limit constant somewhere?

workingjubilee:
we got https://github.com/pgcentralfoundation/pgrx/issues/1351 over that, it's weird.

workingjubilee:
> Replace PgLwLock with PgSpinLock. However, PgSpinLock constructor isn't const, not sure why. I don't see it used anywhere, so might not be ready yet?

Please feel free to make PgSpinLock::new a `const fn`.

levkk:
> > Replace PgLwLock with PgSpinLock. However, PgSpinLock constructor isn't const, not sure why. I don't see it used anywhere, so might not be ready yet?
> 
> Please feel free to make PgSpinLock::new a `const fn`.

Will do!

levkk:
Yes, I should finish this...

workingjubilee:
I just kicked all the open PRs for CI reasons, don't worry~

levkk:
Looks like `cargo fmt` in `develop` needs to be run.

workingjubilee:
Whoops.
But also
> 111 | pub use shmem_hash::*;

levkk:
Ah forgot a `cfg`. One moment.

levkk:
Look at that build queue! $$$

pgrx/src/shmem_hash.rs
```diff
@@ -0,0 +1,239 @@
+//! Shared memory hash map implemented with Postgres' internal `HTAB`,
+//! which is used by other extensions like `pg_stat_statements`.
+use crate::{pg_sys, shmem::PgSharedMemoryInitialization, spinlock::*, PGRXSharedMemory};
+use once_cell::sync::OnceCell;
+use std::ffi::c_void;
+use uuid::Uuid;
+
+#[derive(Debug, Eq, PartialEq)]
+#[non_exhaustive]
+pub enum Error {
```
workingjubilee: A public error type shall be given a name that does not collide with the name of the Error trait.

pgrx/src/shmem_hash.rs
```diff
@@ -0,0 +1,239 @@
...
+impl Default for PgHashMapInner {
+    fn default() -> Self {
+        Self { htab: std::ptr::null_mut(), elements: 0 }
+    }
+}
+
+/// A shared memory HashMap using Postgres' `HTAB`.
+/// This HashMap is used for `pg_stat_statements` and Postgres
+/// internals to store key/value pairs in shared memory.
+pub struct PgHashMap<K: Copy + Clone, V: Copy + Clone> {
```
workingjubilee: The type for new APIs shall not be prefixed with "Pg". It's Postgres, everything's Pg everywhere, all these types are namespaced via `pgrx::`, and this style is inconsistently used at best (Array vs. PgHeapTuple, for instance).

workingjubilee: ShmemHashMap would be fine.

levkk:
Now I feel bad! Maybe we could have just one job sniff testing basics before launching the whole test suite?

pgrx/src/shmem_hash.rs
```diff
@@ -0,0 +1,239 @@
...
+        let value_ptr: *mut Value<K, V> = $entry as *mut Value<K, V>;
+
+        value_ptr
+    }};
+}
+
+impl<K: Copy + Clone, V: Copy + Clone> PgHashMap<K, V> {
+    /// Create new `PgHashMap`. This still needs to be allocated with
+    /// `pg_shmem_init!` just like any other shared memory structure.
+    pub const fn new(size: u64) -> PgHashMap<K, V> {
```
workingjubilee: It's common for `::new()` to be semantically equivalent to `::default()`, so the size parameter's purpose should be documented rather than left implicit, if we require it.

pgrx/src/spinlock.rs
```diff
@@ -28,6 +28,7 @@ use std::{cell::UnsafeCell, marker::PhantomData};
 /// [`storage/spin.h`]:
 ///     https://github.com/postgres/postgres/blob/1f0c4fa255253d223447c2383ad2b384a6f05854/src/include/storage/spin.h
 #[doc(alias = "slock_t")]
+#[derive(Debug)]
```
workingjubilee: This will just say
```
PgSpinLock { item: UnsafeCell { .. }, lock: UnsafeCell { .. } }
```

Please impl Debug explicitly with something more concise.

pgrx/src/shmem_hash.rs
workingjubilee: It should also be made clear why it's a u64 instead of the more common usize. Yes, the difference "doesn't matter", but the difference between those two types informs what purpose it serves.

workingjubilee:
> Now I feel bad! Maybe we could have just one job sniff testing basics before launching the whole test suite?

I certainly would not reject PRs improving our test suite's economy of CPU cycles!

pgrx/src/shmem_hash.rs
```diff
@@ -82,21 +82,30 @@ macro_rules! value_ptr {
...
     /// `pg_shmem_init!` just like any other shared memory structure.
-    pub const fn new(size: u64) -> PgHashMap<K, V> {
-        PgHashMap {
+    ///
+    /// # Arguments
+    ///
+    /// * `size` - Maximum number of elements in the HashMap. This is allocated
+    ///            at server start and cannot be changed. `i64` is the expected type
+    ///            for `pg_sys::ShmemInitHash`, so we don't attempt runtime conversions
+    ///            unnecessarily.
```
workingjubilee, line 94: ...okay, "you only can allocate this once ever" is pretty coherent as to why we need it, yeah. Though I believe a runtime conversion that happens once ever is... fine?

pgrx/src/shmem_hash.rs
```diff
@@ -203,15 +212,17 @@ impl<K: Copy + Clone, V: Copy + Clone> PgHashMap<K, V> {
     }
 
     /// Get the number of elements in the HashMap.
-    pub fn len(&self) -> u64 {
+    pub fn len(&self) -> i64 {
```
workingjubilee, line 215: This reminds me, this needs `capacity(&self)` as well.

pgrx/src/shmem_hash.rs
```diff
@@ -0,0 +1,250 @@
...
+            let value_ptr = value_ptr!(entry);
+            let value = unsafe { std::ptr::read(value_ptr) };
+            return Some(value.value);
+        }
+    }
+
+    /// Remove the value from the `ShmemHashMap` and return it.
+    /// If the key doesn't exist, return None.
+    pub fn remove(&self, key: K) -> Option<V> {
+        if let Some(value) = self.get(key) {
```
workingjubilee, line 192: This should probably use let-else, like so:
```suggestion
        let Some(value) = self.get(key) else { return None };
```

pgrx/src/shmem_hash.rs
```diff
@@ -0,0 +1,250 @@
...
+                std::ptr::null_mut(),
+            )
+        };
+
+        if entry.is_null() {
+            return None;
+        } else {
+            let value_ptr = value_ptr!(entry);
+            let value = unsafe { std::ptr::read(value_ptr) };
+            return Some(value.value);
```
workingjubilee, line 185:
```suggestion
            Some(value.value)
```

pgrx/src/shmem_hash.rs
```diff
@@ -0,0 +1,250 @@
...
+                htab.htab,
+                key_ptr,
+                hash_value,
+                pg_sys::HASHACTION_HASH_FIND,
+                std::ptr::null_mut(),
+            )
+        };
+
+        if entry.is_null() {
+            return None;
```
workingjubilee, line 181:
```suggestion
            None
```

pgrx/src/shmem_hash.rs
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    /// Get a value from the HashMap using the key.
+    /// If the key doesn't exist, return `None`.
+    pub fn get(&self, key: K) -> Option<V> {
+        let htab = self.htab.get().unwrap().lock();
+        let (key_ptr, hash_value) = key!(key, htab);
+
+        let entry = unsafe {
+            pg_sys::hash_search_with_hash_value(
```
workingjubilee, line 171: Please document the preconditions for calling this function somewhere in this file, you can reference them in later `unsafe` calls to the same function.

pgrx/src/shmem_hash.rs
```diff
@@ -0,0 +1,250 @@
...
+                htab.htab,
+                key_ptr,
+                hash_value,
+                pg_sys::HASHACTION_HASH_FIND,
+                std::ptr::null_mut(),
```
workingjubilee, line 176: In particular, it is good to know why passing nullptr is okay here.

pgrx/src/shmem_hash.rs
```diff
@@ -0,0 +1,250 @@
...
+            let value_ptr = value_ptr!(entry);
+            let value = unsafe { std::ptr::read(value_ptr) };
+            Some(value.value)
+        };
+
+        // If we don't do this check, pg will overwrite
+        // some random entry with our key/value pair...
+        if entry.is_null() && htab.elements == self.size {
+            return Err(ShmemHashMapError::HashTableFull);
+        }
```
workingjubilee, line 135: Wait, why do we read our value in and THEN do this early return?

pgrx/src/shmem_hash.rs
```diff
@@ -0,0 +1,250 @@
...
+}
+
+/// Compute the hash for the key and its pointer
+/// to pass to `pg_sys::hash_search_with_hash_value`.
+/// Lock on HTAB should be taken, although not strictly required I think.
+macro_rules! key {
+    ($key:expr, $htab:expr) => {{
+        let key = Key { key: $key };
+        let key_ptr: *const c_void = std::ptr::addr_of!(key) as *const Key<K> as *const c_void;
+        let hash_value = unsafe { pg_sys::get_hash_value($htab.htab, key_ptr) };
```
workingjubilee, line 70: Same comment re: "this has to document the preconditions of this function so we know why it's safe to call". The answer of "there are virtually none and this is an in-principle safe function as long as we call it when loaded in the Postgres context" is fine.

pgrx/src/shmem_hash.rs
```diff
@@ -0,0 +1,250 @@
...
+                    pg_sys::HASHACTION_HASH_REMOVE,
+                    std::ptr::null_mut(),
+                );
+            };
+
+            htab.elements -= 1;
+            return Some(value);
+        } else {
+            return None;
+        }
```
workingjubilee, line 211: But if you do use let-else this should probably be updated.

workingjubilee: Something like this, but then it will want you to hit it with rustfmt.
```suggestion
            Some(value)
```

pgrx/src/shmem_hash.rs
workingjubilee, line 185: Is there a reason this returns `V` and not `&V`? Because we cannot rely on concurrent changes not altering the data in the K/V pair? It would be good to understand the soundness conditions better here and why things look the way they do.

pgrx/src/shmem_hash.rs
```diff
@@ -0,0 +1,250 @@
...
+impl Default for ShmemHashMapInner {
+    fn default() -> Self {
+        Self { htab: std::ptr::null_mut(), elements: 0 }
+    }
+}
+
+/// A shared memory HashMap using Postgres' `HTAB`.
+/// This HashMap is used for `pg_stat_statements` and Postgres
+/// internals to store key/value pairs in shared memory.
+pub struct ShmemHashMap<K: Copy + Clone, V: Copy + Clone> {
```
workingjubilee, line 51: Please weaken the bounds for the struct and instead keep them on the implementation. It should be made clear why certain functions require the bounds they do in order to be sound, instead of plastering the bounds across the entire API. Aside from leaking implementation details, it makes it unclear where the soundness requirements actually start, which means later refactoring that attempts to weaken the bounds can be made unsound.

workingjubilee: Note that it's fine if the bounds are overeagerly applied in cases where it's not clear, but e.g. I have done a lot of this refactoring on `Array<'_, T>` and `List<T>` recently, and in general I tried to pull the scope tighter.

pgrx/src/shmem_hash.rs
```diff
@@ -0,0 +1,250 @@
...
+
+        let mut hash_ctl = pg_sys::HASHCTL::default();
+        hash_ctl.keysize = std::mem::size_of::<Key<K>>();
+        hash_ctl.entrysize = std::mem::size_of::<Value<K, V>>();
+
+        let shm_name =
+            alloc::ffi::CString::new(Uuid::new_v4().to_string()).expect("CString::new() failed");
+
+        let htab_ptr = unsafe {
+            pg_sys::ShmemInitHash(
```
workingjubilee, line 239: Preconditions, postconditions, "why is this sound?", etc.

pgrx/src/shmem_hash.rs
```diff
@@ -0,0 +1,250 @@
...
+        Self { htab: std::ptr::null_mut(), elements: 0 }
+    }
+}
+
+/// A shared memory HashMap using Postgres' `HTAB`.
+/// This HashMap is used for `pg_stat_statements` and Postgres
+/// internals to store key/value pairs in shared memory.
+pub struct ShmemHashMap<K: Copy + Clone, V: Copy + Clone> {
+    /// HTAB protected by a SpinLock.
+    htab: OnceCell<PgSpinLock<ShmemHashMapInner>>,
```
workingjubilee, line 53: Will the use of OnceCell remain correct in the event of this being fixed?
- https://github.com/pgcentralfoundation/pgrx/issues/1223

workingjubilee (commented):
Thank you. I realize reviewing twice on entirely different parts of the PR like this may be slightly unusual(?), and I appreciate you bearing with me here. My first pass often is about smaller nits because once they are addressed, it's easier to focus on the finer details.

pgrx-tests/src/tests/shmem_tests.rs
```diff
@@ -11,22 +11,35 @@ use pgrx::prelude::*;
 use pgrx::{pg_shmem_init, PgAtomic, PgLwLock, PgSharedMemoryInitialization};
 use std::sync::atomic::AtomicBool;
 
+#[cfg(feature = "cshim")]
+use pgrx::PgHashMap;
```
workingjubilee, line 15:
```suggestion
use pgrx::ShmemHashMap;
```

pgrx-tests/src/tests/shmem_tests.rs
```diff
@@ -11,22 +11,35 @@ use pgrx::prelude::*;
...
+
 static ATOMIC: PgAtomic<AtomicBool> = PgAtomic::new();
 static LWLOCK: PgLwLock<bool> = PgLwLock::new();
 
+#[cfg(feature = "cshim")]
+static HASH_MAP: PgHashMap<i64, i64> = PgHashMap::new(500);
```
workingjubilee, line 21:
```suggestion
static HASH_MAP: ShmemHashMap<i64, i64> = ShmemHashMap::new(500);
```

workingjubilee (changes requested):

pgrx/src/shmem_hash.rs
workingjubilee: It's fine if the answer is "no", I just want to check to make sure if this should get a note relating it back to that issue.

eeeebbbbrrrr:
This has been idle for a long time.  @levkk, old buddy old pal, if you wanna bring this back to life, feel free.  If not, I understand.

PR #1386, opened 2023-11-09
Partially clean up pgrx-sql-entity-graph

workingjubilee:
No functional changes.

No, seriously.


thomcc:
You're a hero for actually doing this.

PR #1387, opened 2023-11-09
Cleanup pgrx sql entity graph more

workingjubilee:
A second pass at cleanup. While the first was simply moving code around, this flattens out significant parts of the logic using more streamlined expressions so that the rest is more readable. No functional changes.


workingjubilee:
For my next pass, I'll try (gasp) actually extracting duplicated logic into reused functions!

PR #1325, opened 2023-10-10
Sniff for libclang-version-correct headers

workingjubilee:
Bindgen's strategy of autodetecting headers attempts to use a hacky route used by clang-sys, however, it doesn't take into account that the strategy of clang-sys uses `$CLANG_PATH` instead of `$LIBCLANG_PATH`, and then falls back to searching the entire directory structure. This means it's fairly easy for it to pull the headers for a different version, trying to compile them with the wrong clang!

This adopts a mesh of approaches, using the collected information from pg_config, clang-sys, and rust-bindgen to try to locate the correct compiler headers. If it can't find anything, it falls back to doing what currently happens.

PR #1389, opened 2023-11-11
Stop overriding the host clang during build

workingjubilee:
This should also fix the same thing that #1325 does, but I believe both should be accepted... if it works.


workingjubilee:
Alas it does not.

PR #1391, opened 2023-11-13
Remove `pg_wchar.h`

workingjubilee:
This header is tainted with simd.h


workingjubilee:
This header was added because it "might be convenient" but it's not clear we should use it.

workingjubilee:
Motherfuzzer.

PR #1393, opened 2023-11-14
Support patching Postgres with a given patch file

orf:
This PR adds a `--patch` argument to `cargo-pgrx` that applies a given patch file after downloading Postgres, before building.

This can be quite useful in some situations, such as when testing potential Postgres patches.

### DIFF
--- a/cargo-pgrx/src/command/init.rs
+++ b/cargo-pgrx/src/command/init.rs
@@ -75,6 +75,9 @@ pub(crate) struct Init {
     /// installed, but the resulting build is usable without valgrind.
     #[clap(long)]
     valgrind: bool,
+    /// Patch Postgresql with a given patch file
+    #[clap(long)]
+    patch: Option<PathBuf>,
     #[clap(long, short, help = "Allow N make jobs at once")]
     jobs: Option<usize>,
     #[clap(skip)]
@@ -255,6 +258,7 @@ fn download_postgres(
     let mut buf = Vec::new();
     let _count = http_response.into_reader().read_to_end(&mut buf)?;
     let pgdir = untar(&buf, pgrx_home, pg_config, init)?;
+    patch_postgres(pg_config, &pgdir, init)?;
     configure_postgres(pg_config, &pgdir, init)?;
     make_postgres(pg_config, &pgdir, init)?;
     make_install_postgres(pg_config, &pgdir, init) // returns a new PgConfig object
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr:
What’s the benefit to your workflow with this versus using an already-patched Postgres somewhere on your local filesystem?

What I’d like to understand is just how far cargo-pgrx really needs to go in terms of customization like this.  Not that it shouldn’t evolve, but it wasn’t the original intent.

orf:
For me, the benefit is that it's pretty seamless. I run `cargo run -- pgrx init --patch="some.patch"` and I have 5 versions of Postgres downloaded, patched, compiled and set up to run with `pgrx-pg-sys`. I then just do `PGRX_PG_SYS_GENERATE_BINDINGS_FOR_RELEASE=1 cargo build` in `pgrx-pg-sys` to have any external changes visible to rust.

It's really nice if that's part of your workflow. I guess I could pre-patch them, but having `init` do it is also nice.

One alternative might be to specify a git repository to download from, instead of using the RSS feed?

eeeebbbbrrrr:
We've had other PRs (that I rejected) related to being able to tweak some of the building that `cargo-pgrx init` does, I think related to the `./configure` flags.  I think there was some talk about someone wanting to install specific point release versions too.  I rejected those because I felt like at that point you're probably better off managing your own from-source Postgres installations.  I kinda feel the same way here but am still arguing with myself.  :)

I guess it's interesting to me to see a case where someone is modifying the postgres sources and concurrently writing a pgrx extension against whatever those modifications are.

I'd like to hear what @workingjubilee and @thomcc think.

> One alternative might be to specify a git repository to download from, instead of using the RSS feed?

Well, in the initial days of pgrx this is how getting postgres sources was implemented.  At the time, that introduced more problems than it solved, so I switched to the RSS feed.  That said, I think there's been some discussion around maybe going back to something that's git-based in an effort to aid Windows support (of which we have exactly zero).  So git is an option, if it doesn't complicate things.

pgrx is most definitely a developer tool and we want it to be useable by actual developers and we all do different things differently, so I'd like us to be as accommodating as possible.  I just don't necessarily want to veer too far away from the original intent, which is reducing friction to newcomers getting started.

(As an aside, if we were to merge this, we'd want to see the `cargo-pgrx/README.md` updated, and probably a sentence or two about what this flag does, when one might use it, and why.)

workingjubilee:
> One alternative might be to specify a git repository to download from, instead of using the RSS feed?

I feel like we do want to have an override and we want it to be fairly high-level, so "pg16 is actually located here" feels like the "way to go" rather than supporting ad-hoc patches.

workingjubilee:
oh right, bindgen is still effed, don't know why I even ran the tests.

eeeebbbbrrrr:
This approach isn't something we're going to support.  I really feel it's best for a developer to point cargo-pgrx at a local postgres source tree that's already patched the way they need.

PR #1394, opened 2023-11-16
[NO MERGE] Parallelize cshim builds

workingjubilee:
Figured I might as well run this experiment.


workingjubilee:
Even after some fiddling this doesn't appear to speed anything up, so abandoning this experiment.

workingjubilee:
The "correct" way to do this probably involves turning both the bindgen and the cshim build into tasks and running them in full-parallel with each other.

PR #1396, opened 2023-11-16
Don't run the blender on weekends

workingjubilee:


BradyBonnette (approved):
Looks good :) I wasn't too sure about the `MON-FRI` specification (Github Actions gonna Github Action), but it's in their own documentation so TIL.

PR #1397, opened 2023-11-17
Add trusted option in the control file

liuxueyang:
`trusted` is supported in postgresql 13 and above. Extensions can add it to the control file and also want to support postgresql 12 and below. This PR removes the trusted option in the control file when installing the extension. The default value of `trusted` is false.

Related PR in postgresml https://github.com/postgresml/postgresml/pull/810

Options in the control file: https://www.postgresql.org/docs/13/extend-extensions.html#id-1.8.3.20.11

### DIFF
--- a/pgrx-examples/custom_types/custom_types.control
+++ b/pgrx-examples/custom_types/custom_types.control
@@ -3,3 +3,4 @@ default_version = '@CARGO_VERSION@'
 module_pathname = '$libdir/custom_types'
 relocatable = false
 superuser = false
+trusted = false
--- a/pgrx-sql-entity-graph/src/control_file.rs
+++ b/pgrx-sql-entity-graph/src/control_file.rs
@@ -68,7 +68,7 @@ impl ControlFile {
 
             temp.insert(k, v);
         }
-        Ok(ControlFile {
+        let control_file = ControlFile {
             comment: temp
                 .get("comment")
                 .ok_or(ControlFileError::MissingField { field: "comment" })?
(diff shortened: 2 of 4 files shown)


pgrx-examples/custom_types/custom_types.control
```diff
@@ -3,3 +3,4 @@ default_version = '@CARGO_VERSION@'
 module_pathname = '$libdir/custom_types'
 relocatable = false
 superuser = false
+trusted = false
```
workingjubilee, line 6: In a case where this was, e.g.
```
superuser = false
trusted = true
```
this would do nothing. While Postgres notes this is simply "irrelevant", I think pgrx should error, or at least warn on that case, rather than permitting it to pass by unremarked.

pgrx-sql-entity-graph/src/control_file.rs
```diff
@@ -86,6 +87,10 @@ impl ControlFile {
                 .ok_or(ControlFileError::MissingField { field: "superuser" })?
                 == &"true",
             schema: temp.get("schema").map(|v| v.to_string()),
+            trusted: temp
+                .get("trusted")
+                .ok_or(ControlFileError::MissingField { field: "trusted" })?
+                == &"false",
```
workingjubilee: I don't think we should require this parameter to be set. We should simply treat the absent case as equivalent to false, since it de facto is.

workingjubilee (changes requested):

pgrx-sql-entity-graph/src/control_file.rs
workingjubilee: Doing that would also fix the tests.

pgrx-examples/custom_types/custom_types.control
liuxueyang: Thanks. I added another type of error for this case.

pgrx-sql-entity-graph/src/control_file.rs
liuxueyang: Thanks. This field is optional now.

liuxueyang: The CI stuck for more than 2 hours. 😢

workingjubilee: whoops. accidentally out of budget.

workingjubilee (approved):
Okay, the formatting problem is fixed on develop, and this otherwise looks fine.

workingjubilee:
Thanks!

Issue #1370, opened 2023-11-01
Use `.default_enum_style(EnumVariation::ModuleConsts)`


workingjubilee:
We actually all kinda hate the current enum style.

It's not correct, due to various reasons, to use Rust enums for every case... we want to only do that in cases where we verify it's correct, like NodeTag. Otherwise it's UB.

A nicer version that _looks_ like Rust enums when you use them but has the same types as we currently use and doesn't have weird UB implications is `EnumVariation::ModuleConsts`.

Hypothetically, if people wanted we could make a _slightly_ more graceful changeover of taking every enum and running a callback which then generates the old enum style but tagged with `#[deprecated(message = "use EnumName::Variant instead"]`. I'd accept such a PR. I probably won't write it that way myself. If anyone wants to do that, now is the time, and you'd probably be best-off doing it via `ParseCallbacks::enum_variant_{name,behavior}` (but don't actually do anything, just record things so you can emit the extra constants later).

PR #1399, opened 2023-11-18
Add access hook

orf:
This PR adds support for `object_access_hook` in much the same way as `pgrx_emit_log`.

### DIFF
--- a/pgrx-tests/src/tests/hooks_tests.rs
+++ b/pgrx-tests/src/tests/hooks_tests.rs
@@ -23,6 +23,7 @@ mod tests {
 
         struct TestHook {
             events: u32,
+            accesses: u32,
         }
         impl PgHooks for TestHook {
             /// Hook before the logs are being processed by PostgreSQL itself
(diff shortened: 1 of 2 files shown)


pgrx/src/hooks.rs
```diff
@@ -176,6 +176,24 @@ pub trait PgHooks {
         prev_hook(pstate, query, jumble_state)
     }
 
+    fn object_access_hook(
```
workingjubilee, line 179: Please document when this gets run, in principle. Each of these hooks is in a different header file, so we cannot simply point to a single place in the Postgres documentation or even headers for all of them.

pgrx/src/hooks.rs
```diff
@@ -176,6 +176,24 @@ pub trait PgHooks {
...
+        &mut self,
+        access: pg_sys::ObjectAccessType,
+        class_id: pg_sys::Oid,
+        object_id: pg_sys::Oid,
+        sub_id: ::std::os::raw::c_int,
+        arg: *mut ::std::os::raw::c_void,
```
workingjubilee, line 185: re: docs, What do these arguments control?

pgrx-tests/src/tests/hooks_tests.rs
```diff
@@ -121,15 +122,42 @@ mod tests {
...
+                arg: *mut ::std::os::raw::c_void,
+                prev_hook: fn(
+                    access: pg_sys::ObjectAccessType,
+                    class_id: pg_sys::Oid,
+                    object_id: pg_sys::Oid,
+                    sub_id: ::std::os::raw::c_int,
+                    arg: *mut ::std::os::raw::c_void,
+                ) -> HookResult<()>,
+            ) -> HookResult<()> {
+                if access == pg_sys::ObjectAccessType_OAT_POST_CREATE {
```
workingjubilee, line 141: Since this is now becoming part of our public API, please explicitly specify this enum should be generated as a "module const", so that it uses the appropriate style, ahead of
- https://github.com/pgcentralfoundation/pgrx/issues/1370

pgrx-tests/src/tests/hooks_tests.rs
```diff
@@ -121,15 +122,42 @@ mod tests {
...
+                    self.accesses += 1;
+                }
+                prev_hook(access, class_id, object_id, sub_id, arg)
+            }
         }
 
-        static mut HOOK: TestHook = TestHook { events: 0 };
+        Spi::run("CREATE TABLE test_hooks_table (bar int)").expect("SPI failed");
+
+        static mut HOOK: TestHook = TestHook { events: 0, accesses: 0 };
```
workingjubilee, line 150: I find myself less-than-excited about our use of a `static mut` for this test but this is not a problem you introduced so you do not have to fix it, I simply needed to note this for my own purposes.

workingjubilee (commented):
In-principle, this seems like it should be fine. Even though I want to rewrite most of the hooks API, this one looks like it would probably be identical even after the rewrite. However, new API surface should be thoroughly documented, even if it's primarily pointers to other documentation.

eeeebbbbrrrr:
Hi @orf!  Looks like Jubilee wants a few documentation/comment changes before we merge this PR.  Are you up for that?

pgrx/src/hooks.rs
workingjubilee: Note that pointing to a header for _each_ of them is at least hypothetically fine, and would be okay here.

pgrx/src/hooks.rs
workingjubilee: Note that it doesn't have to go deep into practical applications or sophisticated examples.

eeeebbbbrrrr:
This has been idle for awhile now.  Closing it.  @orf, if you want to bring this back to life, that's fine.

PR #1401, opened 2023-11-18
Fix actions badge

muzimuzhi:
GitHub now needs workflow file name, rather than workflow name in badge url.

see https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge#using-the-workflow-file-name

- Before
  cargo test --all
- After
  GitHub Actions badge

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 > Build Postgres Extensions with Rust!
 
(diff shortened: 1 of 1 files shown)


workingjubilee:
Thanks! I'm going to go ahead and merge this since it only affects docs.

PR #1402, opened 2023-11-21
Begin work on a guide

workingjubilee:
Stubbing out mostly.


workingjubilee:
While I want to avoid adding too much (because at some point I should refactor the existing advisory files, and I want this to be a "move files around" commit for those files that are moved), I should probably address https://github.com/pgcentralfoundation/pgrx/issues/1406 in a skeletal form.

PR #1416, opened 2023-11-30
Simplify logic in lifetimes.rs

workingjubilee:
This cleans up pgrx-sql-entity-graph AGAIN, this time focusing mostly on the lifetimes.rs file and where the related functions are used, now that the entire thing isn't quite such a horrorshow any more. Unfortunately it still needs quite a bit of work just to be able to see what the hell is going on.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -4,6 +4,9 @@
 # adding a rustfmt.toml
 1ded8553fbeff635fcca0cbc8f37034cc124d698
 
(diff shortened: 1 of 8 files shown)


workingjubilee:
With the last commit I have also ripped out the constant duplication of the wrapping function's overall template in `wrapper_func` and lifted it into a shared closure. It is almost looking like it's going to become... dare I say it?... cohesive and maintainable.

PR #1418, opened 2023-11-30
:fix: avoid segfault on background worker pool exhaustion

### LINKED ISSUE #1417: Segfault starting and managing background workers when worker pool exhausted
`BackgroundWorkerBuilder::load_dynamic()` does not check the return value from `pg_sys::RegisterDynamicBackgroundWorker`; in case of failure (e.g. because of `max_worker_processes` being exceeded) the passed `handle` is left set to null.

https://github.com/pgcentralfoundation/pgrx/blob/6a0f825e5fb73d37dfe07229ba97bc1dea69ffe4/pgrx/src/bgworkers.rs#L605C9-L605C9

Later trying to call `DynamicBackgroundWorker::wait_for_shutdown()` causes a segfault due to null pointer dereference.

See [bgworker_segfault](https://github.com/optionfactory/bgworker_segfault) (`cargo pgrx test`) for a test that triggers this bug.

AndreaOddo89:
`BackgroundWorkerBuilder::load_dynamic()` does not check the return value from `pg_sys::RegisterDynamicBackgroundWorker`; in case of failure (e.g. because of `max_worker_processes` being exceeded) the passed `handle` is left set to null.
Later trying to call `DynamicBackgroundWorker::wait_for_shutdown()` causes a segfault due to null pointer dereference.

We make load_dynamic() fallible, returning an `Err` in this case, preventing the null pointer dereference.

For testing, we spawn multiple workers until we exhaust the pool; this needs to be done in isolation, as concurrently-run tests, that spawn bgworkers might fail spuriously otherwise; we do this using an RW advisory lock.

Fixes https://github.com/pgcentralfoundation/pgrx/issues/1417

### DIFF
--- a/pgrx/src/bgworkers.rs
+++ b/pgrx/src/bgworkers.rs
@@ -597,15 +597,20 @@ impl BackgroundWorkerBuilder {
     }
 
     /// Once properly configured, call `load_dynamic()` to get the BackgroundWorker registered and started dynamically.
-    pub fn load_dynamic(self) -> DynamicBackgroundWorker {
+    /// Start up might fail, e.g. if max_worker_processes is exceeded. In that case an Err is returned.
+    pub fn load_dynamic(self: Self) -> Result<DynamicBackgroundWorker, ()> {
         let mut bgw: pg_sys::BackgroundWorker = (&self).into();
         let mut handle: *mut pg_sys::BackgroundWorkerHandle = null_mut();
 
-        unsafe {
-            pg_sys::RegisterDynamicBackgroundWorker(&mut bgw, &mut handle);
+        let success = unsafe {
+            pg_sys::RegisterDynamicBackgroundWorker(&mut bgw, &mut handle)
         };
 
-        DynamicBackgroundWorker { handle, notify_pid: bgw.bgw_notify_pid }
+        if !success {
+            Err(())
+        } else {
+            Ok(DynamicBackgroundWorker { handle, notify_pid: bgw.bgw_notify_pid })
+        }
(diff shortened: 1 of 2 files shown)


pgrx/src/bgworkers.rs
```diff
@@ -597,15 +597,20 @@ impl BackgroundWorkerBuilder {
...
-    pub fn load_dynamic(self) -> DynamicBackgroundWorker {
+    /// Start up might fail, e.g. if max_worker_processes is exceeded. In that case an Err is returned.
+    pub fn load_dynamic(self: Self) -> Result<DynamicBackgroundWorker, ()> {
         let mut bgw: pg_sys::BackgroundWorker = (&self).into();
         let mut handle: *mut pg_sys::BackgroundWorkerHandle = null_mut();
 
-        unsafe {
-            pg_sys::RegisterDynamicBackgroundWorker(&mut bgw, &mut handle);
+        let success = unsafe {
+            pg_sys::RegisterDynamicBackgroundWorker(&mut bgw, &mut handle)
```
workingjubilee: Please add a `// SAFETY` comment to this block so that the preconditions and postconditions of using this function are documented someways.

workingjubilee:
Please run `cargo fmt` on all PRs submitted to this repo (and most any other repo that has a rustfmt.toml).

That formality aside, damn, I really have to get more annoying about asking questions about FFI functions and their invariants, huh? Thanks. I shouldn't have missed this in review.

workingjubilee:
Thank you!

PR #1422, opened 2023-12-05
Simplify void return handling

workingjubilee:
Aside from imposing a few stylistic choices on pgrx-sql-entity-graph, FunctionMetadata had its second generic removed after some consulting with type system experts who expressed it is unlikely to be necessary. Indeed, everything seems to be unaffected by this.

In doing this, it reveals that FunctionMetadataEntity is only using Option to handle a single return type: `()`. Removing this allows simplifying the handling for function pointers that return `-> ()`... of which you might remember there are a lot, given that is any function without an explicit other return type.


thomcc (approved):
I've disliked how we do this for quite a while. Seems to be a definite improvement.

PR #1427, opened 2023-12-06
Stub out pgrx-internal.md

workingjubilee:
Adds ~800 words, though it remains stubs, just in a single file.

### DIFF
--- a/docs/src/contributing/pgrx-internal.md
+++ b/docs/src/contributing/pgrx-internal.md
@@ -58,23 +58,26 @@ an irreplaceable functionality in pgrx, even if its role could be reduced.
 A primary task for pgrx, accomplished via cargo-pgrx and support libraries, is to generate SQL.
 This allows programmers to avoid separately maintaining the SQL that installs the extension when
 `CREATE EXTENSION` runs. This is not merely a convenience: misdeclaring extensions, functions,
-types, and operators can in some cases cause illogical behavior from Postgres. The query executor
-may assume something based on the SQL definitions, causing it to interact with extensions in ways
-that return invalid query results or, depending on the nature of the extension, corrupt a table's
-active rows or index. While Postgres employs various techniques to detect such incorrect behaviors
-and will abort transactions and roll back if it does, and this may not be "undefined behavior" in
-Rust terms, it is at least as equally undesirable!
+types, and operators can in some cases cause illogical behavior from Postgres.
 
-However, Postgres imposes ordering constraints on how extensions are described to it, by rejecting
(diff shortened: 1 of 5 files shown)


docs/src/contributing/pgrx-internal.md
```diff
@@ -1 +1,112 @@
...
+# test suite
+
+This should be run in CI for every merge.
+
+## pgrx-tests
+This currently contains both our test support framework and the actual test suite.
+
+Fortunately, the way this works is _almost_ magic enough to simply happen if you run `cargo test`.
+Unfortunately, due to the way that `#[pg_test]` works, the placement of test code is extremely
+constrained in terms of where it must be in code. Thus we have this additional crate.
```
workingjubilee:
```suggestion
constrained in terms of where it must be in files. This is part of why we have this
additional crate.
```

docs/src/contributing/pgrx-internal.md
```diff
@@ -1 +1,112 @@
...
+the alternative is a horrible pile of bash and regexen.
+
+# test suite
+
+This should be run in CI for every merge.
+
+## pgrx-tests
+This currently contains both our test support framework and the actual test suite.
+
+Fortunately, the way this works is _almost_ magic enough to simply happen if you run `cargo test`.
```
workingjubilee:
```suggestion
Fortunately, the way `#[pg_test]` works is magic enough to simply happen if you run `cargo test`.
```

docs/src/contributing/pgrx-internal.md
```diff
@@ -1 +1,112 @@
...
+
+A primary task for pgrx, accomplished via cargo-pgrx and support libraries, is to generate SQL.
+This allows programmers to avoid separately maintaining the SQL that installs the extension when
+`CREATE EXTENSION` runs. This is not merely a convenience: misdeclaring extensions, functions,
+types, and operators can in some cases cause illogical behavior from Postgres. The query executor
+may assume something based on the SQL definitions, causing it to interact with extensions in ways
+that return invalid query results or, depending on the nature of the extension, corrupt a table's
+active rows or index. While Postgres employs various techniques to detect such incorrect behaviors
+and will abort transactions and roll back if it does, and this may not be "undefined behavior" in
+Rust terms, it is at least as equally undesirable!
```
workingjubilee: Find a way to work in the main citation for this, which is https://www.postgresql.org/docs/current/xoper-optimization.html

workingjubilee: Also technically nasal demons are yet-more-powerful in the arbitrariness of the damage they can do, but anything that amounts to risking almost as much damage as `DROP DATABASE production;` is uh... pretty bad!

PR #1433, opened 2023-12-12
expand readme

Lokathor:
this explains the version match requirements of `cargo-pgrx` and also breaks up the initial "one step" into two steps with slightly more explanation.


eeeebbbbrrrr (approved):
Thanks, @Lokathor!  Welcome to the committers club.

PR #1436, opened 2023-12-15
add bindings for libpq

usamoi:
add two header files:

* libpq/libpq.h
* libpq/pqformat.h

They are useful for a PostgreSQL type to implement `send_function` and `receive_function`.

### DIFF
--- a/pgrx-pg-sys/include/pg12.h
+++ b/pgrx-pg-sys/include/pg12.h
@@ -76,6 +76,7 @@
 #include "executor/tuptable.h"
 #include "foreign/fdwapi.h"
 #include "foreign/foreign.h"
+#include "libpq/pqformat.h"
 #include "mb/pg_wchar.h"
 #include "nodes/execnodes.h"
 #include "nodes/extensible.h"
--- a/pgrx-pg-sys/include/pg13.h
+++ b/pgrx-pg-sys/include/pg13.h
@@ -76,6 +76,7 @@
 #include "executor/tuptable.h"
 #include "foreign/fdwapi.h"
 #include "foreign/foreign.h"
+#include "libpq/pqformat.h"
 #include "mb/pg_wchar.h"
 #include "nodes/execnodes.h"
 #include "nodes/extensible.h"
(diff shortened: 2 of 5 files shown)


eeeebbbbrrrr:
Hi, @usamoi, would you mind rebasing this PR with `develop`?  I just fixed CI there and I want to make sure things still build with the libpq headers included.

Looking at these headers, looks like they're mostly a combination of `#define`s and `static inline`s.  I wonder if what we actually need here is to hand-port these to rust?

What symbols from these headers are you trying to get access to?

usamoi:
Hi, @eeeebbbbrrrr. I need symbols including `pq_begintypsend`, `pq_endtypsend`, `pq_sendbytes` and so on. They are used for send/recevie function of a Postgresql type, eg. https://github.com/postgres/postgres/blob/REL_16_STABLE/src/backend/utils/adt/varlena.c#L617.

eeeebbbbrrrr:
> Hi, @eeeebbbbrrrr. I need symbols including `pq_begintypsend`, `pq_endtypsend`, `pq_sendbytes` and so on.

I see.  Because they're `static inline` this won't pull in some of the other helper functions like `pg_write[int32, int64, int16, int8]`, etc.  :(

So I'll just note here that we may need to port those to run at some point (and would happily accept a PR!).

Once you rebase and we can confirm CI passes, I'll merge this.  I was thinking of doing another release on Monday or Tuesday.

eeeebbbbrrrr:
```
1: clang diagnosed error: /usr/include/postgresql/15/server/libpq/libpq-be.h:34:10: fatal error: 'gssapi/gssapi.h' file not found
```

I was worried about something like this.

I think this header is related to Kerberos. It's being pulled in through the new `libpq.h`.  I would expect `ENABLE_GSS` to not be defined for the Postgres that pgrx compiles/installs.

Do you need `libpq.h` too?

usamoi:
I think I don't need `libpq.h` really. Deleted it.

eeeebbbbrrrr:
> I think I don't need `libpq.h` really. Deleted it.

Yeah. I was getting concerned about potentially adding another development requirement on Kerberos. 

Assuming the latest ci job passes, I’ll merge and get a new release out early next week. 

Thanks for your efforts on this Saturday afternoon!

PR #1438, opened 2023-12-16
does updating apt fix CI?

eeeebbbbrrrr:
yes, it does.

also, telling homebrew to **not** update also fixes CI on MacOS.

I love computers.


eeeebbbbrrrr:
Screenshot 2023-12-16 at 12 38 01 PM

PR #1435, opened 2023-12-14
Address the case when `RUSTFLAGS` environment variable is set but is empty

eeeebbbbrrrr:
Previously, an empty `RUSTFLAGS` would result in an error later on when we invoke `rustc`.

--- 
I could never reproduce it locally but saw it in CI on a different pgrx-based project once.

```
error: multiple input filenames provided (first two filenames are `` and `/home/runner/.pgrx/postmaster_stubs/home/runner/.pgrx/14.10/pgrx-install/bin/postmaster_stub.rs`)

Error:
   0: rustc exited with code 1

Location:
   /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-pgrx-0.11.2/src/command/schema.rs:557

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ SPANTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

   0: cargo_pgrx::command::schema::create_stub with postmaster_path=/home/runner/.pgrx/14.10/pgrx-install/bin/postgres postmaster_stub_dir=/home/runner/.pgrx/postmaster_stubs/home/runner/.pgrx/14.10/pgrx-install/bin
      at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-pgrx-0.11.2/src/command/schema.rs:466
   1: cargo_pgrx::command::schema::generate_schema with pg_version=14.10 profile=Dev test=true path=/home/runner/.pgrx/14.10/pgrx-install/share/postgresql/extension/pronto_pgrx--0.0.1.sql features=["pg14", "pg_test"]
      at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-pgrx-0.11.2/src/command/schema.rs:176
   2: cargo_pgrx::command::install::install_extension with pg_version=14.10 profile=Dev test=true features=["pg14", "pg_test"]
      at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-pgrx-0.11.2/src/command/install.rs:114
   3: cargo_pgrx::command::install::execute
      at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-pgrx-0.11.2/src/command/install.rs:63

[...]

PR #1441, opened 2023-12-18
Handle RUSTFLAGS better

eeeebbbbrrrr:
This cleans up PR #1435 as per
(https://github.com/pgcentralfoundation/pgrx/pull/1435#issuecomment-1859360044).

What we probably really want is a legit "RUSTFLAGS" parser.  I don't feel like writing and testing one and a quick search didn't find anything on crates.io.  So splitting on whitespace will have to be Good Enough.


eeeebbbbrrrr:
I approve my own PR.

workingjubilee:
This looks fine but we should be using [shlex](https://crates.io/crates/shlex) (it is the equivalent of a "legit RUSTFLAGS parser" for our purposes).

Issue #1430, opened 2023-12-10
pgrx init doesn't work with self-signed certificate


tumluliu:
Hi there, thanks for making this amazing framework. I however has hit the wall behind a coporation proxy as we are using our own certificate:

```
❯ cargo pgrx init
Error:
   0: unable to retrieve https://www.postgresql.org/versions.rss
   1: https://www.postgresql.org/versions.rss: Connection Failed: tls connection init failed: invalid peer certificate: UnknownIssuer
   2: invalid peer certificate: UnknownIssuer

Location:
   /home/xx/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-pgrx-0.11.2/src/command/version.rs:42

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ SPANTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

   0: cargo_pgrx::command::init::execute
      at /home/xx/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-pgrx-0.11.2/src/command/init.rs:83
```

I took a very quick look at the code from `version.rs` [here](https://github.com/pgcentralfoundation/pgrx/blob/develop/cargo-pgrx/src/command/version.rs#L38) and it seems it caused by the `ureq` `Agent` you created is using the default `webpki-roots` instead of the OS' trust store (refs: [1](https://docs.rs/ureq/2.8.0/ureq/index.html#trusted-roots), [2](https://github.com/algesten/ureq/issues/386#issuecomment-838851259) ). So I guess it's not very easy to change this behaviour without modifying the `mod.rs` [codes](https://github.com/pgcentralfoundation/pgrx/blob/develop/cargo-pgrx/src/command/mod.rs#L33)? But I'm not sure since I'm neither an expert in security nor in rust. I'm wondering if there is a quick way to bypass this cert complaint when running `init`. Could you provide some hints? Thanks again!

p.s. my env: Ubuntu 22.04 Linux

workingjubilee:
If setting `HTTPS_PROXY` doesn't fix it, then there is no quick way and it does require modifying the code using [rustls-native-certs](https://docs.rs/rustls-native-certs/latest/rustls_native_certs/).

tumluliu:
[...]

PR #1442, opened 2023-12-18
Disable ureq defaults to avoid enabling both native-tls and rustls

jirutka:
ureq defines `default = ["tls", "gzip"]`, so "tls" feature (rustls) is currently always enabled even when native-tls is preferred. This makes disallows building cargo-pgrx on platforms not supported by the ring crate (dependency of rustls).

### DIFF
--- a/cargo-pgrx/Cargo.toml
+++ b/cargo-pgrx/Cargo.toml
@@ -38,7 +38,7 @@ prettyplease = "0.2.15"
 proc-macro2 = { version = "1.0.69", features = [ "span-locations" ] }
 quote = "1.0.33"
 regex = "1.10.0"
-ureq = "2.8.0"
+ureq = { version = "2.8.0", default-features = false, features = [ "gzip" ] }
 url = "2.4.1"
 serde = { version = "1.0", features = [ "derive" ] }
 serde_derive = "1.0"


eeeebbbbrrrr:
I wonder if this helps with (or even resolves) #1430?

eeeebbbbrrrr:
well... that took out CI real quick!  https://github.com/pgcentralfoundation/pgrx/actions/runs/7251777864/job/19755596250

jirutka:
> I wonder if this helps with (or even resolves) #1430?

I guess so.

jirutka:
> well... that took out CI real quick! https://github.com/pgcentralfoundation/pgrx/actions/runs/7251777864/job/19755596250

Are openssl libs and header files available in this environment?

eeeebbbbrrrr:
I don’t know but I wouldn’t want to make that a new requirement for all our users.  This is one of the reasons we defaulted, however improperly, to using rusttls 

Any thoughts here?

workingjubilee:
We are not going to make OpenSSL a hard requirement of building cargo-pgrx. The correct fix is the one that involves writing actual Rust code, instead of a pile of bash scripts to placate OpenSSL into working.

workingjubilee:
If we have to do this for a particular platform, I would like to be able to point to the ring or rustls issue we are working around before we merge any such change, because we would like to remove it upon those issues being resolved.

PR #1443, opened 2023-12-18
use the full `cargo` version string in the version match check.

Lokathor:
Rather than `cargo-pgrx` using just the `rustc` minor version when checking for a toolchain version mismatch, this now uses the full `cargo` version string. `cargo` uses the same version number as `rustc`, so it's a sufficient stand in for checking the version of `rustc`.

By checking the full version string instead of only the minor version, we can now detect a mismatch between two nightly versions of rust issued in the same 6 week period.

Example:
`cargo 1.76.0-nightly (9b13310ca 2023-11-24)`
vs
`cargo 1.76.0-nightly (1aa9df1a5 2023-12-12)`

These detect as the same minor version, but using the new full string check they will correctly show up as different and flag the mismatch error.

Also: the error message now alerts people how they can override the version check when necessary.

### DIFF
--- a/cargo-pgrx/build.rs
+++ b/cargo-pgrx/build.rs
@@ -27,5 +27,5 @@ fn main() {
 fn cargo_version() -> Option<String> {
     let cargo = std::env::var_os("CARGO").expect("`CARGO` env var wasn't set!");
     let output = std::process::Command::new(cargo).arg("--version").output().ok()?;
-    std::str::from_utf8(&output.stdout).map(str::to_string).ok()
+    std::str::from_utf8(&output.stdout).map(|s| s.trim().to_string()).ok()
 }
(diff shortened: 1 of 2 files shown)


workingjubilee:
As memory serves, we had been avoiding this kind of granularity because it doesn't actually matter this much in practice (we had also imagined we would solve the problem that inflicts ABI Hell on ourselves sooner, but then other things came up). However, the improvement to the error message is making me consider accepting this anyways. I'm going to roll this around in my head a little bit.

workingjubilee (approved):
Yeah, I slept on it and I feel confident accepting this is better than the status quo.

thomcc:
This was deliberate at the time, because I update nightly almost every day, and don’t necessarily want to reinstall cargo-pgrx every day (especially given that ABI breaks are fairly rare). That said, I’m not on pgrx that much atm, so I don’t care a ton.

PR #1445, opened 2023-12-24
Add a macro `pg_cast` for constructing PG casts from Rust functions.

xwkuang5:
To use the new macro, add the `pg_cast` macro on a single-argument single-return value Rust function:

```rust
#[pg_cast]
fn test_cast(_value: Json) -> i32 {
    0
}
```

It is possible to modify the cast by adding either `assignment` or `implicit` to the `pg_cast` macro as argument:

```
#[pg_cast(assignment)] /* creates a PG cast that may implicitly be used in assignment */

#[pg_cast(implicit)] /* creates a PG cast that may implicitly be used in any situation */
```

TESTED="cargo test"

### DIFF
--- a/pgrx-macros/src/lib.rs
+++ b/pgrx-macros/src/lib.rs
@@ -21,7 +21,7 @@ use operators::{deriving_postgres_eq, deriving_postgres_hash, deriving_postgres_
 use pgrx_sql_entity_graph as sql_gen;
 use sql_gen::{
     parse_extern_attributes, CodeEnrichment, ExtensionSql, ExtensionSqlFile, ExternArgs,
-    PgAggregate, PgExtern, PostgresEnum, Schema,
+    PgAggregate, PgCast, PgExtern, PostgresEnum, Schema,
 };
 
 mod operators;
@@ -148,11 +148,64 @@ pub fn initialize(_attr: TokenStream, item: TokenStream) -> TokenStream {
     item
 }
 
-/// Declare a function as `#[pg_cast]` to indicate that it represents a Postgres cast
-/// `cargo pgrx schema` will automatically generate the underlying SQL
+/**
+Declare a function as `#[pg_cast]` to indicate that it represents a Postgres [cast](https://www.postgresql.org/docs/current/sql-createcast.html).
+
+* `assignment`: Corresponds to [`AS ASSIGNMENT`](https://www.postgresql.org/docs/current/sql-createcast.html).
+* `implicit`: Corresponds to [`AS IMPLICIT`](https://www.postgresql.org/docs/current/sql-createcast.html).
+
+By default if no attribute is specified, the cast function can only be used in an explicit cast.
+
+Functions MUST accept and return exactly one value whose type MUST be a `pgrx` supported type. `pgrx` supports many PostgreSQL types by default.
+New types can be defined via [`macro@PostgresType`] or [`macro@PostgresEnum`].
+
+Example usage:
+```rust,ignore
+use pgrx::*;
+#[pg_cast(implicit)]
+fn cast_json_to_int(input: Json) -> i32 { todo!() }
+*/
 #[proc_macro_attribute]
-pub fn pg_cast(_attr: TokenStream, item: TokenStream) -> TokenStream {
-    item
+pub fn pg_cast(attr: TokenStream, item: TokenStream) -> TokenStream {
+    fn wrapped(attr: TokenStream, item: TokenStream) -> Result<TokenStream, syn::Error> {
+        use syn::parse::Parser;
+        let mut cast = PgCast::Default;
+        match syn::punctuated::Punctuated::<syn::Path, syn::Token![,]>::parse_terminated.parse(attr)
+        {
+            Ok(paths) => {
+                if paths.len() > 1 {
+                    panic!(
+                        "pg_cast must take either 0 or 1 attribute. Found {}: {}",
+                        paths.len(),
+                        paths.to_token_stream()
+                    )
+                } else if paths.len() == 1 {
+                    match paths.first().unwrap().segments.last().unwrap().ident.to_string().as_str()
+                    {
+                        "implicit" => cast = PgCast::Implicit,
+                        "assignment" => cast = PgCast::Assignment,
+                        other => panic!("Unrecognized pg_cast option: {}. ", other),
+                    }
+                }
+            }
+            Err(err) => {
+                panic!("Failed to parse attribute to pg_cast: {}", err)
+            }
+        }
+        // `pg_cast` does not support other `pg_extern` attributes for now, pass an empty attribute token stream.
+        let pg_extern = PgExtern::new(TokenStream::new().into(), item.clone().into())?.0;
+        Ok(CodeEnrichment(pg_extern.as_cast(cast)).to_token_stream().into())
+    }
+
+    match wrapped(attr, item) {
+        Ok(tokens) => tokens,
+        Err(e) => {
+            let msg = e.to_string();
+            TokenStream::from(quote! {
+              compile_error!(#msg);
+            })
+        }
+    }
 }
 
 /// Declare a function as `#[pg_operator]` to indicate that it represents a Postgres operator
--- a/pgrx-macros/src/lib.rs
+++ b/pgrx-macros/src/lib.rs
@@ -148,6 +148,13 @@ pub fn initialize(_attr: TokenStream, item: TokenStream) -> TokenStream {
     item
 }
 
+/// Declare a function as `#[pg_cast]` to indicate that it represents a Postgres cast
+/// `cargo pgrx schema` will automatically generate the underlying SQL
+#[proc_macro_attribute]
+pub fn pg_cast(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    item
+}
+
 /// Declare a function as `#[pg_operator]` to indicate that it represents a Postgres operator
 /// `cargo pgrx schema` will automatically generate the underlying SQL
 #[proc_macro_attribute]
--- a/pgrx-sql-entity-graph/src/pg_extern/entity/mod.rs
+++ b/pgrx-sql-entity-graph/src/pg_extern/entity/mod.rs
@@ -341,7 +341,7 @@ impl ToSql for PgExternEntity {
             }
         };
 
-        let ext_sql = format!(
+        let mut ext_sql = format!(
             "\n\
             -- {file}:{line}\n\
             -- {module_path}::{name}\n\
@@ -349,7 +349,7 @@ impl ToSql for PgExternEntity {
             {fn_sql}"
         );
 
-        let rendered = if let Some(op) = &self.operator {
+        if let Some(op) = &self.operator {
             let mut optionals = vec![];
             if let Some(it) = op.commutator {
                 optionals.push(format!("\tCOMMUTATOR = {}", it));
@@ -473,8 +473,9 @@ impl ToSql for PgExternEntity {
                                                     maybe_comma = if !optionals.is_empty() { "," } else { "" },
                                                     optionals = if !optionals.is_empty() { optionals.join(",\n") + "\n" } else { "".to_string() },
                                             );
-            ext_sql + &operator_sql
-        } else if let Some(cast) = &self.cast {
+            ext_sql += &operator_sql
+        };
+        if let Some(cast) = &self.cast {
             let target_arg = &self.metadata.retval;
             let target_fn_arg = &self.fn_return;
             let target_arg_graph_index = context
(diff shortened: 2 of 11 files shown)


eeeebbbbrrrr:
Thanks. Best Christmas present ever! 

I won’t have a chance to review it probably until January, so hang in there.

workingjubilee:
why does `pg_cast` not emit `pg_extern` if that is required?

xwkuang5:
> why does `pg_cast` not emit `pg_extern` if that is required?

I considered 3 implementation options:

1. (current): use `pg_extern` + `pg_cast` as a macro for using the generated function in a PG cast
2. make `pg_cast` emit `pg_extern` and add another `castoption` macro for modifying the cast object
3. make `pg_cast` emit `pg_extern` and add `Implicit` and `Assignment` as an option to [Attribute](https://github.com/pgcentralfoundation/pgrx/blob/614456e191a87f6df1b72428c79a01dedfe75516/pgrx-sql-entity-graph/src/pg_extern/attribute.rs#L27)

Option 2 requires 2 more macros which seems unnecessary to me (though this is the approach that `pg_operator` takes). For option 3, I was not sure about whether adding the generic `Implicit` and `Assignment` variant to the `Attribute` enum is a good idea as I'm not sure if there will be any other PG objects that share the same attributes. I decided on option 1 because I felt that `pg_extern` is already exposed as a API that users know how to use. So `pg_cast` can focus on just adding the `CREATE CAST` statement. Currently, using `pg_cast` without `pg_extern` or using `pg_extern` without `pg_cast` both do nothing so the "error" handling seems okay to me.

I may be missing something or not grasping the overall macro design consideration here so curious about what you think!

workingjubilee:
Option 2 does seem bad, yes.

> Currently, using `pg_cast` without `pg_extern` or using `pg_extern` without `pg_cast` both do nothing so the "error" handling seems okay to me.

Oh, we wouldn't want to accept using `#[pg_cast]` without `#[pg_extern]`. In that case, because the user specified an intent to designate a cast, but hasn't designated a valid target of a cast, we should error.

Obviously, that's a one-way constraint. The other way around wouldn't suddenly emit errors, since that would break... uh, every instance of `#[pg_extern]` ever.

In general, people don't actually _write_ the `#[pg_operator]` invocations. It's available, but for the most part they are instead part of the special `derive`s, so it is tolerable if their API is confusing. That might help to keep in mind when interpreting our existing design.

For Option 3, it's fine if we make any number of breaking changes to that enum, so it remains a possibility.

pgrx-sql-entity-graph/src/pg_extern/entity/mod.rs
```diff
@@ -472,6 +474,122 @@ impl ToSql for PgExternEntity {
...
+                                                        \tAS\n\
+                                                        \t{schema_prefix_target}{target_arg_sql} /* {target_name} */\n\
+                                                    )\n\
+                                                    WITH FUNCTION {function_name}{optional};\
+                                                    ",
+                                                    file = self.file,
+                                                    line = self.line,
+                                                    name = self.name,
+                                                    module_path = self.module_path,
+                                                    source_arg_sql = source_arg_sql,
```
workingjubilee: I prefer to avoid redundantly specifying names in formatting.
```suggestion
```

pgrx-sql-entity-graph/src/pg_extern/entity/mod.rs
```diff
@@ -472,6 +474,122 @@ impl ToSql for PgExternEntity {
...
+                                                    schema_prefix_source = context.schema_prefix_for(&source_arg_graph_index),
+                                                    source_name = source_arg.type_name,
+                                                    target_arg_sql = target_arg_sql,
```
workingjubilee:
```suggestion
```

pgrx-sql-entity-graph/src/pg_extern/entity/mod.rs
```diff
@@ -472,6 +474,122 @@ impl ToSql for PgExternEntity {
...
+                                                    schema_prefix_target = context.schema_prefix_for(&target_arg_graph_index),
+                                                    target_name = target_arg.type_name,
+                                                    function_name = self.name,
+                                                    optional = optional,
```
workingjubilee:
```suggestion
```

pgrx-sql-entity-graph/src/pg_extern/mod.rs
```diff
@@ -230,6 +236,36 @@ impl PgExtern {
...
+                                match p.path.segments.last().unwrap().ident.to_string().as_str() {
+                                    "implicit" => cast = PgCast::Implicit,
+                                    "assignment" => cast = PgCast::Assignment,
+                                    _ => eprintln!("Unrecognized option: {}. Using default cast options.", p.path.to_token_stream()),
+                                }
+                            }
+                            _ => eprintln!(
+                                "Unable to parse attribute to pg_cast as a Rust Expr: {}. Using default cast options.",
+                                attr.tokens
+                            ),
```
workingjubilee: I would prefer to not silently accept something that can be validated at compile time. Please return an error or panic instead.

workingjubilee (dismissed):
For the errors I am asking for, please validate they occur using `pgrx-tests/tests/ui` tests.

pgrx-sql-entity-graph/src/pg_extern/entity/mod.rs
```diff
@@ -472,6 +474,122 @@ impl ToSql for PgExternEntity {
...
+                            .ok_or(eyre!("Found a composite type but macro expansion time did not reveal a name, use `pgrx::composite_type!()`"))?;
+                        format!("{composite_type}[]")
+                    } else {
+                        self.fn_args[0].used_ty.composite_type
+                            .ok_or(eyre!("Found a composite type but macro expansion time did not reveal a name, use `pgrx::composite_type!()`"))?.to_string()
+                    }
+                }
+                Ok(SqlMapping::Skip) => {
+                    return Err(eyre!("Found an skipped SQL type in a cast, this is not valid"))
+                }
```
workingjubilee: At least one of these errors should also have a UI test for it. It's fine if you can't contrive a test case for each of these errors, I just want to see one.

pgrx-macros/src/lib.rs
```diff
@@ -148,6 +148,13 @@ pub fn initialize(_attr: TokenStream, item: TokenStream) -> TokenStream {
     item
 }
 
+/// Declare a function as `#[pg_cast]` to indicate that it represents a Postgres cast
+/// `cargo pgrx schema` will automatically generate the underlying SQL
+#[proc_macro_attribute]
+pub fn pg_cast(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    item
+}
```
workingjubilee: This should, if we go with `pg_cast` requiring `pg_extern`, specify that it requires `pg_extern`.

In either case, it should also document the valid arguments to the attribute (implicit, assignment), because these things are not made obvious by tooling, unlike how they would be in e.g. a function (which would emit its arguments and types into rustdoc, at least).

Not all `proc_macro_attribute`s in `pgrx-macros` do this, so you may be short on examples to imitate. It just has to cover the very basics, here.

pgrx-sql-entity-graph/src/pg_extern/entity/mod.rs
xwkuang5: Done

pgrx-sql-entity-graph/src/pg_extern/entity/mod.rs
xwkuang5: Done

pgrx-sql-entity-graph/src/pg_extern/entity/mod.rs
xwkuang5: Done

pgrx-sql-entity-graph/src/pg_extern/mod.rs
xwkuang5: Done. Switched to panic on invalid options

pgrx-sql-entity-graph/src/pg_extern/entity/mod.rs
xwkuang5: I added two basic tests in the `ui` directory but could not figure out how to run them. Could you share a command that I can use to invoke those UI tests? I also tried `cargo test --verbose --features "pg13" --no-default-features -- --list` to list the tests in `pgrx-tests` but could not find any ui test

pgrx-macros/src/lib.rs
xwkuang5: Went with a different approach of having `pg_cast` do the compile time validation and only after validation delegate to `PgExtern` for code-gen for the actual pg functions and casts. This avoids all the impossible states that the previous approach has.

Let me know what you think!

workingjubilee:
Nice! This seems pretty good. Going to take a closer look tomorrow, but it seems mergeable as-is or with perhaps small changes. I'm [considering revising the idiom for these attribute macros to match the SQL they implement](https://github.com/pgcentralfoundation/pgrx/issues/1488) but that would go beyond this PR so it won't affect this, though I'm happy to hear if you have any feelings after having tooled around here.

pgrx-tests/tests/ui/invalid_pgcast_options.rs
```diff
@@ -0,0 +1,8 @@
+use pgrx::prelude::*;
+
+#[pg_cast(invalid_opt)]
+pub fn cast_function(foo: i32) -> i32 {
+    foo
+}
+
+fn main() {}
```
workingjubilee, line 8: This one fails to compile (which is correct), please run:
```shell
TRYBUILD=overwrite cargo test --test ui
```
and then commit the resulting .stderr file.

pgrx-tests/tests/ui/invalid_pgcast_function.rs
```diff
@@ -0,0 +1,8 @@
+use pgrx::prelude::*;
+
+#[pg_cast]
+pub fn cast_function() -> i32 {
+    0
+}
```
workingjubilee: This one successfully compiles (which is incorrect). The implementation needs to validate the argument count, it appears. When you have made this fail to compile, please run the same command as I described for the other test.

pgrx-sql-entity-graph/src/pg_extern/entity/mod.rs
```diff
@@ -472,6 +474,119 @@ impl ToSql for PgExternEntity {
...
+                    "PG cast function must have exactly one argument, got {}",
+                    self.metadata.arguments.len()
+                ));
+            }
+            if self.fn_args.len() != 1 {
+                return Err(eyre!(
+                    "PG cast function must have exactly one argument, got {}",
+                    self.fn_args.len()
+                ));
+            }
```
workingjubilee: Somehow, this does nothing. I'm not sure what's going on there, honestly. My "essentially just instinct" guess is that this is somehow skipped over outright during building the example function used, and the closest reason why seems to be that the `else if let Some` was missed?

Is it correct for it to be an `else if let`?

workingjubilee (changes requested):
Apologies, but this does seem to be unfinished! Thank you for writing the tests! I think it's probably just a missed condition check.

pgrx-tests/tests/ui/invalid_pgcast_options.rs
xwkuang5: Done

pgrx-tests/tests/ui/invalid_pgcast_function.rs
xwkuang5: Removed this one because this is a runtime error that only get triggered when we actually generate some SQL entities. Therefore it does not fit as a ui test.

pgrx-sql-entity-graph/src/pg_extern/entity/mod.rs
xwkuang5: Updated to `if let`. FWIW, this does do what we expect (output below), it's just this is only triggered at SQL generation time (see the other comment in the UI test). 

```
#[pg_cast]
pub fn cast_function() -> bool {
    true
}

    Finished dev [unoptimized + debuginfo] target(s) in 3.28s
 Discovering SQL entities
  Discovered 2 SQL entities: 0 schemas (0 unique), 2 functions, 0 types, 0 enums, 0 sqls, 0 ords, 0 hashes, 0 aggregates, 0 triggers
     Writing SQL entities to /dev/stdout
Error:
   0: Could not write SQL to stdout
   1: PG cast function (cast_function) must have exactly one argument, got 0

Location:
   pgrx-sql-entity-graph/src/pg_extern/entity/mod.rs:515
```

pgrx-tests/tests/ui/invalid_pgcast_function.rs
workingjubilee: Oh okay! I sometimes don't remember what actually happens during compilation and what is technically """runtime""". I'd like to add schema-gen testing to the UI work, but I think that may need a burlier library than trybuild to do it, so it's definitely a separate issue. Thanks!

workingjubilee (approved):
Great!

xwkuang5:
Thank you for the review!

Issue #653, opened 2022-08-25
Incrementally adding `#[deny(unsafe_op_in_unsafe_fn)]`


workingjubilee:
This attribute should be attached to any code blocks that currently can build in practical cases with it in place (macros may make this peculiar and harder to tell).

It is preferable to, inside an `unsafe fn`, have a specific documentation on why entering the `unsafe` block is valid, to go with the documentation on the `unsafe fn` that describes the requirements on the caller and such. And it is useful to know if an `unsafe fn` imposes only requirements but has no actual `unsafe` operations inside!

There are also clippy lints for things like `// SAFETY` blocks we might want to use, though I know less about those.

PR #1447, opened 2023-12-27
Begin `unsafe_op_in_unsafe_fn` conversion.

Lokathor:
Small progress on https://github.com/pgcentralfoundation/pgrx/issues/653, not sure if this is what @workingjubilee actually had in mind


workingjubilee:
Yeah, that was basically the idea. I had been doing it "inside-out" instead, but either direction is fine.

PR #1448, opened 2023-12-27
Allow to build cargo-pgrx either with rustls (default) or native-tls

jirutka:
ureq defines `default = ["tls", "gzip"]`, so "tls" feature (rustls) is currently always enabled even when native-tls is preferred, i.e. both rustls and native-tls is built if openssl headers are available and the user cannot opt-out from using rustls. This also disallows building cargo-pgrx on platforms not supported by the ring crate (dependency of rustls).

This should also solve #1430 if `cargo-pgrx` is built with `--no-default-features --features native-tls`.

Supersedes #1442

### DIFF
--- a/cargo-pgrx/Cargo.toml
+++ b/cargo-pgrx/Cargo.toml
@@ -38,7 +38,7 @@ prettyplease = "0.2.15"
 proc-macro2 = { version = "1.0.69", features = [ "span-locations" ] }
 quote = "1.0.33"
 regex = "1.10.0"
-ureq = "2.8.0"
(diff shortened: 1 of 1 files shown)


workingjubilee:
Ah, I see. That does make more sense. My apologies, I misapprehended what was being fixed due to the conversation drift.

workingjubilee:
Right, this seems fine, and any followup fixes we require (like the one for #1430 with rustls enabled, which I think is actually even easier than I thought, now that I look) are easier to do with the two features properly partitioned out. Thank you!

PR #1449, opened 2023-12-27
Use native-certs with ureq

workingjubilee:
I at first had settled down to write up something using `rustls_native_certs` but then I actually noticed ureq simply enables this with a feature.

This should fully solve #1430 for both features now (rustls and native-tls).


workingjubilee:
I am not entirely sure what an adequate test for this behavior would be, beyond what we are already doing.

Issue #1383, opened 2023-11-08
Handle lifetimes correctly in pgrx-sql-entity-graph


workingjubilee:
Arrays and VariadicArrays and PgHeapTuples use various dirty hacks with staticized lifetimes.

workingjubilee:
This was required by some code which pretended to use variable lifetimes but actually only ever returned `'static` after macro expansion. So as a preliminary step, all code which secretly relies on always returning `'static` will likely have to become code that explicitly returns `'static`. This will make things look much worse before they get better, but then everything will get actually better, in due time.

We may be able to save Arrays, but we are likely going to wind up breaking composite types.

workingjubilee:
For things like the `composite_type!` macro, we also need to invent lifetimes, and it was very convenient to use the `'static` lifetime in all cases. This is because it freed us from thinking about the problem of `fn('a, 'b) -> '???`: any argument accepting two lifetime-bounded objects and returning a third object bound by a lifetime. The compiler cannot infer which lifetime it is supposed to use.

workingjubilee:
For `composite_type!`, that can currently be solved by simply avoiding providing a default `'static` lifetime, and then using https://github.com/pgcentralfoundation/pgrx/pull/1463 to pass all lifetimes through correctly.

workingjubilee:
Closed in #1476.

PR #1457, opened 2023-12-30
Adjust StringInfo wrapper impl to match actual definition

workingjubilee:
A commit that alters the definition of StringInfo was introduced about 2 months ago to Postgres: https://github.com/postgres/postgres/commit/f0efa5aec19358e2282d4968a03db1db56f0ac3f

It contains the following comment:
```
  * As a special case, a StringInfoData can be initialized with a read-only
  * string buffer.  In this case "data" does not necessarily point at a
  * palloc'd chunk, and management of the buffer storage is the caller's
  * responsibility.  maxlen is set to zero to indicate that this is the case.
  * Read-only StringInfoDatas cannot be appended to or reset.
  * Also, it is caller's option whether a read-only string buffer has a
  * terminating '\0' or not.  This depends on the intended usage.
```

This will make the following sequence of Rust code completely unsound **by mere invocation** as of Postgres 17, even assuming the lifetime constraint is correct:
```rust
let stringinfo: *mut pg_sys::StringInfoData = something_that_returns_ptr_to_stringinfodata();
let stringinfo: StringInfo<AllocatedByPostgres> = unsafe { StringInfoData::from_pg(stringinfo) };
let cstr_from_stringinfo: &'static CStr = stringinfo.into();
```

However, that commit has its own remarks:
```
There were various places in our codebase which conjured up a StringInfo
by manually assigning the StringInfo fields and setting the data field
to point to some existing buffer.  There wasn't much consistency here as
to what fields like maxlen got set to and in one location we didn't
correctly ensure that the buffer was correctly NUL terminated at len
bytes, as per what was documented as required in stringinfo.h
```

..."In one location we didn't correctly ensure that the buffer was correctly NUL terminated"?

On actually reviewing the commit, in multiple places, the maximum len was previously set to -1 or other incorrect values to indicate similar sorts of "borrow-like" usages. In other words, the `max_len > len` invariant was not preserved either, so it was not impossible for the pointee buffer to actually end at exactly len N, instead of also having another byte available.

While the only constructor for the StringInfoData wrapper is its `unsafe fn from_pg`, we were most certainly relying on this being true, and making comments as if we expected Postgres to not randomly pass us bogus data. In actuality, we find now that Postgres passes us bogus data all the time. As this guarantee has never actually been enforced, we should remove the implementation of `impl From<StringInfo<_>> for &'static CStr` immediately. This also means our Drop implementation has to go.

A more direct API is offered instead via `StringInfo::leak_cstr`.

### DIFF
--- a/pgrx-macros/src/lib.rs
+++ b/pgrx-macros/src/lib.rs
@@ -840,7 +840,8 @@ fn impl_postgres_type(ast: DeriveInput) -> syn::Result<proc_macro2::TokenStream>
             pub fn #funcname_out #generics(input: #name #generics) -> &'static ::core::ffi::CStr {
                 let mut buffer = ::pgrx::stringinfo::StringInfo::new();
                 ::pgrx::inoutfuncs::JsonInOutFuncs::output(&input, &mut buffer);
-                buffer.leak_cstr()
+                // SAFETY: We just constructed this StringInfo ourselves
+                unsafe { buffer.leak_cstr() }
             }
 
         });
@@ -863,7 +864,8 @@ fn impl_postgres_type(ast: DeriveInput) -> syn::Result<proc_macro2::TokenStream>
             pub fn #funcname_out #generics(input: #name #generics) -> &'static ::core::ffi::CStr {
                 let mut buffer = ::pgrx::stringinfo::StringInfo::new();
                 ::pgrx::inoutfuncs::InOutFuncs::output(&input, &mut buffer);
-                buffer.leak_cstr()
+                // SAFETY: We just constructed this StringInfo ourselves
+                unsafe { buffer.leak_cstr() }
             }
         });
     } else if args.contains(&PostgresTypeAttribute::PgVarlenaInOutFuncs) {
@@ -885,7 +887,8 @@ fn impl_postgres_type(ast: DeriveInput) -> syn::Result<proc_macro2::TokenStream>
             pub fn #funcname_out #generics(input: ::pgrx::datum::PgVarlena<#name #generics>) -> &'static ::core::ffi::CStr {
                 let mut buffer = ::pgrx::stringinfo::StringInfo::new();
                 ::pgrx::inoutfuncs::PgVarlenaInOutFuncs::output(&*input, &mut buffer);
-                buffer.leak_cstr()
+                // SAFETY: We just constructed this StringInfo ourselves
+                unsafe { buffer.leak_cstr() }
             }
         });
     }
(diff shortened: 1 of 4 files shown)


workingjubilee:
Oh, eff. Forgot about the sql annotation... sigh.

workingjubilee:
...okay the "just one more level"-ness of this has faded away so I'm stopping here. The only way this is currently failing tests, as far as I can tell, is if we have _always been unsound_ about this code. Which seems unfortunately plausible.

eeeebbbbrrrr:
I probably want to say something here but I’m not sure what. I’ll circle back in the new year.

workingjubilee:
> ...okay the "just one more level"-ness of this has faded away so I'm stopping here. The only way this is currently failing tests, as far as I can tell, is if we have always been unsound about this code. Which seems unfortunately plausible.

To be clear about my remark here: I mean violating Rust-level requirements. It may be that I'm misunderstanding something about the current code and need to review everything again, but the current panics shouldn't really happen if our code wasn't breaking the `_unchecked` CStr contract.

I think we should allow ourselves to rely on some invariants that Postgres specifies but favor checking those invariants... much more often, especially when it's some sort of basically-one-time check on instantiation of a type.

workingjubilee:
Okay, so the tests actually want to work now that I restart them? Okay, GitHub Actions. I see you.

pgrx-macros/src/lib.rs
```diff
@@ -836,11 +836,12 @@ fn impl_postgres_type(ast: DeriveInput) -> syn::Result<proc_macro2::TokenStream>
...
             #[doc(hidden)]
-            #[::pgrx::pgrx_macros::pg_extern(immutable,parallel_safe)]
-            pub fn #funcname_out #generics(input: #name #generics) -> &#lifetime ::core::ffi::CStr {
+            #[::pgrx::pgrx_macros::pg_extern (immutable,parallel_safe)]
+            pub fn #funcname_out #generics(input: #name #generics) -> &'static ::core::ffi::CStr {
                 let mut buffer = ::pgrx::stringinfo::StringInfo::new();
                 ::pgrx::inoutfuncs::JsonInOutFuncs::output(&input, &mut buffer);
-                buffer.into()
+                // SAFETY: We just constructed this StringInfo ourselves
+                unsafe { buffer.leak_cstr() }
```
workingjubilee, line 844: This is bad, for various reasons, but right now it has to be this way. The proper fix is blocked on https://github.com/pgcentralfoundation/pgrx/issues/1383 but this issue existing blocks solving that, because the previous, worse version of this code requires the staticizing hack and ignoring lifetimes in order to be allowed to ship. In other words:

**I know it's unsound, and what's currently extant is also unsound, so there's no actual change, except that now the unsoundness becomes fixable.**

workingjubilee:
The unsoundness I got confused about (not related to StringInfo, but related to the `from_bytes_with_nul_unchecked` contract) is actually not a problem, the real problem was "CI lol". Or a cache somewhere.

We absolutely are not using `_unchecked` anymore though.

PR #1461, opened 2024-01-04
Manually staticize lifetimes in pgrx tests

workingjubilee:
These all are actually lifetimes bound by their invocation site, which means that they return any lifetime which is named, instead of a lifetime that is bound by the input types in a logical way. This means that these lifetimes might as well be `'static`, for all it would (not) matter.

An important detail is that while this pattern is normally acceptable _and useful_ in Rust code, in pgrx it should be rejected for now because the lifetimes on the actually-called-by-Postgres function are defined, essentially, by a function _pointer_. These lifetimes cannot be coherently described to a function pointer. Thus, they form a sort of impedance mismatch: when you call this function from Rust code, you get a constraint that doesn't actually exist in the code that will run in Postgres. The difference is confusing, allowing you to write code relying on the compiler's lifetime checks and even fail to call it in certain ways, seemingly satisfying the soundness requirements. Then suddenly the assumption those compiler checks are founded on evaporates in the actual runtime execution, leading to undefined behavior.

I hope to fix a number of these problems relatively soon, but first thing's first: The removal of illusions. If this code is unsound now, then in truth it already was.

This effectively serves as a followup to https://github.com/pgcentralfoundation/pgrx/pull/1457 and it is part of what enables solving https://github.com/pgcentralfoundation/pgrx/issues/1383 by not requiring the pervasive use of that hack.

### DIFF
--- a/pgrx-examples/srf/src/lib.rs
+++ b/pgrx-examples/srf/src/lib.rs
@@ -33,15 +33,15 @@ fn random_values(num_rows: i32) -> TableIterator<'static, (name!(index, i32), na
 }
 
 #[pg_extern]
-fn result_table<'a>() -> Result<
-    Option<::pgrx::iter::TableIterator<'a, (name!(a, Option<i32>), name!(b, Option<i32>))>>,
+fn result_table() -> Result<
+    Option<::pgrx::iter::TableIterator<'static, (name!(a, Option<i32>), name!(b, Option<i32>))>>,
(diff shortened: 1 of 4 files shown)


workingjubilee:
For an example of a legitimate use of this pattern: An `'invariant` lifetime would not unify with `'static`. The problem is that you can't describe the lifetime of `Invariant<'lt>` to a function pointer, and the function pointer is what is ultimately truly called.

PR #1466, opened 2024-01-08
fix undefined behaviour when loading control file

usamoi:
It is introduced in https://github.com/pgcentralfoundation/pgrx/pull/911 and it should be a mistake.


workingjubilee:
It seems to predate that.

workingjubilee:
Very "I have no idea how anything actually has been working" hours here, but lgtm.

workingjubilee:
Oh, no, you're right, this was introduced there.

PR #1468, opened 2024-01-08
remove dlopen hack

### LINKED ISSUE #1369: Fixing sql-entity-graph and cargo-pgrx's dlopen hack
I am like 80% sure I know how to fix pgrx-sql-entity-graph and cargo-pgrx in a way that makes it so we no longer have to do dlopen at build time or anything, but it's a fairly beefy refactor I don't have time to get into right now. It involves use of the [macro_magic](https://crates.io/crates/macro_magic) crate, or if necessary, extending it until it can fix the problem. However, I believe it is close to being able to solve the problem almost single-handedly.

I spoke to Ana about it and she agreed that it was basically what she would have done all along if it had been available at the time, instead of twiddling with linker hacks. It doesn't involve affecting anything we consider "public API", though it might affect our preferred design paths for future public APIs, so this should be done before adding any new design that involves extending our proc macros in significant ways that we want people to rely on.

Importantly, this would most likely enable building on Windows, and would definitely resolve issues re: surprising build instability, like
- https://github.com/pgcentralfoundation/pgrx/issues/1173

If anyone wants to do this in the next few months feel free to pepper me with questions.

usamoi:
Resolves https://github.com/pgcentralfoundation/pgrx/issues/1369

Motivation of this PR:
* no need to compile `cargo-pgrx` after upgrading compiler
* make it possible to support `binstall`

This PR:
* recompilation occurs even if we use a nightly toolchain.
* schema is generated by `cargo run --bin pgrx_embed_${ext_name}` , instead of using `libloading` crate (unused symbols are removed by linker, so it compiles)
* removes compiler version check
* removes codes about stub
* removes unnecessary metadata traits

This is a breaking change: users need to write `pgrx::pgrx_embed!();` to `src/pgrx_embed.rs` and add that `[[bin]]` to the `Cargo.toml`, otherwise schema generation will report an error.


eeeebbbbrrrr:
This is... this is exciting!

workingjubilee:
> find all entities using ctor, instead of using object crate (macro_magic does not support items that are not accessible)

Curious, why do we want to support non-accessible items?

workingjubilee:
Using linker sections again would be annoying, but could be plausible.

However:

> Rust's philosophy is that nothing happens before or after main and this library explicitly subverts that. The code that runs in the ctor and dtor functions should be careful to limit itself to libc functions and code that does not rely on Rust's stdlib services.

This sounds like it will eventually prove to be an unreasonable expectation, if it is not one already?

workingjubilee:
Does this support cross-compilation?

cargo-pgrx/src/templates/cargo_toml
```diff
@@ -4,7 +4,7 @@ version = "0.0.0"
 edition = "2021"
 
 [lib]
-crate-type = ["cdylib"]
+crate-type = ["cdylib", "lib"]
```
workingjubilee, line 7: Why does everything need to also be a "lib" now? Can't this be worked around by switching our `cargo build` to `cargo rustc -- --crate-type="lib"`?

workingjubilee:
> Curious, why do we want to support non-accessible items?

To be clearer: when pgrx performs code expansion, even if the programmer doesn't put `pub` on their function, their function is linked into a publicly-accessible wrapper. The fact that we accept non-`pub` functions is mostly a relic of implementation details, they are still publicly-accessible in a fairly meaningful sense.

workingjubilee:
> Does this support cross-compilation?

I suppose schema gen has to happen on the build host anyways, so we're hoping programmers don't do anything silly like combining `pg_extern` with `cfg(target_...)`...

usamoi:
> > Does this support cross-compilation?
> 
> I suppose schema gen has to happen on the build host anyways, so we're hoping programmers don't do anything silly like combining `pg_extern` with `cfg(target_...)`...

They can take advantages of qemu. We had the idea in https://github.com/pgcentralfoundation/pgrx/pull/981#issuecomment-1383113583

cargo-pgrx/src/templates/cargo_toml
usamoi: It's needed by `cargo run --bin schema` and I do not know how to force the lib to be compiled to crate-type lib in this command

workingjubilee: Hmm. I will fuss with that for a bit, later, then, since it should still be approximately as easy but require a slightly more obscure command (or maybe just executing the binary. Either way).

usamoi:
We can use `cargo rustc --lib -- --Zunpretty=expanded` and `syn` to get all `#[no_mangle] extern "Rust" fn __pgrx_internals`, do some codegen for binary `schema` in `cargo-pgrx` and run binary `schema`. Is it an acceptable solution? Non-accessible and private items can be accessed (my code marks many modules containing `pg_extern` private and I do not want to break my code) and `ctor` is removed.

workingjubilee:
> We can use `cargo rustc --lib -- --Zunpretty=expanded` and syn to get all `#[no_mangle] extern "Rust" fn __pgrx_internals`, do some codegen for binary `schema` in cargo-pgrx and run binary `schema`. Is it an acceptable solution? Non-accessible and private items can be accessed (my code marks many modules containing `pg_extern` private and I do not want to break my code) and ctor is removed.

Yes, I think that sounds acceptable. My main concern was that anything based on linker sections is historically prone to dying when it slams into the rocks of platform-specific nonsense. So we simply don't have sufficient contributors to reasonably work through that and fix while also maintaining the other parts of pgrx, and they always cause issues in ways that are hard to even identify as a problem with the platform-specific thing.

I understand the concern about breakage. 0.12 is likely to have quite enough to fix to update to as-is, because it will land fixes for lifetime issues, so I agree we don't want to have even more breakage in this cycle.

usamoi:
CI passed and I will continue to finish unfinished work.

workingjubilee:
This looks very interesting!

I discussed this briefly with the other maintainers. Another note against the ctor crate: @thomcc mentions that rustc has problems with... accidentally discarding ctors. Only `#[no_mangle] pub extern "C" fn` can really be trusted to land in the binary. Hypothetically these ctor issues are bugs and rustc should have fixed them by now, but we feel uncertain that there aren't more bugs underneath, and they seem like good to avoid.

A concern of build performance regression was brought up, but at the same time, we don't really feel a regression that isn't "literally doubling build time" is worth blocking this over: pgrx has a bunch of low-hanging fruit on build time, and having a much less cursed build setup is worth sacrificing at least _some_ performance.

pgrx-examples/aggregate/embed.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
+    panic!("PGRX_EMBED is not override.");
```
workingjubilee: This should be, I think,
```suggestion
    panic!("PGRX_EMBED was not set.");
```

usamoi:
~blocked by https://github.com/rust-lang/cargo/issues/13280~ use cfg as a workaround

workingjubilee:
Hmm. That currently requires a `build.rs` in order to work, for [rerun-if-env-changed](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-env-changed). Is there a chance we could use something like `cfg` options?

usamoi:
`cargo expand` causes recompliation even if we use a nightly toolchain. Also using `cargo expand` is just moving from a trick to another trick since this is unstable. So I just add `object` crate back.

usamoi:
Link fails for `__pgrx_internals_*` symbols built only for `--cfg test`.

I think we can just ignore it (always build schema for tests not activated), since we of course want to test on same schema with a non-test build.

~A possible solution: generate and run schema directly in tests.~

They can use `#[cfg(any(test, feature = "pg_test"))]`.

usamoi:
Ready for merging now.

usamoi:
CI failed last week due to a homebrew issue and I rerun the CI just now.

workingjubilee:
Nice! Taking a closer look.

cargo-pgrx/src/command/schema.rs
```diff
@@ -458,22 +527,112 @@ fn compute_sql(
...
+    let command_str = format!("{:?}", command);
+    tracing::debug!(command = %command_str, "Running");
+    let embed_output = command
+        .output()
+        .wrap_err_with(|| format!("failed to spawn pgrx_embed: {}", command_str))?;
+    tracing::trace!(status_code = %embed_output.status, command = %command_str, "Finished");
+
+    if !embed_output.status.success() {
+        // We explicitly do not want to return a spantraced error here.
+        std::process::exit(1)
```
workingjubilee: I wish I remembered why. Let me look at the blame...

cargo-pgrx/src/command/schema.rs
```diff
@@ -353,47 +403,73 @@ fn compute_codegen(
...
+            #inputs
+            #build
+            #outputs
+        }
+    }
+    .to_string())
 }
 
-fn compute_sql(
+fn second_build(
```
workingjubilee, line 460: Hmm, what actually distinguishes "first build" and "second build"? Is this trying to do a kind of "incremental rebuild", without necessarily recomputing the entire graph? Is there a way we can make this a single idempotent function?

cargo-pgrx/src/templates/pgrx_embed_rs
```diff
@@ -0,0 +1 @@
+pgrx::pgrx_embed!();
```
workingjubilee: In generated code, we prefer to name the `::pgrx` library directly in this manner, unless the macro must be expanded inside the `pgrx` crate itself. This prevents renaming the crate except via controlling Cargo and thus the compiler: occasionally pgrx's tooling is used to build untrusted code.

...but I don't think it matters in this case, and this is purely a stylistic note.

```suggestion
::pgrx::pgrx_embed!();

```

cargo-pgrx/src/command/schema.rs
workingjubilee: ...not sure why. It seems to be about avoiding emitting an error twice.
```suggestion
    if !embed_output.status.success() {
        // We do not want to return a spantraced error here, to
        // (speculative:) reduce the likelihood of emitting errors twice
        std::process::exit(1)
```

cargo-pgrx/src/command/schema.rs
```diff
@@ -376,195 +295,272 @@ pub(crate) fn generate_schema(
...
+        std::process::exit(1)
+    }
+
+    Ok(())
+}
+
+fn compute_sql(profile: &CargoProfile) -> eyre::Result<()> {
+    let mut bin = get_target_dir()?;
+    bin.push(profile.target_subdir());
+    bin.push("pgrx_embed");
```
workingjubilee: We shouldn't be placing the binary at `/target/{profile}/pgrx_embed`. This kind of intermediate artifact should be using something like `env::var("OUT_DIR")`, or at least just not dumping it in the same place as every other build of this. Otherwise, we encounter problems if we want to do any kind of build parallelism and will probably eventually need to roll our own flocking just to stop ourselves from stomping on ourselves.

At minimum this needs to be further namespaced.

workingjubilee (commented):
This unfortunately seems to make using the command line tools to interact with things in this repo a bit... confusing:
```console
$ cargo run --bin strings/pgrx_embed
error: no bin target named `strings/pgrx_embed`.
Available bin targets:
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    cargo-pgrx
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx-version-updater
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
```

This would be understandable if the other concerns were not as big a concern.

cargo-pgrx/src/command/schema.rs
workingjubilee: I mean, I see the use of the `--cfg` and such on the second build, but the relationship of "first" and "second" feels unclear as opposed to naming them for what they actually do. Especially once multiple extensions are being built in the same workspace.

cargo-pgrx/src/command/schema.rs
```diff
@@ -376,195 +295,272 @@ pub(crate) fn generate_schema(
...
     }
 
-    std::fs::write(&postmaster_hash_file, postmaster_bin_hash)
-        .wrap_err("could not write postmaster stub hash")?;
+    command.arg("--");
 
-    Ok(postmaster_stub_built)
+    command.args(["--cfg", "pgrx_embed"]);
+
+    command.env("PGRX_EMBED", embed_path.as_ref());
```
workingjubilee: After examining the problem more closely I think we are better off with something that uses `RUSTFLAGS="--cfg pgrx_embed" PGRX_EMBED="..." cargo run --bin some_binary_name` as the primary command we use to execute the schema generation, if we must set a cfg or whatever, which I think is closer to what you had in the first place. Sorry about the run-around!

workingjubilee:
To address the "suddenly all pgrx extensions look the same to tooling" thing, and to allow developers to have more than one pgrx extension in a workspace... more common than you might think... I (currently) think we're best off using the crate name plus a suffix so as to indicate our schema-generator's binary name. Perhaps just giving up on `cargo run --bin` being easy to type in a workspace is better? I'm not entirely sold on either path.

Either way, we also can help ourselves by adding notes to the Cargo.toml using a `[package.metadata.pgrx]` table if we like. The `[package.metadata]` table is completely unexamined and explicitly "unclaimed" by the cargo tooling, so we can write whatever we want to it. And we can coexist peacefully with other tools that do this as long as we use a sensible namespace, like using a namespaced subtable.

workingjubilee:
I opened https://github.com/rust-lang/cargo/issues/13312 and got some interesting feedback: using the same bin name for all of these can **definitely** cause problems.

cargo-pgrx/src/command/schema.rs
usamoi: `RUSTFLAGS="--cfg pgrx_embed"` leads to recompilation of all deps so I prefer `cargo rustc`, which only sets `cfg` for the bin target.

thomcc:
Just to check, we no longer use `ctor` right? (It won't work because rustc doesn't compile things as whole-archive, so it would lead to silently dropped exports in large crates).

usamoi:
> Just to check, we no longer use `ctor` right? (It won't work because rustc doesn't compile things as whole-archive, so it would lead to silently dropped exports in large crates).

Yes, we no longer use `ctor`.

workingjubilee:
~~It looks like my `todo` tests were a bad idea in actuality, since they have many tiny spurious failures, so the current CI status doesn't reflect on this PR. I will PR an ignore for them (on Monday, probably), if nobody beats me to it.~~ Handled, it was actually just one test that was prone to spuriousness.

pgrx-sql-entity-graph/src/extension_sql/mod.rs
```diff
@@ -196,6 +196,7 @@ impl ToEntityGraphTokens for ExtensionSql {
             syn::Ident::new(&format!("__pgrx_internals_sql_{}", name.value()), Span::call_site());
         quote! {
             #[no_mangle]
+            #[doc(hidden)]
```
workingjubilee, line 199: true.

pgrx/src/lib.rs
```diff
@@ -327,3 +323,16 @@ pub(crate) enum Utf8Compat {
     /// An "extended ASCII" encoding, so we're fine if we only touch ASCII
     Ascii,
 }
+
+#[doc(hidden)]
```
workingjubilee: This is visibly called by a binary we template into their project directory, and they can read it without doing something like `cargo expand`, so it should have documentation. We should explain this macro is part of schema generation and it should not be called outside of the schema-gen binary, because it serves as an entry point for inserting the code our proc macros expand to generate the schema, which will be arbitrary code.

cargo-pgrx/src/command/schema.rs
workingjubilee: Ah, true. That makes sense then. Can you add a comment about the reasons we are choosing `cargo rustc` over its alternatives, then, like that?

pgrx-sql-entity-graph/src/used_type.rs
```diff
@@ -342,9 +342,8 @@ impl UsedType {
                 /// Set via the type being an `Option`.
                 optional: #optional,
                 metadata: {
-                    use ::pgrx::pgrx_sql_entity_graph::metadata::PhantomDataExt;
-                    let marker: core::marker::PhantomData<#resolved_ty> = core::marker::PhantomData;
-                    marker.entity()
+                    use ::pgrx::pgrx_sql_entity_graph::metadata::SqlTranslatable;
+                    <#resolved_ty>::entity()
```
workingjubilee, line 346: Since this seems to take care of both use sites for PhantomDataExt, we can simply delete that trait too, in pgrx-sql-entity-graph/src/metadata/phantomdata_ext.rs and make this a nice -100loc PR.

pgrx/src/lib.rs
workingjubilee: For example:

```suggestion
/// Entry point for cargo-pgrx's schema generation so that PGRX's framework can
/// generate SQL for its types and functions and topographically sort them into
/// an order Postgres will accept. Typically written by the `cargo pgrx new`
/// template, so you probably don't need to worry about this.
```

cargo-pgrx/src/command/schema.rs
workingjubilee: Revisiting this:
> Is there a way we can make this a single idempotent function?

No, I don't think so.

cargo-pgrx/src/command/schema.rs
workingjubilee: Note that the "further namespacing" by using the concatenated binary name of `ext_name_pgrx_embed` (or whatever) is sufficient.

usamoi:
`pgrx_embed` is renamed to `pgrx_embed_{name}` now.

workingjubilee (approved):
re: 5566ceab66365a01fa0aff4db017194251d83c2a yeah that can get a bit long...

I might take a look later at adjusting the exact way that is handled before 0.12.0 actual... We may want to use the package metadata table to record a name, for future evolution purposes, and use that to make it more concise. It also would let people to rearrange their repo how they want, assigning whatever binary and file names they want, without confusing the pgrx tooling.

That's not a blocking concern for this, however. This looks great and a new extension passes `cargo pgrx test all` after doing the `path` hack and rerunning `cargo pgrx init`, which means that all we need to make this reusable is to publish it. This opens up so many possibilities! Thank you.

eeeebbbbrrrr:
@usamoi: Thanks for all your work here.  You too, of course, @workingjubilee.  

It's really special that someone from the community helped solve this problem.  Much appreciation and respect!

PR #1472, opened 2024-01-09
Allow `detail` argument and trailing commas in `ereport!`

hoodiecollin:
I use `ereport!` because I like to include a `detail` entry but it comes with some inconveniences. When I would write the following snippet in my `Err` match arms, compiler would complain about mismatch types being output from the match arms thus requiring the inclusion of `unreachable!` in my code.

```rust
pgrx::ereport!(
    PgLogLevel::ERROR,
    PgSqlErrorCode::ERRCODE_INTERNAL_ERROR,
    format!("{} failed", stringify!(snowflake_pool_stats)),
    err.to_string()
);

// necessary to satisfy the compiler
unreachable!()
```

After looking at the macro src for `ereport!` I realized why I started writing `PgLogLevel::ERROR` rather than simply `ERROR`; the `ERROR` case didn't support a `detail` argument.

Going forward, I would like to use the dedicated macros for the common use cases of logging/aborting so hopefully someone will find these proposed changes to be a worthwhile improvement.

### DIFF
--- a/pgrx-pg-sys/src/submodules/elog.rs
+++ b/pgrx-pg-sys/src/submodules/elog.rs
@@ -336,77 +336,84 @@ macro_rules! function_name {
 /// ```
 #[macro_export]
 macro_rules! ereport {
-    (ERROR, $errcode:expr, $message:expr) => {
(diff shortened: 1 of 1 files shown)


workingjubilee:
This seems reasonable, yep!

PR #1473, opened 2024-01-09
Alternative input syntax for `elog` macros

hoodiecollin:
# tl;dr

I recently proposed some slight adjustments to the `ereport` macro to allow `detail` to be passed in all cases. After thinking it through, I decided to further improve the usability of all the reporting macros.

## Goals

- [x] Allow `detail` and `hint` to be specified in all reporting macros
- [x] No breaking changes
- [x] Support formatting arguments for `detail` and `hint` just like the `message`
- [x] `hint` can be provided without needing to also provide `detail`
- [x] `hint` can be provided before `detail`

## Notes

- I updated the docs for each macro to reflect these changes 

## Examples

```rust
pgrx::error!(
    message = "something went wrong";
    detail = "original error: {}", err.to_string();
    hint = "check your `.env` file";
);
```

### DIFF
--- a/pgrx-pg-sys/src/submodules/elog.rs
+++ b/pgrx-pg-sys/src/submodules/elog.rs
@@ -103,6 +103,36 @@ impl From<i32> for PgLogLevel {
 ///
 /// The output these logs goes to the PostgreSQL log file at `DEBUG5` level, depending on how the
 /// [PostgreSQL settings](https://www.postgresql.org/docs/current/runtime-config-logging.html) are configured.
+///
+/// ## Adding Details and Hints
+///
+/// Additionally, you can use specify a `detail` and/or `hint` to be included with the report.
+/// After the string literal to be used as the message and any formatting arguments, add a semicolon.
+/// Then assign a string literal to `detail` or `hint`, optionally followed by a list of
+/// comma separated formatting arguments. The `detail` and `hint` options can be used together or
+/// separately but `detail` must come before `hint` and both must end with a semicolon.
+///
+/// ## Examples
+///
+/// ```rust,no_run
+/// // just like `println!` and `format!`
+/// debug5!("a simple message");
+/// debug5!("or a formatted message: {:?}", "pgrx rocks!");
+/// // include details and hints
+/// debug5! {
+///     "add details if you want";
+///     detail = "...";
+/// }
+/// debug5! {
+///     "or a message with just a hint";
+///     hint = "...";
+/// }
+/// debug5! {
+///     "put it all together for {} {}", "great", "success!";
+///     detail = "extra {}", "info";
+///     hint = "{} helpful", "very";
+/// }
+/// ```
 #[macro_export]
 macro_rules! debug5 {
     (
@@ -136,6 +166,36 @@ macro_rules! debug5 {
 ///
 /// The output these logs goes to the PostgreSQL log file at `DEBUG4` level, depending on how the
 /// [PostgreSQL settings](https://www.postgresql.org/docs/current/runtime-config-logging.html) are configured.
+///
+/// ## Adding Details and Hints
+///
+/// Additionally, you can use specify a `detail` and/or `hint` to be included with the report.
+/// After the string literal to be used as the message and any formatting arguments, add a semicolon.
+/// Then assign a string literal to `detail` or `hint`, optionally followed by a list of
+/// comma separated formatting arguments. The `detail` and `hint` options can be used together or
+/// separately but `detail` must come before `hint` and both must end with a semicolon.
+///
+/// ## Examples
+///
+/// ```rust,no_run
+/// // just like `println!` and `format!`
+/// debug4!("a simple message");
+/// debug4!("or a formatted message: {:?}", "pgrx rocks!");
+/// // include details and hints
+/// debug4! {
+///     "add details if you want";
+///     detail = "...";
+/// }
+/// debug4! {
+///     "or a message with just a hint";
+///     hint = "...";
+/// }
+/// debug4! {
+///     "put it all together for {} {}", "great", "success!";
+///     detail = "extra {}", "info";
+///     hint = "{} helpful", "very";
+/// }
+/// ```
 #[macro_export]
 macro_rules! debug4 {
     (
@@ -169,6 +229,36 @@ macro_rules! debug4 {
 ///
 /// The output these logs goes to the PostgreSQL log file at `DEBUG3` level, depending on how the
 /// [PostgreSQL settings](https://www.postgresql.org/docs/current/runtime-config-logging.html) are configured.
+///
+/// ## Adding Details and Hints
+///
+/// Additionally, you can use specify a `detail` and/or `hint` to be included with the report.
+/// After the string literal to be used as the message and any formatting arguments, add a semicolon.
+/// Then assign a string literal to `detail` or `hint`, optionally followed by a list of
+/// comma separated formatting arguments. The `detail` and `hint` options can be used together or
+/// separately but `detail` must come before `hint` and both must end with a semicolon.
+///
+/// ## Examples
+///
+/// ```rust,no_run
+/// // just like `println!` and `format!`
+/// debug3!("a simple message");
+/// debug3!("or a formatted message: {:?}", "pgrx rocks!");
+/// // include details and hints
+/// debug3! {
+///     "add details if you want";
+///     detail = "...";
+/// }
+/// debug3! {
+///     "or a message with just a hint";
+///     hint = "...";
+/// }
+/// debug3! {
+///     "put it all together for {} {}", "great", "success!";
+///     detail = "extra {}", "info";
+///     hint = "{} helpful", "very";
+/// }
+/// ```
 #[macro_export]
 macro_rules! debug3 {
     (
@@ -202,6 +292,36 @@ macro_rules! debug3 {
 ///
 /// The output these logs goes to the PostgreSQL log file at `DEBUG2` level, depending on how the
 /// [PostgreSQL settings](https://www.postgresql.org/docs/current/runtime-config-logging.html) are configured.
+///
+/// ## Adding Details and Hints
+///
+/// Additionally, you can use specify a `detail` and/or `hint` to be included with the report.
+/// After the string literal to be used as the message and any formatting arguments, add a semicolon.
+/// Then assign a string literal to `detail` or `hint`, optionally followed by a list of
+/// comma separated formatting arguments. The `detail` and `hint` options can be used together or
+/// separately but `detail` must come before `hint` and both must end with a semicolon.
+///
+/// ## Examples
+///
+/// ```rust,no_run
+/// // just like `println!` and `format!`
+/// debug2!("a simple message");
+/// debug2!("or a formatted message: {:?}", "pgrx rocks!");
+/// // include details and hints
+/// debug2! {
+///     "add details if you want";
+///     detail = "...";
+/// }
+/// debug2! {
+///     "or a message with just a hint";
+///     hint = "...";
+/// }
+/// debug2! {
+///     "put it all together for {} {}", "great", "success!";
+///     detail = "extra {}", "info";
+///     hint = "{} helpful", "very";
+/// }
+/// ```
 #[macro_export]
 macro_rules! debug2 {
     (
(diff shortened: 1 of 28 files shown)


pgrx-pg-sys/src/submodules/elog.rs
```diff
@@ -103,6 +103,36 @@ impl From<i32> for PgLogLevel {
...
+/// debug5! {
+///     "or a message with just a hint";
+///     hint = "...";
+/// }
+/// debug5! {
+///     "put it all together for {} {}", "great", "success!";
+///     detail = "extra {}", "info";
+///     hint = "{} helpful", "very";
+/// }
+/// ```
```
workingjubilee: I recommend `cargo test --doc` to make sure these examples work.

hoodiecollin: Thanks for the tip! There were a couple issues:
- I needed to include `use pgrx_pg_sys::xxx;` in each example
- The compiler was struggling to identify the string literal in the `format!` calls. Passing the components through `format_args!` seems to have made things clearer.

While the implementation is considerably more verbose than what I originally anticipated, the end result delivers the ergonomics I was aiming for.

pgrx-pg-sys/src/submodules/elog.rs
```diff
@@ -103,14 +103,109 @@ impl From<i32> for PgLogLevel {
...
+                alloc::format!("{}", core::format_args!($msg $(, $msg_args)*));
+                $(hint = alloc::format!("{}", core::format_args!($hint $(, $hint_args)*));)?
+            );
+        }
+    );
+    (
+        $msg:literal $(, $msg_args:expr)*;
+        $(hint = $hint:literal $(, $hint_args:expr)*)?;
+        $(detail = $detail:literal $(, $detail_args:expr)*)?;
+    ) => (
```
workingjubilee: This is a somewhat deceptively simple syntax. It looks simple, but it actually has a fair amount of nuance. For instance, this introduces a significant divergence between the behavior in the presence of a `;` versus not, right?

I'll sleep on it, but it definitely can't be merged without UI tests covering pathological inputs, like
```rust
debug5!(hint = "..."; detail = "...";);
debug5!(detail = "...");
```

You can find the existing ui tests in `./pgrx-tests/tests/ui`. They allow us to cover erroneous cases that don't compile. The main reason for these tests is so that the tests fail if they _start_ compiling. I would prefer that the patterns that worked with 0.11.2 continue to work in 0.12.0, and that anything that seems weird or ambiguous is tested to fail.

pgrx-pg-sys/src/submodules/elog.rs
```diff
@@ -120,14 +215,109 @@ macro_rules! debug5 {
...
+        {
+            extern crate core;
+            extern crate alloc;
+            $crate::ereport!(
+                $crate::elog::PgLogLevel::DEBUG4,
+                $crate::errcodes::PgSqlErrorCode::ERRCODE_SUCCESSFUL_COMPLETION,
+                alloc::format!("{}", core::format_args!($msg $(, $msg_args)*));
+                $(detail = alloc::format!("{}", core::format_args!($detail $(, $detail_args)*));)?
+                $(hint = alloc::format!("{}", core::format_args!($hint $(, $hint_args)*));)?
```
workingjubilee: All of this eager allocation is starting to make me slightly unhappy. Perhaps it's unimportant in the grand scheme of things, and I don't have time to look into the impl deep enough to figure it out right now, but I really hope we aren't just allocating here only to palloc and copy it later? I won't block this change on that, but I would like to know.

pgrx-pg-sys/src/submodules/elog.rs
```diff
@@ -137,14 +327,109 @@ macro_rules! debug4 {
...
+            );
+        }
+    );
     ($($arg:tt)*) => (
         {
             extern crate alloc;
             $crate::ereport!($crate::elog::PgLogLevel::DEBUG3, $crate::errcodes::PgSqlErrorCode::ERRCODE_SUCCESSFUL_COMPLETION, alloc::format!($($arg)*).as_str());
         }
-    )
+    );
```
workingjubilee: This code has a rather lot of duplication now. That might have been tolerable when there was less code overall, but I should note that Rust allows you to define a macro that itself mostly serves to define new macros when invoked. It's an extremely "now you have two problems" sort of decision to go down that road, but this is at the point where it seems justified, doesn't it?

There's other solutions of course like a layer of inner macros, recursion, or whatever. I'm not too picky about that, as long as it's slightly more controlled than "there are **5** impls that are a copy-paste of each other, one for each debug level, and each is now about 100 lines".

hoodiecollin: 1. I agree this is too much duplication. I wanted to avoid introducing a new public macro but if you think it is justified then I can refactor to keep things DRY. 
2. When you say "define a macro that itself mostly serves to define new macros", do you mean a private `macro by example` that defines another macro that is public? Or a functional proc macro that expands to a public macro? I was under the impression that it wasn't possible or at least wasn't practical to implement higher-order macros with only `macro by example`.

pgrx-pg-sys/src/submodules/elog.rs
hoodiecollin: I don't think there is any extra allocation that could otherwise be avoided. [`set_detail` and `set_hint`](https://github.com/pgcentralfoundation/pgrx/blob/a5625169297cd378c3d206cbf501025b5b2f6ba8/pgrx-pg-sys/src/submodules/panic.rs#L266C1-L276C6) have the trait bound `S: Into<String>` so the resulting owned `String` from `format` should pass through without being copied. 

If my logic is correct then the original cases should be updated to omit the `.as_str()` to avoid an extra copy.

> Original implementation
```rust
$crate::errcodes::PgSqlErrorCode::ERRCODE_SUCCESSFUL_COMPLETION, alloc::format!($($arg)*).as_str());
```

pgrx-pg-sys/src/submodules/elog.rs
workingjubilee: A new public macro is not required.

Yes, I mean a private decl macro that expands to public decl macros.

It's possible and in my opinion practical, but it can be a bit of a trick, admittedly.

pgrx-pg-sys/src/submodules/elog.rs
workingjubilee: That's correct, but we control all the signatures of everything so we can change those signatures if we want, and the real question is: after the String is allocated, what is it used for?

pgrx-pg-sys/src/submodules/elog.rs
workingjubilee: If you hit me up on [Discord](https://discord.gg/9tAcg7w7), I can help walk you through the tricks more exactly (it's easier for talking a bit faster, and thus dialing in my explanation to a more exact understanding of what you yourself understand, so I can fill in the gaps), but otherwise it will be best if you review [The Little Book of Rust Macros](https://veykril.github.io/tlborm/introduction.html) if you haven't, including the [callback](https://veykril.github.io/tlborm/decl-macros/patterns/callbacks.html) pattern. It may help explain some of the limitations of macros that you encountered.

But if the cleanest solution is to add a new public macro or something, that's actually fine? In particular, I have for some time now thought all the `debugN!` macros should just be a single:
```rust
#[macro_export]
macro_rules! debug {
    (1, $($tok:tt)*) => {};
    (2, $($tok:tt)*) => {};
    (3, $($tok:tt)*) => {};
    (4, $($tok:tt)*) => {};
    (5, $($tok:tt)*) => {};
```

We can match on literals to some extent, so we probably should.

pgrx-pg-sys/src/submodules/elog.rs
hoodiecollin: After considering your feedback and reworking some things, I decided to change it to `message = $msg:literal`. I like this makes the syntax more obvious and helped to avoid some of the ambiguity when matching. I also added ui tests, please take a look and let me know if/how they can be improved.

pgrx-pg-sys/src/submodules/elog.rs
hoodiecollin: I refactored things to get things more DRY. ~When I run tests locally _(including ui tests)_ everything passes but the GH checks have an issue comparing one of the ui snapshots. Can you help me understand what the issue is?~

Any feedback you have on this updated implementation, I would be grateful to hear it. I can hop on a call via Discord tomorrow if it would be easier to discuss over VoIP.

pgrx-pg-sys/src/submodules/elog.rs
hoodiecollin: After looking deeper into the `.report()` implementation I see your point, there is potential to eliminate all Rust allocation in the case where it is known ahead of time that the payload will be handed off to `palloc` immediately. 

My first thoughts on how to achieve this would be:
- define a type that wraps `fmt::Arguments`, call it `ErrorReportData`.
- impl `AsPgCStr` for `ErrorReportData`
- define a type like `ErrorReport`, call it `ErrorReportWithBorrowedData` with fields of `ErrorReportData` rather than `String`.
- define an enum to differentiate between `ErrorReport` and `ErrorReportWithBorrowedData`, call it `PreparedErrorReport`.
- update `do_ereport_impl` functions to take `PreparedErrorReport`
- refactor the elog macros to use `ErrorReportWithBorrowedData` when the macro matches a `format!`-like usage, otherwise fallback to using `ErrorReport`

^ What are your first impressions? If this seems reasonable enough I can put something together _(I might need some help/guidance on creating unit tests for it. I admittedly have limited experience writing tests for unsafe code)_

workingjubilee: I see, that is what I expected. I think that can wait until a later PR since it's _technically_ unrelated. I think I would avoid using quite so many indirections, also, and I can explain how slightly better in another context.

pgrx-tests/tests/ui/elog/argument-order/detail_first.stderr
```diff
@@ -0,0 +1,10 @@
+error: The alternative invocation requires `message` to be specified first.
+  --> tests/ui/elog/argument-order/detail_first.rs:7:5
+   |
+7  | /     log!(
+8  | |         detail = "detail";
+9  | |         message = "...";
+10 | |     );
```
workingjubilee, line 7: Hmm. I don't understand why we would allow assignment at all if it's going to enforce a strict ordering: the main benefit of an assignment-based scheme, like with `format_args!`, is that I don't have to worry about that sort of thing!

pgrx-pg-sys/src/submodules/elog.rs
```diff
@@ -96,195 +96,333 @@ impl From<i32> for PgLogLevel {
...
-            unreachable!()
-        }
-    )
+define_elog! {
+    error = {
+        loglevel = ERROR
+        errcode = ERRCODE_INTERNAL_ERROR
+        docs = "This will abort the current Postgres transaction."
+        extra_docs = false
+        extra_stmts = {unreachable!()}
```
workingjubilee, line 404: We don't need this `extra_stmts` thing. This code is already non-returning and this doesn't emit any optimization hints or anything like that.

pgrx-tests/tests/ui/elog/semicolon-terminators/ereport_detail.stderr
```diff
@@ -0,0 +1,12 @@
+error: The alternative invocation style requires component sections to be terminated with a semicolon.
+  --> tests/ui/elog/semicolon-terminators/ereport_detail.rs:7:5
+   |
+7  | /     ereport!(
+8  | |         loglevel = PgLogLevel::LOG;
+9  | |         errcode = PgSqlErrorCode::ERRCODE_SUCCESSFUL_COMPLETION;
+10 | |         message = "...";
+11 | |         detail = "detail",
+12 | |     );
```
workingjubilee, line 9: The semicolon-oriented style is a fairly novel style, syntactically, and if this is what is necessitated by this approach, I think it is useful to step back and ask why we are requiring it to look like that?

pgrx-pg-sys/src/submodules/elog.rs
```diff
@@ -96,195 +96,333 @@ impl From<i32> for PgLogLevel {
...
+        extra_stmts = $extra_stmts:tt
+        dollar = $_:tt
+    ) => {
+        #[allow(non_snake_case)]
+        #[macro_export]
+        #[doc=concat!(
+            define_elog!(@get_first_doc_line name = $name $doc_str),
+            "\n",
+            "This macro accepts arguments like the [`println`] and [`format`] macros.\n",
+            "See [`fmt`](std::fmt) for information about options.\n",
```
workingjubilee, line 148: I worry about leaning on `fmt` syntax if we're going to use this `assignment = "{}", expr;` syntax which doesn't match the actual way fmt syntax is used?

pgrx-pg-sys/src/submodules/elog.rs
```diff
@@ -96,195 +96,333 @@ impl From<i32> for PgLogLevel {
...
+                        "  - detail: impl Into<String>\n",
+                        "  - hint: impl Into<String>\n",
+                        "\n",
+                        "## Example\n",
+                        "  ", stringify!($name), "! {\n",
+                        "    message = \"...\";\n",
+                        "    detail = \"extra detail\";\n",
+                        "    hint = \"plus a hint\";\n",
+                        "  }\n",
+                        "\n",
```
workingjubilee, line 300: This could stand to be more concise. This is an error message, not documentation.

pgrx-pg-sys/src/submodules/elog.rs
workingjubilee: ...Does this syntax currently support `message = "{message}", message = $expr;`?

hoodiecollin:
@workingjubilee ive taken a few days to step back and think about your feedback. I played around with manually creating ereport and adding hints/details vs my proposed macro syntax. Ultimately I can no longer justify the extra complexity since consumers of `pgrx` can achieve basically the same result with a set of simple constructor functions. For that reason, I am closing this PR.

I do think the optimizations we discussed regarding unnecessary allocations before an inevitable `palloc` are still worth exploring but I will open a separate PR for that.

PR #1476, opened 2024-01-13
Unbox types properly from arrays

workingjubilee:
Introducing `Datum<'src>` allows using it with HRTB fns in schema generation to persist lifetimes all the way through `#[pg_extern]`'s wrapper-fn infrastructure without compromising the integrity of lifetime annotations. This is immediately implemented to prevent many types from exceeding the lifetime of the Array they are unboxed from, unless they are already "self-owned" and thus 'static.

Unfortunately, this does not fix all instances using `pg_getarg`, and those simply are left unsolved for now in some cases. I reorganized the UI tests to make it easier to notice these problems and when they get solved.


workingjubilee:
I can split this down the line of [a0184f4](https://github.com/pgcentralfoundation/pgrx/pull/1476/commits/a0184f4edeafbb52d7ac37da7738e78a512c5d1d) and [e8ead32](https://github.com/pgcentralfoundation/pgrx/pull/1476/commits/e8ead32f62b80805dfbbb37044c20b2d9a4fecac) if you want, and I can add more notes explaining the lifetimes, names, etc. but this seems like it would benefit from a cursory glance to inform that step, at least.

pgrx-sql-entity-graph/src/lifetimes.rs
```diff
@@ -7,76 +7,6 @@
...
-    staticize_lifetimes(&mut ty);
-    match ty {
-        syn::Type::Path(type_path) => type_path,
-
-        // shouldn't happen
-        _ => panic!("not a TypePath"),
-    }
-}
-
-pub fn staticize_lifetimes(value: &mut syn::Type) {
```
eeeebbbbrrrr, line 24: goodby, `staticize_lifetimes`... you will not be missed!

pgrx-examples/bad_ideas/src/lib.rs
```diff
@@ -55,7 +55,7 @@ fn warning(s: &str) -> bool {
 #[pg_extern]
 fn exec<'a>(
     command: &'a str,
-    args: default!(Vec<Option<&'a str>>, "ARRAY[]::text[]"),
+    args: default!(Vec<Option<String>>, "ARRAY[]::text[]"),
```
eeeebbbbrrrr, line 58: Just a question... why this change?

Could it not be `Vec<Option<&'a str>>`?

This was the only example you touched, so either that means this example is the only one that accepts a `&str` and that pattern is no longer valid, or you changed this for some other reason?

eeeebbbbrrrr (approved):
Jubilee, this looks really clean.  It's clear you put a lot of thought into this to basically only have a few lifetime signature changes throughout the code.

Speaking of which, I'm surprised PgHeapTuple didn't require more effort than this.  Cool.

pgrx-examples/bad_ideas/src/lib.rs
workingjubilee: Because this changes FromDatum in the impls in pgrx/src/datum/array.rs to be in terms of `T: UnboxDatum` with certain bounds, this PR currently breaks `FromDatum` specifically for `Vec<Option<T>>` where there is any doubt as to whether `T` implements `UnboxDatum`. Essentially, the trait resolver can't figure it out at that depth. This can be solved by involving additional traits which work differently and either allow `Vec<Option<T>>` to not depend directly on `Array<'_, T>` for its unboxing or to move `pg_getarg` (or its replacement) off `T: FromDatum`.

workingjubilee:
> Speaking of which, I'm surprised PgHeapTuple didn't require more effort than this. Cool.

GitHub won't open it implicitly because it is a "large diff", but unfortunately, pgrx-tests/src/tests/heap_tuple.rs betrays how many cases won't work now, and so this isn't really finished work. I had a PR similar to this a couple months ago, which did (more) damage, but I needed all the work until now to understand whether it would be viable to even fix, and to fix a number of less-complex cases that hinged on that groundwork. I need to add in a bunch more changes to make all of the cases work that need to work properly. Some of those hinge on SPI, which started busted and remains busted.

...Ultimately I am unconcerned about making `PgHeapTuple<'a, WA>` work really _well_, because I now believe it's the wrong abstraction in the long run, so as a warning, I have accepted most of those examples getting broken on a semi-permanent basis.

workingjubilee:
I added the currently-broken tests to the set of ui tests, and reorganized tests slightly, so that it will be easier for us to notice when those problems are fixed.

pgrx-tests/tests/ui/heap-tuples-dont-leak.rs
```diff
@@ -0,0 +1,29 @@
...
+    dog.get_by_name("name").unwrap().unwrap()
+}
+
+#[pg_extern]
+fn gets_name_field_strict(dog: pgrx::composite_type!("Dog")) -> &str {
+    // Gets resolved to:
+    let dog: PgHeapTuple<AllocatedByRust> = dog;
+
+    dog.get_by_name("name").unwrap().unwrap()
+}
```
workingjubilee:
```suggestion
}

```

pgrx-tests/tests/compile-fail/heap-tuples-dont-leak.rs
```diff
@@ -0,0 +1,29 @@
...
+    dog.get_by_name("name").unwrap().unwrap()
+}
+
+#[pg_extern]
+fn gets_name_field_strict(dog: pgrx::composite_type!("Dog")) -> &str {
+    // Gets resolved to:
+    let dog: PgHeapTuple<AllocatedByRust> = dog;
+
+    dog.get_by_name("name").unwrap().unwrap()
+}
```
workingjubilee:
```suggestion
}

```

pgrx-tests/tests/todo/array-serialize-json.rs
```diff
@@ -0,0 +1,32 @@
...
+    #[pg_test]
+    fn test_serde_serialize_array() -> Result<(), pgrx::spi::Error> {
+        let json = Spi::get_one::<Json>(
+            "SELECT serde_serialize_array(ARRAY['one', null, 'two', 'three'])",
+        )?
+        .expect("returned json was null");
+        assert_eq!(json.0, json! {{"values": ["one", null, "two", "three"]}});
+        Ok(())
+    }
+}
```
workingjubilee:
```suggestion
}

```

PR #1480, opened 2024-01-17
Document CPU and OS support expectations

### LINKED ISSUE #1478: Document arch/OS support expectations
We have some things that are explicitly supported, some things that are explicitly unsupported, and some things that are known to not work but we are interested in exploring, or are unknown to work and if you can use them... then... sure?

My current understanding:
- 32-bit architectures are "explicitly unsupported" at the current moment, but we don't `compile_error!` currently because last we heard, someone was poking around with this possibility. We haven't heard back recently, so maybe we should add in a `compile_error!`
- Windows is ???: we aren't able to offer much technical expertise from our current bank of contributors, but we aren't all that averse to increasing support for it.
- We expect a typical Linux distro, with our prescribed configs, to essentially always work.
- We expect macOS to work, but homebrew sometimes is painful enough to make that a "usually".
- We don't explicitly support non-Linux non-Apple Unix OS, but they tend to work.

workingjubilee:
This documents our current state of support, and adds a note for people hoping for 32-bit support: it unfortunately would not be cheap, neither in terms of the incurred technical debt, nor in terms of running the additional build and test infra to test on 32-bit CPUs, which is not something GitHub Actions helps with.

This closes https://github.com/pgcentralfoundation/pgrx/issues/1478

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,8 @@ those remain untested. So far, some of PGRX's build tooling works on Windows, bu
 
  ⹋ PGRX has not been tested to work on 32-bit: the library assumes an 8-byte `pg_sys::Datum`
 which may result in unexpected behavior on 32-bit, like dropping 4 bytes of data from `int8`
-and `double`. This is not "unsound", but may undermine safety assumptions of PGRX extensions.
+and `double`. This may not be "unsound" in itself, as it is "merely" illogical,
+but it may undermine otherwise-reasonable safety assumptions of PGRX extensions.
 We do not plan to add support without considerable ongoing technical and financial contributions.
 
 <details>
--- a/README.md
+++ b/README.md
@@ -66,6 +66,10 @@
 
 ## System Requirements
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -78,9 +82,18 @@
 
  ‡ A local PostgreSQL server installation is not required. `cargo pgrx` can download and compile PostgreSQL versions on its own.
 
+ ⹋ PGRX has not been tested to work on 32-bit: the library assumes an 8-byte `pg_sys::Datum`
+which may result in unexpected behavior on 32-bit, like dropping 4 bytes of data from `int8`
+and `double`. This is not "unsound", but may undermine safety assumptions of PGRX extensions.
```
workingjubilee: I don't like the overstated confidence here:

```suggestion
and `double`. This may not be "unsound" in itself, as it is "merely" illogical,
but it may undermine otherwise-reasonable safety assumptions of PGRX extensions.
```

PR #1485, opened 2024-01-20
Include varlena functions

daamien:
There's already a pgrx::varlena module but this header brings utility fonctions to manipulate GUC variables and identifiers.

### DIFF
--- a/pgrx-pg-sys/include/pg12.h
+++ b/pgrx-pg-sys/include/pg12.h
@@ -158,3 +158,4 @@
 #include "utils/syscache.h"
 #include "utils/typcache.h"
 #include "utils/rangetypes.h"
+#include "utils/varlena.h"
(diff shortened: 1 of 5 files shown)


workingjubilee:
> pgrx::varlena

To clarify, most of that cannot actually be defined via bindgen, because bindgen does not "do" function macros.

@daamien Please find and include the additional headers for varlenas like varatt.h, and explicitly include them even if they are transitively included already.

workingjubilee:
Note that varatt.h is a recent split-off header so this will actually differ from version to version.

daamien:
Hi @workingjubilee 

I don't see any macros in the varlena header file... Am I missing something ?

I can add `varatt.h` for PG16. Should I also add `utils/sortsupport.h` ?

workingjubilee:
Apologies, the comment about macros was only related in my own head, I think.

> Should I also add utils/sortsupport.h?

Sure.

workingjubilee:
Thank you!

PR #1486, opened 2024-01-22
Add bindings for `include/multixact.h`

rebasedming:
Adds header files for `multixact.h`


workingjubilee:
Please explicitly include the transitive inclusions of access/multixact.h (not necessarily the recursive transitive inclusions, just 1 level), for future reference, if you care to use any of the things _those_ expose. I realize some of these are already transitively included, I just need it for documentation reasons. I also realize this is a new standard for these PRs, so my apologies for springing this on you.

For instance, you might want lib/stringinfo.h, and of course, so does pgrx itself.

workingjubilee:
You are allowed to refuse to do this, of course, if you genuinely do not need any of the API those additional headers enable inside include/access/multixact.h, as that will also help me figure some things out soon.

rebasedming:
> use

Ok, happy to do that. Will get on it later today.

workingjubilee:
Thank you!

PR #1487, opened 2024-01-23
Filter `all` to both config && Cargo .tomls

### LINKED ISSUE #1421: unsupported pg version in `~/.pgrx/config.toml` isn't handled
We never actually remove an old version like pg11 from the config.toml, and it's not clear we should. We can tell the user they need to intervene, but we probably should correctly identify when they must. Currently `PG_VERSION=all cargo pgrx test` doesn't work for my local installation.

Sample busted toml:

```toml
[configs]
pg16 = "/home/jubilee/.pgrx/16.0/pgrx-install/bin/pg_config"
pg13 = "/home/jubilee/.pgrx/13.12/pgrx-install/bin/pg_config"
pg12 = "/home/jubilee/.pgrx/12.16/pgrx-install/bin/pg_config"
pg11 = "/home/jubilee/.pgrx/11.21/pgrx-install/bin/pg_config"
pg14 = "/home/jubilee/.pgrx/14.9/pgrx-install/bin/pg_config"
pg15 = "/home/jubilee/.pgrx/15.4/pgrx-install/bin/pg_config"
```

workingjubilee:
Our support for a given Postgres version, for an extension crate, is informed by two things:
- The setting to find `pg_config`s, defaulting to `"${PGRX_HOME}/config.toml"`
- The crate's Cargo.toml for a given pgrx `"pg{major}"` feature.

Interpret `cargo-pgrx {start,stop,test} all` to act only on versions for which both of these are present. We will simply filter silently without note because there are some nuances for when we should (not) emit errors that I am not entirely settled on how to decide, but ultimately this allows e.g. an extension like PL/Rust that never supported the same version minimum as pgrx to still use `cargo pgrx test all`, and causes less confusing errors when `config.toml` includes out-of-date versions unless those are also in Cargo.toml.

Note this doesn't prevent us from emitting errors: instead the error moves to when building the pgrx library. Currently it is first caught by cargo when the extension's Cargo.toml doesn't include the features to for the request from cargo-pgrx, but this is a hidden behavior of cargo-pgrx. The failed build of the library is easier to diagnose.

Closes https://github.com/pgcentralfoundation/pgrx/issues/1421

### DIFF
--- a/cargo-pgrx/README.md
+++ b/cargo-pgrx/README.md
@@ -232,13 +232,22 @@ $ cargo pgrx stop all
     Stopping Postgres v15
 ```
 
-`cargo pgrx` has three commands for managing each Postgres installation: `start`, `stop`, and `status`. Additionally, `cargo pgrx run` (see below) will automatically start its target Postgres instance if not already running.
+`cargo pgrx` has three commands for managing Postgres installations: `start`, `stop`, and `status`.
+Additionally, `cargo pgrx run` (see below) will automatically start its target Postgres instance if
+not already running. Note that when directed to use `"all"` Postgres instances, this is interpreted
+in terms of an extension's `pg{MAJOR}` features in its Cargo.toml, except for `cargo pgrx status`.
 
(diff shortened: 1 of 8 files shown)


eeeebbbbrrrr (approved):
LGTM!

PR #1490, opened 2024-01-24
Replace dependency on `dirs` with `home`

the-kenny:
The `dirs` crate recently [started depending](https://github.com/dirs-dev/dirs-sys-rs/commit/e169da7af901eb621e5d244efe960f4da8ed150d) on the `options-ext` crate
which uses copyleft license (MPL). This (unnecessary) dependency causes
licensing issues for various users by possibly poisoning the dependency
tree of their projects.

This change replaces the `dirs` crate with `home`. The `home` crate is
maintained by the cargo team and offers the same functionality.

As a bonus, this change also results in a slightly smaller dependency
tree.

- https://github.com/artichoke/artichoke/pull/2564
- https://github.com/pyrossh/rust-embed/issues/231
- https://github.com/juhaku/utoipa/issues/834
- https://github.com/harryfei/which-rs/pull/78

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -697,27 +697,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "dirs"
-version = "5.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "44c45a9d03d6676652bcb5e724c7e988de1acad23a711b5217ab9cbecbec2225"
-dependencies = [
- "dirs-sys",
-]
-
-[[package]]
-name = "dirs-sys"
-version = "0.4.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "520f05a5cbd335fae5a99ff7a6ab8627577660ee5cfd6a94a6a929b52ff0321c"
-dependencies = [
- "libc",
- "option-ext",
- "redox_users",
- "windows-sys 0.48.0",
-]
-
-[[package]]
 name = "doc-comment"
 version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


workingjubilee:
It has been 9 months since that change was merged, and it was released quite a while ago.

I do not enjoy the choice of words like "poison" and "tainted", because by depending on dirs, this accusation is currently fully extensible to us. If I refuse to merge this, then perhaps I am now also "poisoning" someone's dependency tree, even though I have already released several versions of cargo-pgrx and pgrx-pg-config and... well, frankly, this function doesn't actually get linked into actual Postgres extensions generated by pgrx, but is only used by the pgrx build system, so I only barely see why anyone _should_ care. It's not like no one uses gcc, a GPL compiler.

Anyways, for these reasons, I would prefer to merge this with a different commit message. I usually simply exercise this power at-will to make a message closer to truth or even just better fit a style, but as this would be exercising a different class of editorial power, I am happy to allow you to rectify it first, if you would prefer.

workingjubilee:
> As a bonus, this change also results in a slightly smaller dependency
tree.

This doesn't really seem to be a meaningful bonus as, as far as I can tell, it just means reduced cross-platform support, and while we don't really support Redox, I don't hate the idea of someone trying to build Postgres with Rust extensions on an operating system written in Rust. It seems rather funny, even.

workingjubilee:
I have resolved the merge conflicts in this PR caused by introducing pervasive use of workspace dependencies. I have toned down the commit message to accurately reflect the actual annoyance of us somehow using the MPL, because it's not exactly like this person is smuggling the AGPL into our source code.

I kind of wish they would have. Then I could have just gone "oh, okay, they're being super-silly" and clicked "merge" instead of feeling like I have to actually understand the case.

workingjubilee:
I have followed up on this in #1502 and #1504.

PR #1494, opened 2024-01-25
Remove atty

workingjubilee:
Requires tweaking owo-color features to reduce exposure via transitive deps.

Addresses a security alert that essentially doesn't apply to pgrx due to being Windows-only, but still annoys me to have on my dashboard.

### DIFF
--- a/cargo-pgrx/src/main.rs
+++ b/cargo-pgrx/src/main.rs
@@ -18,7 +18,7 @@ mod pgrx_pg_sys_stub;
 pub(crate) mod env;
 pub(crate) mod profile;
 
-use atty::Stream;
+use std::io::{self, IsTerminal};
 use clap::Parser;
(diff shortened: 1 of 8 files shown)


cargo-pgrx/src/main.rs
```diff
@@ -61,9 +61,10 @@ impl CommandExecute for CargoSubcommands {
 }
 
 fn main() -> color_eyre::Result<()> {
+    let stderr_is_tty = io::stderr().is_terminal();
     env::initialize();
     color_eyre::config::HookBuilder::default()
-        .theme(if !atty::is(Stream::Stderr) {
+        .theme(if stderr_is_tty {
```
thomcc, line 67: This boolean got flipped.

workingjubilee: oops.

PR #1495, opened 2024-01-25
Use heapless 0.8

workingjubilee:
Note this is an API break for users of heapless because we expose this in our own public API and they need to match the types and such. We should probably constrain this problem more by reexporting the relevant crates, so pgrx's users can simply use the reexported types and traits.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -129,22 +129,13 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


workingjubilee:
I think I am already settled on my picks for the 0.11.3 point release, but for future reference: This should not be backported. It will go into 0.12 just fine.

PR #1499, opened 2024-01-26
cargo-pgrx: reduce trivial dep usages

workingjubilee:
I noticed #1468 would allow us to also remove a dependency or two from cargo-pgrx, so I started work by removing `once_cell` from it and a few other spots in our tree, although we use it so pervasively it's not realistic to remove it entirely. It was actually secretly in our public API, behind a `#[doc(hidden)]`, for some reason.

I then noticed a bunch of other easy removals from cargo-pgrx. In particular:
- `serde_derive` should not be used directly
- flate2, prettyplease, syn, and unescape are not directly used anymore
- the nix crate maintenance has gotten better, but it feels like a big dep for a bool


workingjubilee:
I should note that OnceCell is the wrong abstraction to use in pgrx itself due to its behavior in the face of restarts when combined with other abstractions, see:
- https://github.com/pgcentralfoundation/pgrx/issues/1223

workingjubilee:
Not worth until #1468 lands, as then it drops `once_cell` from cargo-pgrx's first-order dependencies. Alternatively, maybe worth it if I bother pulling OnceCell out from pgrx's impl details: that way it would make us actually build less code for extensions.

workingjubilee:
cc @NotGyro

NotGyro (approved):
Looks good to me, and compiles and passes tests on my machine. The behavior of OnceLock seems to simply be a safer OnceCell, and I do not see anything going on here that could lead to a deadlock, so this looks good!

PR #1501, opened 2024-01-27
Simplify color-eyre initialization

thomcc:
I mentioned that the bool was flipped here: https://github.com/pgcentralfoundation/pgrx/pull/1494#discussion_r1468206496

But it turns out Theme::new() and Theme::default() are the same anyway https://docs.rs/color-eyre/latest/src/color_eyre/config.rs.html#84-86, so the branch can be removed.


workingjubilee (approved):
Nice!

PR #1503, opened 2024-01-29
Remove flaky `serde_serialize_array` todo test

workingjubilee:
This test emits a dependency's source code path in the snapshot, which makes it prone to breaking inaccurately because that is versioned. This complicates contributions unduly, so remove it.


workingjubilee:
Confirmed this fixes https://github.com/pgcentralfoundation/pgrx/pull/1468 locally.

PR #1504, opened 2024-01-30
Add license check

### LINKED ISSUE #1502: Introduce license auditing tooling to CI
The pgrx crates have a lot of transitive dependencies. Our total dependency tree is about 246 entries... that's including pgrx itself, last I checked, so the actual number of dependencies we don't personally control is smaller. But not that much smaller. We should be using some tooling like `cargo vet` or `cargo deny` to check our licensing requirements in our continuous integration workflows, and probably eventually expand that to soundness concerns (though such an expansion would be low priority, as right now we have to reinforce our own glass house). This would make it easier to handle the massive updates to our dependency tree that we like to do for releases, so that our software depends on the latest versions.

Note that because software licenses treat these cases differently, we may have very different licensing requirements for cargo-pgrx and pgrx: one of these gets linked into binaries (pgrx, the library crate), the other is just a build tool (cargo-pgrx). Although pgrx-pg-sys using pgrx-pg-config in its build.rs may make it look like pgrx-pg-config is in pgrx's directly-linked dependency tree, this is not the case, it is in `[build-dependencies]` only. This is relevant because we recently had the very silly "oh no, an upstream maintainer is deviously trying to make software developers ship a download link to source code published on crates.io!" scare due to dirs. Making the license for pgrx itself clearer would be another benefit from
- https://github.com/pgcentralfoundation/pgrx/issues/1451

Any dependents on pgrx may wish to mention what policies they would like to see checked in this issue.

cc @johnrballard re: helping answer this for TCDI.

workingjubilee:
This uses an allow list, mostly-duplicates the deny.toml, and runs a simple shell script instead of using the action, so that the pgrx library can have an independently tracked set of licenses. This is done because the dependencies of that are liable to be actually linked into extensions as dynamic libraries.

In order to let the deny.toml have accurate reporting, the example extension crates are marked as being unpublished. I also tweaked the version on them because I didn't care about the version field in my regex-replace.

Closes #1502.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -64,7 +64,7 @@ allow = [
     "BSL-1.0",
     "ISC",
     "MIT",
-    # "Unicode-DFS-2016",
+    "Unicode-DFS-2016",
     "Unlicense",
     "Zlib",
     "MPL-2.0"
@@ -78,7 +78,7 @@ default = "deny"
 # The confidence threshold for detecting a license from license text.
 # The higher the value, the more closely the license text must be to the
 # canonical license text of a valid SPDX license file.
-confidence-threshold = 0.997 # it isn't six sigma but it'll do
+confidence-threshold = 0.99999 # we got five nines!
 # Allow 1 or more licenses on a per-crate basis.
 exceptions = [
     { allow = ["LicenseRef-Ring"], name = "ring", version = "*" },
--- a/deny.toml
+++ b/deny.toml
@@ -55,7 +55,8 @@ ignore = [
 # The lint level for crates which do not have a detectable license
 unlicensed = "deny"
 # See https://spdx.org/licenses/ for list of possible licenses
-# [possible values: any SPDX 3.11 short identifier (+ optional exception)].
+# This allowlist is a superset of pgrx's allowlist, including the build tools like cargo-pgrx, etc.
+# We are not as concerned about licenses that affect linked code here: most tools aren't shipped.
 allow = [
     "0BSD",
     "Apache-2.0",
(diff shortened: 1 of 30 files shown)


workingjubilee:
cc @johnrballard This prooobably addresses any questions for TCDI around licensing.

cc @the-kenny You expressed concerns about the MPL-2.0 being included in this repository's code, and it looks like, due to our ureq dependency for cargo-pgrx, it will remain in cargo-pgrx for the foreseeable future. The rest of this PR is an attempt to clarify future concerns.

deny.toml
```diff
@@ -0,0 +1,138 @@
...
+[bans]
+multiple-versions = "warn"
+wildcards = "allow"
+# * lowest-version - The path to the lowest versioned duplicate is highlighted
+# * simplest-path - The path to the version with the fewest edges is highlighted
+# * all - Both lowest-version and simplest-path are used
+highlight = "all"
+workspace-default-features = "allow" # for "default" features here
+external-default-features = "allow" # for "default" features in other crates
+allow = [] # "I would prefer not to." — Bartleby the Scrivener
```
thomcc, line 110: ohno

pgrx/deny.toml
thomcc, line 1: I don't love the duplication between this and the other, but I don't realistically see an alternative. We just have to remember if one of these changes, the other might need to.

thomcc (approved):

pgrx/deny.toml
workingjubilee: Well, that's why I bothered with the shell script for CI: it should catch things like that. But yeah, the duplication is annoying.

PR #1512, opened 2024-02-02
Document detoasting reqs for RawArray

workingjubilee:


NotGyro (approved):
This looks good! Thank you for the detailed explanation, including the reason why it only works with de-toasted arrays.

Issue #1511, opened 2024-02-02
Allow self managed postgres instances


rjuju:
Hi,

I see that it's possible to rely on self managed binaries (using either --pgXX or setting $PGXX_PG_CONFIG env var) to avoid recompiling everything, but even in that case pgrx forces new dedicated instances in $PGRX_HOME.  That's quite troublesome as I don't want to have yet another place to look at for configuring postgres, purging old data and all usual tasks that go with any work beyond writing a toy extension.

Would it be possible to similarly use custom data directories (and ports if they're not automatically retrieved) and let me deal with them as I want?  Having additional $PGXX_DATA_DIR and $PGXX_PORT would be ideal.

workingjubilee:
I must say, I personally don't understand the desire people have to deploy their yet-untested extension directly to the same data directory as their usual Postgres installation. I have actually useful data in my Postgres, you know, and Rust doesn't actually prevent anyone from writing an extension that runs DROP DATABASE.

In any case, I am happy to add more configuration to the `"${PGRX_HOME}"/config.toml` which addresses some of these use-cases, but I do not think it is beneficial for us to add any more new environment variables. I in fact intend to remove most of the environment variables we have in the fullness of time, leaving those that are strictly necessary (or _overwhelmingly_ useful).

eeeebbbbrrrr:
I'm inclined to agree here.  I could totally see expanding the per-version variables in `${PGRX_HOME}/config.toml` for each Postgres version, defaulting to what we do now.  This is why that file exists in the first place.

I would **not** want any of this controlled by environment variables.  I think that's generally too risky and would make it super difficult for us to help debug "why isn't this working" problems.

`cargo pgrx install` exists, actually, as a developer tool rather than a production-deploy too.  This is why it defaults to "--debug" mode, actually.  It's intended to let you just use whatever Postgres installation you want.  In addition to the pgrx-managed Postgres versions, I have my own custom-compiled versions installed locally for 9.3, 9.5, 9.6, 10, 11, 12, 13, 14, 15, 16.  And depending on the thing I'm developing, will frequently target one of those (that are supported) with `cargo pgrx install`.

As priorities go, I think we'd happily review and accept a PR that expands `${PGRX_HOME}/config.toml` in the obvious ways, but I don't see the core team doing that work high on our development priorities since `cargo pgrx install` can basically get you there, and do that today.  Plus, I'd rather us not second-guess what such a feature *should* do.

[...]

PR #1513, opened 2024-02-02
Document we compile Postgres with assertions

workingjubilee:
From the conversation with @rjuju in pgcentralfoundation/pgrx#1511. We did not document how we compile Postgres, which results in some confusion when people expect to need to have to go to additional trouble to test against a Postgres with debug symbols and assertions. Let's write that down.

As a matter of spring cleaning, this includes some tangential fixups that crossed my eyes as I rebuilt the docs, and some tangential clarification of the rationales for cargo-pgrx's behavior.

Are there any other advisories regarding using cargo-pgrx (as it currently stands now) that I am missing? Perhaps I could explain the full "override `${PGRX_HOME}` to switch which set of Postgreses you are using" strategy, though I don't do that often enough, so my explanation might be confusing.

### DIFF
--- a/cargo-pgrx/README.md
+++ b/cargo-pgrx/README.md
@@ -110,13 +110,18 @@ $ cargo pgrx init
 `cargo pgrx init` is required to be run once to properly configure the `pgrx` development environment.
 
 As shown by the screenshot above, it downloads the latest releases of supported Postgres versions,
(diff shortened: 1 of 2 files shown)


rjuju:
Thanks for the taking care of that!

PR #1514, opened 2024-02-02
Bounds-checking debug assertions for array access

### LINKED ISSUE #1195: Introduce `debug_assert!`s before ptr deref
Introduce a check which would have caught https://github.com/tcdi/pgrx/pull/1180

Also see: https://gist.github.com/thomcc/b2730e3b1a56c4d8cdebf29814ea2351

NotGyro:
This introduces several `debug_assert!()` calls around any attempt to dereference an element in a Postgres array, in order to catch memory bugs. It was written to resolve pgcentralfoundation/pgrx#1195. 

If RawArray's `end_ptr()` ever pointed at something other than one byte past the end of the array's buffer, this might fail to catch memory bugs. However, it should be pointing to the end of the ArrayType's varlena, which includes the ArrayType header, its null bitmap, and its data, so this should work.

### DIFF
--- a/pgrx/src/datum/array.rs
+++ b/pgrx/src/datum/array.rs
@@ -189,12 +189,14 @@ impl<'mcx, T: UnboxDatum> Array<'mcx, T> {
     #[allow(clippy::option_option)]
     #[inline]
     pub fn get<'arr>(&'arr self, index: usize) -> Option<Option<T::As<'arr>>> {
-        // Technically this should be covered by the null_slice check, 
+        // Technically this should be covered by the null_slice check,
         // but that assumes the null bitmap is well-formed (i.e. equal in
-        // length to the array), which might be worth double-checking in 
-        // debug builds. 
+        // length to the array), which might be worth double-checking in
+        // debug builds.
         #[cfg(debug_assertions)]
-        if index >= self.raw.len() { return None };
+        if index >= self.raw.len() {
+            return None;
+        };
 
         let Some(is_null) = self.null_slice.get(index) else { return None };
         if is_null {
@@ -244,14 +246,12 @@ impl<'mcx, T: UnboxDatum> Array<'mcx, T> {
                 debug_assert!(self.is_within_bounds(ptr));
                 // Prevent a datum that begins inside the array but would end
                 // outside the array from being dereferenced.
-                debug_assert!(
-                    self.is_within_bounds_inclusive(
-                        ptr.wrapping_add(unsafe { self.slide_impl.hop_size(ptr) })
-                    )
-                );
+                debug_assert!(self.is_within_bounds_inclusive(
+                    ptr.wrapping_add(unsafe { self.slide_impl.hop_size(ptr) })
+                ));
 
                 unsafe { self.slide_impl.bring_it_back_now(self, ptr) }
-            },
+            }
         }
     }
 
--- a/pgrx/src/datum/array.rs
+++ b/pgrx/src/datum/array.rs
@@ -189,19 +189,14 @@ impl<'mcx, T: UnboxDatum> Array<'mcx, T> {
     #[allow(clippy::option_option)]
     #[inline]
     pub fn get<'arr>(&'arr self, index: usize) -> Option<Option<T::As<'arr>>> {
-        // Technically this should be covered by the null_slice check,
-        // but that assumes the null bitmap is well-formed (i.e. equal in
-        // length to the array), which might be worth double-checking in
-        // debug builds.
-        #[cfg(debug_assertions)]
-        if index >= self.raw.len() {
-            return None;
-        };
-
         let Some(is_null) = self.null_slice.get(index) else { return None };
         if is_null {
             return Some(None);
         }
+        
+        // This assertion should only fail if null_slice is longer than the
+        // actual array,thanks to the check above.
+        debug_assert!(index < self.raw.len());
 
         // This pointer is what's walked over the entire array's data buffer.
         // If the array has varlena or cstr elements, we can't index into the array.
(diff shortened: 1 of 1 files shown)


workingjubilee:
Thanks! First impression is this looks good, though I have to meditate on some of the possible pointer-inbounds edge cases here and figure out if any of them matter for this check.

pgrx/src/datum/array.rs
```diff
@@ -256,6 +276,21 @@ impl<'mcx, T: UnboxDatum> Array<'mcx, T> {
...
+    #[inline]
+    pub(crate) fn is_within_bounds(&self, ptr: *const u8) -> bool {
+        (self.raw.data_ptr() <= ptr) && (ptr < self.raw.end_ptr())
+    }
+    /// Similar to [`Self::is_within_bounds()`], but also returns true for the
+    /// 1-past-end position.
+    #[inline]
+    pub(crate) fn is_within_bounds_inclusive(&self, ptr: *const u8) -> bool {
+        (self.raw.data_ptr() <= ptr) && (ptr <= self.raw.end_ptr())
+    }
```
workingjubilee: Something I am currently thinking through is whether these are resilient in the face of UB, LLVM optimizations, and arithmetic offset wrapping values around the integer addressing space.

workingjubilee: For `RawArray::end_ptr`, the implementation is:
```rust
    /// "one past the end" pointer for the entire array's bytes
    pub(crate) fn end_ptr(&self) -> *const u8 {
        let ptr = self.ptr.as_ptr().cast::<u8>();
        ptr.wrapping_add(unsafe { crate::varlena::varsize_any(ptr.cast()) })
    }
```

Notably, we should not trust `varsize_any` to be correct, but the `ptr.wrapping_add` returns a pointer that may be invalid, but is not, according to my understanding of LLVM's mechanics, inherently "bad" to compare against (where "bad" is much more nebulous than "UB"). But I am slightly suspicious because pointer comparisons are ill-defined in too many cases...

pgrx/src/datum/array.rs
```diff
@@ -231,7 +240,18 @@ impl<'mcx, T: UnboxDatum> Array<'mcx, T> {
...
-            false => unsafe { self.slide_impl.bring_it_back_now(self, ptr) },
+            false => {
+                // Ensure we are not attempting to dereference a pointer
+                // outside of the array.
+                debug_assert!(self.is_within_bounds(ptr));
+                // Prevent a datum that begins inside the array but would end
+                // outside the array from being dereferenced.
+                debug_assert!(self.is_within_bounds_inclusive(
+                    ptr.wrapping_add(unsafe { self.slide_impl.hop_size(ptr) })
+                ));
```
workingjubilee, line 246: If we used a fn from the `.offset` family instead of `.wrapping_offset` then it's simply UB and any such check is elided, but `hop_size` does not do this inherently...

pgrx/src/datum/array.rs
workingjubilee: Okay!

I think we have to trust the implementation of `varsize_any` anyways, thinking about it. Ideally, we should set up more tests for it. Opened #1539 for that.

After recaching my memory from the original issue and PR, I now remember: What we *really* need is to defend against LLVM deciding pointers with the same address are unequal, as that means we can't rely on comparison semantics being coherent:
- https://github.com/rust-lang/rust/issues/107975
- https://github.com/llvm/llvm-project/issues/45725

This means that we need to cast the pointer addresses to usize, and do the comparisons purely on integral values.

NotGyro:
Opened related issue #1540

workingjubilee (approved):

pgrx/src/datum/array.rs
DoubleHyphen: ### Terse version:
Casting to `usize` won't save you, but I'll bet that you don't need this anyway, for several reasons. In the remote case you do, a `core::hint::black_box()` will serve you much better, and will be much more readable.

-------
### Verbose version:

Forgive me for intruding, but I'm worried you might be misremembering some small details of the bug you're worried about. Chief amongst them is the fact that _casting the pointer addresses to `usize` does nothing to alleviate this problem._ [This example](https://github.com/rust-lang/rust/issues/107975#issuecomment-1430704499) makes this absolutely clear.

That said, I'm not sure how much of a problem this is for your use-case. There are two reasons for this, one incidental and one fundamental.

__The incidental one:__ In every single reproduction of this issue that I've managed to find, _the pointers only miscompare when they're dangling._ This pull-request appears to only compare non-dangling pointers, so it ought to dodge this specific issue.

__The fundamental one:__ The main problem is that LLVM (and GCC, for that matter) think that two pointers can never be equal if they “come from” different allocations. If the pointers you're comparing are both “coming from” the same allocation –as appears to be the case as best as I can tell– then comparison will never misbehave.

__The solution I propose:__ Rust has a specific function whose entire purpose is to tell the compiler “please pessimise this as best as you can”; that function is `black_box`, found in `core::hint` and `std::hint`. I've yet to encounter a reproduction where this didn't immediately fix the issue. (There's also `(a^b).to_ne_bytes().into_iter().map(|x| x.count_ones()).sum() == 0`, but that's both slow and plain unreadable.)

__One thing to note:__ I can't help but notice, due to your title, that this entire discussion happens _for some debug assertions_. Thus far, this issue has only ever been encountered in release mode. If you manage to reproduce it with debug assertions enabled, that would be important information that should be noted in the issue that tracks it.

DoubleHyphen: @workingjubilee …was the above useful, after all?

workingjubilee: @DoubleHyphen 

1. Debug assertions and optimization levels are not values that are intrinsically connected, so I would prefer not to reason on that grounds.
2. `std::hint::black_box` is not "please pessimize this as much as you can", it has a specific and more narrow behavior than that.
3. I have in fact noticed usize-casting interfering, though it has been a while so I don't have the reproduction code for such. Shrug.

PR #1516, opened 2024-02-03
Fix the pgrx links in the aggregate article.

rjuju:
The links were explicitly referencing the then new version 0.3.0 but this version isn't available anymore on docs.rs, so use the "latest" version instead.


workingjubilee:
Thanks! This article predated a small rename due to various reasons, so the crate published for 0.3.0 used a different name! Probably for the best that it points people to the latest version though, yes.

rjuju:
Oh I see.  On docs.rs the version selector doesn't go below 0.7.4 so I thought that older version were just removed.

Thanks for merging!

PR #1517, opened 2024-02-04
chore: remove unneeded namespace in examples

nyurik:
Remove `pgrx::` in `pgrx::pg_module_magic!();` because `use pgrx::prelude::*;` already includes that.

This is part of the `unused_qualifications` build lint. I will submit another PR after running this:

```bash
RUSTFLAGS='-W unused_qualifications' cargo check --workspace --all-targets
```

### DIFF
--- a/cargo-pgrx/src/templates/bgworker_lib_rs
+++ b/cargo-pgrx/src/templates/bgworker_lib_rs
@@ -2,7 +2,7 @@ use pgrx::bgworkers::*;
 use pgrx::prelude::*;
 use std::time::Duration;
 
-pgrx::pg_module_magic!();
+pg_module_magic!();
 
 /*
 In order to use this bgworker with a cargo-pgrx managed database, you'll need to add this line to
--- a/cargo-pgrx/src/templates/lib_rs
+++ b/cargo-pgrx/src/templates/lib_rs
@@ -1,6 +1,6 @@
 use pgrx::prelude::*;
 
-pgrx::pg_module_magic!();
+pg_module_magic!();
 
 #[pg_extern]
 fn hello_{name}() -> &'static str {{
--- a/nix/templates/default/src/lib.rs
+++ b/nix/templates/default/src/lib.rs
@@ -9,7 +9,7 @@
 //LICENSE Use of this source code is governed by the MIT license that can be found in the LICENSE file.
 use pgrx::prelude::*;
 
-pgrx::pg_module_magic!();
+pg_module_magic!();
 
 #[pg_extern]
 fn hello() -> &'static str {
--- a/pgrx-examples/aggregate/src/lib.rs
+++ b/pgrx-examples/aggregate/src/lib.rs
@@ -10,11 +10,11 @@
 use core::ffi::CStr;
 use pgrx::aggregate::*;
 use pgrx::prelude::*;
-use pgrx::{pgrx, PgVarlena, PgVarlenaInOutFuncs, StringInfo};
+use pgrx::{PgVarlena, PgVarlenaInOutFuncs, StringInfo};
 use serde::{Deserialize, Serialize};
 use std::str::FromStr;
 
-pgrx::pg_module_magic!();
+pg_module_magic!();
 
 #[derive(Copy, Clone, PostgresType, Serialize, Deserialize)]
 #[pgvarlena_inoutfuncs]
--- a/pgrx-examples/arrays/src/lib.rs
+++ b/pgrx-examples/arrays/src/lib.rs
@@ -11,7 +11,7 @@ use pgrx::prelude::*;
 use pgrx::Array;
 use serde::*;
 
-pgrx::pg_module_magic!();
+pg_module_magic!();
 
 #[pg_extern]
 fn sq_euclid_pgrx(a: Array<f32>, b: Array<f32>) -> f32 {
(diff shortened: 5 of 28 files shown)


workingjubilee:
You have to enable this lint? Why should we conform to it then?

nyurik:
@workingjubilee thx for reviewing. Technically we do not "have to" conform to any lint :)  This lint has a special history - it is something that should have been in Clippy, but back when it was created Clippy didn't really exist - so it is stuck in a limbo - on one hand, it is not as severe as other compiler warnings to be auto-enabled, on the other, compiler cannot easily remove it because so many projects rely on it (by enabling it explicitly). Clippy doesn't want it because it would duplicate existing functionality... plus the lint is not perfect - I read somewhere that it has some false positives.

That said, I think my change is good for several reasons:
* it makes code more readable by making it shorter (to a degree of course, renaming all vars to a single letter would be taking it too far).  There is a lot more of the code like this
* I noticed this issue because it was part of the generated template code - and it is best to generate the cleanest example possible
* IntelliJ tools automatically highlight these types of issues, making the code appear "unclean" - i.e. instead of focusing on the actual code, devs focus is attracted to code imperfections (from Intellij's perspective of course)

I will be submitting a number of other cleanup PRs mostly to try to make code more readable for newcomers if that's ok with the PGRX community.  Thx!

workingjubilee:
These warnings go silent if the qualifier is made unambiguous, i.e. `::pgrx::pg_module_magic`. I would prefer that change. In general the style used in this workspace's code is, indeed, incapable of getting a job in this market, because they are constantly told they are "overqualified". While I have removed those quals in some cases, I do not actually wish to change that.

workingjubilee:
> IntelliJ tools automatically highlight these types of issues, making the code appear "unclean" - i.e. instead of focusing on the actual code, devs focus is attracted to code imperfections (from Intellij's perspective of course)

Yes. I am familiar with such distraction. I have a low tolerance for unpleasant code. My main concern is that I actually don't want to trammel the diffs in this repo... I am **constantly** performing code archaeology, because these crates are very difficult to understand, and often require tracing through both Rust nuances, their commit history, and even Postgres commit history. This is why I have not "cleaned up" many modules myself.

For that reason, if these are all going to be beautification changes of a similar nature, I would prefer to merge a _larger_ diff as a single PR, so I can add that commit to .git-blame-ignore-revs afterwards. You may add them as individual commits to this PR, as this repo uses a squash-based workflow. Functional changes can be split out to another PR.

> it makes code more readable by making it shorter (to a degree of course, renaming all vars to a single letter would be taking it too far). There is a lot more of the code like this

While I am aware of the virtue of terseness... the `PgMemoryContexts` type will eventually be fully replaced by `MemCx`, which pretty much says it all... the code in this crate is wildly `unsafe` and making it painfully explicit is often preferable.

Please refer to the implementation of `pgrx::list::List` for an example of where such qualifiers are _very_ necessary, and I hope then you will see why this workspace may have an unusual style, especially with regards to using the `pg_sys::` qualifier habitually. I have removed such in the past, and it proved to be a mistake. Please feel free to correct such or otherwise enforce uniformity in this repo with its own style, but you may find it jarringly distinct from the usual Rust style.

workingjubilee:
An example set of PRs that do cleanup work in this repo:
- https://github.com/pgcentralfoundation/pgrx/pull/1386
- https://github.com/pgcentralfoundation/pgrx/pull/1387
- https://github.com/pgcentralfoundation/pgrx/pull/1416

I should probably add the first two commits to .git-blame-ignore-revs?

I hope you see now why I am asking for a Big Diff to merge: there is a nightmarish amount of work to do, so while I do appreciate it your interest, we will have a better time if it is... structured more for bursty effort.

nyurik:
@workingjubilee makes perfect sense, thanks.  Most maintainers tend to want smaller PRs to simplify reviews. Could you check the following PRs I submitted individually, and then I can create one PR that includes all of them at once to simplify merging for you?  #1518 #1522 #1520 #1519

The only exception is probably #1521 because it actually makes somewhat bigger code changes that should be tracked individually.

nyurik:
P.S.   On second thought, #1519 might also have enough logic changes to stay out of the git blame ignore. Perhaps merge it with #1521?

Also, there are many more minor optimizations this code could benefit from.  Even running non-customized `cargo clippy` produces a lot of them, which is a big potential for uncovering minor bugs, simplifying code, and optimizing perf.  I could do some of that optimization - but I don't want to optimize without any hope of merging :)

workingjubilee:
> Most maintainers tend to want smaller PRs to simplify reviews. Could you check the following PRs I submitted individually, and then I can create one PR that includes all of them at once to simplify merging for you?

If any two of those PRs touch the same file, I would prefer they be combined in advance. In the case of multiple changes applied to a file, I need to review them together to make sure they are correct when combined.

workingjubilee:
Right, to clarify, I would prefer changes to touch only one of these sets:

- tests:
  - pgrx-examples
  - pgrx-tests/src/tests
- build:
  - cargo-pgrx
  - pgrx-pg-config
  - pgrx-pg-sys/build
- lib:
  - pgrx
  - pgrx-tests
  - pgrx-pg-sys
- macros
  - pgrx-macros
  - pgrx-sql-entity-graph
  - pgrx-tests/tests/ui

This helps me dial-in the attention I am giving. I should mention this general rule in the contributing advice. Changes to the last two sets are increasingly radioactive as they go "deeper" into this weird-ass compiler.

workingjubilee:
Mooted.

nyurik:

PR #1518, opened 2024-02-04
chore: inline format args

nyurik:
Inlining format args makes code more readable, shorter, and in some cases can gain some performance advantages (when accidentally using unneeded references as params)

I ran this command, and manually reverted some questionable cases:

```bash
cargo clippy --workspace --allow-dirty --fix --benches --tests --bins --all-features -- -A clippy::all -W clippy::uninlined_format_args
```

Note that I specifically did **not** inline format args in cases where some args are inlinable and some args are not (i.e. expressions).   I did notice a few cases like this in the code though, so it is clear the project is OK with such cases.


nyurik:

PR #1519, opened 2024-02-04
chore: perf optimize arg formatting

nyurik:
Format args should almost never be passed as references to the `format!` because that causes some performance (boxing) degradation due to how Rust handles it internally (unable to optimize).

In this PR I also removed a few trailing commas where they seemed to be incorrect.

Lastly, in `pgrx-sql-entity-graph/src/pg_extern/mod.rs` I removed `to_string()` as it seemed redundant (and also perf cost)


nyurik:

@workingjubilee should I delete that line or not?  It is still present in one of these

PR #1520, opened 2024-02-04
chore: inline non-core formatting args

nyurik:
Manually inlining format args to make the code more a bit more readable and shorter -- all these args are calling non-core formatting functions like logging, so extra care must be taken to check - just in case the macros were incorrectly implemented (e.g. if a macro has a special case for a literal string without params -- the way `panic!` was implemented in edition 2018)

### DIFF
--- a/pgrx-examples/bad_ideas/src/lib.rs
+++ b/pgrx-examples/bad_ideas/src/lib.rs
@@ -19,7 +19,7 @@ pgrx::pg_module_magic!();
 #[pg_extern]
 fn panic(s: &str) -> bool {
     catch_unwind(|| {
-        PANIC!("{}", s);
+        PANIC!("{s}");
     })
     .ok();
     true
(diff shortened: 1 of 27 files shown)


pgrx-examples/bad_ideas/src/lib.rs
```diff
@@ -113,7 +113,7 @@ fn random_abort() {
 pub unsafe extern "C" fn _PG_init() {
     #[pg_guard]
     extern "C" fn random_abort_callback(event: pg_sys::XactEvent, _arg: *mut std::os::raw::c_void) {
-        // info!("in global xact callback: event={}", event);
+        // info!("in global xact callback: event={event}");
```
workingjubilee, line 116: ??? why is this line even here?

nyurik: someone was debugging something?  i can delete it if you want

nyurik:

@workingjubilee should I delete that line or not?  It is still present in one of these

workingjubilee:
It can go, yes.

PR #1521, opened 2024-02-04
perf: avoid mem alloc in formatting, minor bug fix

nyurik:
Optimize writing formatted string to Serde, bypassing string allocations when possible.

Note that `pgrx-sql-entity-graph/src/extension_sql/entity.rs` seemed to have a bug for Function printing - it was printing `Function name)` without the opening parenthesis.

Also, in a few spots, I replaced `buffer.write_fmt(format_args!(...))` with `write!(buffer, ...)` -- as that's exactly what `write!` macro is for.

### DIFF
--- a/pgrx-sql-entity-graph/src/extension_sql/entity.rs
+++ b/pgrx-sql-entity-graph/src/extension_sql/entity.rs
@@ -134,13 +134,13 @@ impl Display for SqlDeclaredEntity {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             SqlDeclaredEntity::Type(data) => {
-                f.write_str(&(String::from("Type(") + &data.name + ")"))
+                write!(f, "Type({})", data.name)
             }
             SqlDeclaredEntity::Enum(data) => {
-                f.write_str(&(String::from("Enum(") + &data.name + ")"))
+                write!(f, "Enum({})", data.name)
             }
             SqlDeclaredEntity::Function(data) => {
-                f.write_str(&(String::from("Function ") + &data.name + ")"))
+                write!(f, "Function({})", data.name)
             }
         }
     }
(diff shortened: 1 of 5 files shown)


workingjubilee:
> Note that pgrx-sql-entity-graph/src/extension_sql/entity.rs seemed to have a bug for Function printing - it was printing Function name) without the opening parenthesis.

I don't see how the code could function with that being the case, so it seems possible this has just moved the bug around. I will have to take a closer look.

pgrx-sql-entity-graph/src/extension_sql/entity.rs
```diff
@@ -134,13 +134,13 @@ impl Display for SqlDeclaredEntity {
...
-                f.write_str(&(String::from("Type(") + &data.name + ")"))
+                write!(f, "Type({})", data.name)
             }
             SqlDeclaredEntity::Enum(data) => {
-                f.write_str(&(String::from("Enum(") + &data.name + ")"))
+                write!(f, "Enum({})", data.name)
             }
             SqlDeclaredEntity::Function(data) => {
-                f.write_str(&(String::from("Function ") + &data.name + ")"))
+                write!(f, "Function({})", data.name)
```
workingjubilee, line 143: I can't actually find a case where this actually diffs a schema, and that is Worrying, because that suggests it is Simply Dead Code.

workingjubilee: I did not exhaustively cover every extension in reach but I have reached the "I need to script out the schema generation delta check" part, which means I might as well bash out a v0.1 solution to https://github.com/pgcentralfoundation/pgrx/pull/1521

NotGyro (approved):
This looks good to me. In addition to cutting out unnecessary allocations, this also makes the affected code more readable.

nyurik:
any updates on this?

PR #1414, opened 2023-11-29
Fix `date_bin` fn someways (like a cat?)

workingjubilee:
This neither is used internally nor is it publicly available, due to invalid cfg attributes on it. Fixing the cfg on it breaks compilation, so the code inside it isn't valid either.

PR #1523, opened 2024-02-04
chore: lots of lints and fixes

nyurik:
* it is an anti-pattern for a fn to take a reference, and clone it right away.  If a fn needs to own a value, a value ownership should be passed in, not cloned inside the fn.  This way the caller can decide to clone something, or to pass it in as is
* use `&Path` instead of `&PathBuf` (same as &str instead of &String)
* use `.cmp` for comparison
* some spelling
* lots of `match` statements can be simplified
* use `.is_null()` instead of comparing with the null ptr
* do not impl `ToString` trait, use `Display` instead, and get ToString for free - perf gain too.

### DIFF
--- a/pgrx/src/bgworkers.rs
+++ b/pgrx/src/bgworkers.rs
@@ -467,7 +467,7 @@ impl BackgroundWorkerBuilder {
     /// `startup` allows specifying shared memory initialization startup hook. Ignored
     /// if [`BackgroundWorkerBuilder::load_dynamic`] is used.
     pub fn enable_shmem_access(mut self, startup: Option<unsafe extern "C" fn()>) -> Self {
-        self.bgw_flags = self.bgw_flags | BGWflags::BGWORKER_SHMEM_ACCESS;
+        self.bgw_flags |= BGWflags::BGWORKER_SHMEM_ACCESS;
         self.shared_memory_startup_fn = startup;
         self
     }
@@ -600,7 +600,7 @@ impl BackgroundWorkerBuilder {
 
     /// Once properly configured, call `load_dynamic()` to get the BackgroundWorker registered and started dynamically.
     /// Start up might fail, e.g. if max_worker_processes is exceeded. In that case an Err is returned.
-    pub fn load_dynamic(self: Self) -> Result<DynamicBackgroundWorker, ()> {
+    pub fn load_dynamic(self) -> Result<DynamicBackgroundWorker, ()> {
         let mut bgw: pg_sys::BackgroundWorker = (&self).into();
         let mut handle: *mut pg_sys::BackgroundWorkerHandle = null_mut();
 
--- a/pgrx/src/bgworkers.rs
+++ b/pgrx/src/bgworkers.rs
@@ -467,7 +467,7 @@ impl BackgroundWorkerBuilder {
     /// `startup` allows specifying shared memory initialization startup hook. Ignored
     /// if [`BackgroundWorkerBuilder::load_dynamic`] is used.
     pub fn enable_shmem_access(mut self, startup: Option<unsafe extern "C" fn()>) -> Self {
-        self.bgw_flags |= BGWflags::BGWORKER_SHMEM_ACCESS;
+        self.bgw_flags = self.bgw_flags | BGWflags::BGWORKER_SHMEM_ACCESS;
         self.shared_memory_startup_fn = startup;
         self
     }
--- a/pgrx/src/datum/array.rs
+++ b/pgrx/src/datum/array.rs
@@ -186,7 +186,7 @@ impl<'mcx, T: UnboxDatum> Array<'mcx, T> {
 
     #[allow(clippy::option_option)]
     #[inline]
-    pub fn get(&self, index: usize) -> Option<Option<T::As<'_>>> {
+    pub fn get<'arr>(&'arr self, index: usize) -> Option<Option<T::As<'arr>>> {
         let Some(is_null) = self.null_slice.get(index) else { return None };
         if is_null {
             return Some(None);
@@ -226,7 +226,11 @@ impl<'mcx, T: UnboxDatum> Array<'mcx, T> {
     /// # Safety
     /// This assumes the pointer is to a valid element of that type.
     #[inline]
-    unsafe fn bring_it_back_now(&self, ptr: *const u8, is_null: bool) -> Option<T::As<'_>> {
+    unsafe fn bring_it_back_now<'arr>(
+        &'arr self,
+        ptr: *const u8,
+        is_null: bool,
+    ) -> Option<T::As<'arr>> {
         match is_null {
             true => None,
             false => {
@@ -441,9 +445,9 @@ mod casper {
         ///
         /// This function is unsafe as it cannot guarantee that `ptr` points to the proper bytes
         /// that represent a `T`, or even that it belongs to `array`.  Both of which must be true
-        unsafe fn bring_it_back_now<'arr>(
+        unsafe fn bring_it_back_now<'arr, 'mcx>(
             &self,
-            array: &'arr Array<'_, T>,
+            array: &'arr Array<'mcx, T>,
             ptr: *const u8,
         ) -> Option<T::As<'arr>>;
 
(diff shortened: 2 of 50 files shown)


workingjubilee:
> there was a typo in pgrx/src/datum/datetime_support/ctor.rs - incorrectly declared conditional compilation for fn date_bin. Fixing it does not compile.

This is https://github.com/pgcentralfoundation/pgrx/pull/1414

workingjubilee:
No rustfmt?

Be aware this repo has a rustfmt.toml.

pgrx/src/datum/date.rs
```diff
@@ -151,8 +151,8 @@ impl Date {
     /// This function will panic, aborting the current transaction, if any part is out-of-bounds.
     pub fn new_unchecked(year: isize, month: u8, day: u8) -> Self {
         let year: i32 = year.try_into().expect("invalid year");
-        let month: i32 = month.try_into().expect("invalid month");
-        let day: i32 = day.try_into().expect("invalid day");
+        let month: i32 = month.into();
+        let day: i32 = day.into();
```
workingjubilee: These implementations existing are simply incorrect, do not use them.

pgrx/src/datum/array.rs
```diff
@@ -521,9 +515,9 @@ mod casper {
 
     impl<T: UnboxDatum> ChaChaSlide<T> for PassByFixed {
         #[inline]
-        unsafe fn bring_it_back_now<'arr, 'mcx>(
+        unsafe fn bring_it_back_now<'arr>(
             &self,
-            _array: &'arr Array<'mcx, T>,
+            _array: &'arr Array<'_, T>,
```
workingjubilee: In the various places that lifetimes are specified in this file, they should not be elided, even if Rust permits such a transformation, because the documentation of the real objects these deliberately named lifetime bounds are referencing is very important for understanding this code, and any errors lead to *incredibly severe soundness problems*.

pgrx/src/datum/array.rs
```diff
@@ -913,7 +907,7 @@ where
 
     fn composite_type_oid(&self) -> Option<pg_sys::Oid> {
         // the composite type oid for a vec of composite types is the array type of the base composite type
-        self.get(0)
+        self.first()
```
workingjubilee: The use of `first()` presents a false certainty here: it's not actually relevant that it be the first.

pgrx/src/trigger_support/pg_trigger_when.rs
```diff
@@ -38,13 +40,12 @@ impl TryFrom<TriggerEvent> for PgTriggerWhen {
...
-    fn to_string(&self) -> String {
+impl Display for PgTriggerWhen {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         match self {
-            PgTriggerWhen::Before => "BEFORE",
-            PgTriggerWhen::After => "AFTER",
-            PgTriggerWhen::InsteadOf => "INSTEAD OF",
+            PgTriggerWhen::Before => f.write_str("BEFORE"),
+            PgTriggerWhen::After => f.write_str("AFTER"),
+            PgTriggerWhen::InsteadOf => f.write_str("INSTEAD OF"),
```
workingjubilee: Should use 
```rust
f.write_str(match self {
```

pgrx/src/spi/tuple.rs
```diff
@@ -330,6 +330,7 @@ impl<'conn> SpiHeapTupleData<'conn> {
         unsafe {
             // SAFETY:  we know tupdesc is not null
             let natts = (*tupdesc).natts;
+            // BUG: usize::try_from will always succeed here, and is not needed.
```
workingjubilee: There is no bug. The type can change between versions. Some of this repo's code is simply mistakes, but in other cases, it is deliberately written to be resilient against version-based changes.

pgrx/src/bgworkers.rs
```diff
@@ -466,7 +466,7 @@ impl BackgroundWorkerBuilder {
     /// `startup` allows specifying shared memory initialization startup hook. Ignored
     /// if [`BackgroundWorkerBuilder::load_dynamic`] is used.
     pub fn enable_shmem_access(mut self, startup: Option<unsafe extern "C" fn()>) -> Self {
-        self.bgw_flags = self.bgw_flags | BGWflags::BGWORKER_SHMEM_ACCESS;
+        self.bgw_flags |= BGWflags::BGWORKER_SHMEM_ACCESS;
```
workingjubilee: I simply don't like this style, that's why I haven't committed it.

workingjubilee (commented):
Essentially none of the datetime types should actually implement From, even if they currently do, which is why so many dicey implementations and use of `try_into()` everywhere. I have to create more tests to verify the correctness of their implementations, and any change which removes these is breaking API, and the lifetimes problem has distracted me, which is also why I haven't shipped more of them:
- https://github.com/pgcentralfoundation/pgrx/pull/1258

pgrx/src/datum/array.rs
workingjubilee: ( It's actually doubly important that the lifetimes remain named if the assumptions are incorrect. )

pgrx/src/pg_sys.rs
```diff
@@ -39,7 +39,7 @@ pub unsafe fn rt_fetch(index: Index, range_table: *mut List) -> *mut RangeTblEnt
 #[inline]
 pub unsafe fn planner_rt_fetch(index: Index, root: *mut PlannerInfo) -> *mut RangeTblEntry {
     unsafe {
-        if (*root).simple_rte_array != core::ptr::null_mut() {
+        if !(*root).simple_rte_array.is_null() {
```
workingjubilee: I tend to dislike the `!value.access.access.is_null()` style because I lose track of the `!` by the time I get to the test method call.

pgrx/src/trigger_support/pg_trigger_when.rs
nyurik: good catch.  Would it make sense to use https://crates.io/crates/enum-display ?

pgrx/src/datum/array.rs
nyurik: i'll revert all lifetime changes

pgrx/src/datum/date.rs
nyurik: would it make sense to perhaps implement `into()` as `try_into().unwrap()` ?  In that case it would be consistent, and can be replaced.

pgrx/src/trigger_support/pg_trigger_when.rs
workingjubilee: No. Display implementations should be deliberate.

pgrx/src/spi/tuple.rs
nyurik: @workingjubilee How can `usize::try_from(natts as usize).unwrap_or_default()` be resilient/helpful?  It looks like a double cast, first unconditional with `as usize`, and then `usize::try_from` which will always succeed.  Would it make sense to rewrite it as `usize::try_from(natts).unwrap_or_default()` or even `usize::from(natts)` ?  The compiler would then usefully report any compilation errors.

nyurik:
Thx for the feedback, addressed most of it in 582a0ac. Does it look better now?

pgrx/src/spi/tuple.rs
workingjubilee: answer: ...my eyes are failing me, I suppose? My apologies.

nyurik: all good, how would you like it modified? :)

nyurik:
> No rustfmt?
> 
> Be aware this repo has a rustfmt.toml.

not sure what you mean - `cargo fmt` should pick up that file automatically.  I did run `cargo fmt` on the whole repo.  Also, just did a rebase to fix merge conflicts.

workingjubilee:
> not sure what you mean - cargo fmt should pick up that file automatically.

I dunno, wasn't sure if it was simple omission or tooling or what.

pgrx/src/spi/tuple.rs
workingjubilee: > Would it make sense to rewrite it as `usize::try_from(natts).unwrap_or_default()

Thinking about it, it should be `usize::try_from(natts).unwrap()` probably.

nyurik: thx, fixed

pgrx/src/list/flat_list.rs
```diff
@@ -113,7 +113,7 @@ impl<'cx, T: Enlist> List<'cx, T> {
                 let list_size = 128;
                 unsafe {
                     let list: *mut pg_sys::List = mcx.alloc_bytes(list_size).cast();
-                    assert_ne!(list, ptr::null_mut());
+                    assert!(!list.is_null());
```
workingjubilee:
```suggestion
                    assert!(list.is_non_null());
```

workingjubilee (commented):
Looks good, will doublecheck to see if it still looks good when I wake up.

workingjubilee:
Ah, this had conflicts with #1468, sorry!

nyurik:
rebasing...

nyurik:
seems to be passing ок

workingjubilee (approved):
Yep, still looks good!

PR #1528, opened 2024-02-05
Clean up library with `is_non_null`

workingjubilee:
This adds `PointerExt::is_non_null` and cleans up some of the list code. I am told this is more readable?

### DIFF
--- a/pgrx/src/lib.rs
+++ b/pgrx/src/lib.rs
@@ -88,6 +88,7 @@ pub use once_cell;
 
 /// Not ready for public exposure.
 mod layout;
+mod ptr;
 mod slice;
 mod toast;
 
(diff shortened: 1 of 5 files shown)


nyurik:
Definitely!  I will remove all `!= null` from the #1523

NotGyro (approved):
This does seem clearer. Particularly, `null_mut` could be misunderstood as some mutable null - for example, if a C programmer in a hurry thought it was equivalent to a `void*`. 

`is_non_null()` has none of that ambiguity, and seems much better to me. This change looks good.

nyurik:
I searched for `is_non_null` and `is_not_null` in the rust codebase - seems the `is_not_null` is mentioned more frequently - i think that would be a better "more obvious" choice.

nyurik:
(i was trying to figure out why it wasn't in the core already - apparently there are some discussions about `*T` and `Option<NonNull<*T>>` and similar

workingjubilee:
:bike: :derelict_house: :paintbrush:

Issue #690, opened 2022-09-14
Refactoring ExternArgs -> DDL gen for type safety


workingjubilee:
As noted during discussion in #683, there's currently situations where we are using purely programmatic logic with vectors of enums rather than type-driven logic, when types are fairly good at encoding logical invariants. This results in errors like the one mentioned in https://github.com/tcdi/pgx/pull/640, unfortunately.

I believe we should start with tokenizing the `ExternArgs` more or less read off "as-is", and then use them to build `struct CreateFn` or similar that can only `fmt::Display` into a valid DDL statement. By taking in the ExternArgs "directly", and only verifying they are valid arguments in the first step, we effectively "lex" them, and then parse them in a separate step that will discern if the combined arguments can logically parse into a correct DDL statement. This is good compiler design (and PGX is more and more that: a compiler) and allows distinguishing between individually invalid and combined invalid inputs, providing more precise and useful error messages.

PR #1532, opened 2024-02-05
macros: inline `format_args!` idents

nyurik:
Per https://github.com/pgcentralfoundation/pgrx/pull/1517#issuecomment-1925899219 consolidate all macros-related PRs


workingjubilee (commented):
...This PR is mostly not an improvement to my eyes, unfortunately. The formatting for pgrx-sql-entity-graph needs to be entirely reworked, because I figured out a couple weeks ago that it is incorrect from the root of how it is going about it. Well, more like remembered how https://github.com/pgcentralfoundation/pgrx/issues/690 should work. Accordingly, I would prefer not to accept formatting-impl changes that don't significantly move away from the current implementation.

I can accept the diffs to pgrx-macros.

nyurik:
@workingjubilee I moved it to #1541, and only left pgrx-macros here

PR #1533, opened 2024-02-05
Simplify errors with thiserror

nyurik:
Thiserror is far better for most of these, and reduces accidental bugs / mismatches with descriptions.

### DIFF
--- a/pgrx-pg-config/src/lib.rs
+++ b/pgrx-pg-config/src/lib.rs
@@ -12,13 +12,13 @@ use eyre::{eyre, WrapErr};
 use owo_colors::OwoColorize;
 use serde_derive::{Deserialize, Serialize};
 use std::collections::{BTreeMap, HashMap};
-use std::error::Error;
 use std::ffi::OsString;
 use std::fmt::{self, Debug, Display, Formatter};
 use std::io::ErrorKind;
 use std::path::PathBuf;
 use std::process::{Command, Stdio};
 use std::str::FromStr;
+use thiserror::Error;
 use url::Url;
 
 pub mod cargo;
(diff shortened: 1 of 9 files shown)


pgrx-pg-config/src/lib.rs
```diff
@@ -491,17 +491,14 @@ impl<'a> PgConfigSelector<'a> {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Error)]
 pub enum PgrxHomeError {
+    #[error("You don't seem to have a home directory")]
     NoHomeDirectory,
+    #[error("$PGRX_HOME does not exist")]
```
nyurik, line 499: note that the `PathBuf` here is not printed. Should it?

workingjubilee: Hm, no. We want to be careful about printing paths or otherwise enumerating the environment in a way that could automatically pass-through to IO, because this crate doesn't know what the security context is for this error being formatted.

nyurik: thx, i will add a comment about that

pgrx-pg-config/src/lib.rs
```diff
@@ -495,6 +495,7 @@ impl<'a> PgConfigSelector<'a> {
 pub enum PgrxHomeError {
     #[error("You don't seem to have a home directory")]
     NoHomeDirectory,
+    // Do not print PathBuf due to potential security concerns
```
workingjubilee: perhaps "allow caller to decide whether it is safe to enumerate paths"?

nyurik: Done, my turn:  :bicyclist: :house: :paintbrush:

workingjubilee (approved):

workingjubilee:
Thank you, that does look better, and indeed we might as well use thiserror everywhere if we use thiserror at all.

PR #1537, opened 2024-02-06
docs: improve quotations

nyurik:
Several quotations needed improvement. One quote was not a quote. Another needed a citation. For the remainder, the `text` code block does not format very well for free-flow text. Quotes should use `> ...` instead.

### DIFF
--- a/pgrx/src/pg_catalog/pg_proc.rs
+++ b/pgrx/src/pg_catalog/pg_proc.rs
@@ -218,6 +218,7 @@ impl PgProc {
         self.get_attr(pg_sys::Anum_pg_proc_proconfig)
     }
 
+    /// From <https://www.postgresql.org/docs/current/catalog-pg-proc.html>:
     /// > An array of the modes of the function arguments, encoded as i for IN arguments, o for OUT
     /// > arguments, b for INOUT arguments, v for VARIADIC arguments, t for TABLE arguments. If all
     /// > the arguments are IN arguments, this field will be null. Note that subscripts correspond to
--- a/pgrx/src/pg_catalog/pg_proc.rs
+++ b/pgrx/src/pg_catalog/pg_proc.rs
@@ -218,12 +218,10 @@ impl PgProc {
         self.get_attr(pg_sys::Anum_pg_proc_proconfig)
     }
 
-    /// ```text
-    /// An array of the modes of the function arguments, encoded as i for IN arguments, o for OUT
-    /// arguments, b for INOUT arguments, v for VARIADIC arguments, t for TABLE arguments. If all
-    /// the arguments are IN arguments, this field will be null. Note that subscripts correspond to
-    /// positions of proallargtypes not proargtypes.
-    /// ```
+    /// > An array of the modes of the function arguments, encoded as i for IN arguments, o for OUT
+    /// > arguments, b for INOUT arguments, v for VARIADIC arguments, t for TABLE arguments. If all
+    /// > the arguments are IN arguments, this field will be null. Note that subscripts correspond to
+    /// > positions of proallargtypes not proargtypes.
     ///
     /// In our case, if all the arguments are `IN` arguments, the returned Vec will have the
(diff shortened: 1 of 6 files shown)


pgrx/src/trigger_support/pg_trigger.rs
```diff
@@ -17,13 +17,11 @@ use crate::trigger_support::{
 };
 use std::ffi::c_char;
 
-/**
-The datatype accepted by a trigger
-
-A safe structure providing the an API similar to the constants provided in a PL/pgSQL function.
-
-Usage examples exist in the module level docs.
-*/
```
workingjubilee: hm. there are a lot of block doc comments in this repo.

nyurik: oh, oops, i began hacking on this using v0.11, this was a merge issue, reverted

workingjubilee: Ah, that makes sense.

pgrx/src/spi.rs
```diff
@@ -200,20 +200,18 @@ impl Spi {
...
-    /// ```text
-    ///    Within a transaction, it's fine to execute Spi commands as `read_only = true` until the
-    ///    first mutable statement (DDL or DML).  From that point forward **all** statements
-    ///    must be executed as `read_only = false`.
-    /// ```
+    ///
+    /// > Within a transaction, it's fine to execute Spi commands as `read_only = true` until the
+    /// > first mutable statement (DDL or DML).  From that point forward **all** statements
+    /// > must be executed as `read_only = false`.
+    ///
```
workingjubilee: This shouldn't really be a quote, it's our interpretation.

pgrx/src/pg_catalog/pg_proc.rs
```diff
@@ -218,12 +218,10 @@ impl PgProc {
...
-    /// ```text
-    /// An array of the modes of the function arguments, encoded as i for IN arguments, o for OUT
-    /// arguments, b for INOUT arguments, v for VARIADIC arguments, t for TABLE arguments. If all
-    /// the arguments are IN arguments, this field will be null. Note that subscripts correspond to
-    /// positions of proallargtypes not proargtypes.
-    /// ```
+    /// > An array of the modes of the function arguments, encoded as i for IN arguments, o for OUT
+    /// > arguments, b for INOUT arguments, v for VARIADIC arguments, t for TABLE arguments. If all
+    /// > the arguments are IN arguments, this field will be null. Note that subscripts correspond to
+    /// > positions of proallargtypes not proargtypes.
```
workingjubilee, line 225: Hmm. Reviewing this, it's not clear what this is even quoting or the context.

pgrx/src/spi.rs
nyurik: My only point is that we shouldn't use code blocks for regular texts.  I can make this into a regular block if you want.

workingjubilee: I agree, I just think that this should be changed to be what it should be instead of an alternatively incorrect form. It should be merged into the sentence "pgrx interprets this to mean".

pgrx/src/pg_catalog/pg_proc.rs
workingjubilee: Oh, it's from https://www.postgresql.org/docs/current/catalog-pg-proc.html

Can you add a link to that somewhere in this type's docs?

pgrx/src/spi.rs
nyurik: thx, fixed

pgrx/src/pg_catalog/pg_proc.rs
nyurik: let me know how it should be reworded. Perhaps just use a `suggest` github feature?

nyurik: thx, added a link

workingjubilee (approved):
Thank you!

PR #1544, opened 2024-02-08
Remove unnecessary nightly hack

workingjubilee:
cc @NotGyro

It was done to suppress the `unstable_name_collisions` lint, which was independently fixed in 69e9b5b and 3e4b188. Propagate Thom's fix to remove this foolishness once and for all.


NotGyro (approved):
This looks good to me.

PR #1545, opened 2024-02-09
add new catalog and cache APIs

usamoi:
Provide a safe wrapper of PostgreSQL catalog and cache.


usamoi:
Failed compile-fail tests make me really confused...

Screenshot_20240210_143228

pgrx/src/pg_catalog.rs
```diff
@@ -0,0 +1,866 @@
...
+    }
+}
+
+unsafe impl GetStruct<pgrx::pg_sys::oidvector> for &[Oid] {
+    unsafe fn get_struct(raw: *const pgrx::pg_sys::oidvector) -> Self {
+        unsafe { (*raw).values.as_slice((*raw).dim1 as usize) }
+    }
+}
+
+macro_rules! _macro_1 {
```
workingjubilee: Okay, these macros definitely need names, not ordinals.

pgrx/src/pg_catalog.rs
```diff
@@ -0,0 +1,866 @@
...
+            unsafe impl GetStruct<c_char> for [<$table:camel $column:camel>] {
+                unsafe fn get_struct(raw: *const c_char) -> Self {
+                    unsafe { Self::from_c_char(raw.read()) }
+                }
+            }
+        }
+    };
+}
+
+macro_rules! _macro_0 {
```
workingjubilee: These macros definitely need names, not ordinals.

pgrx/src/pg_catalog.rs
```diff
@@ -0,0 +1,866 @@
...
+                $(#[$m_column])*
+                pub fn $column(&self) -> $type {
+                    self.get_attr::<$type>(pgrx::pg_sys::[<Anum_ $table _ $column>]).unwrap()
+                }
+            }
+        }
+    };
+}
+
+macro_rules! define {
```
workingjubilee: Define what? A catalog table?

pgrx/src/pg_catalog.rs
```diff
@@ -0,0 +1,866 @@
...
+                        let inner = std::ptr::NonNull::new(entry).unwrap();
+                        Some([<$catalog:camel SearchList>] { inner, cache_id })
+                    }
+                }
+            }
+        }
+    };
+}
+
+macro_rules! defines {
```
workingjubilee: Defines what?

workingjubilee (changes requested):
This appears to completely remove the docs from the functions, so I can't accept a strict regression on our documentation like this.

pgrx-tests/tests/todo/composite-types-broken-on-spi.stderr
```diff
@@ -46,26 +46,26 @@ note: required by a bound in `pg_getarg`
...
     |
 125 |         for dog in dogs {
-    |                    ^^^^ the trait `IntoIterator` is not implemented for `Array<'_, pgrx::prelude::PgHeapTuple<'_, pgrx::AllocatedByRust>>`
+    |                    ^^^^ the trait `IntoIterator` is not implemented for `pgrx::Array<'_, pgrx::prelude::PgHeapTuple<'_, pgrx::AllocatedByRust>>`
     |
-    = help: the trait `IntoIterator` is implemented for `Array<'_, pgrx::prelude::PgHeapTuple<'_, pgrx::AllocatedByRust>>`
+    = help: the trait `IntoIterator` is implemented for `pgrx::Array<'_, pgrx::prelude::PgHeapTuple<'_, pgrx::AllocatedByRust>>`
 
-error[E0277]: the trait bound `Array<'_, pgrx::prelude::PgHeapTuple<'_, pgrx::AllocatedByRust>>: IntoIterator` is not satisfied
+error[E0277]: the trait bound `pgrx::Array<'_, pgrx::prelude::PgHeapTuple<'_, pgrx::AllocatedByRust>>: IntoIterator` is not satisfied
```
workingjubilee: Not sure what happened here... odd.

pgrx/src/pg_catalog.rs
usamoi: Renamed to `define_character`.

pgrx/src/pg_catalog.rs
usamoi: Renamed to `define_catalog_column`.

pgrx/src/pg_catalog.rs
usamoi: Renamed to `define_catalog_or_cache`.

pgrx/src/pg_catalog.rs
usamoi: Renamed to `define_catalogs_and_caches`.

usamoi:
Added documents copied from PostgreSQL website.

workingjubilee:
Ah, most of them were copied, huh? Hm. Will have to think a little about that.

pgrx/src/pg_catalog.rs
workingjubilee: Thank you!

pgrx/src/pg_catalog.rs
```diff
@@ -0,0 +1,1212 @@
...
+        paste::paste! {
+            impl [<$table:camel>]<'_> {
+                $(#[$column_meta])*
+                pub fn $column(&self) -> $column_type {
+                    unsafe {
+                        let start = self.inner.t_data.cast::<u8>();
+                        let offset = (*self.inner.t_data).t_hoff as usize;
+                        let p = start.add(offset).cast::<pgrx::pg_sys::[<FormData_ $table>]>();
+                        GetStruct::get_struct(std::ptr::addr_of!((*p).$column))
+                    }
```
workingjubilee: I would prefer all the macros that expand into nontrivial `unsafe {}` blocks, i.e. that are what you might call "generic implementations of `unsafe` code" as a result, come with a comment explaining why they are safe for **all** the things they are used on in these macros.

Postgres, for instance, has multiple structs with a field named `t_hoff`, with a u8 each time. Do those all have the same semantics? If so, that's great to know.

pgrx/src/pg_catalog.rs
```diff
@@ -0,0 +1,1212 @@
...
+        $(#[$column_meta:meta])* column ($table:ident, $column:ident, $column_type:ty, get_struct)
+    } => {
+        paste::paste! {
+            impl [<$table:camel>]<'_> {
+                $(#[$column_meta])*
+                pub fn $column(&self) -> $column_type {
+                    unsafe {
+                        let start = self.inner.t_data.cast::<u8>();
+                        let offset = (*self.inner.t_data).t_hoff as usize;
+                        let p = start.add(offset).cast::<pgrx::pg_sys::[<FormData_ $table>]>();
```
workingjubilee: Idiomatic Rust normally avoids type annotations, but I would prefer these macros enforce their expectations by declaring relevant types like we were back in the C Mines and tending JavaBean Farms, especially before casting the type away and feeding it into a pointer offset.
```suggestion
                        let offset: u8 = (*self.inner.t_data).t_hoff;
                        let p = start.add(offset as usize).cast::<pgrx::pg_sys::[<FormData_ $table>]>();
```

pgrx/src/pg_catalog.rs
```diff
@@ -0,0 +1,1212 @@
...
+                inner: &'a pgrx::pg_sys::HeapTupleData,
+                cache_id: i32,
+            }
+
+            impl<'a> [<$table:camel>]<'a> {
+                #[inline]
+                #[allow(dead_code)]
+                fn get_attr<T: pgrx::datum::FromDatum>(&self, attribute: u32) -> Option<T> {
+                    unsafe {
+                        let mut is_null = true;
```
workingjubilee: ...We can still skip the type annotation on a literal.

pgrx/src/pg_catalog.rs
```diff
@@ -0,0 +1,1212 @@
...
+
+macro_rules! define_catalog {
+    {
+        $(#[$table_meta:meta])* catalog ($table:ident) { $($(#[$column_meta:meta])* ($column:ident, $column_type:ty, $($x:tt)*))* }
+    } => {
+        paste::paste!{
+            #[doc = concat!("Safe wrapper for pg_catalog.", stringify!($table), ".")]
+            $(#[$table_meta])*
+            pub struct [<$table:camel>]<'a> {
+                inner: &'a pgrx::pg_sys::HeapTupleData,
```
workingjubilee: Normally I avoid holding references to variably-sized structs due to [UCG#256](https://github.com/rust-lang/unsafe-code-guidelines/issues/256) still being unresolved. I think it doesn't matter in this case.

pgrx/src/pg_catalog.rs
```diff
@@ -0,0 +1,1212 @@
...
+                /// Returns true if the search result is empty.
+                pub fn is_empty(&self) -> bool {
+                    self.len() == 0
+                }
+                /// Get the `i`-th search result.
+                pub fn get(&self, i: usize) -> Option<[<$table:camel>]> {
+                    unsafe {
+                        let inner = self.inner.as_ref();
+                        let slice = inner.members.as_slice(inner.n_members as usize);
+                        let member = *slice.get(i)?;
```
workingjubilee: This is an example of where the types get sufficiently confusing without any annotations. I understand if there are cases where the types are, due to the macro expansion, sufficiently generic that it is basically implausible to type annotate them at that point, however. Only add them where it is easy and clarifying to do so.

workingjubilee (commented):
I can follow the macro expansion patterns now, but I'm somewhat uneasy about the code and its durability in the future. I would prefer that this came with more ways of enforcing the expansion was correct, and more explanation as to why the `unsafe` code is correct, so that it is easier to evaluate changes of it in terms of soundness. Especially if e.g. Postgres changes the size of a value and invalidates a safety assumption we are tacitly relying on.

pgrx/src/pg_catalog.rs
```diff
@@ -0,0 +1,1212 @@
...
+
+use crate as pgrx;
+use pgrx::datum::Array;
+use pgrx::pg_sys::Oid;
+use pgrx::pg_sys::Oid as Regproc;
+use std::ffi::{c_char, CStr};
+
+unsafe trait GetStruct<T> {
+    unsafe fn get_struct(raw: *const T) -> Self;
+}
```
workingjubilee: Oh, and what's the safety contract of this trait? i.e. what is the promise that `unsafe impl GetStruct` must uphold?

pgrx/src/pg_catalog.rs
```diff
@@ -0,0 +1,1212 @@
...
+    unsafe fn get_struct(raw: *const pgrx::pg_sys::nameData) -> Self {
+        unsafe { CStr::from_ptr(raw.cast::<c_char>()) }
+    }
+}
+
+unsafe impl GetStruct<pgrx::pg_sys::int2vector> for &[i16] {
+    unsafe fn get_struct(raw: *const pgrx::pg_sys::int2vector) -> Self {
+        unsafe { (*raw).values.as_slice((*raw).dim1 as usize) }
+    }
+}
```
workingjubilee: I'm guessing the reason we need this is because sometimes we want to read the data, sometimes we want to get a CStr, and sometimes we need a slice, huh? But what about the other types that terminate in VLAs?

pgrx/src/pg_catalog.rs
usamoi: It's a limit of Rust decl macros: if tokens are matched with `$t:type`, they cannot be matched with any tokens in next macro call.

pgrx/src/pg_catalog.rs
usamoi: Done.

pgrx/src/pg_catalog.rs
usamoi: Removed `unsafe` in line `17` and added safety comments for `fn get_struct`.

pgrx/src/pg_catalog.rs
usamoi: All members in `FromData_*` are fixed-sized. There are only 3 exceptions: `oidvector`, `int2vector` and `bytea`. If there is `oidvector`, `int2vector` or `bytea` (only in `pg_largeobject`) in `FromData_*`, they must be the last struct member. Someone who needs `bytea` or there are more variable-sized types in future PostgreSQL versions could add it by themselves.

`nameData` is fixed-sized 64-byte array in fact. It's guaranteed that it's nul-padded so we can get `CStr` from it and we expose `&CStr` to users since it's more friendly.

usamoi:
The type of a pg_catalog column is unlikely to change. But if it changes,

1. if it's in `FromData_*` (at most cases), type checking fails.
2. if it's not in `FromData_*`, it's unsound.

If we want to check the second condition, we can get oid of the attribute in run-time (or test codes) and compare it with what we record in macro-call statements. Will the check make you more confident?

workingjubilee:
@usamoi Oh, I'm less concerned about runtime types, and more concerned about a later code change by someone else making something subtly wrong that nonetheless the macros allow, and it being hard to tell.

usamoi:
Unsafe code is mostly moved out of `macro_rules` now.

There are still several unsafe blocks like `unsafe { call_xyz() }` in `macro_rules`.

usamoi:
@workingjubilee Could you give me some advice about this patch? If it could not be merged, I feel it's better to move code into a downstream package.

workingjubilee:
Sorry about not getting back to you about this! I was consumed by trying to write down as much of #1661 and #1701 as possible. I am going to try to take another look this week.

pgrx/src/pg_catalog.rs
```diff
@@ -0,0 +1,1284 @@
...
+        /// Right-hand input data type of associated operator
+        (amprocrighttype, Oid, get_struct)
+        /// Support function number
+        (amprocnum, i16, get_struct)
+        /// OID of the function
+        (amproc, Regproc, get_struct)
+    }
+}
+
+define_catalog! {
```
eeeebbbbrrrr, line 611: Just picking one of these at random...

You hand-wrote all of these `define_catalog!()` declarations?  I don't see that as sustainable.  Is there any way we can generate this, via `build.rs`, from the Postgres sources?

Even if that meant we'd need our own equivalent of `genbki.pl` (in Rust, obviously), that will be infinitely better than trying to manually keep these up-to-date.  Postgres does change catalogs from time to time.

eeeebbbbrrrr (changes requested):
I'm marking this "Request Changes", but really I just mean "Lets discuss how we can generate the catalog declarations in `pg_catalog.rs`...

pgrx/src/pg_catalog.rs
usamoi: All declarations are hand-written. If a new column is added in a newer version of PostgreSQL, who uses it could add it again, so I thought it's not a problem.

It looks that this process could be done by parsing `catalog/pg_*.h` and `catalogs.sgml`. I will try it later.

eeeebbbbrrrr: > who uses it could add it again

The problem with that is it's not guaranteed anyone would notice, which might lead to miscompilations or runtime UB or who-knows-what.

In `build.rs`, we ultimately have the struct definitions as provided by bindgen.  Maybe those can be traversed to generate the remaining code?

Ultimately, this would get us 100% coverage of all the catalogs.

workingjubilee: That does seem like it would be best, although it might be a bit toilsome to implement the first time and occasionally to keep up to date.

Issue #1465, opened 2024-01-07
Switch to `bindgen::NonCopyUnionStyle::ManuallyDropUnion`


workingjubilee:
The `__BindgenUnionField` shim generates a bunch of bizarre-looking and hard-to-read code that is, honestly, probably incorrect (why the hell are two union fields always Eq??? I mean, I know it's "because that doesn't affect bit-equality", but...). We should eliminate it by throwing the switch that moves away from it. Unfortunately, this breaks some code internally, so it's not automatically doable. That code probably needs a review anyways.

workingjubilee:
This was addressed in #1547, nice!

PR #1547, opened 2024-02-09
Switch to `ManuallyDropUnion`

NotGyro:
Out of the available methods of wrapping C unions in Bindgen, `__BindgenUnionField` generates code with questionable correctness. This was spotted by @workingjubilee in #1465. This pull request adds a `.default_non_copy_union_style(NonCopyUnionStyle::ManuallyDrop)` line to `pgrx-pg-sys`'s `build.rs`. Since the semantics around these types are different, I have also changed some methods to work with the new type signature (particularly `set_varsize_4b()`).


pgrx/src/datum/into.rs
```diff
@@ -366,26 +366,22 @@ impl<'a> IntoDatum for &'a [u8] {
...
             // SAFETY: `varlena` can properly cast into a `varattrib_4b` and all of what it contains is properly
             // allocated thanks to our call to `palloc` above
-            let varattrib_4b = varlena
-                .cast::<pg_sys::varattrib_4b>()
-                .as_mut()
-                .unwrap_unchecked()
-                .va_4byte
-                .as_mut();
+            let varattrib_4b: *mut _ =
+                &mut varlena.cast::<pg_sys::varattrib_4b>().as_mut().unwrap_unchecked().va_4byte;
```
workingjubilee, line 370: Hm, okay... so we cast the varlena pointer, then call `as_mut()`, then `unwrap_unchecked` it, then deref `.va_4byte`...  since we're just derefing anyways, can't this just be
```rust
&mut (*varlena.cast::<pg_sys::varattrib_4b>()).va_4byte;
```
or something like that?

...also why is it a pointer? If that is important for soundness, this should actually be written with `core::ptr::addr_of_mut!` without `&mut` or `as_mut`. I think when I first reviewed this code, I didn't have my feet under me on the exact semantics of place expressions the way I do now: we want to create a place expression without expressing the strong semantics of `&mut`, and get a pointer to that place, which is what the macro I mentioned allows.

pgrx/src/datum/into.rs
```diff
@@ -366,26 +366,22 @@ impl<'a> IntoDatum for &'a [u8] {
...
 
             // This is the same as Postgres' `#define SET_VARSIZE_4B` (which have over in
             // `pgrx/src/varlena.rs`), however we're asserting that the input string isn't too big
             // for a Postgres varlena, since it's limited to 32bits -- in reality it's about half
             // that length, but this is good enough
-            varattrib_4b.va_header = <usize as TryInto<u32>>::try_into(len)
+            (*varattrib_4b).va_header = <usize as TryInto<u32>>::try_into(len)
                 .expect("Rust string too large for a Postgres varlena datum")
                 << 2u32;
```
workingjubilee: So this is doing some stuff involving reserving some initial bits... I don't quite know why this doesn't offset the usize first and then call `.try_into()`, as that is what will make the `.try_into()` fail if it isn't big enough for fitting in the varlena's sneaky encoding.

pgrx/src/datum/into.rs
```diff
@@ -366,26 +366,22 @@ impl<'a> IntoDatum for &'a [u8] {
...
 
             // SAFETY: src and dest pointers are valid, exactly `self.len()` bytes long,
             // and the `dest` was freshly allocated, thus non-overlapping
             std::ptr::copy_nonoverlapping(
                 self.as_ptr().cast(),
-                varattrib_4b.va_data.as_mut_ptr(),
+                (*varattrib_4b).va_data.as_mut_ptr(),
```
workingjubilee: this must be calling the `__IncompleteArrayField::as_mut_ptr`.

pgrx/src/varlena.rs
```diff
@@ -21,7 +21,10 @@ pub unsafe fn set_varsize_4b(ptr: *mut pg_sys::varlena, len: i32) {
     // 	(((varattrib_4b *) (PTR))->va_4byte.va_header = (((uint32) (len)) << 2))
 
     // SAFETY:  A varlena can be safely cast to a varattrib_4b
-    (*ptr.cast::<pg_sys::varattrib_4b>()).va_4byte.as_mut().va_header = (len as u32) << 2;
+    let header = &mut (*ptr.cast::<pg_sys::varattrib_4b>()).va_4byte.deref_mut().va_header;
```
workingjubilee, line 24: Ah, I see, it needs ManuallyDrop because Rust doesn't require `!Drop` to put something in a union, but instead requires `Copy`, and we can't really impl `Copy` for a dynamically-sized length... got it. Yeah, that makes sense. And then we get this weird special-case of auto-DerefMut being disabled, so we have to call it explicitly.

Okay, my comment about `core::ptr::addr_of_mut!` remains the same here, if you can get that place-expression to work instead, , but otherwise this is fine.

...wait, this is open-coded in `impl<'a> IntoDatum for &'a [u8]`.

...that should be replaced by this, and the `try_into` should be turned into a debug assertion.

workingjubilee (commented):
This should be good after a little bit of cleanup.

## Before:

```rust
#[repr(C)]
pub struct varattrib_4b {
    pub va_4byte: __BindgenUnionField<varattrib_4b__bindgen_ty_1>,
    pub va_compressed: __BindgenUnionField<varattrib_4b__bindgen_ty_2>,
    pub bindgen_union_field: [u32; 2usize],
}
```

## After:

```rust
#[repr(C)]
pub union varattrib_4b {
    pub va_4byte: ::std::mem::ManuallyDrop<varattrib_4b__bindgen_ty_1>,
    pub va_compressed: ::std::mem::ManuallyDrop<varattrib_4b__bindgen_ty_2>,
}
```

pgrx/src/varlena.rs
NotGyro: This should be resolved now - `IntoDatum`'s separate implementation of this behavior has been replaced with `set_varsize_4b()`, and there's no longer an invocation of the zero-sized-type's `.as_mut_ptr()`. Let me know if it still does not look correct.

workingjubilee (approved):
Great!

PR #1557, opened 2024-02-13
Update 2 syn

workingjubilee:
We have not been receiving updates to syn for a while, because we are on syn 1 still. Update to syn 2 and lift it into the workspace, which requires revamping how several `proc_macro_attribute`s are parsed. In the process, sprinkle pgrx-sql-entity-graph with paranoid amounts of `track_caller`, expects, and compound errors on the unhappy path, to try to make the code start producing useful error spans.

Then also drop unused features from some of our dependencies to completely erase syn 1 from the dependency tree. We don't meddle with any compressible sections, because we just look at exported symbols, not debuginfo, so I believe this is inconsequential. It's acceptable to revert that part of this change, however, if it doesn't actually work out.

Removes 4 crates from the dependency tree

The syn changes include:
- `syn::LifetimeDef` is now `LifetimeParam`
- `syn::ImplItem::Method` is now `syn::ImplItem::Fn`
- `syn::ImplItemMethod` is now `syn::ImplItemFn`
- Type ascription && `syn::Expr::Type` no longer exist
- `syn::MetaNameValue` replaces its `lit` field with `value`, as it accepts non-literals
- `syn::GenericArgument::Binding` is now `GenericArgument::AssocType`
- `syn::Attribute` hid its `path` field as `.path()`
- `syn::Attribute` hid its `token` field, so there is only `ToTokens`
- `syn::Attribute` changed how it structures access to its args. We now parse inner args of a `syn::Attribute` for our proc-macro attributes like `#[pg_operator]`, without considering parentheses.

### DIFF
--- a/cargo-pgrx/Cargo.toml
+++ b/cargo-pgrx/Cargo.toml
@@ -29,7 +29,6 @@ pgrx-sql-entity-graph.workspace = true
 
 cargo_toml.workspace = true
 libc.workspace = true
-
 regex.workspace = true
 
 cargo_metadata = "0.17.0"
--- a/cargo-pgrx/Cargo.toml
+++ b/cargo-pgrx/Cargo.toml
@@ -29,6 +29,7 @@ pgrx-sql-entity-graph.workspace = true
 
 cargo_toml.workspace = true
 libc.workspace = true
+
 regex.workspace = true
 
 cargo_metadata = "0.17.0"
@@ -40,9 +41,9 @@ tempfile = "3.8.0"
 toml = "0.8.2" # our config files
 
 # SQL schema generation
-object = "0.32.1"
-proc-macro2 = { version = "1.0.69", features = [ "span-locations" ] }
-quote = "1.0.33"
+object = { version = "0.32.1", default-features = false, features = [ "std" ] }
+proc-macro2.workspace = true
+quote.workspace = true
 
 # emit better diagnostics
 color-eyre = "0.6.2"
--- a/pgrx-macros/src/lib.rs
+++ b/pgrx-macros/src/lib.rs
@@ -133,7 +133,7 @@ pub fn pg_test(attr: TokenStream, item: TokenStream) -> TokenStream {
                 thing.span(),
                 "#[pg_test] can only be applied to top-level functions",
             )
-            .to_compile_error()
+            .into_compile_error()
             .into()
         }
     }
@@ -197,12 +197,7 @@ pub fn pg_cast(attr: TokenStream, item: TokenStream) -> TokenStream {
         Ok(CodeEnrichment(pg_extern.as_cast(cast)).to_token_stream().into())
     }
 
-    wrapped(attr, item).unwrap_or_else(|e| {
-        let msg = e.to_string();
-        TokenStream::from(quote! {
-          compile_error!(#msg);
-        })
-    })
+    wrapped(attr, item).unwrap_or_else(|e: syn::Error| e.into_compile_error().into())
 }
 
 /// Declare a function as `#[pg_operator]` to indicate that it represents a Postgres operator
@@ -288,12 +283,7 @@ pub fn pg_schema(_attr: TokenStream, input: TokenStream) -> TokenStream {
         Ok(pgrx_schema.to_token_stream().into())
     }
 
-    wrapped(input).unwrap_or_else(|e| {
-        let msg = e.to_string();
-        TokenStream::from(quote! {
-          compile_error!(#msg);
-        })
-    })
+    wrapped(input).unwrap_or_else(|e: syn::Error| e.into_compile_error().into())
 }
 
 /**
@@ -428,12 +418,7 @@ pub fn extension_sql(input: TokenStream) -> TokenStream {
         Ok(ext_sql.to_token_stream().into())
     }
 
-    wrapped(input).unwrap_or_else(|e| {
-        let msg = e.to_string();
-        TokenStream::from(quote! {
-          compile_error!(#msg);
-        })
-    })
+    wrapped(input).unwrap_or_else(|e: syn::Error| e.into_compile_error().into())
 }
 
 /**
@@ -470,12 +455,7 @@ pub fn extension_sql_file(input: TokenStream) -> TokenStream {
         Ok(ext_sql.to_token_stream().into())
     }
 
-    wrapped(input).unwrap_or_else(|e| {
-        let msg = e.to_string();
-        TokenStream::from(quote! {
-          compile_error!(#msg);
-        })
-    })
+    wrapped(input).unwrap_or_else(|e: syn::Error| e.into_compile_error().into())
 }
 
 /// Associated macro for `#[pg_extern]` or `#[macro@pg_operator]`.  Used to set the `SEARCH_PATH` option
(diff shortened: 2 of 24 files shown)


workingjubilee:
cc @NotGyro

eeeebbbbrrrr (commented):
This looks largely mechanical, which is good. I’m sure it was a process tho!

pgrx-sql-entity-graph/src/extern_args.rs
```diff
@@ -116,6 +116,8 @@ impl ToTokens for ExternArgs {
     }
 }
 
+// This horror-story should be returning result
+#[track_caller]
 pub fn parse_extern_attributes(attr: TokenStream) -> HashSet<ExternArgs> {
```
NotGyro, line 121: Could improving error-handing in this function be a part of this PR, or is that out of scope? Leaving this todo here is probably fine I'm just checking.

pgrx-sql-entity-graph/src/pg_extern/mod.rs
```diff
@@ -83,12 +83,17 @@ pub struct PgExtern {
...
+    #[track_caller]
     pub fn new(attr: TokenStream2, item: TokenStream2) -> Result<CodeEnrichment<Self>, syn::Error> {
         let mut attrs = Vec::new();
         let mut to_sql_config: Option<ToSqlConfig> = None;
 
         let parser = Punctuated::<Attribute, Token![,]>::parse_terminated;
-        let punctuated_attrs = parser.parse2(attr)?;
+        let punctuated_attrs = parser.parse2(attr).map_err(|e| {
+            let mut e = e.clone();
+            e.combine(syn::Error::new(Span::call_site(), "lol here we go"));
```
NotGyro: Probably shouldn't ship this error, haha. I get it though.

NotGyro (changes requested):
Other than the small error message issue and the comment I think this is good. Reading through Syn's changelog, there shouldn't be any silent breaking changes (i.e. if there's a breaking change that we haven't taken into account it won't let us compile) and this builds and passes tests on my machine so it looks like it should be all set. The most common edit I'm seeing here is the [parse_terminated()](https://docs.rs/syn/latest/syn/parse/struct.ParseBuffer.html#method.parse_terminated) change and that all looks correct. 

As an aside, I'm glad `Binding` got split into `AssocType` and `AssocConst`.

pgrx-sql-entity-graph/src/extern_args.rs
workingjubilee: It's technically out of scope.

pgrx-macros/src/lib.rs
```diff
@@ -1177,12 +1173,7 @@ pub fn pg_aggregate(_attr: TokenStream, item: TokenStream) -> TokenStream {
     }
 
     let parsed_base = parse_macro_input!(item as syn::ItemImpl);
-    wrapped(parsed_base).unwrap_or_else(|e| {
-        let msg = e.to_string();
-        TokenStream::from(quote! {
-          compile_error!(#msg);
-        })
-    })
+    wrapped(parsed_base).unwrap_or_else(|e| e.to_compile_error().into())
```
workingjubilee: This diff is applied inconsistently: I could be hitting all the call sites that originally were
```rust
 wrapped(parsed_base).unwrap_or_else(|e| {
        let msg = e.to_string();
        TokenStream::from(quote! {
          compile_error!(#msg);
        })
    })
```

pgrx-sql-entity-graph/src/pg_extern/mod.rs
workingjubilee: Ah, yeah, I'll give that a real error message. It also shouldn't be using `Span::call_site()`, but rather pulling a span from the TokenStream.

pgrx-sql-entity-graph/src/pg_extern/mod.rs
```diff
@@ -83,12 +83,17 @@ pub struct PgExtern {
...
+            e
+        })?;
```
workingjubilee: Hey Jubilee, you do this a lot with `syn::Result`s in this PR, have you considered factoring out this error-combining pattern into its own function, since dtolnay is probably not going to make combining one `syn::Error` with another more ergonomic than it already is anytime soon?

pgrx-macros/src/lib.rs
```diff
@@ -624,18 +624,14 @@ fn example_return() -> pg_sys::Oid {
...
         Ok(pg_extern_item.to_token_stream().into())
     }
 
-    wrapped(attr, item).unwrap_or_else(|e| {
-        let msg = e.to_string();
-        TokenStream::from(quote! {
-          compile_error!(#msg);
-        })
-    })
+    wrapped(attr, item).unwrap_or_else(|e: syn::Error| e.to_compile_error().into())
```
workingjubilee: I think you were unconsciously programming against syn 1.0 here: this is borrowing, but there's an `into_compile_error()` we could be using here.

workingjubilee (commented):
Eliminating 4 dependencies is nice, but we shouldn't pass up easy points on which to make the nightmarish code here more maintainable.

pgrx-macros/src/lib.rs
workingjubilee: Ah, no, I think I copied it from elsewhere in that file. The `.into_compile_error()` is definitely new as of 2.0 though.

pgrx-sql-entity-graph/src/fmt.rs
```diff
@@ -1,3 +1,33 @@
+macro_rules! lazy_err {
+    ($span_expr:expr, $lit:literal $(, $tokens:tt),*) => {
+        {
+            let spanning = $span_expr;
```
workingjubilee, line 6: This is so that the expr can resolve instead of being captured into the closure. This make the name `lazy_err!` slightly confusing, perhaps, since this part is eagerly-evaluated, but a lot of exprs die instead.

pgrx-sql-entity-graph/src/fmt.rs
```diff
@@ -1,3 +1,33 @@
+macro_rules! lazy_err {
```
workingjubilee, line 3:
```suggestion
/// Evaluate an expression that resolves to any `impl ToTokens`, then produce a closure
/// for lazily combining errors only on the unhappy path of `syn::Result`
macro_rules! lazy_err {
```

pgrx-sql-entity-graph/src/pg_extern/mod.rs
workingjubilee: Done!

pgrx-sql-entity-graph/src/fmt.rs
workingjubilee: Fixed by adding docs.

cargo-pgrx/Cargo.toml
```diff
@@ -29,6 +29,7 @@ pgrx-sql-entity-graph.workspace = true
 
 cargo_toml.workspace = true
 libc.workspace = true
+
```
workingjubilee: was there a reason for this...?
```suggestion
```

NotGyro (approved):
Looks good to me after the changes! 

Per Syn's docs, the replacement of `parse2()` with `parse_args()` should be fine, since: 

> This is similar to pulling out the TokenStream from Meta::List and doing syn::parse2::<T>(meta_list.tokens), except that using parse_args the error message has a more useful span when tokens is empty.

(quote from [here](https://docs.rs/syn/latest/syn/struct.Attribute.html#method.parse_args) )

PR #1560, opened 2024-02-17
Composite type result

### LINKED ISSUE #1470: `#[pg_extern]` functions cannot return `composite_type!` wrapped in `Result`
Based on this [snippet from the srf example](https://github.com/pgcentralfoundation/pgrx/blob/634c32ebeed2e3188f5ecebdfde5dd219df6ef9f/pgrx-examples/srf/src/lib.rs#L35C1-L41C2) _(included below)_, I assumed that a composite type could also be wrapped in a `Result` but I get an esoteric error. After looking at the offending code, `pgrx` tries to unwrap an option that should contain the name of the composite type.

Ideally I would like to use `Result` return types and `?` to keep my code easier to read rather than manually causing a `panic`.

```rust
#[pg_extern]
fn result_table() -> Result<
    Option<::pgrx::iter::TableIterator<'static, (name!(a, Option<i32>), name!(b, Option<i32>))>>,
    Box<dyn std::error::Error + Send + Sync + 'static>,
> {
    Ok(Some(TableIterator::new(vec![(Some(1), Some(2))])))
}
```

Below is a minimal example with the error log included too.

```rust

extension_sql!(
    r#"
    CREATE TYPE pool_stats_entry AS (
        max_open_connections int,
        open_connections int,
        in_use int,
        idle int,
        wait_count bigint,
        wait_duration bigint,
        max_idle_closed bigint,
        max_idle_time_closed bigint,
        max_lifetime_closed bigint
    );
    "#,
    name = "create_external_objects",
    bootstrap
);

const POOL_STATS_ENTRY: &str = "pool_stats_entry";

// This doesn't compile???
#[pg_extern(
    requires = [ "create_external_objects" ]
)]
pub fn pool_stats() -> Result<
    composite_type!('static, POOL_STATS_ENTRY),
    Box<dyn std::error::Error + Send + Sync + 'static>,
> {
    /*
    
    The application panicked (crashed).
    Message:  called `Option::unwrap()` on a `None` value
    Location: /Users/collinkokotas/.cargo/registry/src/index.crates.io-6f17d22bba15001f/pgrx-sql-entity-graph-0.11.2/src/pg_extern/entity/mod.rs:261
    
      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ SPANTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
    
       0: cargo_pgrx::command::schema::generate_schema with pg_version=16.1 profile=Dev test=false path=/Users/collinkokotas/.pgrx/16.1/pgrx-install/share/postgresql/extension/pgmayan--0.0.0.sql features=["pg16"]
          at /Users/collinkokotas/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-pgrx-0.11.2/src/command/schema.rs:176
       1: cargo_pgrx::command::install::install_extension with pg_version=16.1 profile=Dev test=false features=["pg16"]
          at /Users/collinkokotas/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-pgrx-0.11.2/src/command/install.rs:114
       2: cargo_pgrx::command::run::run with pg_version=16.1 profile=Dev
          at /Users/collinkokotas/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-pgrx-0.11.2/src/command/run.rs:94
       3: cargo_pgrx::command::run::execute
          at /Users/collinkokotas/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-pgrx-0.11.2/src/command/run.rs:55
    
    Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
    Run with RUST_BACKTRACE=full to include source snippets.
    
    */
    todo!()
}


// But this does... 
#[pg_extern(
    requires = [ "create_external_objects" ]
)]
pub fn pool_stats_2() -> composite_type!('static, POOL_STATS_ENTRY) {
    todo!()
}

```

NotGyro:
This set of changes allows `#[pg-extern]` functions to return `Result<composite_type!(..), E>`. To get it working, I've added another type-specific case to match statements in `UsedType::new()`. That case then calls `fn resolve_result_inner()`, which I wrote for this purpose.

This is intended to resolve #1470.

### Future work 

* Although I've tested this behavior by hand and confirmed it's working correctly, I haven't yet written a test case for this code path. If you'd prefer me to put that in *this* PR, request changes and I'll take care of it. Otherwise, I'll open an issue for it shortly.
* #1561 : While working on this implementation, I ran into a naming ambiguity. Specifically, `pgrx::pg_sys::panics::ErrorReportable::report()` looks an awful lot like `std::process::Termination::report()`, and they are both traits that are implemented on `Result<T, E>` so if  you import both traits you'll have a name collision. I spoke to @workingjubilee about this and she agreed it should be renamed to `unwrap_or_report()` as part of the breaking changes in `0.12`. 

### A note on something esoteric
A note on something that was difficult to dig up answers to: Some concerns surrounding correctness jumped out at me while I was looking at macro expansions for this pull request. In the generated `UsedTypeEntity`, `ty_source` and `full_path` were definitely correct, but for a `Result<T, E>` the ty_id is `core::any::TypeId::of<T>`, whereas for `Option<T>` the macro expands to `core::any::TypeId::of<Option<T>>`.  I asked around, and after some thinking on the part of our senior engineers it looks like this is actually correct. This type information is used to communicate to Postgres, and while `Option<T>` is used as the Rust equivalent to translate to and from nullable Postgres types, so the Option behavior is special, whereas Postgres doesn't have a concept of a Result type and so translating `Result<T>` will either give Postgres an instance of `T` or an exit code and an error message. 

I just thought I'd clarify that to make sure my reasoning on the correctness of this code is written down.

### DIFF
--- a/pgrx-sql-entity-graph/src/used_type.rs
+++ b/pgrx-sql-entity-graph/src/used_type.rs
@@ -143,7 +143,7 @@ impl UsedType {
             }
             original => (original, None),
         };
-        
+
         // In this  step, we go look at the resolved type and determine if it is a variadic, optional, result, etc.
         let (resolved_ty, variadic, optional, result) = match resolved_ty {
             syn::Type::Path(type_path) => {
@@ -626,35 +626,31 @@ fn resolve_result_inner(
     // TODO - more robust / idiomatic way of handling this?
     let mut without_type_args = original.path.clone();
     without_type_args.segments.last_mut().unwrap().arguments = syn::PathArguments::None;
- 
-    let (ok_ty, err_ty) = { 
-        if let syn::PathArguments::AngleBracketed(path_arg) = &last.arguments { 
+
+    let (ok_ty, err_ty) = {
+        if let syn::PathArguments::AngleBracketed(path_arg) = &last.arguments {
             if let Some(first_ty) = path_arg.args.first() {
                 // Since `pub type Result<T> = std::error::Result<T, OurError>
-                // is a common pattern, 
+                // is a common pattern,
                 // we should support single-argument Result<T> style.
                 if path_arg.args.len() == 1 {
                     (first_ty.clone(), None)
-                }
-                else if path_arg.args.len() == 2 { 
+                } else if path_arg.args.len() == 2 {
                     (first_ty.clone(), Some(path_arg.args[1].clone()))
-                }
-                else { 
+                } else {
                     return Err(syn::Error::new(
                         last.arguments.span(),
                         "Result<..> with more than two type arguments not supported.",
                     ));
                 }
-            }
-            else { 
+            } else {
                 // Return early, Result<> with no type args.
                 return Err(syn::Error::new(
                     last.arguments.span(),
                     "Cannot return a Result without type generic arguments.",
                 ));
             }
-        }
-        else {
+        } else {
             // Return early, invalid signature for Result<T,E>
             return Err(syn::Error::new(
                 last.arguments.span(),
@@ -666,18 +662,19 @@ fn resolve_result_inner(
     // Inner / nested function for getting a type signature for a Result from
     // the tuple of (ok_type, Option<error_type>)
     fn type_for_args(
-            no_args_path: syn::Path,
-            first_ty: syn::Type,
-            err_ty: Option<GenericArgument>) -> syn::Type {
-        match err_ty { 
-            Some(e) => { 
+        no_args_path: syn::Path,
+        first_ty: syn::Type,
+        err_ty: Option<GenericArgument>,
+    ) -> syn::Type {
+        match err_ty {
+            Some(e) => {
                 syn::parse_quote! {
                     #no_args_path<#first_ty, #e>
                 }
             }
             None => {
                 // Since `pub type Result<T> = std::error::Result<T, OurError>
-                // is a common pattern, 
+                // is a common pattern,
                 // we should support single-argument Result<T> style.
                 syn::parse_quote! {
                     #no_args_path<#first_ty>
--- a/pgrx-sql-entity-graph/src/used_type.rs
+++ b/pgrx-sql-entity-graph/src/used_type.rs
@@ -698,8 +698,10 @@ fn resolve_result_inner(
                             let ty = type_for_args(without_type_args, comp_ty, err_ty);
                             Ok((ty, sql))
                         },
-                        // Option<default!(composite_type!(..))> isn't valid. If the user wanted the default to be `NULL` they just don't need a default.
-                        "default" => Err(syn::Error::new(mac.span(), "`Result<default!(T, \"my_default\"), E>` not supported, choose `Result<T, E>` for a default of `NULL`, or `default!(T, default)` for a non-NULL default")),
+                        // Result<default!(composite_type!(..)), E> 
+                        "default" => {
+                            Err(syn::Error::new(mac.span(), "`Result<default!(T, default), E>` not supported, choose `default!(Result<T, E>, ident)` instead"))
+                        },
                         _ => Ok((syn::Type::Path(original), None)),
                     }
                 }
@@ -737,8 +739,7 @@ fn resolve_result_inner(
                             let wrapped_ty = type_for_args(without_type_args, inner_ty, err_ty);
                             Ok((wrapped_ty, expr))
                         }
-                        // Result<..>
-                        // TODO - evaluate, should this error rather than proceeding?
+                        // Result<T> where T is plain-old-data and not a (supported) container type.
                         _ => Ok((syn::Type::Path(original), None)),
                     }
                 }
--- a/pgrx-sql-entity-graph/src/used_type.rs
+++ b/pgrx-sql-entity-graph/src/used_type.rs
@@ -627,27 +627,22 @@ fn resolve_result_inner(
     without_type_args.segments.last_mut().unwrap().arguments = syn::PathArguments::None;
 
     let (ok_ty, err_ty) = {
-        if let syn::PathArguments::AngleBracketed(path_arg) = &last.arguments {
-            if let Some(first_ty) = path_arg.args.first() {
+        if let syn::PathArguments::AngleBracketed(path_arg) = last.arguments.clone() {
+            let mut iter = path_arg.args.into_iter();
+            match (iter.next(), iter.next()) {
+                (None, _) => {
+                    // Return early, Result<> with no type args.
+                    return Err(syn::Error::new(
+                        last.arguments.span(),
+                        "Cannot return a Result without type generic arguments.",
+                    ));
+                }
                 // Since `pub type Result<T> = std::error::Result<T, OurError>
                 // is a common pattern,
                 // we should support single-argument Result<T> style.
-                if path_arg.args.len() == 1 {
-                    (first_ty.clone(), None)
-                } else if path_arg.args.len() == 2 {
-                    (first_ty.clone(), Some(path_arg.args[1].clone()))
-                } else {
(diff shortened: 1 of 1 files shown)


pgrx-sql-entity-graph/src/used_type.rs
```diff
@@ -608,6 +613,146 @@ fn resolve_option_inner(
...
+ 
+    let (ok_ty, err_ty) = { 
+        if let syn::PathArguments::AngleBracketed(path_arg) = &last.arguments { 
+            if let Some(first_ty) = path_arg.args.first() {
+                // Since `pub type Result<T> = std::error::Result<T, OurError>
+                // is a common pattern, 
+                // we should support single-argument Result<T> style.
+                if path_arg.args.len() == 1 {
+                    (first_ty.clone(), None)
+                }
```
workingjubilee: Does this break if some sicko uses `type Result<E> = Result<SomeType, E>;`?

It's fine if it does, I'm just curious.

pgrx-sql-entity-graph/src/used_type.rs
```diff
@@ -608,6 +613,146 @@ fn resolve_option_inner(
...
+    original: syn::TypePath,
+) -> syn::Result<(syn::Type, Option<CompositeTypeMacro>)> {
+    let segments = &original.path;
+    let last = segments
+        .segments
+        .last()
+        .ok_or(syn::Error::new(original.span(), "Could not read last segment of path"))?;
+
+    // Get the path of our Result type, to handle crate::module::Result pattern
+    // TODO - more robust / idiomatic way of handling this?
```
workingjubilee: Robust against?

pgrx-sql-entity-graph/src/used_type.rs
```diff
@@ -608,6 +613,146 @@ fn resolve_option_inner(
...
+                    ));
+                }
+            }
+            else { 
+                // Return early, Result<> with no type args.
+                return Err(syn::Error::new(
+                    last.arguments.span(),
+                    "Cannot return a Result without type generic arguments.",
+                ));
+            }
```
workingjubilee: Can this just be
```rust
let mut iter = args.iter();
match (iter.next(), iter.next(), iter.next()) {
```

So that we can have the compiler check that we're exhaustively handling all cases we want to handle, and we don't have the problem of trying to track whether the indexing we do later is correct? Seeing `.first()` and then `first_ty` and then `[1]` is lexically confusing.

pgrx-sql-entity-graph/src/used_type.rs
```diff
@@ -608,6 +613,143 @@ fn resolve_option_inner(
...
+                        "composite_type" => {
+                            let composite_mac = handle_composite_type_macro(mac)?;
+                            let comp_ty = composite_mac.expand_with_lifetime();
+                            let sql = Some(composite_mac);
+
+                            let ty = type_for_args(without_type_args, comp_ty, err_ty);
+                            Ok((ty, sql))
+                        },
+                        // Option<default!(composite_type!(..))> isn't valid. If the user wanted the default to be `NULL` they just don't need a default.
+                        "default" => Err(syn::Error::new(mac.span(), "`Result<default!(T, \"my_default\"), E>` not supported, choose `Result<T, E>` for a default of `NULL`, or `default!(T, default)` for a non-NULL default")),
```
workingjubilee: For the SQL NULL default case, they want `Result<Option<T>, E>`, right?

pgrx-sql-entity-graph/src/used_type.rs
```diff
@@ -608,6 +613,143 @@ fn resolve_option_inner(
...
+                        // Result<Array<composite_type!(..)>>
+                        // Result<Array<Option<composite_type!(..)>>>
+                        "Array" => {
+                            let (inner_ty, expr) = resolve_array_inner(arg_type_path)?;
+                            let wrapped_ty = type_for_args(without_type_args, inner_ty, err_ty);
+                            Ok((wrapped_ty, expr))
+                        }
+                        // Result<..>
+                        // TODO - evaluate, should this error rather than proceeding?
+                        _ => Ok((syn::Type::Path(original), None)),
```
workingjubilee: Hmm, what's the `..` a stand-in for? This is... probably fine? If we're in doubt, however, we should be returning an error.

pgrx-sql-entity-graph/src/used_type.rs
workingjubilee: Sadly, `syn` doesn't expose a slice type here, so we can't just
```rust
match &slice_of_generic_args {
    [a, b, c, ..] => return Err(..),
    [a, b] => { /* code */ },
    [a] => { /* code */ },
    [] => return Err(..),
}
```

pgrx-sql-entity-graph/src/used_type.rs
NotGyro: It most likely would. I've never seen that pattern in the wild but it'd be nice if that could be accounted for - unfortunately I'm not sure if there's a good way to programmatically extract what type is used for the `Ok(T)` variant vs what type is used for the `Err(E)` variant at the level of a procedural macro.

workingjubilee:
Also, can you rebase this on `develop`? It has a lot of merge commits that will go away, I think, when I push `squash`, but it weirds me out.

pgrx-sql-entity-graph/src/used_type.rs
NotGyro: I wasn't fully understanding the design of UsedType's codegen at this time, and I was making some incorrect guesses about how this would fit together.  I can remove this comment.

pgrx-sql-entity-graph/src/used_type.rs
workingjubilee: I don't think there's a realistic way since we aren't guaranteed to see the `type Whatever<?> = Result<T, E>;` decl.

workingjubilee: We would have to demand it is annotated by a macro.

pgrx-sql-entity-graph/src/used_type.rs
workingjubilee: I prefer more actionable TODOs, yes.

NotGyro: removed in b2555d0

pgrx-sql-entity-graph/src/used_type.rs
NotGyro: After digging deeper, this should be the case selected for any `Result<SomePlainOldData>` or `Result<SomePlainOldData, E>`, i.e. the `T` in `Result<T,E>` is not a container type.

workingjubilee: Yeah, this is fine then.

pgrx-sql-entity-graph/src/used_type.rs
NotGyro: Honestly this is just something I forgot to change. I have fixed it now.

NotGyro:
> Also, can you rebase this on `develop`? It has a lot of merge commits that will go away, I think, when I push `squash`, but it weirds me out.

I've rebased it now, the diff should be cleaner.

pgrx-sql-entity-graph/src/used_type.rs
NotGyro: This should be cleaner now, let me know if the new implementation fits.

workingjubilee (approved):
Thank you!

PR #1562, opened 2024-02-17
add bindings of subscript function

usamoi:
Add bindings of supporting subscript access for user-defined types, eg. `udf_type_value[1]`. It's supported since PostgreSQL 14.

Ref: https://www.postgresql.org/docs/current/sql-createtype.html

> The optional `subscript_function` allows the data type to be subscripted in SQL commands. Specifying this function does not cause the type to be considered a “true” array type; for example, it will not be a candidate for the result type of `ARRAY[]` constructs. But if subscripting a value of the type is a natural notation for extracting data from it, then a `subscript_function` can be written to define what that means. The subscript function must be declared to take a single argument of type `internal`, and return an `internal` result, which is a pointer to a struct of methods (functions) that implement subscripting. The detailed API for subscript functions appears in src/include/nodes/subscripting.h. It may also be useful to read the array implementation in src/backend/utils/adt/arraysubs.c, or the simpler code in contrib/hstore/hstore_subs.c. Additional information appears in [Array Types] below.

[Array Types]: https://www.postgresql.org/docs/current/sql-createtype.html#SQL-CREATETYPE-ARRAY


pgrx-pg-sys/include/pg12.h
```diff
@@ -86,6 +87,8 @@
 #include "nodes/makefuncs.h"
 #include "nodes/nodeFuncs.h"
```
workingjubilee, line 88: To avoid relying on transitive inclusion, for my purposes:
```suggestion
#include "nodes/miscnodes.h"
#include "nodes/nodeFuncs.h"
```

pgrx-pg-sys/include/pg13.h
```diff
@@ -86,6 +87,8 @@
 #include "nodes/makefuncs.h"
 #include "nodes/nodeFuncs.h"
```
workingjubilee, line 88:
```suggestion
#include "nodes/miscnodes.h"
#include "nodes/nodeFuncs.h"
```

pgrx-pg-sys/include/pg14.h
```diff
@@ -86,8 +87,11 @@
 #include "nodes/makefuncs.h"
 #include "nodes/nodeFuncs.h"
```
workingjubilee, line 88:
```suggestion
#include "nodes/miscnodes.h"
#include "nodes/nodeFuncs.h"
```

pgrx-pg-sys/include/pg15.h
```diff
@@ -87,8 +88,11 @@
 #include "nodes/makefuncs.h"
 #include "nodes/nodeFuncs.h"
```
workingjubilee, line 89:
```suggestion
#include "nodes/miscnodes.h"
#include "nodes/nodeFuncs.h"
```

pgrx-pg-sys/include/pg16.h
```diff
@@ -88,8 +89,11 @@
 #include "nodes/makefuncs.h"
 #include "nodes/nodeFuncs.h"
```
workingjubilee, line 91:
```suggestion
#include "nodes/miscnodes.h"
#include "nodes/nodeFuncs.h"
```

workingjubilee (commented):
This seems fine, but I would prefer to lift to the top anything that we would otherwise need, from transitive inclusions, to make these useful, and `ErrorSaveContext` shows up in a _lot_ of places.

usamoi:
Added but `nodes/miscnodes.h` and `ErrorSaveContext` are only present in PostgreSQL 16.

workingjubilee:
Oh, whoops.

workingjubilee:
Looks good, thanks!

PR #1563, opened 2024-02-19
Try to reenable workflows using ubuntu-latest

workingjubilee:
Also reduces Will It Blend frequency again to once per week. This will make bisecting the subtly breaking changes harder, but we only benefit from running it about 2 or 3 times per week, anyways, as we often have a day I don't merge anything.


workingjubilee:
I should note that Will It Blend is not really the sole culprit here, since it's more costly than other CI jobs but guaranteed to only run once per day. It's only x3 the on-PR CI.

PR #1564, opened 2024-02-19
Give CI a makeover

workingjubilee:
Various fixups to our CI infra
- fix "Will It Blend?" to use the free CI
- enable the beta aarch64-darwin jobs
- split out lints that don't need `cargo pgrx init` into a lintck job
- make most jobs depend on lintck passing
- use `{checkout,cache}@v4` to avoid deprecation
- update the nix CI template

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -116,9 +116,9 @@ jobs:
     - name: Clippy -Dwarnings sql-entity-graph
       run: cargo clippy -p pgrx-sql-entity-graph -- -Dwarnings
 
-    - name: Check doc-links
+    - name: Check doc-links in pgrx-macros
       run: |
-        cargo rustdoc -p pgrx --features pg$PG_VER -- \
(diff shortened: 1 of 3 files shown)


workingjubilee:
cc @NotGyro Does this look okay?

workingjubilee:
Maybe I should just keep the deny-sql-entity-graph clippy job, which would allow skipping the build.rs of pgrx-pg-sys, which would allow skipping building cargo pgrx?

NotGyro (approved):
Looks good! Nice catch on the `set-env` deprecation

workingjubilee:
Yeah, I'm removing everything that requires `cargo pgrx init` here.

BradyBonnette (approved):
Looks like it works to me, and I don't see anything glaringly wrong with it. At some point we can talk about de-duping some of the stuff that's in here, if possible. But that's a task for another day :)

workingjubilee:
Deduplication is a good point! I deleted the setup in the lintck job that was copied from the other ones, at least, since it was unnecessary.

PR #1567, opened 2024-02-19
Cleanup List impl

workingjubilee:
I noticed there was some garbage floating around the List impl, like unnecessary `unsafe` code, so I cleaned that up. Also, adds some warnings because people sometimes do not think clearly about pointers.


NotGyro (approved):
This looks good to me

Issue #1561, opened 2024-02-17
Rename `ErrorReportable::report()` to avoid collision with `std`'s `Termination::report()`


NotGyro:
While working on the implementation for #1560, I ran into a naming ambiguity. Specifically, `pgrx::pg_sys::panics::ErrorReportable::report()` looks an awful lot like `std::process::Termination::report()`, and they are both traits that are implemented on `Result<T, E>` so if you import both traits you'll have a name collision. I spoke to @workingjubilee about this and she agreed it should be renamed to `unwrap_or_report()` as part of the breaking changes in `0.12`.

NotGyro:
Looks like it's all set!

PR #1568, opened 2024-02-20
Rename `ErrorReportable::report()` to `unwrap_or_report()`

NotGyro:
As noted in #1561, `pgrx::pg_sys::panics::ErrorReportable::report()` is named identically to `std::process::Termination::report()`, and they are both traits that are implemented on `Result<T, E>` so if you import both traits you'll have a name collision. Aside from the name collision, it's very easy to see `.report()` called on a `Result` type and make the incorrect assumption this is the `Termination` trait when one is unfamiliar with PGRX's codebase.

To avoid ambiguity, this set of changes renames that method to `unwrap_or_report()`, and is intended to be included in the set of breaking changes for 0.12.

### DIFF
--- a/pgrx-pg-sys/src/submodules/panic.rs
+++ b/pgrx-pg-sys/src/submodules/panic.rs
@@ -28,7 +28,7 @@ pub trait ErrorReportable {
     type Inner;
 
     /// Raise a Postgres ERROR if appropriate, otherwise return a value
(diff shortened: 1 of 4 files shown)


workingjubilee (approved):
This looks great! (assuming CI also approves)

Just to remind though, we have the **best** versioning scheme: https://0ver.org/

PR #1569, opened 2024-02-20
Add "installation" section to docs

### LINKED ISSUE #1538: Document necessity of `shared_preload_libraries` for singletons
> this sounds about right, you can't guarantee that if an extension is loaded that it won't appear in multiple procs because that's just how Postgres works[^1]. you can only guarantee that if you are loaded early enough that you have a single instance of _something immutable_ visible from all procs.

This is also required for shmem, and thus for mutable singletons.

[^1]: as in, "it forks into multiple procs"

workingjubilee:
This addresses a few questions that keep coming up.

Resolves https://github.com/pgcentralfoundation/pgrx/issues/1538

Also quite a bit of rearrangement was done while I was thinking about this, because our documentation is a bit hard to navigate even if you build it, as it relies on the summary bar to the left instead of being navigable via in-doc hyperlinks. This makes it easy to forget where you are supposed to be going if you are referencing the Markdown itself.

I focused on trying to make it so the SUMMARY.md can be repeated... copy-pasted, really... into each section for the relevant parts.


workingjubilee:
cc @NotGyro

docs/src/extension/write.md
```diff
@@ -1,3 +1,33 @@
 # Writing Extensions with PGRX
 
 <!-- TODO: write all of this -->
```
NotGyro: This TODO looks done already, right? Or, at least, it'd become ``<!-- TODO: write more of this -->``

NotGyro (approved):
This all looks good to me! There's a "TODO" in there which I think no longer applies since you've now done the task requested, but all of these docs read clearly and seem comprehensive on the topics they're describing.

NotGyro (approved):
Looks good, I'm liking that new `varlena` section also

workingjubilee:
Thank you!

PR #1571, opened 2024-02-21
pfree toasted Arrays on drop

workingjubilee:
For tables with many rows, or that include enough large arrays, being unable to pfree the entire array can cause needless memory pressure. As pgrx used to not bind objects to the lifetime of backing arrays, the assigned lifetime of any 'obj could exceed `&'arr Array<'_, T>`. This rendered any Drop implementation for Array or any of its parts unsound. The references, expecting to be valid for the underlying 'mcx or even longer, would yield clobbered memory when actually examined.

This has finally become untrue with the introduction of UnboxDatum. It has an `unsafe trait` obligation that the implementation not allow lifetimes to exceed the Datum's lifetime except if the data is always inherently 'static or Copy. This allows deallocating detoasted items, as Arrays now only yield an object of an appropriate lifetime. It is possible for us to get this wrong, but only by implementing UnboxDatum in an incorrect, unsound way, so the fix would be to that code instead.

### DIFF
--- a/pgrx/src/toast.rs
+++ b/pgrx/src/toast.rs
@@ -27,18 +27,7 @@ impl<T: Toasty> Drop for Toast<T> {
     fn drop(&mut self) {
         match self {
             Toast::Stale(_) => {}
-            Toast::Fresh(_t) => {
-                //
-                // issue #971 (https://github.com/pgcentralfoundation/pgrx/issues/971) points out a UAF with Arrays
-                // of `&str`s.  This happens because ultimately we free the detoasted array.  Rather
-                // than allowing outstanding borrows to become freed, we'll just not drop the detoasted
-                // Datum pointer, which will leak that Postgres-allocated memory.
-                //
-                // In general tho, this will be happening within the confines of a #[pg_extern] call
-                // and Postgres will just free the CurrentMemoryContext when the function is finished
-                //
-                // unsafe { t.drop_toast() }
-            }
+            Toast::Fresh(t) => unsafe { t.drop_toast() },
         }
     }
 }


workingjubilee:
~~My local is giving me "too many clients already" errors, so hopefully that's just an artifact of the testing infra and I don't wind up with egg on my face.~~ Yeah, everything passes on rerun.

workingjubilee:
cc @eeeebbbbrrrr @NotGyro

eeeebbbbrrrr:
I’m unsure what you could have done to use up all the connection slots?  Hanging processes or something from prior test runs?  Very unlikely.

workingjubilee:
It happens every now and then and usually clears on a rerun or clean-then-rerun. I think there's a problem in the test framework where it doesn't anticipate "CPU go brrr".

NotGyro (approved):
Looks good to me, and your reasoning is sound. This passes tests on my machine also.

eeeebbbbrrrr:
I forget offhand what the default connection limit is. Part of me thinks it’s 100 but the other part thinks it’s 10. We should look that up and adjust our postmaster options accordingly

eeeebbbbrrrr:
It’s 100*

https://www.postgresql.org/docs/current/runtime-config-connection.html

*unless your kernel is dumb.

eeeebbbbrrrr:
Rust wouldn’t be generating a test binary that fires off more threads than you have CPUs, would it?

you got 128 threads?!

workingjubilee:
> you got 128 threads?!

No.

> Rust wouldn’t be generating a test binary that fires off more threads than you have CPUs, would it?

I have no idea how many threads it generates! I have only about 16 logical threads locally.

Issue #1455, opened 2023-12-29
Use shlex whenever we touch shell-argument-shaped-things directly


workingjubilee:
in:
- https://github.com/pgcentralfoundation/pgrx/pull/1435
- https://github.com/pgcentralfoundation/pgrx/pull/1441

we do some things which would be better if we used [shlex](https://crates.io/crates/shlex).

workingjubilee:
Note https://github.com/advisories/GHSA-r7qv-8r2h-pg27 which is now patched as of shlex 1.3.0 and is otherwise further proof of why we shouldn't be hand-rolling our own shell-touching code.

NotGyro:
It looks like this has been solved in #1468. Let me know if this needs further changes.

workingjubilee:
Oh, yeah! It seems done!

PR #1572, opened 2024-02-21
Use shlex rather than split_whitespace()

NotGyro:
Potential implementation of #1455, in an attempt to resolve #1435 and #1441. As I said in our conversation in the private discord channel, actually implementing this change was very easy but verifying its correctness has been tricky. To my knowledge there's no test that would touch this codepath yet.

After looking over Cargo and Rustc's source and speaking with @workingjubilee, it looks like Cargo reads RUSTFLAGS and passes it to `rustc` invocations, and `rustc` doesn't touch RUSTFLAGS directly, so this shouldn't duplicate arguments.

### DIFF
--- a/pgrx/src/toast.rs
+++ b/pgrx/src/toast.rs
@@ -27,18 +27,7 @@ impl<T: Toasty> Drop for Toast<T> {
     fn drop(&mut self) {
         match self {
             Toast::Stale(_) => {}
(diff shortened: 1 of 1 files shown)


workingjubilee:
@NotGyro Oh, you shouldn't close and then force-push, that breaks the PR. You'll have to open a new one.

PR #1575, opened 2024-02-22
Give `WithTypeIds` its own module

workingjubilee:
All of this code is intimately related and clearly not part of all Datum usage. Put it in its own module. Also, hide the ~~bodies~~ ~~evidence~~ assertions which are only for private doctest purposes about unstable API details. :^)


NotGyro (approved):
Looks good to me.

Issue #1536, opened 2024-02-06
Usability: bad error message for unsigned integer arguments


nyurik:
This example produces an un-reasonably complex error, and only during `cargo pgrx test` phase.  I also tried it with `Option<u32>`

```rust
#[pg_extern]
fn hello_postile(value: u32) -> String {
    format!("hello {value}")
}

#[cfg(any(test, feature = "pg_test"))]
#[pg_schema]
mod tests {
    use pgrx::prelude::*;

    #[pg_test]
    fn test_hello_postile() {
        assert_eq!("Hello 10", crate::hello_postile(10));
    }
}
```

### Expected
* If `u32` is an invalid type parameter, compilation should fail
* Even if compilation succeeds, error message should point to the problem and explain solutions

### Actual results
* compilation succeeds
* test error is nearly impossible to understand

```
<compilation succeeded>
    Finished dev [unoptimized + debuginfo] target(s) in 7.49s



 Discovering SQL entities
  Discovered 4 SQL entities: 1 schemas (1 unique), 3 functions, 0 types, 0 enums, 0 sqls, 0 ords, 0 hashes, 0 aggregates, 0 triggers
     Writing SQL entities to /home/nyurik/.pgrx/13.13/pgrx-install/share/postgresql/extension/postile--0.0.0.sql
Error: 
   0: Could not write SQL to /home/nyurik/.pgrx/13.13/pgrx-install/share/postgresql/extension/postile--0.0.0.sql
   1: Macro expansion time suggested a source only mapping in return

Location:
   /home/nyurik/.cargo/registry/src/index.crates.io-6f17d22bba15001f/pgrx-sql-entity-graph-0.11.3/src/pg_extern/entity/mod.rs:200

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ SPANTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

   0: cargo_pgrx::command::schema::generate_schema with pg_version=13.13 profile=Dev test=true path=/home/nyurik/.pgrx/13.13/pgrx-install/share/postgresql/extension/postile--0.0.0.sql features=["pg_test", "pg13"]
      at /home/nyurik/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-pgrx-0.11.3/src/command/schema.rs:173
[...]

Issue #951, opened 2022-12-11
Obscure schema generator error for `pg_extern` function returning u32


EdMcBane:
Creating a `[pg_extern]` function that returns u32 compiles fine, but the schema generator reports a very obscure error:

```rust
#[pg_extern]
fn foobar() -> u32 {
    42
}
```

```
Error: 
   0: Could not write SQL to /Users/edmcbane/.pgx/13.9/pgx-install/share/postgresql/extension/rewind--0.0.0.sql
   1: Got non-plain mapped/composite return variant SQL in what macro-expansion thought was a type, got: One(Source { array_brackets: false })
```

workingjubilee:
The argument form of this was also reported in https://github.com/pgcentralfoundation/pgrx/issues/1536 and the diagnostics have significantly improved, see https://github.com/pgcentralfoundation/pgrx/pull/1577

I still intend to improve this a bit, but this issue is a duplicate now, even though it came first.

PR #1577, opened 2024-02-22
u32 has bad diagnostics with `#[pg_extern]`

workingjubilee:
Check in a test that demonstrates this so that we are aware of improvements, and to prevent any regressions, as they used to be significantly worse.

This is a step towards resolving these issues:
- https://github.com/pgcentralfoundation/pgrx/issues/951
- https://github.com/pgcentralfoundation/pgrx/issues/1536

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -187,15 +187,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9d297deb1925b89f2ccc13d7635fa0714f12c87adce1c75356b39ca9b7178567"
 
 [[package]]
-name = "basic-toml"
-version = "0.1.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2db21524cad41c5591204d22d75e1970a2d1f71060214ca931dc7d5afe2c14e5"
-dependencies = [
- "serde",
-]
-
-[[package]]
 name = "bgworker"
 version = "0.0.0"
 dependencies = [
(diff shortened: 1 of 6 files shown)


workingjubilee:
@nyurik This is a bit indirect still, isn't it? The spans are all fucked up.

nyurik:
Thx for working on it!  But yeah, trait not satisfied error is a bit of a rabbit hole...  Perhaps it might be better to handle it explicitly?  I am sure there will be plenty of surprises for such a common use case of an incorrect data type - so juts look at the input/return params in the `PgExtern::new` or nearby, and raise an error?

workingjubilee:
nondet diagnostic output, adding cshim alters the hash I think

workingjubilee:
> Thx for working on it! But yeah, trait not satisfied error is a bit of a rabbit hole... Perhaps it might be better to handle it explicitly? I am sure there will be plenty of surprises for such a common use case of an incorrect data type - so juts look at the input/return params in the PgExtern::new or nearby, and raise an error?

Well, I'd rather not special-case the strings of any types (at least, any further than our code already does...), because given the sheer amount of FFI pgrx involves, people are very likely to do something like:

```rust
#[pg_extern]
fn pg_fn(arg: ffi::c_uint) -> ffi::c_ulong {
```

workingjubilee:
This is not to say I'm against adding more diagnostic code here, I'd just rather do so in a more underhanded / generic way, like e.g. emitting structs with the name of the error I would want to emit and embedding the argument and return types in the struct.

workingjubilee:
Nope, last one didn't fix it. Guess I'm waiting for better test infra or finding a way to emit a better diagnostic or rustc being less random about ordering here.

nyurik:
How about just handle the well known simple types, e.g. `u16`, `u32`, ...?

workingjubilee:
> How about just handle the well known simple types, e.g. `u16`, `u32`, ...?

It feels like it lacks a satisfactory boundary.

nyurik:
Well, it is not perfect, i agree, but if it will make the most common usecase more helpful... at least maybe give a warning?

workingjubilee:
> Well, it is not perfect, i agree,

It isn't about perfection. The problem in this case is that there are three type systems involved (Rust's, C's, and Postgres's), and people are going to have different opinions about what is "obvious" and "common". Without usage statistics (and `cargo pgrx` doesn't include telemetry), I can't rule anything out.

workingjubilee:
I'm going to do some things that will significantly affect the diagnostics and might make them significantly worse or better, so this PR needs to serve its original purpose.

nyurik:
Would the recently released diagnostics support help in this case?  https://blog.rust-lang.org/2024/05/02/Rust-1.78.0.html#diagnostic-attributes

PR #1584, opened 2024-02-23
VARHDRSZ `const`s are not `fn`

workingjubilee:
These were defined as functions at a time, I presume, that the `offset_of!` macro we use was a bit less reliable at compile time? That's long since not been the case anymore, if ever. Redefine them as constants.


workingjubilee:
Technically breaking, I guess.

PR #1587, opened 2024-02-26
Make `RawArray::from_ptr` a simple cast

workingjubilee:
I have been working on the DetoastDatum implementation, and because our ArrayType wrappers happen to exercise a lot of parts of pgrx while being something I understand very well, I am aiming to ship its MVP with Yet Another improvement on `pgrx::datum::Array` that will be a model for future safe abstractions of Datum-derived types. This will involve quite a lot of work with RawArray and the code that is currently living next to RawArray, so I decided to factor that out, modularize the ported code, and overall just spruce things up a bit.

As part of that, this ports [`pg_sys::ArrayGetNItems`] to simply live inside the implementation of `RawArray::len`. This allows us to perform the calculation entirely in Rust, so there is now no more FFI overhead for creating a RawArray, and we can delete the `ARR_NELEMS` port. This also means we can simply stop caching the total element count, as we aren't saving enough math to be worth it. Now casting to RawArray from a simple reference or other raw pointer is a zero-overhead pointer cast, allowing reusing its already-defined accessors without incurring needless math or calls to Postgres.

[`pg_sys::ArrayGetNItems`]: https://github.com/postgres/postgres/blob/6979ea2638a51c40acf6d04b816550b2c35b3e55/src/backend/utils/adt/arrayutils.c#L46-L102

### DIFF
--- a/pgrx/src/array.rs
+++ b/pgrx/src/array.rs
@@ -9,12 +9,12 @@
 //LICENSE Use of this source code is governed by the MIT license that can be found in the LICENSE file.
 #![allow(clippy::precedence)]
 use crate::datum::Array;
-use crate::pg_sys;
 use crate::toast::{Toast, Toasty};
+use crate::{layout, pg_sys, varlena};
 use bitvec::prelude::*;
-use bitvec::ptr::{bitslice_from_raw_parts_mut, BitPtr, BitPtrError, Mut};
-use core::ptr::{slice_from_raw_parts_mut, NonNull};
-use core::slice;
+use bitvec::ptr::{self as bitptr, BitPtr, BitPtrError, Mut};
+use core::ptr::{self, NonNull};
+use core::{mem, slice};
 
 #[allow(non_snake_case)]
 #[inline(always)]
@@ -38,10 +38,8 @@ const fn MAXALIGN(len: usize) -> usize {
 unsafe fn ARR_NDIM(a: *mut pg_sys::ArrayType) -> usize {
     // #define ARR_NDIM(a)				((a)->ndim)
 
-    unsafe {
-        // SAFETY:  caller has asserted that `a` is a properly allocated ArrayType pointer
-        (*a).ndim as usize
-    }
+    // SAFETY:  caller has asserted that `a` is a properly allocated ArrayType pointer
+    unsafe { (*a).ndim as usize }
 }
 
 /// # Safety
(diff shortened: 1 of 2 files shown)


pgrx/src/array.rs
```diff
@@ -332,15 +181,21 @@ impl RawArray {
...
     #[inline]
     pub fn len(&self) -> usize {
-        self.len
+        // Calculating the product with i32 mirrors the Postgres implementation,
+        // except we can use checked_mul instead of trying to cast to 64 bits and
+        // hoping that doesn't also overflow on multiplication.
+        self.dims()
+            .into_iter()
+            .fold(Some(1i32), |prod, &d| prod.and_then(|m| m.checked_mul(d)))
+            .unwrap() as usize
```
workingjubilee: This isn't right, 0 dimensions should return 0 len.

workingjubilee:
cc @NotGyro

pgrx/src/array.rs
```diff
@@ -332,15 +181,28 @@ impl RawArray {
...
-            slice::from_raw_parts(ARR_DIMS(self.ptr.as_ptr()), ndim)
+            slice::from_raw_parts(port::ARR_DIMS(self.ptr.as_ptr()), ndim)
         }
     }
 
     /// The flattened length of the array over every single element.
     /// Includes all items, even the ones that might be null.
+    ///
+    /// # Panics
+    /// Panics if the Array's dimensions, multiplied together, exceed sizes Postgres can handle.
```
workingjubilee, line 192: This is *really silly*, but we're aiming to expose the ability to mutate an ArrayType eventually, which means that while I expect this to never happen for a Postgres-created array, they *do* assert on it for a reason! People would need to be able to check array invariants during constructing one, so...

pgrx/src/array/port.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+/// # Safety
+/// Does a field access, but doesn't deref out of bounds of ArrayType
+#[inline(always)]
+pub(super) unsafe fn ARR_HASNULL(a: *mut pg_sys::ArrayType) -> bool {
+    // #define ARR_HASNULL(a)			((a)->dataoffset != 0)
+
+    // SAFETY:  caller has asserted that `a` is a properly allocated ArrayType pointer
+    unsafe { (*a).dataoffset != 0 }
```
NotGyro, line 41: As confirmed in DMs, this will definitely always be zero if there is no null bitmap. The spec backs this up: https://doxygen.postgresql.org/array_8h_source.html

**This code is good so this doesn't need to be a comment,** but I'm just leaving this here in case anybody else reviewing this PR has the same question.

NotGyro (approved):
Looks good to me, this should be safe unless I'm missing something important.

pgrx/src/array.rs
```diff
@@ -332,15 +181,28 @@ impl RawArray {
...
+        // Calculating the product with i32 mirrors the Postgres implementation,
+        // except we can use checked_mul instead of trying to cast to 64 bits and
+        // hoping that doesn't also overflow on multiplication.
+        let dims = self.dims();
+        if dims.len() == 0 {
+            0
+        } else {
+            dims.into_iter()
+                .fold(Some(1i32), |prod, &d| prod.and_then(|m| m.checked_mul(d)))
+                .expect("Product of array dimensions should be <= i32::MAX") as usize
```
workingjubilee: This should probably be actually asserting on the size being less than `pg_sys::MaxArraySize`, since now that I look, that's smaller than `i32::MAX`... so tiresome.

workingjubilee:
I may wind up adding `RawArray::try_len` but... later.

Issue #1546, opened 2024-02-09
FR: Support extension names not matching the crate names


netthier:
I have encountered an issue while developing a regular Rust library and a PostgreSQL extension module for it.
Both are crates in the same workspace, named `foo` and `foo_psql` respectively.
It seems that `pgrx` expects the `.control` file to always be named `<CRATE_NAME>.control`, meaning I have to create a `foo_psql.control`, leading to users having to execute `CREATE EXTENSION foo_psql;`.
I would prefer if the extension were named `foo` instead, but am unsure how to achieve that (beyond manually renaming the `.control` file in the compiled artifacts), as naming the extension crate `foo` would lead to a conflict.
I would also be happy if the `.control` file in my source code stayed named the same, but the file produced by `cargo-pgrx` commands was renamed.

workingjubilee:
I asked @NotGyro to take a look at this. Unfortunately, it seems to be a bit tougher than expected. Implementing the patch in cargo-pgrx is easy, and that was done in https://github.com/NotGyro/pgrx/tree/extension-name-independent but pgrx features this, even on current develop:

https://github.com/pgcentralfoundation/pgrx/blob/3e530b6fdc99f2bbeab168dc48819a7810351254/pgrx/src/lib.rs#L272-L285

It's not clear how to work around this part so we'll probably be spinning our wheels for a bit on this.

eeeebbbbrrrr:
> I asked @NotGyro to take a look at this. Unfortunately, it seems to be a bit tougher than expected. Implementing the patch in cargo-pgrx is easy, and that was done in https://github.com/NotGyro/pgrx/tree/extension-name-independent but pgrx features this, even on current develop:
> 
> https://github.com/pgcentralfoundation/pgrx/blob/3e530b6fdc99f2bbeab168dc48819a7810351254/pgrx/src/lib.rs#L272-L285
> 
> It's not clear how to work around this part so we'll probably be spinning our wheels for a bit on this.

We're addressing this mess in PR #1591.  From there I suppose it should be fairly straightforward to figure out how to solve the OP's request.

PR #1591, opened 2024-02-29
Remove the `pg_sql_graph_magic!()` macro.

eeeebbbbrrrr:
The `pg_sql_graph_magic!()` did some shenanigans related to loading the extension ".control" file at compile-time and embedded it in the compiled artifact.  This was convoluted and actually unnecessary as the location where that information was used (in `schema.rs`) can easily access the ".control" file at runtime.

This PR removes that macro and its usages.

The `fn __pgrx__marker()` function is still required, however.  It's necessary to ensure that the `rustc` build we do for the `pgrx_embed` binary for schema generation properly links so that the other (dynamically referenced) `__pgrx_internals_XXX` symbols can be found.  The function, however, is now a noop.

This is related to #1546, but is not a solution.  It should just make solving that request easier.

### DIFF
--- a/cargo-pgrx/src/command/schema.rs
+++ b/cargo-pgrx/src/command/schema.rs
@@ -183,7 +183,14 @@ pub(crate) fn generate_schema(
         out_dot = Some(x.to_string());
     };
 
-    let codegen = compute_codegen(package_manifest_path, &symbols, &lib_name, out_path, out_dot)?;
+    let codegen = compute_codegen(
+        control_file,
+        package_manifest_path,
+        &symbols,
+        &lib_name,
+        out_path,
+        out_dot,
+    )?;
 
     let embed = {
         let mut embed = tempfile::NamedTempFile::new()?;
@@ -381,6 +388,7 @@ fn first_build(
 }
 
 fn compute_codegen(
+    control_file_path: impl AsRef<Path>,
     package_manifest_path: impl AsRef<Path>,
     symbols: &[String],
     lib_name: &str,
@@ -391,9 +399,19 @@ fn compute_codegen(
     let lib_name_ident = Ident::new(&lib_name, Span::call_site());
 
     let inputs = {
+        // read the .control file and do the required variable replacements
+        let control_file_contents = std::fs::read_to_string(control_file_path)?
+            .replace("@CARGO_VERSION@", env!("CARGO_PKG_VERSION"));
+
         let mut out = quote::quote! {
+            // call the marker.  Primarily this ensures that rustc will actually link to the library
+            // during the "pgrx_embed" build initiated by `cargo-pgrx schema` generation
+            #lib_name_ident::__pgrx_marker();
+
             let mut entities = Vec::new();
-            let control_file_entity = ::pgrx::pgrx_sql_entity_graph::SqlGraphEntity::ExtensionRoot(::#lib_name_ident::__pgrx_marker());
+            let control_file = ::pgrx::pgrx_sql_entity_graph::ControlFile::try_from(#control_file_contents).expect(".control file should properly formatted");
+            let control_file_entity = ::pgrx::pgrx_sql_entity_graph::SqlGraphEntity::ExtensionRoot(control_file);
+
             entities.push(control_file_entity);
         };
         for name in symbols.iter() {
--- a/cargo-pgrx/src/command/schema.rs
+++ b/cargo-pgrx/src/command/schema.rs
@@ -569,10 +569,10 @@ fn compute_sql(
     command.stderr(Stdio::inherit());
 
     // pass the package version through as an environment variable
-    command.env(
-        "PGRX_PKG_VERSION",
-        manifest.package_version().expect("`Cargo.toml` is missing the package version property"),
-    );
+    let cargo_pkg_version = std::env::var("CARGO_PKG_VERSION").unwrap_or_else(|_| {
+        manifest.package_version().expect("`CARGO_PKG_VERSION` should be set, and barring that, `Cargo.toml` should have a package version property")
+    });
+    command.env("CARGO_PKG_VERSION", cargo_pkg_version);
 
     let command_str = format!("{:?}", command);
     tracing::debug!(command = %command_str, "Running");
--- a/cargo-pgrx/src/command/schema.rs
+++ b/cargo-pgrx/src/command/schema.rs
@@ -400,8 +400,11 @@ fn compute_codegen(
 
     let inputs = {
         // read the .control file and do the required variable replacements
-        let control_file_contents = std::fs::read_to_string(control_file_path)?
-            .replace("@CARGO_VERSION@", env!("CARGO_PKG_VERSION"));
+        let control_file_contents = std::fs::read_to_string(control_file_path)?.replace(
+            "@CARGO_VERSION@",
+            &std::env::var("CARGO_PKG_VERSION")
+                .expect("`CARGO_PGK_VERSION` environment variable should be set"),
+        );
 
         let mut out = quote::quote! {
             // call the marker.  Primarily this ensures that rustc will actually link to the library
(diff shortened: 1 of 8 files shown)


pgrx-tests/src/framework.rs
```diff
@@ -312,7 +312,7 @@ fn install_extension() -> eyre::Result<()> {
         .arg("--pg-config")
         .arg(pg_config.path().ok_or(eyre!("No pg_config found"))?)
         .stdout(Stdio::inherit())
-        .stderr(Stdio::piped())
+        .stderr(Stdio::inherit())
```
eeeebbbbrrrr, line 315: maybe out of place in this PR, but I burned well over an hour trying to figure out why the `cargo-pgrx` output was missing, leading me to believe schema generation wasn't even happening during `cargo test`.

pgrx/src/lib.rs
```diff
@@ -182,7 +182,13 @@ const _: () = {
 macro_rules! pg_module_magic {
     () => {
         $crate::pg_magic_func!();
-        $crate::pg_sql_graph_magic!();
+
+        // A marker function which must exist in the root of the extension for proper linking by the
+        // "pgrx_embed" binary during `cargo-pgrx schema` generation.
+        #[doc(hidden)]
+        pub fn __pgrx_marker() {
```
eeeebbbbrrrr, line 190: this should probably be `#[inline(never)`].

I haven't been able to cause it to happen, but it seems to me that the compiler's DCE passes could somehow cause the `pgrx_embed` binary to not link to the .so if the function and its call point in `pgrx_embed::main()` were removed.

EDIT:  It does **NOT** also need to be `#[no_mangle]` as it's referenced directly, not indirectly through a `extern "Rust" {}` declaration like all the other `__pgrx_internals_XXX` functions are.

pgrx-sql-entity-graph/src/control_file.rs
```diff
@@ -83,13 +96,13 @@ impl ControlFile {
             relocatable: temp
                 .get("relocatable")
                 .ok_or(ControlFileError::MissingField { field: "relocatable" })?
-                == &"true",
+                == "true",
```
eeeebbbbrrrr, line 116: `cargo clippy` got upset once I touched this file.  Same below on lines 103 and 105.

pgrx-tests/src/framework.rs
workingjubilee: Any diagnostics changes necessary to uhhh write the PR are appropriate, I think, barring like a complete rewrite of framework.rs

pgrx/src/lib.rs
workingjubilee: That is very astute, as Saethlin has been working extensively on making the compiler do _very aggressive_ inlining, because idiomatic Rust uses lots of tiny functions.

workingjubilee:
For the audience's benefit: This cannot be landed as-is because it breaks `cargo pgrx schema`, which is somewhat necessary for diagnostic purposes.

eeeebbbbrrrr:
> For the audience's benefit: This cannot be landed as-is because it breaks `cargo pgrx schema`, which is somewhat necessary for diagnostic purposes.

ee63bf9 fixes that.  ;)

cargo-pgrx/src/command/schema.rs
```diff
@@ -545,6 +568,12 @@ fn compute_sql(profile: &CargoProfile, package_name: &str) -> eyre::Result<()> {
     command.stdout(Stdio::inherit());
     command.stderr(Stdio::inherit());
 
+    // pass the package version through as an environment variable
+    command.env(
+        "PGRX_PKG_VERSION",
+        manifest.package_version().expect("`Cargo.toml` is missing the package version property"),
+    );
```
workingjubilee: I would prefer if this checked if `std::env::var("CARGO_PKG_VERSION")` is set and only when it is unset do the needful, and when we do that, **lie**: set Cargo's env vars for it, and only set the env vars we need. This avoids adding another quasi-stable quasi-unstable undocumented interface to our code.

eeeebbbbrrrr: okay.

eeeebbbbrrrr:
It’s green!

workingjubilee:
Holy shit!

pgrx-sql-entity-graph/src/control_file.rs
```diff
@@ -43,18 +45,46 @@ pub struct ControlFile {
...
+
+            // endeavor to not require external values if they're not used by the input
+            if input.contains(CARGO_VERSION) {
+                input = input.replace(
+                    CARGO_VERSION,
+                    &std::env::var("CARGO_PKG_VERSION").map_err(|_| {
+                        ControlFileError::MissingEnvvar("CARGO_PKG_VERSION".to_string())
+                    })?,
+                );
+            }
```
workingjubilee, line 83: This does introduce a slight behavior divergence, but that seems fine, since the main validation effort is done via cargo-pgrx.

cargo-pgrx/src/command/schema.rs
```diff
@@ -391,9 +399,20 @@ fn compute_codegen(
...
 
     let inputs = {
+        let control_file_path = control_file_path
+            .as_ref()
+            .to_str()
+            .expect(".control file filename should be valid UTF8");
         let mut out = quote::quote! {
+            // call the marker.  Primarily this ensures that rustc will actually link to the library
+            // during the "pgrx_embed" build initiated by `cargo-pgrx schema` generation
+            #lib_name_ident::__pgrx_marker();
```
workingjubilee, line 409: Perhaps we should use `pub extern "C" fn` on this marker? rustc is more reluctant to play around with obviously-for-external-consumption interfaces. It's not something we need to fix now, however, just musing aloud.

cargo-pgrx/src/command/schema.rs
```diff
@@ -391,9 +399,20 @@ fn compute_codegen(
...
+
             let mut entities = Vec::new();
-            let control_file_entity = ::pgrx::pgrx_sql_entity_graph::SqlGraphEntity::ExtensionRoot(::#lib_name_ident::__pgrx_marker());
+            let control_file_path = std::path::PathBuf::from(#control_file_path);
```
workingjubilee, line 412: I wonder if there's some cursed set of cases that results in this being a malformed UTF-8 string? ...but if someone seriously asks for supporting non-UTF-8 paths in our greenfield extension build system, I'm probably going to tell them exactly where they can shove that request. Maybe we should be simplifying our code by using [camino](https://docs.rs/camino/latest/camino/).

pgrx-sql-entity-graph/src/control_file.rs
```diff
@@ -43,18 +45,46 @@ pub struct ControlFile {
...
+    ///
+    /// Returns a `ControlFileError` if any of the required fields are missing from the input string
+    /// or if any required environment variables (for dynamic variable substitution) are missing
     ///
     /// ```rust
     /// use pgrx_sql_entity_graph::ControlFile;
     /// # fn main() -> eyre::Result<()> {
-    /// let context = include_str!("../../pgrx-examples/custom_types/custom_types.control");
+    /// # // arrays.control chosen because it does **NOT** use the @CARGO_VERSION@ variable
+    /// let context = include_str!("../../pgrx-examples/arrays/arrays.control");
```
workingjubilee, line 65: Yeah this seems fair. We have other tests to guarantee this works.

pgrx/src/lib.rs
```diff
@@ -259,39 +266,6 @@ macro_rules! pg_magic_func {
...
-            ))
-            .replace("@CARGO_VERSION@", package_version);
-
-            let control_file =
-                $crate::pgrx_sql_entity_graph::ControlFile::try_from(context.as_str())
-                    .expect("Could not parse control file, is it valid?");
-            control_file
-        }
-    };
-}
```
workingjubilee, line 293: Excellent.

workingjubilee (approved):
Just some comments for future reference if I ever have to come back to this PR. It looks good! Moving build complexity out of the library is great.

PR #1595, opened 2024-02-29
Fix unsound `&` and `&mut` in `fcinfo.rs`

workingjubilee:
There were many unsound details in the implementation of this code, like using `*mut T` and `&mut T` at the same time, creating slices using bindgen-provided code that is of dubious soundness, or taking `&mut [T]` to what should be `&mut [MaybeUninit<T>]`.

They have been fixed. The primary strategy used is to interact only with `*mut T` if there is any doubt, and to constrain `&mut` to very narrow scopes.


workingjubilee:
Not done yet. I have to fix the `direct_function_call_as_datum_internal` code.

pgrx/src/fcinfo.rs
```diff
@@ -332,30 +335,30 @@ unsafe fn direct_function_call_as_datum_internal(
...
-    fcinfo.isnull = false;
-    fcinfo.nargs = nargs;
+    (*fcinfo).flinfo = std::ptr::null_mut();
+    (*fcinfo).context = std::ptr::null_mut();
+    (*fcinfo).resultinfo = std::ptr::null_mut();
+    (*fcinfo).fncollation = pg_sys::InvalidOid;
+    (*fcinfo).isnull = false;
+    (*fcinfo).nargs = nargs;
 
-    let arg_slice = fcinfo.args.as_mut_slice(args.len());
```
workingjubilee, line 349: This? Using a palloc, which creates uninit memory, and taking `&mut` to an initialized type? Worse, one with illegal bitpatterns (bool)? This was simply UB. We could do `&mut [MaybeUninit<NullableDatum>]`, but not this.

pgrx/src/fcinfo.rs
```diff
@@ -332,30 +335,30 @@ unsafe fn direct_function_call_as_datum_internal(
...
-    for (i, &arg) in args.iter().enumerate() {
-        arg_slice[i].isnull = arg.is_none();
-        arg_slice[i].value = arg.unwrap_or(pg_sys::Datum::from(0));
+    let args_ptr: *mut pg_sys::NullableDatum = ptr::addr_of_mut!((*fcinfo).args).cast();
+    let arg_slice = slice::from_raw_parts_mut(args_ptr, args.len());
+    for (n_datum, arg) in arg_slice.iter_mut().zip(args) {
+        n_datum.isnull = arg.is_none();
+        n_datum.value = arg.unwrap_or(pg_sys::Datum::from(0));
```
workingjubilee: I believe this pattern is more idiomatic and optimizes better.

pgrx/src/fcinfo.rs
```diff
@@ -332,30 +335,30 @@ unsafe fn direct_function_call_as_datum_internal(
...
 ) -> Option<pg_sys::Datum> {
     let nargs: i16 = args.len().try_into().expect("too many args passed to function");
-    let fcinfo_ptr = pg_sys::palloc(
+    let fcinfo = pg_sys::palloc0(
         std::mem::size_of::<pg_sys::FunctionCallInfoBaseData>()
             + std::mem::size_of::<pg_sys::NullableDatum>() * args.len(),
     )
     .cast::<pg_sys::FunctionCallInfoBaseData>();
 
-    let fcinfo = fcinfo_ptr.as_mut().unwrap_unchecked();
```
workingjubilee, line 341: This is a dubious-soundness action due to later interactions with the args slice being somewhat incompatible with Stacked Borrows.

pgrx/src/fcinfo.rs
```diff
@@ -332,30 +335,30 @@ unsafe fn direct_function_call_as_datum_internal(
...
+        n_datum.isnull = arg.is_none();
+        n_datum.value = arg.unwrap_or(pg_sys::Datum::from(0));
     }
 
-    let result = func(fcinfo_ptr);
-    let result = if fcinfo.isnull { None } else { Some(result) };
+    let result = func(fcinfo);
+    let result = if (*fcinfo).isnull { None } else { Some(result) };
 
-    pg_sys::pfree(fcinfo_ptr.cast());
```
workingjubilee, line 358: **This was UB. It has always been UB. Even under Tree Borrows, it will always be UB.** This interacts with `&mut T` and `*mut T` as if they have equal standing. **They do not.** The `&mut T`, even under the permissive models, "quiets" the `*mut T` while it is still in existence, and this direct overlapping interaction with the two is and always was simply unsound.

workingjubilee: In fact, we probably should `drop` the slice before we call `func`?

pgrx/src/fcinfo.rs
```diff
@@ -162,10 +163,12 @@ pub unsafe fn pg_get_nullable_datum(
     fcinfo: pg_sys::FunctionCallInfo,
     num: usize,
 ) -> pg_sys::NullableDatum {
-    let fcinfo = unsafe { fcinfo.as_mut() }.unwrap();
+    let _nullptr_check = ptr::NonNull::new(fcinfo).unwrap();
```
NotGyro: Nit: Would this be better as a `.expect("some informative error")` instead of `.unwrap()`?

NotGyro (approved):
Looks good to me, this looks like an improvement in soundness. Compiles and passes tests on my machine.

pgrx/src/fcinfo.rs
workingjubilee: Probably!

workingjubilee: Thank you! Addressed in 87b253e.

PR #1596, opened 2024-03-01
Address pg16 rename of `*_tree_walker` fn

### LINKED ISSUE #1583: Missing parameters in the `arg_walker` callback of `raw_expression_tree_walker` ?
Hi !

This may not be a bug but it is at least very weird :)

I'm trying  to use the `raw_expression_tree_walker` function, but I don't understant why the `arg_walker` callback function as no parameters... It's signature is 

`unsafe extern C fn() -> bool`

https://docs.rs/pgrx/latest/pgrx/pg_sys/fn.raw_expression_tree_walker.html

This callback seems useless if I can't inspect the node itself...

I would expect the walker callback signature to be 

`unsafe extern C fn(* mut Node, * c_void) -> bool`

Looking at the example below in the PostgreSQL code, I see that the tree_walker_callback used in raw_expression_tree_walker does have 2 parameters....

https://github.com/postgres/postgres/blob/master/src/backend/parser/analyze.c#L3576

Am I missing something obvious ?

workingjubilee:
These functions had been renamed in Postgres 16, and they are now fn macros. That means this `extern "C"` declaration is simply incorrect in Postgres 16. Implement a relatively logical version-by-version handling.

Fixes #1583.

### DIFF
--- a/pgrx-pg-sys/src/port.rs
+++ b/pgrx-pg-sys/src/port.rs
@@ -1,7 +1,6 @@
 use crate as pg_sys;
 
 use memoffset::*;
-use pgrx_macros::pg_guard;
 use std::str::FromStr;
 
 /// this comes from `postgres_ext.h`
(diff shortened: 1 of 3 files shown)


workingjubilee:
cc @daamien

pgrx-pg-sys/src/port.rs
```diff
@@ -229,29 +228,70 @@ pub unsafe fn heap_tuple_get_struct<T>(htup: super::HeapTuple) -> *mut T {
...
         query: *mut super::Query,
-        walker: ::std::option::Option<
-            unsafe extern "C" fn(*mut super::Node, *mut ::std::os::raw::c_void) -> bool,
+        walker: ::core::option::Option<
+            unsafe extern "C" fn(*mut super::Node, *mut ::core::ffi::c_void) -> bool,
         >,
-        context: *mut ::std::os::raw::c_void,
-        flags: ::std::os::raw::c_int,
+        context: *mut ::core::ffi::c_void,
+        flags: ::core::ffi::c_int,
```
workingjubilee, line 245: The lack of conditions in the original version made this potentially unsound when called on Postgres 16, since then you just... jump to nothing, I guess? Hope the dynamic loader errors properly! "You mean the linker." **We defer symbol resolution to runtime on some platforms!**

Honestly, redeclaring arguments like this is also pretty fucking UB, but I guess everything sucks!

daamien:
@workingjubilee Thanks you so much !

Issue #1586, opened 2024-02-25
Adding `mod null`


workingjubilee:
There are two different ways that Postgres describes a series of "null" values, and several types like `NullableDatum` of special interest. A set of shared abstractions, traits and types, with appropriate (minimal) genericity should unite them. This is basically https://github.com/pgcentralfoundation/pgrx/issues/620 but in programmatic terms. This will have enough going on that it should get its own module.

Note that the _main_ reason for this is that we have `Option<Option<T>>` as a "reasonable" signature for `Array::get`. It gets confusing as to which layer of that type represents what. One of them represents "is null", the other represents "not in bounds"!

workingjubilee:
The second-most important part is to have a way to easily iterate a set of maybe-null values, where the encoding will be either:
```rust
struct IterNullableValuesWithBitmap<T> {
    nulls: *mut [bits8],
    values: *mut [T], // or some incredibly accursed variant "inline" encodings
    // don't worry too much about the inline encodings
}
```
or this:
```rust
struct IterNullableValuePtrsWithBools {
    nulls: *mut [bool],
    values: *mut [*mut ffi::c_void], // AKA *mut [Datum]
}
```
and have a way to write this generically, something kiiinda like
```rust
struct: IterNullable<Nulls, Values> {
    nulls: Nulls,
    values: Values,
}
```

workingjubilee:
cc @NotGyro

Issue #620, opened 2022-08-12
References to "null" should be unambiguous


workingjubilee:
There are four different kinds of "null" in PGX:

The SQL Null type.
The pointer to `NULL` (`0x0` on most machines, but sometimes also `usize::MAX`).
The null byte aka `\0`.
`is_null` boolean markers.
These all have wildly different implications.
PGX should not, in source or documentation, use the term "null" unqualified. Anywhere. Ever.

workingjubilee:
Postgres also has the concept of a "nullbitmap" and indices into it, which work like the `is_null: [bool]` periodically witnessed elsewhere, except bitpacked.

workingjubilee:
Also see https://github.com/tcdi/pgx/issues/792

PR #1602, opened 2024-03-06
Nullable type

NotGyro:
Add an `Option`-like type to distinguish "This element is out-of-bounds" from "This element is a Postgres `null`", as well as genericizing over NullLayout types that mark where null elements are in a container (for example, the null bitslice on arrays). 

This is meant to implement the behavior requested in #620 and #1586 

For iterators, a separate pull request has been split off: NotGyro/pgrx#2


pgrx/src/nullable.rs
```diff
@@ -0,0 +1,523 @@
...
+    fn count_nulls(&self) -> usize {
+        0
+    }
+}
+// No skipping when there are no nulls.
+impl ContiguousNullLayout<usize> for StrictNulls {}
+
+// Intended to represent any one of the three nullable container layouts used
+// by Postgres.
+pub enum AnyNullLayout<'a> {
```
workingjubilee: I don't think this needs to actually exist right now, so I think we should eschew premature abstraction.

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,523 @@
...
+    /// \[value, value, value, value, null, value value, null\]
+    /// However, the underlying data buffer would be:
+    /// \[value, value, value, value, value, value\]
+    /// because of the skip behavior
+    Bitmap(BitSliceNulls<'a>),
+    // TODO: Find implementation details on this
+    BoolSlice(BoolSliceNulls<'a>),
+    /// Bool map, simply an array of booleans telling you
+    /// No nulls
+    Strict(StrictNulls),
```
workingjubilee: The reason that the Strict variant of NullKind exists is somewhat of an artifact of other decisions, if we can get rid of it that would be preferable.

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,523 @@
...
+}
```
workingjubilee: The case where we want to allow this will probably be the case where we want to allow interacting with either the "Expanded" or "Flat" array representation, and for various reasons we will probably not need an enum of null layouts for that.

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,523 @@
...
+        result_value
+    }
+}
+
+/// Iterator for Strict containers,
+/// which is to say, containers which cannot contain nulls
+pub struct NoNullsIter {
+    /// Will be set to container.len() and so a 0 value means we've reached the end.
+    remaining: usize,
+}
```
workingjubilee: An alternative possibility is to simply make this iterator infinite.

workingjubilee (commented):
Apologies, thoughts are incomplete and what is left over after sneezing my brains out all day. Wanted to post what I did have.

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,523 @@
...
+    fn next(&mut self) -> Option<Self::Item> {
+        // is_null() should handle bounds checking internally
+        let result_value = self.layout.is_null(self.current).map(|b| match b {
+            false => Nullable::Valid(()),
+            true => Nullable::Null,
+        });
+        self.current += 1;
+        result_value
+    }
+}
```
workingjubilee: cool.

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,523 @@
...
+    I: Iterator<Item = Nullable<()>>,
+{
+    nulls: I,
+    container_ref: &'mcx A,
+    valid_idx: usize,
+    // Required to keep a reference to NullableContainer<T>
+    __marker: PhantomData<T>,
+}
+
+// SKIPPING (i.e. non-contiguous) implementation
```
workingjubilee: hmm?

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,523 @@
...
+// Intended to represent any one of the three nullable container layouts used
+// by Postgres.
+pub enum AnyNullLayout<'a> {
+    /// Bitmap of where null slots are in this container.
+    /// For example, 00001001 would represent:
+    /// \[value, value, value, value, null, value value, null\]
+    /// However, the underlying data buffer would be:
+    /// \[value, value, value, value, value, value\]
+    /// because of the skip behavior
+    Bitmap(BitSliceNulls<'a>),
```
workingjubilee: other way around, actually, `[valid, valid, valid, valid, null, valid, valid, null]` is `0b11110110`. you implement this correctly elsewhere, this is just a doc bug.

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,523 @@
...
+            Some(val) => Self::Valid(val),
+            None => Self::Null,
+        }
+    }
+}
+
+/// This type isn't directly related to [`pg_sys::NullableDatum`], but is
+/// intended to reference the same underlying data as a pg_sys::NullableDatum
+/// after undergoing several layers of PGRX's Rust memory safety.
+pub type NullableDatum<'src> = Nullable<Datum<'src>>;
```
workingjubilee: So for reference, and the only thing I am not going to budge on, here: if this is providing some ergonomic improvement in your opinion then I would strongly prefer that we find a way to provide it without using a type name that matches something  in `pg_sys::` but **cannot** be soundly `mem::transmute`d to its corresponding `pg_sys::` type. I have engaged in Considerable Trickery to make that possible in the relevant cases, so as to minimize the surprise to `unsafe` code.

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,523 @@
...
+    fn len(&self) -> Idx;
+
+    /// Returns true if this container has any nulls in it presently.
+    /// # Performance
+    /// Different implementors of this type might have wildly varying
+    /// performance costs for checking this - for a null bitslice -style
+    /// container, it should complete in O(log(n)) at worst, but for a
+    /// null boolean array -style implementation, it could be O(n),
+    /// where n is the length of the container.
+    fn has_nulls(&self) -> bool;
```
workingjubilee: People might expect this function to match the answer of `ARR_HASNULL`, which is an O(1) fn that answers whether containing a null value is _possible_, not whether it has one. I would prefer the name `any_nulls` (matching the associated fn on RawArray) or `contains_nulls` (matching the `array_contains_nulls` fn Postgres uses) for the fn that answers this question.

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,523 @@
...
+}
+
+impl<T> Into<Option<T>> for Nullable<T> {
+    fn into(self) -> Option<T> {
+        match self {
+            Nullable::Valid(val) => Some(val),
+            Nullable::Null => None,
+        }
+    }
+}
```
workingjubilee: Can't we `impl From<Nullable<T>> for Option<T>`?

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,523 @@
...
+where
+    Idx: PartialEq + PartialOrd,
+{
+    fn len(&self) -> Idx;
+
+    /// Returns true if this container has any nulls in it presently.
+    /// # Performance
+    /// Different implementors of this type might have wildly varying
+    /// performance costs for checking this - for a null bitslice -style
+    /// container, it should complete in O(log(n)) at worst, but for a
```
workingjubilee: Hmm? Why do you say that? Isn't it a constant factor divisor?

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,523 @@
...
+    /// null boolean array -style implementation, it could be O(n),
+    /// where n is the length of the container.
+    fn has_nulls(&self) -> bool;
+
+    fn count_nulls(&self) -> usize;
+
+    /// Returns Some(true) if the element at `idx`` is valid (non-null),
+    /// or `None` if `idx` is out-of-bounds
+    /// Implementors should handle bounds-checking.
```
workingjubilee: There is no need to admonish people to implement the trait's described contract correctly when it is a safe fn and this is not a trusted (i.e. `unsafe`) trait.
```suggestion
```

...on the other hand, this may need to be an `unsafe trait`. It's not clear to me if other code needs to rely on its implementation being correct, at the moment.

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,523 @@
...
+pub struct BitSliceNulls<'a>(pub &'a BitSlice<u8>);
+
+impl<'a> NullLayout<usize> for BitSliceNulls<'a> {
+    fn len(&self) -> usize {
+        BitSlice::<u8>::len(&self.0)
+    }
+
+    fn has_nulls(&self) -> bool {
+        self.0.not_all()
+    }
```
workingjubilee: Should doc-comment on cost of impl.

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,523 @@
...
+pub struct StrictNulls(pub usize);
+
+impl NullLayout<usize> for StrictNulls {
+    fn len(&self) -> usize {
+        self.0
+    }
+
+    fn has_nulls(&self) -> bool {
+        false
+    }
```
workingjubilee: Should doc-comment on cost of impl.

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,523 @@
...
+}
+
+pub struct BoolSliceNulls<'a>(pub &'a [bool]);
+
+impl<'a> NullLayout<usize> for BoolSliceNulls<'a> {
+    fn len(&self) -> usize {
+        <[bool]>::len(&self.0)
+    }
+
+    fn has_nulls(&self) -> bool {
```
workingjubilee: Should doc-comment on cost of impl.

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,523 @@
...
+
+impl<T> Eq for Nullable<T> where T: PartialEq + Eq {}
+
+impl<T> Nullable<T> {
+    pub fn into_option(self) -> Option<T> {
+        match self {
+            Nullable::Valid(val) => Some(val),
+            Nullable::Null => None,
+        }
+    }
```
workingjubilee, line 79: Seems like this can just be our `From`/`Into` impls.

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,523 @@
...
+    pub fn as_option_mut<'a>(&'a mut self) -> Option<&'a T> {
+        match self {
+            Nullable::Valid(val) => Some(val),
+            Nullable::Null => None,
+        }
+    }
```
workingjubilee: These are good! We can probably elide the lifetime though?

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,523 @@
...
+            Nullable::Valid(_) => true,
+            Nullable::Null => false,
+        }
+    }
+    pub fn is_null(&self) -> bool {
+        !self.is_valid()
+    }
+    pub fn unwrap(self) -> T {
+        self.into_option().unwrap()
+    }
```
workingjubilee: Yep, these convenience fn will be useful. Thank you!

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,523 @@
...
+        match (self, other) {
+            (Self::Valid(l_value), Self::Valid(r_value)) => l_value == r_value,
+            _ => core::mem::discriminant(self) == core::mem::discriminant(other),
+        }
+    }
+}
+
+impl<T> Eq for Nullable<T> where T: PartialEq + Eq {}
+
+impl<T> Nullable<T> {
```
workingjubilee, line 72: Consider `map` and `and_then`?

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,523 @@
...
+    }
+    pub fn is_null(&self) -> bool {
+        !self.is_valid()
+    }
+    pub fn unwrap(self) -> T {
+        self.into_option().unwrap()
+    }
+    pub fn expect(self, msg: &'static str) -> T {
+        self.into_option().expect(msg)
+    }
```
workingjubilee: I wonder if we want a different signature than std has for `expect`... I often find myself wishing it would take format-args, somehow...

pgrx/src/nullable.rs
NotGyro: I could see that but I'm wondering if there's any context in which the null-layout iter will be used to do bounds-checking, given that the null slice is already used to do that outside of the context of iterators.

workingjubilee: The main reason that is done the way it is was because otherwise we were storing the len like 20 times (once in RawArray, once in NullKind, once in Array itself, and another place...) so I was trying to reduce the number of places that was stored. You can comfortably assume anything that isn't _inherent_ to the type doesn't need to be tracked by these implementations, and other code can simply always do
```rust
nulls.take(len).zip(values).map(|(null, datum)| match null {
        /* etc... */
    }
)
```

pgrx/src/nullable.rs
NotGyro: Oh absolutely, these feel like a bit of a formality but I can take care of that yeah.

pgrx/src/nullable.rs
NotGyro: My worry would be people might be surprised by the different signature. However, something `&'static str` satisfies, but which is more flexible, might work out well - `T: Into<Cow<'a, str>>` perhaps? The standard library has from/into implementations from `&'a str` to `Cow<'a, str>`. I'll give that a shot.

pgrx/src/nullable.rs
NotGyro: Resolved in 1a3d5719b77c233fd96440597e3aa43d794847d4

pgrx/src/nullable.rs
NotGyro: Added as requested in 1a3d5719b77c233fd96440597e3aa43d794847d4

pgrx/src/nullable.rs
NotGyro: Resolved - this type has been refactored away, since it was over-calibrated on the current Array implementation.

pgrx/src/nullable.rs
NotGyro: I think it's reasonable for the Null Layout to know what is and isn't out-of-bounds, but I could be convinced otherwise. It might be unnecessary.

pgrx/src/nullable.rs
NotGyro: Resolved, AnyNullLayout has been cut out. You're right that it was too specific to Array's implementation.

pgrx/src/nullable.rs
NotGyro: Oh that's a good point, I might try to cut out unnecessary copies of the length tag tomorrow.

pgrx/src/nullable.rs
NotGyro: Thanks for the heads-up!

pgrx/src/nullable.rs
NotGyro: Resolved, I cut this out since Nullable<Datum<'src>> isn't that bothersome (ergonomically) to type out.

pgrx/src/nullable.rs
NotGyro: Resolved, thanks for the suggestion. I rewrote it in that form.

pgrx/src/nullable.rs
NotGyro: Done

pgrx/src/nullable.rs
NotGyro: This is also done.

pgrx/src/nullable.rs
NotGyro: The reason I wrote it this way is for ergonomics - sometimes it's easier to simply type `variable.into_option()`, rather than, in an ambiguous context, needing to type something like `<variable as Into<Option<InnerType>>>::into()`. However, I'll defer to your judgement on this - is that a good reason to keep these methods in this arrangement?

pgrx/src/nullable.rs
NotGyro: Looks like eliding the lifetime is fine! Thanks for the tip.

pgrx/src/nullable.rs
NotGyro: Implemented in 1a3d5719b77c233fd96440597e3aa43d794847d4

pgrx/src/nullable.rs
NotGyro: Renamed to `contains_nulls()`

pgrx/src/nullable.rs
workingjubilee: Oh, that's fine.

NotGyro: Thanks!

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,520 @@
...
+        match self {
+            Self::Valid(arg0) => Self::Valid(arg0.clone()),
+            Self::Null => Self::Null,
+        }
+    }
+}
+
+impl<T> PartialEq for Nullable<T>
+where
+    T: Eq,
```
workingjubilee:
```suggestion
impl<T> PartialEq for Nullable<T>
where
    T: PartialEq,
```

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,520 @@
...
+            (Self::Valid(l_value), Self::Valid(r_value)) => l_value == r_value,
+            _ => core::mem::discriminant(self) == core::mem::discriminant(other),
+        }
+    }
+}
+impl<T> Eq for Nullable<T> where T: PartialEq + Eq {}
+
+impl<T> PartialOrd for Nullable<T>
+where
+    T: PartialOrd + PartialEq + Eq,
```
workingjubilee: I don't think we need to have a stricter bound than necessary?

```suggestion
    T: PartialOrd
```

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,520 @@
...
+{
+    #[inline]
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Self::Valid(l_value), Self::Valid(r_value)) => l_value == r_value,
+            _ => core::mem::discriminant(self) == core::mem::discriminant(other),
+        }
+    }
+}
+impl<T> Eq for Nullable<T> where T: PartialEq + Eq {}
```
workingjubilee: Hesitating slightly on this because nulls normally implement `null != null`.

workingjubilee (commented):
I think it would be best to split things up so that we can land the Nullable type and the internal Array refactor and then work on extending the idea further in a followup.

pgrx/src/nullable.rs
NotGyro: [Postgres does treat null as not equal to null,](https://www.postgresql.org/docs/current/functions-comparison.html) and this is intended to represent a Postgres null, so it might be wise to replicate that behavior, yeah.

pgrx/src/nullable.rs
NotGyro: PartialOrd requires PartialEq, but not Eq.
 I removed the Eq bound in 01e64fc3e01489ce402571f3504f80a96d2fe406

pgrx/src/nullable.rs
NotGyro: Resolved in 01e64fc3e01489ce402571f3504f80a96d2fe406

pgrx/src/nullable.rs
NotGyro: Done, `MaybeStrictNulls` now does not have a length tag for the strict case. Array's bounds'-checking is done by accessing Array. See 39f5ad9

pgrx/src/nullable.rs
NotGyro: Removed the length-tag from MaybeStrictNulls

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,466 @@
...
+}
+impl<T> Into<Nullable<T>> for Option<T> {
+    #[inline]
+    fn into(self) -> Nullable<T> {
+        match self {
+            Some(val) => Nullable::Valid(val),
+            None => Nullable::Null,
+        }
+    }
+}
```
workingjubilee: Wait, I thought we had a pair of From implementations here?

pgrx/src/datum/array.rs
```diff
@@ -170,28 +145,61 @@ impl<'mcx, T: UnboxDatum> Array<'mcx, T> {
         ArrayIterator { array: self, curr: 0, ptr }
     }
 
+    /// Return an iterator of `Nullable<T>`
+    /// Behavior is identical to ArrayIterator, but with `Nullable<T>`'s,
+    /// syntax, meaning `next()` returns `Option<Nullable<T>>`
+    /// instead of `Option<Option<T>>`.
+    pub fn iter_nullable(&self) -> NullableArrayIterator<'_, T> {
+        NullableArrayIterator { inner: self.iter() }
+    }
```
workingjubilee: I would rather there be no new public functions on Array.

pgrx/src/nullable.rs
```diff
@@ -0,0 +1,466 @@
...
+{
+}
+
+/// All non-skipping null layouts. Marker trait for nullable containers that
+/// are nonetheless safe to iterate over linearly.
+pub trait ContiguousNullLayout<Idx>: NullLayout<Idx>
+where
+    Idx: PartialEq + PartialOrd,
+{
+}
```
workingjubilee: I don't think we need these marker traits, as they should be fully simply described by how the implementer implements the trait.

workingjubilee:
As I mentioned while discussing this offsides, I have some other thoughts about the way the traits carve up the world that I think are better proposed as a refactoring since it's hard to really think them through in the review/talking format.

pgrx/src/nullable.rs
NotGyro: Resolved in 5792cade6c0b84840d10f8b9ef5394fc12df25a1

pgrx/src/nullable.rs
NotGyro: Fixed in f503f442ed9d88d2d51070db320a41d3b21e3a8d

pgrx/src/datum/array.rs
NotGyro: Resolved in 8bc1648

workingjubilee (approved):

PR #1603, opened 2024-03-10
feat(pgrx): add chrono feature for easy conversions

t3hmrman:
This commit adds a `chrono` feature flag to `pgrx`, which enables conversions between `pgrx` native date/time (with or without timezone) and `chrono` types like `NaiveDateTime`.


eeeebbbbrrrr:
I think you're probably on the right track here.  I suppose we'd want to also have conversions from the `chrono` types into the Postgres types.

I don't know how far @workingjubilee might want to take this in terms of also supporting `IntoDatum`/`FromDatum` for the `chrono` types (not clear to which SQL type they'd map to), so that's an open question...

workingjubilee:
Yeah, this seems like the right direction.

I don't think we want to support IntoDatum/FromDatum for these types. I would rather move towards making it easier to distinguish when something is a "zero-overhead" or "direct" conversion that is natively supported by Postgres, and when something requires additional logic like these types always will.

t3hmrman:
Thanks for the check @workingjubilee @eeeebbbbrrrr and for the heads up -- just to confirm I do plan to add the conversions both ways.

Will finish up the rest of it in this pattern and turn this to ready for review hopefully soon!

t3hmrman:
Hey @workingjubilee @eeeebbbbrrrr just finished the first version of this, would love a review whenever you can find some time!

I've tried to be pretty explicit about where the microseconds-only limitation of pg comes into play. I'm also equally happy to put this functionality in some sort of external crate if you'd prefer to not upstream this functionality as well!

daamien:
Hi !

Is there's a way for me to help moving this PR forward ? It would be very useful in my own extension...

t3hmrman:
Rebased just to make sure that's not blocking progress here :)

workingjubilee:
Sorry, all my focus was being consumed by https://github.com/pgcentralfoundation/pgrx/pull/1731 since that was one of our last release blockers so reviewing anything that wasn't incredibly trivial was backburnered.

pgrx/src/datum/datetime_support/chrono.rs
```diff
@@ -0,0 +1,194 @@
...
+            // Since pgrx counts the fractional seconds (between 1_000_000 and 2_000_000),
+            // the microseconds value will be 1_000_000 * seconds + fractional.
+            //
+            // - at 12:01:01 => hour=12, minute=1, second=2, microseconds=2_000_000
+            // - at 12:01:59 => hour=12, minute=1, second=59, microseconds=59_000_000
+            // - at 12:01:59 => hour=12, minute=1, second=59, microseconds=59_000_000
+            //
+            // Since chrono *does* support leap seconds (representing them as 59 seconds & >1_000_000 microseconds)
+            // we can strip the microseconds in that case to zero and pretend they're not there,
+            // since Postgres does not support leap seconds
```
workingjubilee, line 73: Thank you for documenting these little decisions.

pgrx-tests/src/tests/chrono_tests.rs
```diff
@@ -0,0 +1,75 @@
...
+        let original = pgrx::TimestampWithTimeZone::with_timezone(1970, 1, 1, 1, 1, 1.0, "utc")?;
+        let d = chrono::DateTime::<Utc>::try_from(original)?;
+        assert_eq!(d.hour(), 1, "hours match");
+        assert_eq!(d.minute(), 1, "minutes match");
+        assert_eq!(d.second(), 1, "seconds match");
+        assert_eq!(d.nanosecond(), 0, "nanoseconds are zero (pg only supports microseconds)");
+        let backwards = pgrx::TimestampWithTimeZone::try_from(d)?;
+        assert_eq!(backwards, original);
+        Ok(())
+    }
```
workingjubilee, line 74: Can you, _in addition_ to these simpler tests, set up a proptest like in https://github.com/pgcentralfoundation/pgrx/blob/77a28242fd89372a6fd7cc4ded1fcae9a63319f8/pgrx-tests/src/tests/proptests.rs so that these conversions are fairly invariant?

It _doesn't_ have to roundtrip stuff through SPI, since that's not really what we'd be testing, so it doesn't need to use the exact pattern I used there. It can do something simpler with just handling https://github.com/pgcentralfoundation/pgrx/blob/77a28242fd89372a6fd7cc4ded1fcae9a63319f8/pgrx-tests/src/proptest.rs

workingjubilee (changes requested):
Sorry for letting this sit for a while. It looks good and I think my only request is for property testing. It's been very good at finding edge-cases in our datetime handling in particular.

t3hmrman:
@workingjubilee absolutely no problem -- thanks for taking a look (P.S. I saw the [other thing you were working on :+1: :+1: :+1:](https://github.com/pgcentralfoundation/pgrx/pull/1731)).

Will add the additional tests!

pgrx-tests/src/tests/chrono_tests.rs
t3hmrman: Absolutely -- yeah I didn't think the roundtripping was necessary, but a prop test would definitely make a ton of sense here.

- [ ] Add property test for conversions

pgrx/src/datum/datetime_support/chrono.rs
t3hmrman: Absolutely -- did this mostly so that people would at least have a better chance of finding my bugs if there are some hiding in here.

Time is one of the places where programming goes off the rails, I'm just *hoping* I mostly got it right.

LucaCappelletti94:
Hi - any particular reason for closing down this PR?

PR #1610, opened 2024-03-14
Break Time on the anvil of proptest

### LINKED ISSUE #1525: Proptest datetime types like Date
In https://github.com/pgcentralfoundation/pgrx/pull/1258 I added support for running property tests inside Postgres, including testing for Date. I then broke API on Date, because the test showed that the `From<i32>` implementation is the wrong semantics for what Postgres will actually accept. This same testing, and the consequent API breaks, should be landed for:
- [x] Timestamp
- [x] TimestampWithTimeZone
- [x] Time
- [x] TimeWithTimeZone

workingjubilee:
Our handling of the various datetime types is extremely dubious. In particular, we `impl From` for these types to allow a conversion from their "raw repr", the underlying integer type. This was justified on the idea that this is how Postgres reasons about them anyways, and at worst it was an application bug. However, there are two problems:

- The conversions are not infallible, as `From` requires[^1]! At various points of the conversions, they may `ereport` (which we convert to a `panic!`, naturally).
- For many possible integer values, they return **non-equal** values when passed through the native Postgres (de)serialization functions and converted back to the original type, which we should not accept as a reasonable outcome. It permits data that flows through pgrx extensions to become corrupt or invalid by the time it reaches the database, causing errors when read out again. In this sense, it is another way the conversion is fallible, except the realization of the error may become delayed for hours or weeks after the data is actually committed to the database, as it is only upon (de)serialization that it is guaranteed to error.

Accordingly, this PR breaks the `From` impls for Time, TimeTz, and Timestamp, for their integer representations. The implementations are replaced by convenience functions that _do_ implement a possible valid interpretation of conversion from the integer types, but are lossy, and thus inappropriate for `impl From`. Conversion to-and-from these integer types is no longer lossless, and thus it is not "perfect".

The bounds selected for the conversion functions reflect the bounds that Postgres implements for these types.

Fixes https://github.com/pgcentralfoundation/pgrx/issues/1525

[^1]: https://doc.rust-lang.org/std/convert/trait.From.html


workingjubilee:
This is missing TimeTz, which can be added in a followup.

cc https://github.com/pgcentralfoundation/pgrx/issues/1525

workingjubilee:
Oh, adding TimeTz testing was easier than I thought since the relevant From impl is just a tuple. Same issues tho'.

workingjubilee:
`Time` and `TimeTz` should be `modular_from_raw`, I think.

workingjubilee:
@t3hmrman These changes may affect your implementation work. Sorry about any conflicts.

t3hmrman:
Thanks for the heads up @workingjubilee !

pgrx/src/datum/time.rs
```diff
@@ -68,6 +56,22 @@ impl From<TimeWithTimeZone> for Time {
...
+/// It is incorrect to convert a raw i64 to Time, due to out-of-bounds values
+/// resulting in severe logic errors, including database crashes, if used.
+impl TryFrom<pg_sys::TimeADT> for Time {
+    type Error = pg_sys::TimeADT;
+
+    #[inline]
+    fn try_from(raw: pg_sys::TimeADT) -> Result<Self, Self::Error> {
+        match raw {
+            0..=86400000 => Ok(Time(raw)),
+            i64::MIN..=-1 | 86400001.. => Err(raw),
```
workingjubilee:
```suggestion
    fn try_from(raw: pg_sys::TimeADT) -> Result<Self, Self::Error> {
        const MORE_THAN_A_DAY: i64 = MICROSECONDS_PER_DAY + 1;
        match raw {
            0..=MICROSECONDS_PER_DAY => Ok(Time(raw)),
            i64::MIN..=-1 | MORE_THAN_A_DAY.. => Err(raw),
```

pgrx/src/datum/time.rs
```diff
@@ -24,18 +24,6 @@ use std::num::TryFromIntError;
 #[repr(transparent)]
 pub struct Time(pg_sys::TimeADT);
```
workingjubilee, line 25:
```suggestion
pub struct Time(pg_sys::TimeADT);

const MICROSECONDS_PER_DAY: pg_sys::TimeADT = 86_400_000;
```

pgrx/src/datum/time.rs
```diff
@@ -153,6 +157,10 @@ impl Time {
         }
     }
 
+    pub fn modular_from_raw(time: i64) -> Self {
+        Self(time.rem_euclid(86_400_000))
```
workingjubilee:
```suggestion
        Self(time.rem_euclid(MICROSECONDS_PER_DAY))
```

pgrx/src/datum/time_with_timezone.rs
```diff
@@ -25,17 +25,6 @@ use std::panic::{RefUnwindSafe, UnwindSafe};
 #[repr(transparent)]
 pub struct TimeWithTimeZone(pg_sys::TimeTzADT);
```
workingjubilee:
```suggestion
const MICROSECONDS_PER_DAY: pg_sys::TimeADT = 86_400_000;
```

pgrx/src/datum/time_with_timezone.rs
```diff
@@ -147,6 +145,13 @@ impl TimeWithTimeZone {
         }
     }
 
+    pub fn modular_from_raw(tuple: (i64, i32)) -> Self {
+        let time = tuple.0.rem_euclid(86_400_000);
```
workingjubilee:
```suggestion
        let time = tuple.0.rem_euclid(MICROSECONDS_PER_DAY);
```

pgrx/src/datum/time_stamp.rs
```diff
@@ -72,9 +72,16 @@ impl From<Timestamp> for pg_sys::Timestamp {
...
-        Timestamp(value)
+impl TryFrom<pg_sys::Timestamp> for Timestamp {
+    type Error = pg_sys::Timestamp;
+    #[inline]
+    fn try_from(ts: pg_sys::Timestamp) -> Result<Self, Self::Error> {
+        //  #define IS_VALID_TIMESTAMP(t)  (MIN_TIMESTAMP <= (t) && (t) < END_TIMESTAMP)
+        const TS_MAX_INCLUSIVE: pg_sys::Timestamp = 9223371331200000000 - 1;
+        match ts {
+            i64::MIN | i64::MAX | -211813488000000000..=TS_MAX_INCLUSIVE => Ok(Timestamp(ts)),
+            _ => Err(ts),
```
workingjubilee:
```suggestion
        // defined by Postgres, but bindgen doesn't evaluate them for some reason:
        const MIN_TIMESTAMP: pg_sys::Timestamp = -211_813_488_000_000_000;
        const END_TIMESTAMP: pg_sys::Timestamp = 9_223_371_331_200_000_000;
        //  #define IS_VALID_TIMESTAMP(t)  (MIN_TIMESTAMP <= (t) && (t) < END_TIMESTAMP)
        const MAX_TIMESTAMP: pg_sys::Timestamp = TIMESTAMP_END - 1;
        match ts {
            i64::MIN | i64::MAX | MIN_TIMESTAMP..=MAX_TIMESTAMP => Ok(Timestamp(ts)),
            _ => Err(ts),
```

pgrx-tests/src/tests/proptests.rs
```diff
@@ -1,72 +1,103 @@
...
-                prop_assert_eq!(date, spi_ret.unwrap());
-                Ok(())
-            })
-            .unwrap();
     }
 }
+
+pg_proptest_datetime_types! {
+    Date = prop::num::i32::ANY.prop_map(Date::saturating_from_raw);
+    // 00:00..=24:00
```
workingjubilee:
```suggestion
```

pgrx-tests/src/tests/proptests.rs
```diff
@@ -1,72 +1,103 @@
 use crate::proptest::PgTestRunner;
+use core::ffi;
+use paste::paste;
 use pgrx::prelude::*;
 use proptest::prelude::*;
+use TimeWithTimeZone as TimeTz;
+
+macro_rules! pg_proptest_datetime_roundtrip_tests {
```
workingjubilee, line 14:
```suggestion
/// Generate the roundtrip property tests for a datetime type
///
/// This macro relies on the fact Postgres has a very regular naming format
/// for serialization/deserialization functions for most types, always using
/// - `type_in`
/// - `type_out`
macro_rules! pg_proptest_datetime_roundtrip_tests {
```

pgrx/src/datum/time.rs
```diff
@@ -68,6 +58,23 @@ impl From<TimeWithTimeZone> for Time {
     }
 }
 
+/// TimeADT (i64) to Time conversion
```
workingjubilee:
```suggestion
/// [`pg_sys::TimeADT`] (i64) to Time conversion
```

pgrx-tests/src/tests/proptests.rs
```diff
@@ -1,72 +1,103 @@
...
+
+paste! {
+// A property test consists of
+// 1. Posing a hypothesis
+#[doc = concat!("A value of ", stringify!($datetime_ty), "should be able to be passed to Postgres and back.")]
+#[pg_test]
+pub fn [<$datetime_ty:lower _spi_roundtrip>] () {
+    // 2. Constructing the Postgres-adapted test runner
+    let mut proptest = PgTestRunner::default();
+    // 3. A strategy to create and refining values, which is a somewhat aggrandized function.
```
workingjubilee:
```suggestion
    // 3. A strategy to create and refine values, which is a somewhat aggrandized function.
```

pgrx-tests/src/tests/proptests.rs
```diff
@@ -1,72 +1,103 @@
...
+    //    In some cases it actually can be replaced directly by a closure, or, in this case,
+    //    it involves using a closure to `prop_map` an existing Strategy for producing
+    //    "any kind of i32" into "any kind of in-range value for a Date".
+    let strat = $prop_strat;
+    proptest
```
workingjubilee, line 31:
```suggestion
    // 4. The runner invocation
    proptest
```

pgrx/src/datum/time_stamp.rs
```diff
@@ -72,9 +72,19 @@ impl From<Timestamp> for pg_sys::Timestamp {
...
-        Timestamp(value)
+impl TryFrom<pg_sys::Timestamp> for Timestamp {
+    type Error = pg_sys::Timestamp;
+    #[inline]
+    fn try_from(ts: pg_sys::Timestamp) -> Result<Self, Self::Error> {
+        // defined by Postgres, but bindgen doesn't evaluate them for some reason:
+        const MIN_TIMESTAMP: pg_sys::Timestamp = -211_813_488_000_000_000;
+        const END_TIMESTAMP: pg_sys::Timestamp = 9_223_371_331_200_000_000;
+        //  #define IS_VALID_TIMESTAMP(t)  (MIN_TIMESTAMP <= (t) && (t) < END_TIMESTAMP)
+        const MAX_TIMESTAMP: pg_sys::Timestamp = TIMESTAMP_END - 1;
```
workingjubilee:
```suggestion
        const MAX_TIMESTAMP: pg_sys::Timestamp = END_TIMESTAMP - 1;
```

NotGyro (approved):
Looks good to me.

PR #1620, opened 2024-03-21
Consistently test on latest rustc stable

workingjubilee:
Consistently use latest rustc stable for the test infrastructure by replacing a number of ad hoc lines of bash-in-YAML with a single script. This allows us to use rustc 1.77, for which:
- `offset_of!` in stdlib lets us drop the `memoffset` crate
- some compile-fail tests require blessing, as diagnostics changed

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -54,10 +54,6 @@ jobs:
           cat $GITHUB_ENV
           echo ""
           
-          echo "----- Output Cargo version -----"
-          cargo --version
-          echo ""
-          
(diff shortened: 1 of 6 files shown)


workingjubilee:
I guess it doesn't realize it's rust 1.77 yet.

workingjubilee:
clearing caches.

workingjubilee:
This became an "update CI in general" PR, but whatever.

We now use a single consistent rustc configuration across all test suites, so we can expect everything to test using the same version now.

eeeebbbbrrrr:
Cool

PR #1621, opened 2024-03-22
Use Ceescape's `sigsetjmp()` wrapper

### LINKED ISSUE #1144: Use asm-based shim for sigsetjmp
See https://github.com/rust-lang/unsafe-code-guidelines/issues/404#issuecomment-1560245592 for a proof of concept. This should make our setjmp handling less UB (without requiring unstable features), at the cost of needing some inline asm.

I'd have to write one for aarch64 too, and we'd use the current approach as a fallback for targets without a shim.

NotGyro:
The [cee-scape](https://crates.io/crates/cee-scape) crate uses a hand-rolled assembly wrapper around libc's `sigsetjmp()`. This has better correctness guarantees than calling `sigsetjmp()` directly, because LLVM is not allowed to reason about / attempt to optimize inside hand-written assembly. `sigsetjmp()` is used for `pg_guard_ffi_boundary_impl()`, which is our wrapper around Postgres' exception handling, since Postgres uses `sigsetjmp()` and `siglongjmp()` to handle multiple return points. Rust's paradigm doesn't understand this - in Rust, a a function call can only return to its caller at one point in time: when it has finished executing. Whereas, sigsetjmp can return to two places _chronologically_ - and so this needs to be handled carefully.

Fixes https://github.com/pgcentralfoundation/pgrx/issues/1144

### DIFF
--- a/pgrx-pg-sys/src/submodules/ffi.rs
+++ b/pgrx-pg-sys/src/submodules/ffi.rs
@@ -9,6 +9,8 @@
 //LICENSE Use of this source code is governed by the MIT license that can be found in the LICENSE file.
 #![deny(unsafe_op_in_unsafe_fn)]
 
+use cee_scape::SigJmpBufFields;
+
 /**
 Given a closure that is assumed to be a wrapped Postgres `extern "C"` function, [pg_guard_ffi_boundary]
 works with the Postgres and C runtimes to create a "barrier" that allows Rust to catch Postgres errors
@@ -86,6 +88,7 @@ being converted into a transaction-aborting Postgres `ERROR` by PGRX.
 use crate as pg_sys;
 use crate::panic::{CaughtError, ErrorReport, ErrorReportLocation, ErrorReportWithLevel};
 use core::ffi::CStr;
+use std::mem::MaybeUninit;
 
 #[inline(always)]
 #[track_caller]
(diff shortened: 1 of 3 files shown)


eeeebbbbrrrr:
What I worry about here is cee-scape’s correctness. 

I don’t know if our CI passing all tests is a robust enough exercise. I also don’t have helpful suggestions to really ensure correctness. 

Don’t let this stand in the way of approval. Just wanted to voice a concern.

NotGyro:
> What I worry about here is cee-scape’s correctness.
> 
> I don’t know if our CI passing all tests is a robust enough exercise. I also don’t have a helpful suggestions to really sure correctness.
> 
> Don’t let this stand in the way of approval. Just wanted to voice a concern.

Yeah, that's reasonable. I have similar worries. Unfortunately, I'm not trained in the theory and practice of formal verification, so I can't guarantee this is correct.

Would it be possible to introduce tests that would more thoroughly check the validity of our sigsetjmp usage?

workingjubilee:
in practice all of our "try/catch" suite does that.

eeeebbbbrrrr:
> in practice all of our "try/catch" suite does that.

Aren’t we also assuming this prevents possible compiler optimizations?

eeeebbbbrrrr:
On the flip side, I’m super jazzed this isn’t an invasive change!  A++

workingjubilee:
> > in practice all of our "try/catch" suite does that.
> 
> Aren’t we also assuming this prevents possible compiler optimizations?

Yes? The fact that rustc cannot elide an asm block that control flow passes through, and must execute it, without reasoning about its contents except where annotated (`pure` and the like) is a documented property of the Rust abstract machine:

https://doc.rust-lang.org/reference/inline-assembly.html?highlight=black%20box#rules-for-inline-assembly

This is heavily relied on for allowing the Rust abstract machine to ignore machine-specific details in many cases and instead implementing those in assembly.

workingjubilee:
...looks like the macOS machines are using rustc 1.77 and the Linux ones are using rustc 1.76...

eeeebbbbrrrr:
Okay. That closes the primary concern I had and it’s good to have that linked here. 

Then, I guess the only concern is just “what if cee-scape is bugged?”  But that’s not really any different than any other dependency we have. And the upstream folks seem receptive to patches.

ship it!

workingjubilee:
@NotGyro can you update the PR and/or commit message with that link and adjusted appropriately? i.e. that `sigsetjmp` is called via the assembly, and the assembly is opaque.

workingjubilee:
The tests should pass after rebase.

pgrx-pg-sys/src/submodules/ffi.rs
```diff
@@ -109,23 +112,27 @@ unsafe fn pg_guard_ffi_boundary_impl<T, F: FnOnce() -> T>(f: F) -> T {
...
+            pg_sys::PG_exception_stack = std::mem::transmute(jump_buffer as *const SigJmpBufFields);
 
             // execute the closure, which will be a wrapped internal Postgres function
-            let result = f();
+            result.write(f());
+            0
+        });
 
-            // restore Postgres' understanding of where its next longjmp should go
+        if jump_value == 0 {
```
workingjubilee, line 126: I wonder if we'll need to tweak the scope this includes... maybe. This should be good for now, though.

workingjubilee (approved):
Ultimately if needed, we can always fork/patch or even vendor cee-scape. Having this wired in means we can actually own whatever charming fuckups in control flow ensue, instead of being at the mercy of whether the compiler likes us that day or not.

tucnak:
This seems to have broken builds on ppc64el which is an architecture supported by Postgres and Rust. Is there a mitigation?

Related #1897

PR #1629, opened 2024-04-02
Redeclare additional *_walker fn

daamien:
As a follow up to the discussion here 
https://github.com/pgcentralfoundation/pgrx/issues/1583#issuecomment-2030105513

Here's a patch to declare correctly the following functions:

* query_or_expression_tree_walker
* range_table_entry_walker
* planstate_tree_walker
* range_table_walker

This is a shameless copy of PR #1596 , I'm not entirely sure what I am actually doing here :)

Quick questions regarding the format:

-  I see some arguments have the `arg_` prefix while others don't... Like  `arg_node` vs `node`  in the query_tree_walker function... Is there a naming rule regarding this ?

- I'm unsure how to sort the functions... alphabetical order ?

### DIFF
--- a/pgrx-pg-sys/build.rs
+++ b/pgrx-pg-sys/build.rs
@@ -741,7 +741,8 @@ fn add_blocklists(bind: bindgen::Builder) -> bindgen::Builder {
     bind.blocklist_type("Datum") // manually wrapping datum for correctness
         .blocklist_type("Oid") // "Oid" is not just any u32
         .blocklist_function("varsize_any") // pgrx converts the VARSIZE_ANY macro, so we don't want to also have this function, which is in heaptuple.c
-        .blocklist_function("(?:raw_)?(?:expression|planstate|query|query_or_expression)_tree_walker")
+        .blocklist_function("(?:raw_)?(?:expression|query|query_or_expression)_tree_walker")
+        .blocklist_function("planstate_tree_walker")
         .blocklist_function("range_table_(?:entry_)?walker")
         .blocklist_function(".*(?:set|long)jmp")
         .blocklist_function("pg_re_throw")
--- a/pgrx-pg-sys/build.rs
+++ b/pgrx-pg-sys/build.rs
@@ -741,9 +741,7 @@ fn add_blocklists(bind: bindgen::Builder) -> bindgen::Builder {
     bind.blocklist_type("Datum") // manually wrapping datum for correctness
         .blocklist_type("Oid") // "Oid" is not just any u32
         .blocklist_function("varsize_any") // pgrx converts the VARSIZE_ANY macro, so we don't want to also have this function, which is in heaptuple.c
-        .blocklist_function("(?:raw_)?(?:query|expression)_tree_walker")
-        .blocklist_function("query_or_expression_tree_walker")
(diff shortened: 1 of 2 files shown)


workingjubilee:
>    I see some arguments have the arg_ prefix while others don't... Like arg_node vs node in the query_tree_walker function... Is there a naming rule regarding this ?

We should generally copy the names from the C source except when they deviate from Rust convention. I'm not sure why `raw_tree_expression_walker` is like that, it's a deviation from that rule. Probably should be fixed.

>    I'm unsure how to sort the functions... alphabetical order ?

That will do fine for now.

pgrx-pg-sys/build.rs
```diff
@@ -742,6 +742,9 @@ fn add_blocklists(bind: bindgen::Builder) -> bindgen::Builder {
         .blocklist_type("Oid") // "Oid" is not just any u32
         .blocklist_function("varsize_any") // pgrx converts the VARSIZE_ANY macro, so we don't want to also have this function, which is in heaptuple.c
         .blocklist_function("(?:raw_)?(?:query|expression)_tree_walker")
+        .blocklist_function("query_or_expression_tree_walker")
+        .blocklist_function("planstate_tree_walker")
+        .blocklist_function("range_table_(?:entry_)?walker")
```
workingjubilee: So I wouldn't mind this being two or three lines but four is out:

```suggestion
        .blocklist_function("(?:raw_)?(?:expression|planstate|query|query_or_expression|range_table)(?:_entry)?_tree_walker")
```

workingjubilee (dismissed):
Would you be so kind as to make sure the arg names line up with the relevant `_impl` functions? (including in previously-added fn)

daamien:
> Would you be so kind as to make sure the arg names line up with the relevant `_impl` functions? (including in previously-added fn)

Sure !

Sorry I misunderstood your previous answer..... I will rename the arguments...

workingjubilee:
I think all the ones you added here are correct?

pgrx-pg-sys/build.rs
daamien: Disclaimer: I hate regular expressions as much as they hate me... So I'm always inclined to write a bunch of simple regexps rather than a single sophisticated one. 

But I get your point of keeping the blocklist small... 

However, if i'm correct your regexp would block a lot of false positives such as  `raw_planstate_entry_tree_walker`. There's little chance such function would ever be implemented in Postgres, but by principle shouldn't we match the strict minimum list of possibilities ?

Also it doesn't catch `range_table_entry_walker`

How about those 3 lines ?

```rust
        .blocklist_function("(?:raw_)?(?:query|expression)_tree_walker")
        .blocklist_function("(?:planstate|query_or_expression)_tree_walker")
        .blocklist_function("range_table_(?:entry_)?walker")
```

workingjubilee: Ah, I didn't notice `range_table_entry_walker` didn't have the terminal `tree`.

That's fine.

workingjubilee:
```suggestion
        .blocklist_function("(?:raw_)?(?:expression|planstate|query|query_or_expression)_tree_walker")
        .blocklist_function("range_table_(?:entry_)?walker")
```

PR #1633, opened 2024-04-04
Add a docker image for CI

daamien:
Following a discussion on the discord channel:

Here's a basic docker image with all the requirements to build a PGRX project in a CI job. This saves a lot of CI minutes, especially the `cargo pgrx init` step which is quite long.

By default, the image is based on the latest version available but it can be overriden with `--build-arg PGRX_VERSION=0.11.3`

This could be completed with a CI action to build and push the image automatically on dockerhub, like this

https://github.com/daamien/docker-pgrx/blob/latest/.github/workflows/build-push.yaml

This would require to add [secrets variables in the Github account] for `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN`

[secrets variables in the Github account]: https://docs.github.com/fr/actions/security-guides/using-secrets-in-github-actions

This would also provide an easy way to test against alpha/beta versions by simply pulling any version like this

docker pull pgcentralfoundation/pgrx:0.12.0-alpha.1

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -1,0 +1,22 @@
+FROM rust:1
+
+ARG PGRX_VERSION=0.12.0-alpha.1
+
+RUN apt-get update && apt-get install -y \
+    gettext-base \
+    libclang-dev \
+    postgresql-server-dev-all
+
+USER postgres
+
+ENV USER=postgres
+
+ENV PATH="${PATH}:/usr/local/cargo/bin/:~postgres/.cargo/bin"
+
+RUN rustup component add clippy && \
+    cargo install --locked --version "${PGRX_VERSION}" cargo-pgrx && \
+    cargo pgrx init
+
+WORKDIR /pgrx
+
+VOLUME /pgrx


workingjubilee:
@daamien Could you add a workflow in that actually would use this? It's not enough to just put it in the CI folder, gotta actually make it part of the tests.

workingjubilee:
And I don't mean a workflow to put it on DockerHub or whatever, I mean something that would use it to run the test suite on top of it.

thomcc:
@daamien We don't run the slow `cargo pgrx init` in CI. We use a prebuilt PG, which means the cargo pgrx init step is very fast: image

The downside of using a prebuilt postgres is that it doesn't have debug-asserts enabled. In practice this is kind of fine but is a real deficiency in our CI flow. Unfortunately fixing this requires (as you mentioned) uploading the container to something like dockerhub or (more likely for our case, for trust reasons) GHCR. All this isn't trivial.

workingjubilee:
does it *require* the upload first? 😵‍💫 gah.

daamien:
Thanks for the feedback !

What is the problem with uploading the image to a registry ? The image will be quite big ( roughly 1,5Go ) but the upload would happen only when a tag is pushed on the PGRX repo. So once a month, or so. That seems acceptable.  Anyway I think it's also possible to let dockerhub build the image (via a webhook) and in that case there's no upload at all....

Regarding the examples of CI configuration, I work mostly on gitlab.com and I have some custom jobs that wouldn't apply for a general usage, but I guess that the job below would be a good example to start with: 

```yaml
pgrx:test:
  stage: test
  image: pgcentralfoundation/pgrx:0.11.3
  script:
    - cargo pgrx test $PGVER --verbose
  parallel:
    matrix:
      - PGVER: pg12
      - PGVER: pg13
      - PGVER: pg14
      - PGVER: pg15
      - PGVER: pg16
````
I'm less familiar with github actions, but it would look like this 

```yaml
name: PGRX tests
on: [push]
jobs:
  test:
    strategy:
      matrix:
        pgver: [pg12, pg13, pg14, pg15, pg16 ]
    name: PGRX tests on ${{ matrix.pgver }}
    runs-on: ubuntu-latest
    container:  pgcentralfoundation/pgrx:0.11.3
    steps:
      - name: Check out the repo
        uses: actions/checkout@v4
      - name: Test on  ${{ matrix.pgver }}
        run: cargo pgrx test ${{ matrix.pgver }} --verbose
```

theory:
Unfortunately I don't think this image will work for GitHub actions due to /actions/checkout#956: You can't checkout a repo unless you run as `root` in the image, and you can't run Postgres as `root`. @eeeebbbbrrrr is working on a workaround to allow pgrx to run Postgres as a different system user. [Discussion thread](https://discordapp.com/channels/710918545906597938/1228127048770261062).

theory:
There's also [this bit of documentation](https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions#user):

> Docker actions must be run by the default Docker user (root). Do not use the `USER` instruction in your `Dockerfile`, because you won't be able to access the `GITHUB_WORKSPACE` directory. For more information, see "[Variables](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)" and [USER reference](https://docs.docker.com/engine/reference/builder/#user) in the Docker documentation.

Ultimately we're going to have to create a GitHub action in JavaScript I think. :-(

theory:
With the new `--runas` and `--pgdata` optoins in v0.11.3, I've updated [pgxn-tools](/pgxn/docker-pgxn-tools) with a new command, `pgrx-build-test`, does the full build/test/install/installcheck dance (`installcheck` only if it appears to be in a `Makefile`. I'm now using it in the [pg-jsonschema-boon CI workflow](https://github.com/tembo-io/pg-jsonschema-boon/blob/main/.github/workflows/lint-and-test.yml) ([sample run](https://github.com/tembo-io/pg-jsonschema-boon/actions/runs/8809394356)).

PR #1634, opened 2024-04-04
fix: use `memoffset` for `offset_of`

rebasedming:
As of Rust 1.77.0, `offset_of` is still unstable, and compiling `pgrx` fails for us because of this. We can use `memoffset::offset_of` as a workaround.


workingjubilee:
@rebasedming Huh? This code compiles fine with rustc 1.77.0, what are you talking about?

rebasedming:
Woops, I was on an old version of Rust, offset_of became stable in 1.77.0. I'll close this PR, thanks!

Issue #1624, opened 2024-03-27
Time-type changes break tests on future rustcs


workingjubilee:
Original PR:
- https://github.com/pgcentralfoundation/pgrx/pull/1610

Breaking `rustc --version`s:
- rustc 1.78.0-beta.2 (277d06bc9 2024-03-23)
- rustc 1.79.0-nightly (5f2c7d2bf 2024-03-25)

Broken command:
```shell
#!/bin/sh
for rustver in beta nightly; do
    rustup default "$rustver"
    cargo clean
    cargo install --path cargo-pgrx --locked
    for pgver in 12 13 14 15 16; do
        cargo test -p pgrx-tests \
            --features "pg$pgver" --no-default-features \
            1>"$rustver-pg$pgver.stdout" 2>"$rustver-pg$pgver.stderr";
    done
done
```

Output differs between versions, not sure if on different commits yet, I found the discrepancy first on nightly but did the beta bisect first.

workingjubilee:
I have confirmed the nightly breakage predates 0.12.0-alpha.0, it seems. I suspect they're a rustc problem, but the beta problem seems like ours.

workingjubilee:
Output for beta is:

```
test tests::zero_datum_edge_cases::tests::pg_test_zero_i32_is_some_zero ... ok
test tests::array_tests::tests::pg_test_arr_sort_uniq_with_null - should panic ... ok
test tests::attributes_tests::tests::pg_test_for_should_panic_attribute - should panic ... ok
test tests::fcinfo_tests::tests::pg_test_null_error_type - should panic ... ok
test tests::fn_call_tests::tests::pg_fn_raises_error - should panic ... ok
test tests::shmem_tests::tests::pg_test_behaves_normally_when_elog_while_holding_lock - should panic ... ok
test tests::roundtrip_tests::tests::pg_test_rt_array_time ... FAILED
test tests::struct_type_tests::tests::pg_test_complex_storage_and_retrieval ... FAILED

failures:

---- tests::roundtrip_tests::tests::pg_test_rt_array_time stdout ----
thread 'tests::roundtrip_tests::tests::pg_test_rt_array_time' panicked at pgrx-tests/src/framework.rs:164:9:


Postgres Messages:
[...]

PR #1636, opened 2024-04-04
Fix cargo pgrx finding cdylibs

workingjubilee:
Required for, but does not fix on its own, https://github.com/pgcentralfoundation/pgrx/issues/1624


workingjubilee:
I'd like to demonstrate the necessity of this changeset by enabling test workflows on beta but if I enabled tests on beta they would break so...

PR #1643, opened 2024-04-09
only build library in cargo-pgrx commands

usamoi:
Only build the package’s library in `cargo-pgrx` commands since only `dylib` is needed.


workingjubilee:
Sure, thanks!

PR #1652, opened 2024-04-12
`cargo pgrx test --runas`

eeeebbbbrrrr:
```shell
$ cargo pgrx test pg14 --runas testuser --pgdata=/tmp/pgrxtest
```

This adds two new arguments to `cargo pgrx test`: `--runas <username>` and `--pgdata </writable/path>`

Together, these will initdb, createdb, dropdb, and run postgres as the "runas" user.

Extension artifacts are copied with `cargo pgrx install --sudo`, which will also require the current user be able to become root.

In general, these options are designed for the unusual situation where one needs to be "root" in order to run extension tests, but, of course, Postgres is not allowed to be root.

### DIFF
--- a/.github/workflows/runas.yml
+++ b/.github/workflows/runas.yml
@@ -24,5 +24,5 @@ jobs:
       - name: Report version
         run: cargo --version
 
-      - name: Test pronto-as-a-library
+      - name: Test cargo pgrx test --runas
         run: cd pgrx-examples/arrays && cargo pgrx test --runas postgres --pgdata=/tmp/pgdata
--- a/.github/workflows/runas.yml
+++ b/.github/workflows/runas.yml
@@ -24,5 +24,8 @@ jobs:
       - name: Report version
         run: cargo --version
 
+      - name: Install cargo pgrx
+        run: cd cargo-pgrx && cargo install --path . --debug
+
       - name: Test cargo pgrx test --runas
         run: cd pgrx-examples/arrays && cargo pgrx test --runas postgres --pgdata=/tmp/pgdata
--- a/.github/workflows/runas.yml
+++ b/.github/workflows/runas.yml
@@ -1,4 +1,4 @@
-name: Test cargo pgrx test --runas
+name: cargo pgrx test --runas
 
 on:
   push:
--- a/.github/workflows/runas.yml
+++ b/.github/workflows/runas.yml
@@ -1,0 +1,28 @@
+name: Test cargo pgrx test --runas
+
+on:
+  push:
+    branches: [ develop ]
+  pull_request:
+    branches: [ develop ]
+
+env:
+  CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: 1
+  RUSTFLAGS: "-C link-args=-Wl,-undefined,dynamic_lookup"   # we don't want target-cpu=native from `.cargo/config` b/c we're caching binaries but we want the other stuff
+
+jobs:
+  ubuntu:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: Swatinem/rust-cache@v2
+        with:
+          prefix-key: "v1-rust"
+
+      - name: Report version
+        run: cargo --version
+
+      - name: Test pronto-as-a-library
+        run: cd pgrx-examples/arrays && cargo pgrx test --runas postgres --pgdata=/tmp/pgdata
--- a/.github/workflows/runas.yml
+++ b/.github/workflows/runas.yml
@@ -23,8 +23,8 @@ jobs:
 
       - name: Install Postgres deps
         run: |
-          apt-get update -y -qq --fix-missing
-          apt-get install -y postgresql-server-dev-14
+          sudo apt-get update -y -qq --fix-missing
+          sudo apt-get install -y postgresql-server-dev-14
 
       - name: Report version
         run: cargo --version
--- a/.github/workflows/runas.yml
+++ b/.github/workflows/runas.yml
@@ -9,7 +9,6 @@ on:
 env:
   CARGO_TERM_COLOR: always
   RUST_BACKTRACE: 1
-  RUSTFLAGS: "-C link-args=-Wl,-undefined,dynamic_lookup"   # we don't want target-cpu=native from `.cargo/config` b/c we're caching binaries but we want the other stuff
 
 jobs:
   ubuntu:
--- a/.github/workflows/runas.yml
+++ b/.github/workflows/runas.yml
@@ -27,5 +27,8 @@ jobs:
       - name: Install cargo pgrx
         run: cd cargo-pgrx && cargo install --path . --debug
 
+      - name: cargo pgrx init
+        run: cargo pgrx init --pg14=$(which pg_config)
+
       - name: Test cargo pgrx test --runas
-        run: cd pgrx-examples/arrays && cargo pgrx test --runas postgres --pgdata=/tmp/pgdata
+        run: cd pgrx-examples/arrays && cargo pgrx test pg14 --runas postgres --pgdata=/tmp/pgdata
--- a/.github/workflows/runas.yml
+++ b/.github/workflows/runas.yml
@@ -19,7 +19,12 @@ jobs:
       - uses: actions/checkout@v2
       - uses: Swatinem/rust-cache@v2
         with:
-          prefix-key: "v1-rust"
+          prefix-key: "v1-cargo-pgrx-test--runas"
+
+      - name: Install Postgres deps
+        run: |
+          apt-get update -y -qq --fix-missing
+          apt-get install -y postgresql-server-dev-14
 
       - name: Report version
         run: cargo --version
--- a/cargo-pgrx/README.md
+++ b/cargo-pgrx/README.md
@@ -494,8 +494,8 @@ make to the database are not preserved.
 
 An administrative note is that the `--runas` and `--pgdata` options can be used to control the operating-system user used
 to run the separate `postmaster` process for test execution.  Likely, if `--runas` is used, then `--pgdata` will also need
-to be set to a base directory that is readable and writable by that user -- the default PGDATA directory of `./target/` may
-not be permissive enough.
+to be set to a base directory that is readable and writable by that user -- the default PGDATA directory at `./target/pgrx-test-pgdata` 
+will have the permissions of the user running `cargo pgrx test` and won't be chown-able to the `--runas` user.
 
 ```console
 $ cargo pgrx test --help
--- a/cargo-pgrx/README.md
+++ b/cargo-pgrx/README.md
@@ -514,8 +514,8 @@ Options:
   -r, --release                        compile for release mode (default is debug)
       --profile <PROFILE>              Specific profile to use (conflicts with `--release`)
   -n, --no-schema                      Don't regenerate the schema
-      --runas <RUNAS>                  Use `sudo` to initialize and run the Postgres test instance as this system user
-      --pgdata <PGDATA>                Initialize the test database cluster here, instead of the default location.  If used with `--runas`, then it must be writable by the user
+      --runas <USER>                   Use `sudo` to initialize and run the Postgres test instance as this system user
+      --pgdata <DIR>                   Initialize the test database cluster here, instead of the default location.  If used with `--runas`, then it must be writable by the user
       --all-features                   Activate all available features
       --no-default-features            Do not activate the `default` feature
   -F, --features <FEATURES>            Space-separated list of features to activate
--- a/cargo-pgrx/README.md
+++ b/cargo-pgrx/README.md
@@ -492,6 +492,11 @@ Rust `#[test]` functions behave normally, while `#[pg_test]` functions are run *
 Additionally, a `#[pg_test]` function runs in a transaction that is aborted when the test is finished. As such, any changes it might
 make to the database are not preserved.
 
+An administrative note is that the `--runas` and `--pgdata` options can be used to control the operating-system user used
+to run the separate `postmaster` process for test execution.  Likely, if `--runas` is used, then `--pgdata` will also need
+to be set to a base directory that is readable and writable by that user -- the default PGDATA directory of `./target/` may
+not be permissive enough.
+
 ```console
 $ cargo pgrx test --help
 Run the test suite for this crate
@@ -509,6 +514,8 @@ Options:
   -r, --release                        compile for release mode (default is debug)
       --profile <PROFILE>              Specific profile to use (conflicts with `--release`)
   -n, --no-schema                      Don't regenerate the schema
+      --runas <RUNAS>                  Use `sudo` to initialize and run the Postgres test instance as this system user
+      --pgdata <PGDATA>                Initialize the test database cluster here, instead of the default location.  If used with `--runas`, then it must be writable by the user
       --all-features                   Activate all available features
       --no-default-features            Do not activate the `default` feature
   -F, --features <FEATURES>            Space-separated list of features to activate
--- a/cargo-pgrx/src/command/test.rs
+++ b/cargo-pgrx/src/command/test.rs
@@ -41,10 +41,10 @@ pub(crate) struct Test {
     #[clap(long, short)]
     no_schema: bool,
     /// Use `sudo` to initialize and run the Postgres test instance as this system user
-    #[clap(long)]
+    #[clap(long, value_name = "USER")]
     runas: Option<String>,
     /// Initialize the test database cluster here, instead of the default location.  If used with `--runas`, then it must be writable by the user
-    #[clap(long)]
+    #[clap(long, value_name = "DIR")]
     pgdata: Option<PathBuf>,
     #[clap(flatten)]
     features: clap_cargo::Features,
--- a/cargo-pgrx/src/command/test.rs
+++ b/cargo-pgrx/src/command/test.rs
@@ -40,6 +40,12 @@ pub(crate) struct Test {
     /// Don't regenerate the schema
     #[clap(long, short)]
     no_schema: bool,
+    /// Use `sudo` to initialize and run the Postgres test instance as this system user
+    #[clap(long)]
+    runas: Option<String>,
+    /// Initialize the test database cluster here, instead of the default location.  If used with `--runas`, then it must be writable by the user
+    #[clap(long)]
+    pgdata: Option<PathBuf>,
     #[clap(flatten)]
     features: clap_cargo::Features,
     #[clap(from_global, action = clap::ArgAction::Count)]
@@ -75,6 +81,8 @@ impl CommandExecute for Test {
                 me.no_schema,
                 &features,
                 me.testname,
+                me.runas,
+                me.pgdata,
             )?;
 
             Ok(())
@@ -115,6 +123,8 @@ pub fn test_extension(
     no_schema: bool,
     features: &clap_cargo::Features,
     testname: Option<impl AsRef<str>>,
+    runas: Option<String>,
+    pgdata: Option<PathBuf>,
 ) -> eyre::Result<()> {
     if let Some(ref testname) = testname {
         tracing::Span::current().record("testname", &tracing::field::display(&testname.as_ref()));
@@ -139,6 +149,14 @@ pub fn test_extension(
         .env("PGRX_ALL_FEATURES", if features.all_features { "true" } else { "false" })
         .env("PGRX_BUILD_PROFILE", profile.name())
         .env("PGRX_NO_SCHEMA", if no_schema { "true" } else { "false" });
+
+    if let Some(runas) = runas {
+        command.env("CARGO_PGRX_TEST_RUNAS", runas);
+    }
+
+    if let Some(pgdata) = pgdata {
+        command.env("CARGO_PGRX_TEST_PGDATA", pgdata);
+    }
 
     if let Ok(rust_log) = std::env::var("RUST_LOG") {
         command.env("RUST_LOG", rust_log);
--- a/pgrx-tests/src/framework.rs
+++ b/pgrx-tests/src/framework.rs
@@ -19,6 +19,7 @@ use pgrx_pg_config::{
 use postgres::error::DbError;
 use std::collections::HashMap;
 use std::env::VarError;
+use std::ffi::OsStr;
 use std::io::{BufRead, BufReader, Write};
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex, OnceLock};
@@ -397,10 +398,7 @@ fn maybe_make_pgdata<P: AsRef<Path>>(pgdata: P) -> eyre::Result<bool> {
         //
         // In order to do that, we must become that user to create it
 
-        let mut mkdir = Command::new("sudo");
-        mkdir
-            .arg("-u")
-            .arg(&runas)
+        let mut mkdir = sudo_command(&runas)
             .arg("mkdir")
             .arg("-p")
             .arg(pgdata)
@@ -444,8 +442,8 @@ fn initdb(postgresql_conf: Vec<&'static str>) -> eyre::Result<()> {
         let initdb_path = pg_config.initdb_path().wrap_err("unable to determine initdb path")?;
 
         let mut command = if let Some(runas) = get_runas() {
-            let mut cmd = Command::new("sudo");
-            cmd.arg("-u").arg(runas).arg(initdb_path);
+            let mut cmd = sudo_command(runas);
+            cmd.arg(initdb_path);
             cmd
         } else {
             Command::new(initdb_path)
@@ -500,9 +498,7 @@ fn modify_postgresql_conf(pgdata: PathBuf, postgresql_conf: Vec<&'static str>) -
     let postgresql_auto_conf = pgdata.join("postgresql.auto.conf");
 
     if let Some(runas) = get_runas() {
-        let mut sudo_command = Command::new("sudo")
-            .arg("-u")
-            .arg(&runas)
+        let mut sudo_command = sudo_command(&runas)
             .arg("tee")
             .arg(postgresql_auto_conf)
             .stdin(Stdio::piped())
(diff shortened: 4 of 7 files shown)


pgrx-tests/src/framework.rs
```diff
@@ -381,19 +387,56 @@ fn initdb(postgresql_conf: Vec<&'static str>) -> eyre::Result<()> {
     let pgdata = get_pgdata_path()?;
 
     if !pgdata.is_dir() {
+        let base = pgdata.parent().unwrap();
+        if !base.exists() {
+            std::fs::create_dir(base)?;
+        }
+
+        if let Some(runas) = get_runas() {
+            // NB:  This command has to be run as root to change the permissions from us to the `runas` user
```
eeeebbbbrrrr: ~~oh, this is no longer necessary~~

Incorrect.  It is necessary.  The user running `cargo pgrx test --runas` creates this directory, then we need to `sudo -u root` to `chown` that directory to the `runas` user.  Yes.  This is required.

cargo-pgrx/src/command/test.rs
```diff
@@ -140,6 +150,14 @@ pub fn test_extension(
         .env("PGRX_BUILD_PROFILE", profile.name())
         .env("PGRX_NO_SCHEMA", if no_schema { "true" } else { "false" });
 
+    if let Some(runas) = runas {
+        command.env("CARGO_PGRX_TEST_RUNAS", runas);
```
theory, line 154: Oh environment variable support is handy

pgrx-tests/src/framework.rs
```diff
@@ -381,19 +387,59 @@ fn initdb(postgresql_conf: Vec<&'static str>) -> eyre::Result<()> {
...
+        let base = pgdata.parent().unwrap();
+        if !base.exists() {
+            std::fs::create_dir(base)?;
+        }
+
+        if let Some(runas) = get_runas() {
+            // we've been asked to run as a different user.  As such, the PGDATA directory we just created
+            // needs to be owned by that user.
+            //
+            // In order to do that, we have to become "root" to change its ownership.
```
theory: Unless you're already running as root.

theory (approved):

pgrx-tests/src/framework.rs
eeeebbbbrrrr: Yes, but it doesn't seem worth it to not use sudo in that case.  Complicating a code path that switches users to root doesn't seem like a win.

I am happy to change it tho, if you insist.

cargo-pgrx/src/command/test.rs
eeeebbbbrrrr: Yeah, you could turn this dial in the case where you're not using cargo-pgrx.  Say if you wanted to test a complex rust workspace crate that happens to include one or more pgrx extensions too:

```shell
$ CARGO_PGRX_TEST_PGDATA=/tmp/pgrx-pgdata \
   CARGO_PGRX_TEST_RUNAS=theory \
   cargo test --all --features pg14 --no-default-features
```

pgrx-tests/src/framework.rs
```diff
@@ -678,8 +777,18 @@ pub(crate) fn get_pg_dbname() -> &'static str {
...
+            .unwrap_or_else(|_| panic!("USER environment var is unset or invalid UTF-8"))
+    })
+}
+
+fn get_runas() -> Option<String> {
+    std::env::var("CARGO_PGRX_TEST_RUNAS").ok()
+}
+
+fn requires_runas() -> bool {
+    get_runas().is_some()
```
workingjubilee, line 816: Single-callsite functions should have their contents inlined into their caller.

pgrx-tests/src/framework.rs
```diff
@@ -381,19 +387,59 @@ fn initdb(postgresql_conf: Vec<&'static str>) -> eyre::Result<()> {
...
+        }
+
+        if let Some(runas) = get_runas() {
+            // we've been asked to run as a different user.  As such, the PGDATA directory we just created
+            // needs to be owned by that user.
+            //
+            // In order to do that, we have to become "root" to change its ownership.
+            let mut chown = Command::new("sudo");
+            chown
+                .args(&["-u", "root", "chown", &runas])
```
workingjubilee: Pardon if I am being thick, but what exactly prevents this from being just
```suggestion
            let mut chown = Command::new("chown");
            chown
                .arg(&runas)
```

pgrx-tests/src/framework.rs
```diff
@@ -381,19 +387,59 @@ fn initdb(postgresql_conf: Vec<&'static str>) -> eyre::Result<()> {
...
+        let mut command = if let Some(runas) = get_runas() {
+            let mut cmd = Command::new("sudo");
+            cmd.arg("-u").arg(runas).arg(initdb_path);
+            cmd
+        } else {
+            Command::new(initdb_path)
+        };
 
         command
+            // .current_dir(pgdata.parent().unwrap())
```
workingjubilee: what was he cooking?

pgrx-tests/src/framework.rs
```diff
@@ -420,26 +466,42 @@ fn initdb(postgresql_conf: Vec<&'static str>) -> eyre::Result<()> {
...
+    }
+
+    let postgresql_auto_conf = pgdata.join("postgresql.auto.conf");
+
+    if let Some(runas) = get_runas() {
+        let mut sudo_command = Command::new("sudo")
+            .arg("-u")
+            .arg(&runas)
+            .arg("tee")
+            .arg(postgresql_auto_conf)
```
workingjubilee: We use `sudo` at various points instead of setting the user more pervasively for our execution of these steps. Perhaps we should simply reexecute one of our commands but with `sudo` in order to accomplish this goal instead of fallibly setting it at every step? Would that work? It would be nice if this could be a design more resilient to how things change inside pgrx.

pgrx-tests/src/framework.rs
theory: Nah fine with me if it's harmless, which I would expect it to be.

pgrx-tests/src/framework.rs
```diff
@@ -530,13 +597,27 @@ fn monitor_pg(mut command: Command, cmd_string: String, loglines: LogLines) -> S
...
-            let message_string = std::ffi::CString::new(
-                format!("stopping postgres (pid={pid})\n").bold().blue().to_string(),
-            )
-            .unwrap();
-            // IMPORTANT: Rust string literals are not naturally null-terminated
-            libc::printf("%s\0".as_ptr().cast(), message_string.as_ptr());
+            if let Some(_runas) = get_runas() {
+                Command::new("sudo")
+                    .arg("-u")
+                    .arg("root") // NB:  we must be "root" to kill the `sudo` process we spawned to start postgres
```
workingjubilee: Is there an alternative to using sudo to spawn it?

pgrx-tests/src/framework.rs
```diff
@@ -530,13 +597,27 @@ fn monitor_pg(mut command: Command, cmd_string: String, loglines: LogLines) -> S
...
+                    .arg("kill")
+                    .arg("-s")
+                    .arg("SIGTERM")
+                    .arg(pid.to_string())
+                    .spawn()
+                    .expect("failed to spawn `sudo kill`")
+                    .wait()
+                    .expect("`sudo kill` didn't work");
```
workingjubilee, line 622: ...thinking about how this could get weird. the extension we just loaded could set a SIGTERM handler, I suppose?

pgrx-tests/src/framework.rs
```diff
@@ -530,13 +597,27 @@ fn monitor_pg(mut command: Command, cmd_string: String, loglines: LogLines) -> S
...
+            } else {
+                libc::kill(pid as libc::pid_t, libc::SIGTERM);
+                let message_string = std::ffi::CString::new(
+                    format!("stopping postgres (pid={pid})\n").bold().blue().to_string(),
+                )
+                .unwrap();
+                // IMPORTANT: Rust string literals are not naturally null-terminated
+                libc::printf("%s\0".as_ptr().cast(), message_string.as_ptr());
```
workingjubilee, line 630: I believe this can use a CStr literal.
```suggestion
                libc::printf(c"%s".as_ptr().cast(), message_string.as_ptr());
```

pgrx-tests/src/framework.rs
```diff
@@ -678,8 +777,18 @@ pub(crate) fn get_pg_dbname() -> &'static str {
...
-    std::env::var("USER")
-        .unwrap_or_else(|_| panic!("USER environment var is unset or invalid UTF-8"))
+    get_runas().unwrap_or_else(|| {
+        std::env::var("USER")
+            .unwrap_or_else(|_| panic!("USER environment var is unset or invalid UTF-8"))
+    })
+}
+
+fn get_runas() -> Option<String> {
+    std::env::var("CARGO_PGRX_TEST_RUNAS").ok()
```
workingjubilee: `VarError` has two cases. `VarError::NotPresent` and `VarError::NotUnicode`. Please either make this function return OsString (and so, use `env::var_os`, probably), or panic on encountering `VarError::NotUnicode`.

Also, this is a string, but we're probably assuming something about the shape of the contents. For instance, is whitespace allowed? Should it not be? How about `"$VARIABLE"`? That... *should* be harmless, but

.github/workflows/runas.yml
```diff
@@ -0,0 +1,39 @@
...
+on:
+  push:
+    branches: [ develop ]
+  pull_request:
+    branches: [ develop ]
+
+env:
+  CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: 1
+  RUSTFLAGS: "-C link-args=-Wl,-undefined,dynamic_lookup"   # we don't want target-cpu=native from `.cargo/config` b/c we're caching binaries but we want the other stuff
```
workingjubilee: :raised_eyebrow: `-Ctarget-cpu=native` is not in our repo's `.cargo/config.toml`?

cargo-pgrx/README.md
```diff
@@ -509,6 +514,8 @@ Options:
   -r, --release                        compile for release mode (default is debug)
       --profile <PROFILE>              Specific profile to use (conflicts with `--release`)
   -n, --no-schema                      Don't regenerate the schema
+      --runas <RUNAS>                  Use `sudo` to initialize and run the Postgres test instance as this system user
```
workingjubilee: This should look more like
```suggestion
      --runas <USER>                  Use `sudo` to initialize and run the Postgres test instance as this system user
```

cargo-pgrx/src/command/test.rs
```diff
@@ -40,6 +40,12 @@ pub(crate) struct Test {
     /// Don't regenerate the schema
     #[clap(long, short)]
     no_schema: bool,
+    /// Use `sudo` to initialize and run the Postgres test instance as this system user
+    #[clap(long)]
+    runas: Option<String>,
```
workingjubilee:
```suggestion
    #[arg(long, value_name = "USER")]
    runas: Option<String>,
```

cargo-pgrx/src/command/test.rs
```diff
@@ -40,6 +40,12 @@ pub(crate) struct Test {
...
+    /// Initialize the test database cluster here, instead of the default location.  If used with `--runas`, then it must be writable by the user
+    #[clap(long)]
+    pgdata: Option<PathBuf>,
```
workingjubilee:
```suggestion
    #[arg(long, value_name = "DIR")]
    pgdata: Option<PathBuf>,
```

cargo-pgrx/src/command/test.rs
```diff
@@ -40,6 +40,12 @@ pub(crate) struct Test {
     /// Don't regenerate the schema
     #[clap(long, short)]
     no_schema: bool,
+    /// Use `sudo` to initialize and run the Postgres test instance as this system user
+    #[clap(long)]
```
workingjubilee: I don't believe the `#[clap]` macro is correct here, I believe it's supposed to be `#[arg]`?

workingjubilee (commented):
I have several questions! Hopefully not all of them are relevant

pgrx-tests/src/framework.rs
eeeebbbbrrrr: You gotta be root in order to change the ownership of a file to a different user.

pgrx-tests/src/framework.rs
eeeebbbbrrrr: good catch.

pgrx-tests/src/framework.rs
eeeebbbbrrrr: It'd be nice if this would work, but it would require the user we're becoming to also have (at least) `cargo-pgrx` installed/available to them.  And if that were the case, then the caller could just run `sudo -u otheruser cargo pgrx test` directly.

We also have to su to `root` in a few places to get files copied.  So we become two different users, essentially.

pgrx-tests/src/framework.rs
eeeebbbbrrrr: Not that I can see.  The `sudo` process we're trying to kill is owned by root.

cargo-pgrx/README.md
eeeebbbbrrrr: this is what clap generates for us.  I don't see how to change that.

cargo-pgrx/src/command/test.rs
eeeebbbbrrrr: oh

pgrx-tests/src/framework.rs
eeeebbbbrrrr: We're trying to shutdown the Postgres instance we started.  Normally, the user running `cargo pgrx test` owns that process.  In the case of `--runas`, it's actually "root" that owns that process.  Hence this dance to exec `sudo kill ...`

pgrx-tests/src/framework.rs
eeeebbbbrrrr: This has been this way for quite some time.  It's showing up in the diff due to an indentation change.

It probably could use the CStr literal, now that it exists, but I'm trying to limit the changes of this PR.

pgrx-tests/src/framework.rs
eeeebbbbrrrr: Fixed error handling.

GIGO I think is fine.  Because of the `Command` type, we'll at least be passing whatever the user provides to `--runas` through as-is as a proper argument.  iow, we don't have to worry about shell escaping or anything like that.

cargo-pgrx/src/command/test.rs
eeeebbbbrrrr: I don't follow?

workingjubilee: Maybe the docs I was referencing were dated? maybe it's an alias? idk lol

pgrx-tests/src/framework.rs
workingjubilee: Ah, yeah, I hadn't parsed out the flow fully of having to jump between two users at that point, and still wasn't sure if my model was correct after I did... :sob: So much for that idea.

pgrx-tests/src/framework.rs
workingjubilee: Hm. For some reason I thought you could chown files you own to another user. Guess not.

pgrx-tests/src/framework.rs
```diff
@@ -381,19 +388,59 @@ fn initdb(postgresql_conf: Vec<&'static str>) -> eyre::Result<()> {
...
+            let output = child.wait_with_output()?;
+            if !output.status.success() {
+                panic!(
+                    "failed to change ownership of the PGDATA directory at `{}`:\n{}{}",
+                    base.display().yellow(),
+                    String::from_utf8(output.stdout).unwrap(),
+                    String::from_utf8(output.stderr).unwrap()
+                );
+            }
+        }
```
workingjubilee: Hmm. This is still giving me a bad vibe. The prickles on the back of my neck are trying to spell out [TOCTOU](https://cwe.mitre.org/data/definitions/367.html). Couldn't we
- check if we should `runas` first
- then, in that case, `sudo -u $runas mkdir -p $pgdata` without checking
- then simply refuse to `chown`, reducing the number of times we assume `root` by 1?

I think it's reasonable to classify "you didn't set up the directory ownership correctly for running the tests, despite controlling both the `runas` and `pgdata` parameters" as an error. If it is absolutely critical for @theory's use-case that we do this for him, sure, but it seems like the answer is "no", from my understanding?

I realize this warps the control flow a bit more, but it seems significantly more correct.

.github/workflows/runas.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: cargo --version
+
+      - name: Install cargo pgrx
+        run: cd cargo-pgrx && cargo install --path . --debug
+
+      - name: cargo pgrx init
+        run: cargo pgrx init --pg14=$(which pg_config)
+
+      - name: Test cargo pgrx test --runas
+        run: cd pgrx-examples/arrays && cargo pgrx test pg14 --runas postgres --pgdata=/tmp/pgdata
```
workingjubilee, line 38: Should we have a clone of this that uses just the env vars?

cargo-pgrx/src/command/test.rs
```diff
@@ -140,6 +150,14 @@ pub fn test_extension(
         .env("PGRX_BUILD_PROFILE", profile.name())
         .env("PGRX_NO_SCHEMA", if no_schema { "true" } else { "false" });
 
+    if let Some(runas) = runas {
+        command.env("CARGO_PGRX_TEST_RUNAS", runas);
+    }
+
+    if let Some(pgdata) = pgdata {
+        command.env("CARGO_PGRX_TEST_PGDATA", pgdata);
+    }
```
workingjubilee, line 159: What should win if they do

`CARGO_PGRX_TEST_PGDATA="/some/path/somewhere" cargo pgrx test --pgdata="/another/path/elsewhere"`?

note that as far as I am concerned "cargo pgrx should explode in their face and tell them not to do that" is a fine answer, but if we want, we can take the value from the environment by adding `, env = "VARIABLE"` to the list of "arguments" to the `#[clap]` attribute for that CLI arg.

cargo-pgrx/README.md
```diff
@@ -509,6 +514,8 @@ Options:
   -r, --release                        compile for release mode (default is debug)
       --profile <PROFILE>              Specific profile to use (conflicts with `--release`)
   -n, --no-schema                      Don't regenerate the schema
+      --runas <USER>                   Use `sudo` to initialize and run the Postgres test instance as this system user
+      --pgdata <DIR>                   Initialize the test database cluster here, instead of the default location.  If used with `--runas`, then it must be writable by the user
```
workingjubilee, line 518: re: the chown thing, note that this *says* that it has to be writable by the user already, so that seems like our intended behavior if we believe the docs?

pgrx-tests/src/framework.rs
```diff
@@ -467,6 +530,11 @@ fn start_pg(loglines: LogLines) -> eyre::Result<String> {
             }
         }
 
+        cmd.arg(postmaster_path);
+        cmd
+    } else if let Some(runas) = get_runas() {
+        let mut cmd = Command::new("sudo");
+        cmd.arg("-u").arg(runas);
```
workingjubilee: Maybe, instead of using `sudo` here, we could execute `id -u "$runas"`, and then, having obtained the relevant UID, use [`<Command as std::os::unix::process::CommandExt>::uid`](https://doc.rust-lang.org/std/os/unix/process/trait.CommandExt.html#required-methods)? Then we wouldn't need to `sudo` just to kill our child, later.

If this doesn't work, then the current pattern will do for now.

pgrx-tests/src/framework.rs
```diff
@@ -315,6 +314,14 @@ fn install_extension() -> eyre::Result<()> {
         .stderr(Stdio::inherit())
         .env("CARGO_TARGET_DIR", get_target_dir()?);
 
+    if requires_runas() {
+        // if we're running tests as a different operating-system user we actually need to
```
workingjubilee: Minor legibility nit
```suggestion
        // if we're running tests as a different operating-system user, then we actually need to
```

workingjubilee (commented):
I unfortunately have more questions about the implementation now that my first batch of "do we have any alternatives?" questions has been exhausted.

pgrx-tests/src/framework.rs
workingjubilee: Is there any way for us to, at this point, obtain intelligence on whether our problem was "the `kill` was ignored" or "`sudo` failed"?

I'm wondering if we should try to escalate to SIGKILL after. It seems... more important to consider now.

pgrx-tests/src/framework.rs
```diff
@@ -673,13 +773,34 @@ fn get_pid_file() -> eyre::Result<PathBuf> {
...
+    })
+}
+
+#[inline]
+fn get_runas() -> Option<String> {
+    match std::env::var("CARGO_PGRX_TEST_RUNAS") {
+        Ok(s) => Some(s),
+        Err(e) => match e {
+            VarError::NotPresent => None,
+            VarError::NotUnicode(e) => {
```
workingjubilee, line 804: just noticed a warning. alternately we can print it after doing a `.to_string_lossy()`.
```suggestion
            VarError::NotUnicode(_) => {
```

pgrx-tests/src/framework.rs
eeeebbbbrrrr: SIGKILL is a pretty bad idea to do against a running postgres cluster.  You could argue it probably doesn't matter here since our testing cluster is ephemeral, but still.

I don't really see a case where `sudo -u root kill $pid` is going to fail tho.

pgrx-tests/src/framework.rs
eeeebbbbrrrr: > then, in that case, sudo -u $runas mkdir -p $pgdata without checking

Yeah, this is a better idea for sure.  I'll do this.  I never thought of that.

cargo-pgrx/src/command/test.rs
eeeebbbbrrrr: Note that `cargo pgrx test` doesn't actually use the environment variables.  It sets them as part of the `cargo test ...` Command it runs.  As such, it's `pgrx-tests/src/framework.rs` that reads them.  

So... I think `cargo pgrx test` should just continue to ignore it, like it does every other envar it doesn't use.

pgrx-tests/src/framework.rs
eeeebbbbrrrr: > Sets the child process’s user ID. This translates to a setuid call in the child process. Failure in the setuid call will cause the spawn to fail.

You'd have to already be root in order for that to work.  Or `cargo-pgrx.exe` would need to be setuid, and that's a bad idea.

workingjubilee: That's fair. Then yes, let's not use a setuid binary.

pgrx-tests/src/framework.rs
workingjubilee: All right then.

cargo-pgrx/src/command/test.rs
workingjubilee: That's fair.

Hm. Maybe the test vars introduced here should be named `PGRX_TEST_*`? Not a serious concern, just wondering.

pgrx-tests/src/framework.rs
workingjubilee: Great! That addresses our issues here then.

pgrx-tests/src/framework.rs
```diff
@@ -420,26 +488,42 @@ fn initdb(postgresql_conf: Vec<&'static str>) -> eyre::Result<()> {
...
+        contents.push_str(&format!("{setting}\n"));
+    }
+
+    let postgresql_auto_conf = pgdata.join("postgresql.auto.conf");
+
+    if let Some(runas) = get_runas() {
+        let mut sudo_command = Command::new("sudo")
+            .arg("-u")
+            .arg(&runas)
+            .arg("tee")
```
workingjubilee: You mentioned making a `SudoCommand` type and I think what I'd like to *actually* see is these ~4 lines made into a function/whatever that returns `Command`.

workingjubilee (approved):
You can make the final changes I mentioned. Or not. I'm good with this now.

workingjubilee:
I should note that I am "good with this" in the sense that I have made my peace with the horrors of the world that is Unix access control.

I do encourage you to bother the actions/checkout people, @theory, and try to fix things on their end. If this code turns out to have a security flaw of any significant import... anything that's not "you can pwn a computer that's already been pwned", [which sometimes gets reported as a "vuln" anyways](https://devblogs.microsoft.com/oldnewthing/20060508-22/?p=31283), then I will not hesitate to rip it out. Likely `--pgdata` will remain, I suppose.

theory:
Fair enough, I've gone and complained in actions/checkout#956.

theory:
There's also [this bit of documentation](https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions#user), though:

> Docker actions must be run by the default Docker user (root). Do not use the `USER` instruction in your `Dockerfile`, because you won't be able to access the `GITHUB_WORKSPACE` directory. For more information, see "[Variables](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)" and [USER reference](https://docs.docker.com/engine/reference/builder/#user) in the Docker documentation.

This is for action containers, but I think it applies to workflows steps that run with `container`.

workingjubilee:
Since Docker does support setting a user, I believe the problem is that GHA doesn't support it, so deescalating privileges seems like something GHA could support nonetheless.

pgrx-tests/src/framework.rs
```diff
@@ -420,26 +482,40 @@ fn initdb(postgresql_conf: Vec<&'static str>) -> eyre::Result<()> {
...
-        .open(format!("{}/postgresql.auto.conf", pgdata.display()))
-        .wrap_err("couldn't open postgresql.auto.conf")?;
-    postgresql_conf_file
-        .write_all("log_line_prefix='[%m] [%p] [%c]: '\n".as_bytes())
-        .wrap_err("couldn't append log_line_prefix")?;
+    let mut contents = String::new();
 
+    contents.push_str("log_line_prefix='[%m] [%p] [%c]: '\n");
+    contents
+        .push_str(&format!("unix_socket_directories = '{}'\n", pgdata.parent().unwrap().display()));
```
workingjubilee, line 489: I probably should have asked earlier, but why are we adding this `unix_socket_directories` field? We did not do this before.

workingjubilee: wait, yes we did. no, the problem is now it's different somehow...

workingjubilee: Sorry for the noise, figured out what was up here: adding `"pg"` to the string used up someone's last two bytes. :smiling_face_with_tear:

Issue #1647, opened 2024-04-09
Add `cargo pgrx upgrade` type functionality


workingjubilee:
`cargo pgrx` packages a minimum amount of toml-inspecting and toml-manipulating code of: "a lot". We should just allow updating the pgrx versions for someone.

workingjubilee:
Bigger changes could be made automagical-ish in the future, but this would be the basic one.

NotGyro:
Implemented in https://github.com/pgcentralfoundation/pgrx/pull/1687

workingjubilee:
Yes, this has been completely fixed!

Issue #1645, opened 2024-04-09
gracefully handle `cargo pgrx new $EXISTING_DIR_WITH_FILES`


jyn514:
i just did the following things:
1. `cargo pgrx new` with pgrx 0.11.3.
2. `cargo pgrx run` with pgrx 0.11.3. that gave an error that i installed pgrx with 1.76, but the current toolchain is 1.79-nightly. i remembered @workingjubilee saying that restriction was removed on `develop`, so i
3. `cargo install --git https://github.com/pgcentralfoundation/pgrx --locked cargo-pgrx`
4. `cargo pgrx run` with the new binary. that gives the following error:

```
Error: 
   0: The installed `cargo-pgrx` v0.12.0-alpha.1 is not compatible with the `pgrx = 0.11.3`, `pgrx-macros = 0.11.3`, `pgrx-sql-entity-graph = 0.11.3`, `pgrx-tests = 0.11.3` dependencies in `./Cargo.toml`. `cargo-pgrx` and pgrx dependency versions must be identical.
                  help: cargo install --version `pgrx = 0.11.3`, `pgrx-macros = 0.11.3`, `pgrx-sql-entity-graph = 0.11.3`, `pgrx-tests = 0.11.3` --locked

Location:
   cargo-pgrx/src/metadata.rs:68

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ SPANTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

   0: cargo_pgrx::metadata::validate
      at cargo-pgrx/src/metadata.rs:30
   1: cargo_pgrx::command::run::execute
      at cargo-pgrx/src/command/run.rs:54
```
now the error makes some amount of sense to me, but the `help:` suggestion does not look right. those are dependency versions, not the version of cargo-pgrx itself. maybe https://github.com/pgcentralfoundation/pgrx/blob/82fe8b1ae32cb474a9e1819493725b1efb695d4b/cargo-pgrx/src/metadata.rs#L72 should be using `cargo_pgrx_version` instead of `mismatches`?

jyn514:
alternatively it could suggest to update the dependencies in Cargo.toml, which is what i'm about to do now, but that's hard to make copy/pasteable unless they already have cargo-edit installed.

jyn514:
huh. apparently `cargo pgrx new` does not error if there's already a directory there. so that's another (albeit destructive) option.

[...]

PR #1659, opened 2024-04-16
Only suggest `cargo install` with 1 version

workingjubilee:
I believe this addresses @jyn514's initially-reported bug in https://github.com/pgcentralfoundation/pgrx/issues/1645 by avoiding the help suggestion if it doesn't make sense for pgrx's architecture, and we make sure it's a single version we're suggesting, not... some weirdness.

An alternative help suggestion would only be reasonable with https://github.com/pgcentralfoundation/pgrx/issues/1647

### DIFF
--- a/cargo-pgrx/src/metadata.rs
+++ b/cargo-pgrx/src/metadata.rs
@@ -43,8 +43,14 @@ pub fn validate(
     });
 
     let mut mismatches = BTreeMap::new();
+    let (mut unified, mut universion) = (true, None);
     for package in pgrx_packages {
         let package_semver = package.version.clone();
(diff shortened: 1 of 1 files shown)


jyn514:
i am confused: this is still suggesting `cargo install` on a library package, no? i thought the intended help message was to install a different version of cargo-pgrx, not the pgrx libraries.

workingjubilee:
brain?

workingjubilee:
Should be fixed now.

workingjubilee:
> i am confused: this is still suggesting `cargo install` on a library package, no? i thought the intended help message was to install a different version of cargo-pgrx, not the pgrx libraries.

the briefly-implemented behavior simply didn't pass a suggested crate at all, actually!

PR #1662, opened 2024-04-17
Use `CARGO_TARGET_DIR` if available

### LINKED ISSUE #1625: Whatever we're doing with `get_target_dir` is absolutely unnecessary
Trying to be clever instead of referencing `CARGO_TARGET_DIR` breaks integration with other tools, but I am far too tired to fix it tonight.

workingjubilee:
Fixes https://github.com/pgcentralfoundation/pgrx/issues/1625


workingjubilee:
I think I was hoping I would imagine a useful test, is why I didn't PR it immediately.

I did not.

PR #1663, opened 2024-04-17
Use supports-colors feature

theory:
To enable the `NO_COLOR` and `FORCE_COLOR` environment variables, as well as detect when running in a tty to suppress colors.

Ideally pgrx would resect [`CARGO_TERM_COLOR`](https://doc.rust-lang.org/cargo/reference/config.html#termcolor), but this is a start.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -139,6 +139,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "atty"
+version = "0.2.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
+dependencies = [
+ "hermit-abi",
+ "libc",
+ "winapi",
+]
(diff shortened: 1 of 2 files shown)


workingjubilee:
can you run `cargo update -p owo-colors` for the lockfile update as well?

theory:
Done in 82aa4b30 (and wayward space removed).

Cargo.lock
```diff
@@ -138,6 +138,17 @@ dependencies = [
...
+[[package]]
+name = "atty"
+version = "0.2.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
+dependencies = [
+ "hermit-abi",
+ "libc",
+ "winapi",
+]
```
workingjubilee, line 150: You again! Sigh. I'll deal with you later.

workingjubilee:
Thanks!

PR #1665, opened 2024-04-19
Pipe forward spans by formatting idents

workingjubilee:
Apply a simple change to -sql-entity-graph and -macros:

Use `format_ident!` for every case we format... an identifier... and we don't change the identifier, and reuse the existing span or worse, drop the span on the floor by calling `Span::call_site`. This will, in general, improve the error-reporting of spans so that errors refer back to the original input that provoked the expansion.

It is possible we should by relocating synthesized spans, see https://github.com/rust-lang/rust/issues/124145 for more details. This change is still preferred because it makes it easier to refactor these again.

### DIFF
--- a/pgrx-sql-entity-graph/src/pg_extern/mod.rs
+++ b/pgrx-sql-entity-graph/src/pg_extern/mod.rs
@@ -38,7 +38,7 @@ use crate::enrich::CodeEnrichment;
 use crate::enrich::ToEntityGraphTokens;
 use crate::enrich::ToRustCodeTokens;
 use proc_macro2::{Ident, Span, TokenStream as TokenStream2};
-use quote::{quote, quote_spanned, ToTokens};
+use quote::{format_ident, quote, quote_spanned, ToTokens};
 use syn::parse::{Parse, ParseStream, Parser};
 use syn::punctuated::Punctuated;
 use syn::spanned::Spanned;
@@ -337,8 +337,7 @@ impl PgExtern {
             .collect::<Vec<_>>();
         let hrtb = if lifetimes.is_empty() { None } else { Some(quote! { for<#(#lifetimes),*> }) };
 
-        let sql_graph_entity_fn_name =
-            syn::Ident::new(&format!("__pgrx_internals_fn_{}", ident), Span::call_site());
+        let sql_graph_entity_fn_name = format_ident!("__pgrx_internals_fn_{}", ident);
         quote_spanned! { self.func.sig.span() =>
             #[no_mangle]
             #[doc(hidden)]
@@ -374,10 +373,7 @@ impl PgExtern {
     }
 
     fn finfo_tokens(&self) -> TokenStream2 {
-        let finfo_name = syn::Ident::new(
-            &format!("pg_finfo_{}_wrapper", self.func.sig.ident),
-            Span::call_site(),
-        );
+        let finfo_name = format_ident!("pg_finfo_{}_wrapper", self.func.sig.ident);
         quote_spanned! { self.func.sig.span() =>
             #[no_mangle]
             #[doc(hidden)]
@@ -390,10 +386,7 @@ impl PgExtern {
 
     pub fn wrapper_func(&self) -> TokenStream2 {
         let func_name = &self.func.sig.ident;
-        let func_name_wrapper = Ident::new(
-            &format!("{}_wrapper", &self.func.sig.ident.to_string()),
-            self.func.sig.ident.span(),
-        );
+        let func_name_wrapper = format_ident!("{}_wrapper", &self.func.sig.ident);
         let func_generics = &self.func.sig.generics;
         // the wrapper function declaration may contain lifetimes that are not used, since our input type is `FunctionCallInfo` mainly and return type is `Datum`
         let unused_lifetimes = match func_generics.lifetimes().next() {
@@ -407,10 +400,7 @@ impl PgExtern {
         let fcinfo_ident = syn::Ident::new("_fcinfo", self.func.sig.ident.span());
 
         let args = &self.inputs;
-        let arg_pats = args
-            .iter()
-            .map(|v| syn::Ident::new(&format!("{}_", &v.pat), self.func.sig.span()))
-            .collect::<Vec<_>>();
+        let arg_pats = args.iter().map(|v| format_ident!("{}_", &v.pat)).collect::<Vec<_>>();
         let arg_fetches = args.iter().enumerate().map(|(idx, arg)| {
             let pat = &arg_pats[idx];
             let resolved_ty = &arg.used_ty.resolved_ty;
--- a/pgrx-sql-entity-graph/src/pg_trigger/mod.rs
+++ b/pgrx-sql-entity-graph/src/pg_trigger/mod.rs
@@ -22,7 +22,7 @@ use crate::enrich::{ToEntityGraphTokens, ToRustCodeTokens};
 use crate::{CodeEnrichment, ToSqlConfig};
 use attribute::PgTriggerAttribute;
 use proc_macro2::{Span, TokenStream as TokenStream2};
-use quote::quote;
+use quote::{format_ident, quote};
 use syn::{ItemFn, Token};
 
 #[derive(Debug, Clone)]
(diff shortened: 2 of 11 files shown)


workingjubilee:
... `format_ident!` has existed since quote 1.0,

pgrx-sql-entity-graph/src/pg_extern/mod.rs
```diff
@@ -407,10 +400,7 @@ impl PgExtern {
         let fcinfo_ident = syn::Ident::new("_fcinfo", self.func.sig.ident.span());
 
         let args = &self.inputs;
-        let arg_pats = args
-            .iter()
-            .map(|v| syn::Ident::new(&format!("{}_", &v.pat), self.func.sig.span()))
-            .collect::<Vec<_>>();
+        let arg_pats = args.iter().map(|v| format_ident!("{}_", &v.pat)).collect::<Vec<_>>();
```
workingjubilee, line 403: Despite being named `pat`, this is in fact an Ident.

workingjubilee (commented):
Besides the cases where it was not obvious we were using the same identifier, I skipped a number of locations because the identifier was lowercased, requiring more special handling anyways.

@eeeebbbbrrrr As far as you know, is any of the lowercasing of idents actually important?

workingjubilee:
~~Apparently so.~~[^0]

[^0]: EDIT: Not for the reason I just saw, actually, just an expected UI test variation.

pgrx-sql-entity-graph/src/pg_trigger/mod.rs
```diff
@@ -121,18 +115,15 @@ impl PgTrigger {
...
         let func_sig_ident = &self.func.sig.ident;
+        let sql_graph_entity_fn_name = format_ident!("__pgrx_internals_trigger_{}", func_sig_ident);
         let function_name = func_sig_ident.to_string();
         let to_sql_config = &self.to_sql_config;
 
         quote! {
             #[no_mangle]
             #[doc(hidden)]
-            #[allow(unknown_lints, clippy::no_mangle_with_rust_abi)]
+            #[allow(unknown_lints, clippy::no_mangle_with_rust_abi, nonstandard_style)]
```
workingjubilee, line 126: This should be factored out, but I am leaving that for a future PR.

pgrx-sql-entity-graph/src/pg_trigger/mod.rs
```diff
@@ -121,18 +115,15 @@ impl PgTrigger {
 
 impl ToEntityGraphTokens for PgTrigger {
     fn to_entity_graph_tokens(&self) -> TokenStream2 {
-        let sql_graph_entity_fn_name = syn::Ident::new(
-            &format!("__pgrx_internals_trigger_{}", self.func.sig.ident),
-            self.func.sig.ident.span(),
-        );
         let func_sig_ident = &self.func.sig.ident;
+        let sql_graph_entity_fn_name = format_ident!("__pgrx_internals_trigger_{}", func_sig_ident);
```
workingjubilee, line 119: Even this part could be factored out.

workingjubilee (commented):
It may be that we want to use a **synthetic** span that is actually from `Span::mixed_site` or `Span::call_site` but I am going to merge this refactoring because it allows us to more easily change all these instances of `format_ident` to use something else now that they are a distinctively-named macro and not just ad-hoc string formatting into `Ident::new`, and will hopefully give better error spans when we need to generate those.

workingjubilee:
Oh, and this is an example of the actual error reporting improvements, but it's only fully visible on beta right now:

```diff
diff --git a/pgrx-tests/tests/todo/composite-types-broken-on-spi.stderr b/pgrx-tests/tests/todo/composite-types-broken-on-spi.stderr
index e407435c..b73ac689 100644
--- a/pgrx-tests/tests/todo/composite-types-broken-on-spi.stderr
+++ b/pgrx-tests/tests/todo/composite-types-broken-on-spi.stderr
@@ -74,28 +74,28 @@ error[E0277]: `Array<'_, pgrx::prelude::PgHeapTuple<'_, pgrx::AllocatedByRust>>`
     = help: the trait `IntoIterator` is implemented for `Array<'_, pgrx::prelude::PgHeapTuple<'_, pgrx::AllocatedByRust>>`
 
 error[E0308]: mismatched types
-  --> tests/todo/composite-types-broken-on-spi.rs:12:5
+  --> tests/todo/composite-types-broken-on-spi.rs:13:9
    |
-12 |     fn sum_scritches_for_names(
-   |     ^^ one type is more general than the other
+13 |         dogs: Option<Vec<pgrx::composite_type!(DOG_COMPOSITE_TYPE)>>,
+   |         ^^^^ one type is more general than the other
    |
    = note: expected struct `pgrx::prelude::PgHeapTuple<'arr, _>`
               found struct `pgrx::prelude::PgHeapTuple<'_, _>`
 
 error[E0308]: mismatched types
-  --> tests/todo/composite-types-broken-on-spi.rs:29:5
+  --> tests/todo/composite-types-broken-on-spi.rs:30:9
    |
-29 |     fn sum_scritches_for_names_default(
-   |     ^^ one type is more general than the other
+30 |         dogs: pgrx::default!(Vec<pgrx::composite_type!("Dog")>, "ARRAY[ROW('Nami', 0)]::Dog[]"),
+   |         ^^^^ one type is more general than the other
    |
    = note: expected struct `pgrx::prelude::PgHeapTuple<'arr, _>`
               found struct `pgrx::prelude::PgHeapTuple<'_, _>`
 
 error[E0308]: mismatched types
-  --> tests/todo/composite-types-broken-on-spi.rs:45:5
+  --> tests/todo/composite-types-broken-on-spi.rs:45:39
    |
 45 |     fn sum_scritches_for_names_strict(dogs: Vec<pgrx::composite_type!("Dog")>) -> i32 {
-   |     ^^ one type is more general than the other
+   |                                       ^^^^ one type is more general than the other
    |
    = note: expected struct `pgrx::prelude::PgHeapTuple<'arr, _>`
               found struct `pgrx::prelude::PgHeapTuple<'_, _>`
diff --git a/pgrx-tests/tests/todo/vec-option-composite_type-nesting-problems.stderr b/pgrx-tests/tests/todo/vec-option-composite_type-nesting-problems.stderr
index f5df88b0..cc9e3271 100644
--- a/pgrx-tests/tests/todo/vec-option-composite_type-nesting-problems.stderr
+++ b/pgrx-tests/tests/todo/vec-option-composite_type-nesting-problems.stderr
@@ -47,28 +47,28 @@ note: required by a bound in `pg_getarg`
    |                            ^^^^^^^^^ required by this bound in `pg_getarg`
 
 error[E0308]: mismatched types
- --> tests/todo/vec-option-composite_type-nesting-problems.rs:9:1
+ --> tests/todo/vec-option-composite_type-nesting-problems.rs:9:28
   |
 9 | fn sum_scritches_for_names(dogs: Option<Vec<pgrx::composite_type!(DOG_COMPOSITE_TYPE)>>) -> i32 {
-  | ^^ one type is more general than the other
+  |                            ^^^^ one type is more general than the other
   |
   = note: expected struct `pgrx::prelude::PgHeapTuple<'arr, _>`
              found struct `pgrx::prelude::PgHeapTuple<'_, _>`
 
 error[E0308]: mismatched types
-  --> tests/todo/vec-option-composite_type-nesting-problems.rs:24:1
+  --> tests/todo/vec-option-composite_type-nesting-problems.rs:25:5
    |
-24 | fn sum_scritches_for_names_default(
-   | ^^ one type is more general than the other
+25 |     dogs: pgrx::default!(Vec<pgrx::composite_type!("Dog")>, "ARRAY[ROW('Nami', 0)]::Dog[]"),
+   |     ^^^^ one type is more general than the other
    |
    = note: expected struct `pgrx::prelude::PgHeapTuple<'arr, _>`
               found struct `pgrx::prelude::PgHeapTuple<'_, _>`
 
 error[E0308]: mismatched types
-  --> tests/todo/vec-option-composite_type-nesting-problems.rs:40:1
+  --> tests/todo/vec-option-composite_type-nesting-problems.rs:40:35
    |
 40 | fn sum_scritches_for_names_strict(dogs: Vec<pgrx::composite_type!("Dog")>) -> i32 {
-   | ^^ one type is more general than the other
+   |                                   ^^^^ one type is more general than the other
    |
    = note: expected struct `pgrx::prelude::PgHeapTuple<'arr, _>`
               found struct `pgrx::prelude::PgHeapTuple<'_, _>`
```

PR #1668, opened 2024-04-19
Ad hoc span fixups

workingjubilee:
A suite of random cases where we drop spans on the floor will no longer do so. As with the `format_ident!` change in 5c719e5 the same caveat of "perhaps we should synthesize spans?" applies. A relevant FIXME applies to each case so we can find them later, in the absence of a more-unique string.

The error-reporting changes get needs-test FIXMEs, as none of them diffed any UI tests and that lack is preexisting. I am confident of the improvement due to seeing the improvement from similar changes, and we really must recover basic debuggability for this code.


workingjubilee:
See https://github.com/pgcentralfoundation/pgrx/pull/1665#issuecomment-2067234862 for an example of such an "improvement from similar changes".

PR #1673, opened 2024-04-19
Factor out `finfo_tokens`

workingjubilee:
An initial toehold on refactoring work to partially unify wrapper-fn expansions in their implementations, especially where it is correctness-relevant.


workingjubilee:
Something for me to remember to review on Monday.

PR #1674, opened 2024-04-22
`cargo pgrx test --runas` envar passing

eeeebbbbrrrr:
This teaches the test `framework.rs` to pass through to `sudo` all the various cargo/rust environment variables.

Turns out this might be necessary for `#[pg_test]` tests that use `std::env::var(...)`.  When using `--runas` we spawn the postmaster with `sudo`, and its default policy is to reset the environment (`env_reset` flag in the sudoers file).

### DIFF
--- a/pgrx-tests/src/framework.rs
+++ b/pgrx-tests/src/framework.rs
@@ -546,7 +546,31 @@ fn start_pg(loglines: LogLines) -> eyre::Result<String> {
         cmd.arg(postmaster_path);
         cmd
     } else if let Some(runas) = get_runas() {
+        #[inline]
+        fn accept_envar(var: &str) -> bool {
+            // taken from https://doc.rust-lang.org/cargo/reference/environment-variables.html
+            var.starts_with("CARGO")
+                || var.starts_with("RUST")
+                || var.starts_with("DEP_")
+                || ["OUT_DIR", "TARGET", "HOST", "NUM_JOBS", "OPT_LEVEL", "DEBUG", "PROFILE"]
+                    .contains(&var)
+        }
+
         let mut cmd = sudo_command(runas);
+
+        // when running the `postmaster` process via `sudo`, we need to copy the cargo/rust-related
+        // environment variables and pass as arguments to sudo, ahead of the `postmaster` command itself
+        //
+        // This ensures that any in-processs #[pg_test]s that might otherwise expect some kind of
+        // `CARGO_xxx` envar to actually exist.
+        for (var, value) in std::env::vars() {
+            if accept_envar(&var) {
+                let env_as_arg = format!("{var}={}", shlex::try_quote(&value)?);
+                cmd.arg(env_as_arg);
+            }
+        }
+
+        // now we can add the `postmaster` as the command for `sudo` to execute
         cmd.arg(postmaster_path);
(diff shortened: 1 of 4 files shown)


eeeebbbbrrrr:
we're now doing the latter version:

```shellscript
$ FOO=bar sudo -u testuser printenv FOO
$ sudo -u testuser FOO=bar printenv FOO
bar
```

pgrx-tests/src/framework.rs
```diff
@@ -546,7 +546,30 @@ fn start_pg(loglines: LogLines) -> eyre::Result<String> {
...
         cmd
     } else if let Some(runas) = get_runas() {
+        #[inline]
+        fn accept_envar(var: &str) -> bool {
+            // taken from https://doc.rust-lang.org/cargo/reference/environment-variables.html
+            var.starts_with("CARGO")
+                || var.starts_with("RUST")
+                || var.starts_with("DEP_")
+                || ["OUT_DIR", "TARGET", "HOST", "NUM_JOBS", "OPT_LEVEL", "DEBUG", "PROFILE"]
+                    .contains(&var)
```
workingjubilee, line 556: hmm, regex maybe?

pgrx-tests/src/framework.rs
```diff
@@ -546,7 +546,30 @@ fn start_pg(loglines: LogLines) -> eyre::Result<String> {
...
+
         let mut cmd = sudo_command(runas);
+
+        // when running the `postmaster` process via `sudo`, we need to copy the cargo/rust-related
+        // environment variables and pass as arguments to sudo, ahead of the `postmaster` command itself
+        //
+        // This ensures that in-process #[pg_test]s will see the `CARGO_xxx` envars they expect
+        for (var, value) in std::env::vars() {
+            if accept_envar(&var) {
+                let env_as_arg = format!("{var}={}", shlex::try_quote(&value)?);
```
workingjubilee, line 567: if we use shlex, we should just skip the env vars that we can't quote the values of correctly, *or* we should panic more noisily, or we should stop iterating the env vars entirely. here is one of those options.
```suggestion
                let Ok(quoted_val) = shlex::try_quote(&value) else {
                    continue
                };
                let env_as_arg = format!("{var}={quoted_val}");
```

pgrx-tests/src/framework.rs
```diff
@@ -546,7 +546,30 @@ fn start_pg(loglines: LogLines) -> eyre::Result<String> {
...
+                cmd.arg(env_as_arg);
+            }
+        }
```
workingjubilee, line 570: alternatively, instead of splitting and then reapplying these, why not use sudo's `--preserve-env=var,var,var,...`? are we worried about not having the required privilege?

workingjubilee (commented):
this looks plausible but it seems like we could avoid most of this logic if we use sudo's flags more fully.

pgrx-tests/src/framework.rs
eeeebbbbrrrr: I didn’t see that in the man page. Let me investigate that and make sure we can rely on it even being there here in 2023.

workingjubilee:
Needs a rebase, unforch.

PR #1675, opened 2024-04-23
feat: add bindings for WAL-related functions

rebasedming:
Added some header files that I needed while implementing a custom WAL resource manager.


workingjubilee:
Sure!

PR #1677, opened 2024-04-25
Run cargo-diet on things

thomcc:
I was curious about cargo diet so I ran it on the PGRX crates. (The other crates from the ones I ran it on didn't have any changes)


workingjubilee:
Thanks!

PR #1678, opened 2024-04-25
Factor out `extern_c_wrapper`

workingjubilee:
This briefly existed as a little closure inside `pg_extern/mod.rs`, but this is also more logic shared between `pg_extern` and `pg_trigger`. Handle them uniformly.

In so doing, notice an ABI violation for `void` returns and fix it!


eeeebbbbrrrr (approved):
This looks cool. And good catch on the ABI problem!

PR #1679, opened 2024-04-26
Believe in the typechecker in `mod pg_extern`

workingjubilee:
Do some assorted cleanup because Rust is a language with an actual typechecker, fairly limited coercion sites, and minimal coercions at those sites.

### DIFF
--- a/pgrx-sql-entity-graph/src/pg_extern/mod.rs
+++ b/pgrx-sql-entity-graph/src/pg_extern/mod.rs
@@ -547,7 +547,7 @@ impl PgExtern {
 }
 
 fn last_ident_is(ty: &syn::Type, id: &str) -> bool {
-    let syn::Type::Path(ty_path) = ty else { unimplemented!("queried a non-path!") };
+    let syn::Type::Path(ty_path) = ty else { return false };
     let syn::TypePath { path, .. } = ty_path;
     path.segments.last().is_some_and(|segment| segment.ident == id)
 }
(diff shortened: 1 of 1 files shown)


pgrx-sql-entity-graph/src/pg_extern/mod.rs
```diff
@@ -387,20 +383,16 @@ impl PgExtern {
...
-                quote_spanned! {pat.span()=>
+            match resolved_ty {
+                // There's no danger of misinterpreting the type, as pointer coercions must typecheck!
+                ty @ Type::Path(_) if last_ident_is(ty, "FunctionCallInfo") => quote_spanned! {pat.span()=>
                     let #pat = #fcinfo_ident;
-                }
-            } else if arg.used_ty.resolved_ty.to_token_stream().to_string() == quote!(()).to_token_stream().to_string() {
-                quote_spanned! {pat.span()=>
+                },
+                Type::Tuple(tup) if /* () */ tup.elems.is_empty() => quote_spanned! {pat.span()=>
```
thomcc: nit: delete the `/* () */`, and add spaces around `pat.span()`.

thomcc (commented):
Seems a bit fragile but I suppose you're right, that can't cause any problems since they'd just fail to typecheck.

pgrx-sql-entity-graph/src/pg_extern/mod.rs
workingjubilee:
```suggestion
                Type::Tuple(tup) if tup.elems.is_empty() => quote_spanned! { pat.span() =>
```

workingjubilee:
Yeah. Triple-checking type names for funsies might be okay if it was for better diagnostics, I guess? But as-is...

PR #1684, opened 2024-04-29
Memory management for Datum to String conversion

sardination:
Since `FromDatum` for `String` creates a Rust-allocated copy of the string data, it makes sense to free the palloc'ed varlena created when detoasting the Datum for string conversion. 

This PR also includes a function rename to match the function name in Postgres.

### DIFF
--- a/pgrx/src/datum/from.rs
+++ b/pgrx/src/datum/from.rs
@@ -412,22 +412,20 @@ impl FromDatum for String {
         is_null: bool,
         typoid: pg_sys::Oid,
     ) -> Option<String> {
-        if varlena::varatt_is_1b_e(datum.cast_mut_ptr::<pg_sys::varlena>())
-            || varlena::varatt_is_4b_c(datum.cast_mut_ptr::<pg_sys::varlena>())
+        let varlena = pg_sys::pg_detoast_datum_packed(datum.cast_mut_ptr());
+        let converted_varlena = convert_varlena_to_str_memoized(varlena);
+        let ret_string = converted_varlena.to_owned();
+
+        // If the datum is EXTERNAL or COMPRESSED, then detoasting creates a pfree-able chunk
+        // that needs to be freed. We can free it because `to_owned` above creates a Rust copy
+        // of the string.
+        if varlena::varatt_is_1b_e(datum.cast_mut_ptr())
+            || varlena::varatt_is_4b_c(datum.cast_mut_ptr())
         {
-            let varlena = pg_sys::pg_detoast_datum_packed(datum.cast_mut_ptr());
-            let converted_varlena = convert_varlena_to_str_memoized(varlena);
-            let ret_string = converted_varlena.to_owned();
-
-            // If the datum is EXTERNAL or COMPRESSED, then detoasting creates a pfree-able chunk
-            // that needs to be freed. We can free it because `to_owned` above creates a Rust copy
-            // of the string.
             pg_sys::pfree(varlena.cast());
-
-            Some(ret_string)
-        } else {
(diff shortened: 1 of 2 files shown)


pgrx/src/varlena.rs
```diff
@@ -169,7 +169,7 @@ pub unsafe fn varatt_is_4b_u(ptr: *const pg_sys::varlena) -> bool {
 /// ((((varattrib_1b *) (PTR))->va_header & 0x03) == 0x02)
 /// ```
 #[inline]
-pub unsafe fn varatt_is_b8_c(ptr: *const pg_sys::varlena) -> bool {
+pub unsafe fn varatt_is_4b_c(ptr: *const pg_sys::varlena) -> bool {
```
workingjubilee, line 172: Thanks for catching this!

pgrx/src/datum/from.rs
```diff
@@ -412,7 +412,22 @@ impl FromDatum for String {
...
+            || varlena::varatt_is_4b_c(datum.cast_mut_ptr::<pg_sys::varlena>())
+        {
+            let varlena = pg_sys::pg_detoast_datum_packed(datum.cast_mut_ptr());
+            let converted_varlena = convert_varlena_to_str_memoized(varlena);
+            let ret_string = converted_varlena.to_owned();
+
+            // If the datum is EXTERNAL or COMPRESSED, then detoasting creates a pfree-able chunk
+            // that needs to be freed. We can free it because `to_owned` above creates a Rust copy
+            // of the string.
+            pg_sys::pfree(varlena.cast());
```
workingjubilee: I'm not feeling at my best right now so perhaps doing code review is a bit ambitious, but I had half-forgotten we were nested inside this `if` by the point of this pfree, when I started to double-check the correctness of this pfree. I had to look back up. Not all code can be so "tidy" but I prefer it when simple implementations with high correctness concerns are in fact that.

So for legibility reasons, I'd prefer if this implementation proceeds *unconditionally* until the pfree, and only *then* checks a condition and then does the pfree based on that condition. Usually Postgres implements this as a pointer-equality check with a comment about how detoasting can duplicate, which also means it doesn't need to rely much on the implementation details of `pg_detoast_datum_packed`, so Postgres can add more conditionals or whatever, but I don't really care.

pgrx/src/datum/from.rs
```diff
@@ -412,7 +412,22 @@ impl FromDatum for String {
...
+
+            Some(ret_string)
+        } else {
+            FromDatum::from_polymorphic_datum(datum, is_null, typoid).map(|s: &str| s.to_owned())
```
workingjubilee: It also helps that by deferring the `if` until the pfree, we then don't have to cover an `else` by deferring to another implementation. We'd already be done two lines ago.

pgrx/src/datum/from.rs
```diff
@@ -412,7 +412,22 @@ impl FromDatum for String {
         is_null: bool,
         typoid: pg_sys::Oid,
     ) -> Option<String> {
-        FromDatum::from_polymorphic_datum(datum, is_null, typoid).map(|s: &str| s.to_owned())
+        if varlena::varatt_is_1b_e(datum.cast_mut_ptr::<pg_sys::varlena>())
+            || varlena::varatt_is_4b_c(datum.cast_mut_ptr::<pg_sys::varlena>())
```
workingjubilee: I believe these two shouldn't need to be turbofished, as the inference should be effectively infallible due to [function arguments being coercion sites](https://doc.rust-lang.org/reference/type-coercions.html#coercion-sites), which you'll notice is what happens on line 418. It's fine if they are, though.

workingjubilee (changes requested):

workingjubilee (approved):
Thanks!

PR #1686, opened 2024-04-29
Pass i64 and f64 by reference on 32-bit systems

### LINKED ISSUE #1685: Pass 64-bit types by reference on 32-bit platforms
Int64, UInt64 and Float8 are all passed by reference when the pointer size is less than 64 bit. In Postgres this is controlled by the `USE_FLOAT8_BYVAL` macro defined in [pg_config_manual.h](https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/include/pg_config_manual.h;h=f941ee2faf86b84a9d76b74269bec88d421fcbb7;hb=HEAD#l76), which is used when defining `DatumGetInt64`/`UIn64`/`Float8` and `Int64`/`UInt64`/`Float8GetDatum` in [postgres.h](https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/include/postgres.h;h=5d5fd7813e8086ae5ec148c9ba391b1ee836d040;hb=HEAD#l378) (and propagated to `FLOAT8PASSBYVAL`). The same logic should be implemented in `IntoDatum` and `FromDatum` for `i64`, `u64` and `f64`.

CoderPuppy:
Fixes #1685.

This determines whether to pass by value or by reference based on the size of `usize`, this should replicate the behavior of the `USE_FLOAT8_BYVAL` macro defined in [pg_config_manual.h](https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/include/pg_config_manual.h;h=f941ee2faf86b84a9d76b74269bec88d421fcbb7;hb=HEAD#l76).

### DIFF
--- a/pgrx/src/datum/from.rs
+++ b/pgrx/src/datum/from.rs
@@ -291,7 +291,13 @@ impl FromDatum for i64 {
         if is_null {
             None
         } else {
-            Some(datum.value() as _)
+            if std::mem::size_of::<usize>() >= 8 {
+              Some(datum.value() as _)
+            } else {
+              unsafe {
+                Some(*datum.cast_mut_ptr::<i64>())
+              }
+            }
         }
     }
 }
@@ -323,7 +329,14 @@ impl FromDatum for f64 {
         if is_null {
             None
         } else {
-            Some(f64::from_bits(datum.value() as _))
+            let value: u64 = if std::mem::size_of::<usize>() >= 8 {
+              datum.value() as _
+            } else {
+              unsafe {
+                *datum.cast_mut_ptr::<u64>()
+              }
+            };
+            Some(f64::from_bits(value))
         }
     }
 }
--- a/pgrx/src/datum/into.rs
+++ b/pgrx/src/datum/into.rs
@@ -175,7 +175,15 @@ impl IntoDatum for i32 {
 impl IntoDatum for i64 {
     #[inline]
     fn into_datum(self) -> Option<pg_sys::Datum> {
-        Some(pg_sys::Datum::from(self))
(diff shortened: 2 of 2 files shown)


workingjubilee:
You need to change far more than this.

workingjubilee (changes requested):
Requires tests that validate that pgrx works on a 32-bit system. There is zero point in merging something that breaks in a week because almost everyone who uses pgrx, especially the primary developers of it, does so on 64-bit systems and thus there is almost no testing on 32-bit systems.

workingjubilee:
I also kinda want to know exactly why you want to run a real-ass database on a 32-bit system, much less with custom extensions. Postgres does... many things... that assume that there's an entire gigabyte just lying around of addressable memory, which is often not true on 32-bit systems, or can only be true due to swap files paging in and out, which makes the process slow and inefficient.

workingjubilee:
I'm sorry, but I still can't merge this without adding a regression test for it. We're in the middle of a big change to the codebase that is going to plow right down the middle of these traits: they are not easy to use soundly in complex cases, and the codebase uses them as a bound for safe code in places where they do **not** assure soundness.

No matter how simple this change may seem, there are reasons I am asking for a test.

Even for this change, an example of how we have multiple locations that rely on this assumption is here:

https://github.com/pgcentralfoundation/pgrx/blob/bffc71d23f5a3a8a141f187da93a0eda981a8ac7/pgrx/src/datum/array.rs#L562-L571

I don't want to merge this and then have people complain every five seconds that I broke their code due to ghosts, because they managed to get a bit farther even though it was really never adequately supported in the first place.

PR #1687, opened 2024-05-01
Cargo pgrx upgrade

NotGyro:
This provides a `cargo pgrx upgrade` command that updates pgrx crate versions to the latest (optionally prerelease versions) or to a specified string.

### DIFF
--- a/cargo-pgrx/src/command/upgrade.rs
+++ b/cargo-pgrx/src/command/upgrade.rs
@@ -138,30 +138,39 @@ impl Upgrade {
             }
         };
 
-        if let Some(source) = parsed_dep.source().cloned() {
-            debug!(
-                "Found dependency {dep_name} with current \
-                source {source:#?}"
-            );
-            if let cargo_edit::Source::Workspace(_) = &source {
-                error!(
-                    "Cannot upgrade the version of \
-                        {dep_name} because it is set in the \
-                        workspace, please run `cargo pgrx \
-                        upgrade` in the workspace directory."
-                );
-            } else if get_dep_source_version(&source).is_some() {
-                replace_version(
-                    target_version.as_str(),
-                    path.as_path(),
-                    &mut key,
-                    dep,
-                    parsed_dep,
-                    source,
-                )?;
-            } else {
-                info!("No version specified for {dep_name}, not upgrading.");
+        // As of cargo-edit version 0.12.2, dep.source will always be a Some()
+        // if the Dependency struct was instantiated via from_toml().
+        let source = match parsed_dep.source().cloned() {
+            Some(src) => src,
+            None => {
+                return Err(eyre!(
+                    "Dependency {dep_name}'s source was \
+                parsed as None by cargo-edit."
+                ))
             }
+        };
+        debug!(
+            "Found dependency {dep_name} with current \
+            source {source:#?}"
+        );
+        if let cargo_edit::Source::Workspace(_) = &source {
+            error!(
+                "Cannot upgrade the version of \
+                    {dep_name} because it is set in the \
+                    workspace, please run `cargo pgrx \
+                    upgrade` in the workspace directory."
+            );
+        } else if get_dep_source_version(&source).is_some() {
+            replace_version(
+                target_version.as_str(),
+                path.as_path(),
+                &mut key,
+                dep,
+                parsed_dep,
+                source,
+            )?;
+        } else {
+            info!("No version specified for {dep_name}, not upgrading.");
         }
         Ok(())
     }
--- a/cargo-pgrx/src/command/upgrade.rs
+++ b/cargo-pgrx/src/command/upgrade.rs
@@ -43,7 +43,7 @@ pub(crate) struct Upgrade {
     /// Dry-run - if this flag is set, Cargo.toml will not be modified.
     /// Instead, this command will print the text of the new Cargo.toml
     /// that would have been generated if it was modified.
-    #[clap(long = "dry-run", short = "n")]
+    #[clap(long = "dry-run", short = 'n')]
     pub(crate) dry_run: bool,
 }
 
@@ -125,7 +125,7 @@ impl Upgrade {
             .map_err(|e| eyre!("Unable to fetch registry URL for path: {e}"))?;
         update_registry_index(&reg_url, false)
             .map_err(|e| eyre!("Unable to update registry index: {e}"))?;
-        let target_version = match self.target_version {
+        let target_version = match self.to {
             Some(ref ver) => Some(ver.clone()),
             None => cargo_edit::get_latest_dependency(
                 dep_name,
@@ -208,7 +208,7 @@ impl Upgrade {
         Ok(())
     }
     fn process_manifest(&self, path: &PathBuf, manifest: &mut LocalManifest) -> eyre::Result<()> {
-        const RELEVANT_PACKAGES: [&str; 8] = [
+        const RELEVANT_PACKAGES: [&str; 6] = [
             "pgrx",
             "pgrx-macros",
             "pgrx-pg-config",
@@ -290,8 +290,7 @@ impl CommandExecute for Upgrade {
                                 "Failed to canonicalize path, no \
                                 parent directory found."
                             ))?;
-                            let member_path =
-                                root_path.join(PathBuf::from(member)).join("Cargo.toml");
+                            let member_path = root_path.join(member).join("Cargo.toml");
                             debug!("Generated child path {:#?}", &member_path);
                             child_path_maybe = Some(member_path);
                         }
--- a/cargo-pgrx/src/command/upgrade.rs
+++ b/cargo-pgrx/src/command/upgrade.rs
@@ -9,8 +9,8 @@
 //LICENSE Use of this source code is governed by the MIT license that can be found in the LICENSE file.
 use cargo_edit::{registry_url, Dependency};
 use eyre::eyre;
-use toml_edit::KeyMut;
 use std::path::{Path, PathBuf};
+use toml_edit::KeyMut;
 use tracing::{debug, error, info, warn};
 
 use crate::CommandExecute;
@@ -97,19 +97,24 @@ where
     parsed_dep.update_toml(crate_root, key, dep);
     Ok(())
 }
-impl Upgrade { 
-    fn update_dep(&self, path: &PathBuf, mut key: KeyMut, dep: &mut toml_edit::Item) -> eyre::Result<()> { 
+impl Upgrade {
+    fn update_dep(
+        &self,
+        path: &PathBuf,
+        mut key: KeyMut,
+        dep: &mut toml_edit::Item,
+    ) -> eyre::Result<()> {
         let dep_name_string = key.get().to_string();
         let dep_name = dep_name_string.as_str();
         let parsed_dep: Dependency = match Dependency::from_toml(path.as_path(), dep_name, dep) {
-                Ok(dependency) => dependency,
-                Err(e) => {
-                    return Err(eyre!(
-                        "Could not parse dependency \
+            Ok(dependency) => dependency,
+            Err(e) => {
+                return Err(eyre!(
+                    "Could not parse dependency \
                 entry for {dep_name} due to error: {e}"
-                    ))
-                }
-            };
+                ))
+            }
+        };
         let reg_url = registry_url(path, parsed_dep.registry())
             .map_err(|e| eyre!("Unable to fetch registry URL for path: {e}"))?;
         let target_version = match self.target_version {
@@ -140,7 +145,7 @@ impl Upgrade {
                 ))
             }
         };
-    
+
         if let Some(source) = parsed_dep.source().cloned() {
             debug!(
                 "Found dependency {dep_name} with current \
@@ -175,7 +180,8 @@ impl CommandExecute for Upgrade {
     fn execute(self) -> eyre::Result<()> {
         const RELEVANT_PACKAGES: [&str; 3] = ["pgrx", "pgrx-macros", "pgrx-tests"];
         // Canonicalize because cargo-edit does not accept relative paths.
-        let path = std::fs::canonicalize(self.path.clone().unwrap_or(PathBuf::from("./Cargo.toml")))?;
+        let path =
+            std::fs::canonicalize(self.path.clone().unwrap_or(PathBuf::from("./Cargo.toml")))?;
 
         let mut manifest = cargo_edit::LocalManifest::find(Some(&path))
             .map_err(|e| eyre!("Error opening manifest: {e}"))?;
@@ -187,14 +193,10 @@ impl CommandExecute for Upgrade {
                     self.update_dep(&path, key, dep)?;
                     // Workaround since update_toml() doesn't preserve comments
                     dep_table.key_decor_mut(dep_name).map(|dec| {
-                        if let Some(prefix) =
-                            decor.as_ref().and_then(|val| val.prefix().cloned())
-                        {
+                        if let Some(prefix) = decor.as_ref().and_then(|val| val.prefix().cloned()) {
                             dec.set_prefix(prefix)
                         }
-                        if let Some(suffix) =
-                            decor.as_ref().and_then(|val| val.suffix().cloned())
-                        {
+                        if let Some(suffix) = decor.as_ref().and_then(|val| val.suffix().cloned()) {
                             dec.set_suffix(suffix)
                         }
                     });
--- a/cargo-pgrx/src/command/upgrade.rs
+++ b/cargo-pgrx/src/command/upgrade.rs
@@ -1,3 +1,12 @@
+//LICENSE Portions Copyright 2019-2021 ZomboDB, LLC.
+//LICENSE
+//LICENSE Portions Copyright 2021-2023 Technology Concepts & Design, Inc.
+//LICENSE
+//LICENSE Portions Copyright 2023-2023 PgCentral Foundation, Inc. <contact@pgcentral.org>
+//LICENSE
+//LICENSE All rights reserved.
+//LICENSE
+//LICENSE Use of this source code is governed by the MIT license that can be found in the LICENSE file.
 use cargo_edit::{registry_url, Dependency};
 use eyre::eyre;
 use toml_edit::KeyMut;
--- a/cargo-pgrx/src/command/upgrade.rs
+++ b/cargo-pgrx/src/command/upgrade.rs
@@ -1,4 +1,8 @@
-use crate::CommandExecute;
+use std::{fs::OpenOptions, io::Read, path::PathBuf};
+use eyre::eyre;
+use toml_edit::{Document, DocumentMut};
+
+use crate::{manifest, CommandExecute};
 
 
 /// Upgrade pgrx crate versions in `Cargo.toml`.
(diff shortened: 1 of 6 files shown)


cargo-pgrx/src/command/upgrade.rs
```diff
@@ -0,0 +1,216 @@
...
+pub(crate) struct Upgrade {
+    /// Specify a version to upgrade to, rather than defaulting to the latest
+    /// version.
+    #[clap(long, short)]
+    pub(crate) target_version: Option<String>, //TODO: typed version not string
+
+    /// Path to the manifest file (usually Cargo.toml). Defaults to
+    /// "./Cargo.toml" in the working directory.
+    #[clap(long, short)]
+    pub(crate) path: Option<PathBuf>,
```
jyn514: the other subcommands, and cargo itself, call this `--manifest-path` - let's do the same here for consistency. i thought this was referring to the package directory at first.

cargo-pgrx/src/command/upgrade.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    // Flag to permit pre-release builds
+    #[clap(long, short)]
+    pub(crate) allow_prerelease: bool,
```
jyn514: hmm. so, cargo-upgrade doesn't have a flag that corresponds to this - it will only search stable versions unless you explicitly say the version of the pre-release you want to upgrade to. i kinda like that behavior, i think people should either know which pre-release they're going to use or just use stable. can you talk more about the use case for "the latest pre-release"?

cargo-pgrx/src/command/upgrade.rs
```diff
@@ -0,0 +1,216 @@
...
+
+/// Upgrade pgrx crate versions in `Cargo.toml`.
+/// Defaults to latest.
+#[derive(clap::Args, Debug)]
+#[clap(author)]
+pub(crate) struct Upgrade {
+    /// Specify a version to upgrade to, rather than defaulting to the latest
+    /// version.
+    #[clap(long, short)]
+    pub(crate) target_version: Option<String>, //TODO: typed version not string
```
jyn514: oh interesting - so here we don't have a direct analogue to `cargo-upgrade`'s flag, it uses `-p, --package <PKGID[@<VERSION>]>`, but we already know we're upgrading all the pgrx crates at once, so -p doesn't make sense for us.

i wonder if it makes sense to just have an optional positional argument here instead? so instead of `cargo pgrx upgrade -t 0.12` it would be `pgrx upgrade 0.12`. just a little quality of life thing.

cargo-pgrx/src/command/upgrade.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    // Flag to permit pre-release builds
+    #[clap(long, short)]
+    pub(crate) allow_prerelease: bool,
+}
+trait DependencySource {
+    fn set_version<A: AsRef<str>>(&mut self, new_version: A);
+    fn get_version(&self) -> Option<&String>;
+}
+impl DependencySource for cargo_edit::Source {
```
jyn514: nit: this is the only implementation for the trait, and you only use it once below (in `replace_version`). i assume it's a trait because you can't add inherent functions to `Source` since it's in another crate. i kinda would prefer to make `set_version` and `get_version` free functions instead to make it clear it's not generic over the source.

cargo-pgrx/src/command/upgrade.rs
```diff
@@ -0,0 +1,216 @@
...
+        const RELEVANT_PACKAGES: [&str; 3] = ["pgrx", "pgrx-macros", "pgrx-tests"];
+        // Canonicalize because cargo-edit does not accept relative paths.
+        let path =
+            std::fs::canonicalize(self.path.clone().unwrap_or(PathBuf::from("./Cargo.toml")))?;
+
+        let mut manifest = cargo_edit::LocalManifest::find(Some(&path))
+            .map_err(|e| eyre!("Error opening manifest: {e}"))?;
+
+        for dep_table in manifest.get_dependency_tables_mut() {
+            for dep_name in RELEVANT_PACKAGES {
```
jyn514, line 221: hmm. so, i tried this on a workspace that has pgrx in a *member*, but not in the workspace *root*. right now the way `pgrx upgrade` treats that is to silently do nothing. maybe we could add a warning about that? or even add a `-p` flag, analogous to `cargo build -p`, that says "update the dependencies for this member of the workspace"?

cargo-pgrx/src/command/upgrade.rs
```diff
@@ -0,0 +1,216 @@
...
+                return Err(eyre!(
+                    "Could not parse dependency \
+                entry for {dep_name} due to error: {e}"
+                ))
+            }
+        };
+        let reg_url = registry_url(path, parsed_dep.registry())
+            .map_err(|e| eyre!("Unable to fetch registry URL for path: {e}"))?;
+        let target_version = match self.target_version {
+            Some(ref ver) => Some(ver.clone()),
```
jyn514: this will happily let the user choose a version that doesn't exist:
```
$ cargo pgrx upgrade -t 0.12.0
$ cargo check
error: failed to select a version for the requirement `pgrx = "^0.12.0"`
candidate versions found which didn't match: 0.12.0-alpha.1, 0.12.0-alpha.0, 0.11.4, ...
location searched: crates.io index
required by package `example v0.1.0 (/home/jyn/work/example)`
if you are looking for the prerelease package it needs to be specified explicitly
    pgrx = { version = "0.12.0-alpha.1" }
```
can we add a check that it's in the registry and error out if not? that way we don't write files to disk that won't compile.

cargo-pgrx/src/command/upgrade.rs
```diff
@@ -0,0 +1,216 @@
...
+            None => {
+                return Err(eyre!(
+                    "Unable to update {dep_name} \
+                , no provided crate version and a latest version \
+                could not be retrieved from the registry."
+                ))
+            }
+        };
+
+        if let Some(source) = parsed_dep.source().cloned() {
```
jyn514: hmm, when is it possible for the source to be `None`?

cargo-pgrx/src/command/upgrade.rs
```diff
@@ -0,0 +1,216 @@
...
+trait DependencySource {
+    fn set_version<A: AsRef<str>>(&mut self, new_version: A);
+    fn get_version(&self) -> Option<&String>;
+}
+impl DependencySource for cargo_edit::Source {
+    fn set_version<A: AsRef<str>>(&mut self, new_version: A) {
+        match self {
+            cargo_edit::Source::Registry(reg) => reg.version = new_version.as_ref().to_string(),
+            cargo_edit::Source::Path(path) => path.version = Some(new_version.as_ref().to_string()),
+            cargo_edit::Source::Git(git) => git.version = Some(new_version.as_ref().to_string()),
```
jyn514: hmm. i am not sure the behavior for the registry and git sources is what we want. take for instance a git dependency like this:
```toml
pgrx = { git = "https://github.com/pgcentralfoundation/pgrx/" }
```
upgrade says this about it:
```
$ pgrx upgrade
INFO cargo_pgrx::command::upgrade: No version specified for pgrx, not upgrading.
```
that seems ok. but if there is a version specified, it does this:
```diff
-pgrx = { git = "https://github.com/pgcentralfoundation/pgrx/", version = "0.12.0-alpha.1" }
+pgrx = { git = "https://github.com/pgcentralfoundation/pgrx/", version = "0.11.3" }
```
and then the next time we run cargo, it fails:
```
error: failed to select a version for the requirement `pgrx = "^0.11.3"`
candidate versions found which didn't match: 0.12.0-alpha.1
```
maybe we should just error out for git or path dependencies, because we can't control their version? admittedly using a version requirement with `git` is already kinda weird and maybe the current behavior is what the user expects.

jyn514 (commented):
this is neat! good call on using cargo-edit, that's doing most of the heavy lifting for us :)

overall i like this approach, but i had a few comments about the user interface. anything prefixed with "nit:" is non-blocking and i don't feel strongly about it.

cargo-pgrx/src/command/upgrade.rs
NotGyro: Well, the first use-case I can think of is dogfooding - I can imagine while testing our own projects it could be handy. However, I can also understand why most end-users would be unlikely to use this feature. Is there any downside to offering it as a feature?

cargo-pgrx/src/command/upgrade.rs
NotGyro: I was considering adding an `-r` as in recursive flag along those lines. It might be a good idea. That's a larger refactor of this PR but I'll look into what would be needed.

cargo-pgrx/src/command/upgrade.rs
NotGyro: That sounds reasonable, I'll give it a look.

cargo-pgrx/src/command/upgrade.rs
NotGyro: That is a bit of an oddity, yes. I could also see it as reasonable to use the current version found in Cargo.toml at the source path - parse the manifest at the path specified, and then use the package version found in that file. 

Whereas it should probably print an error upon encountering git dependencies rather than trying to upgrade them.

cargo-pgrx/src/command/upgrade.rs
NotGyro: Resolved in 4dbf3e6cfdefd55d7fb6154801593c840a0fa68a

cargo-pgrx/src/command/upgrade.rs
NotGyro: Resolved in 4dbf3e6cfdefd55d7fb6154801593c840a0fa68a

cargo-pgrx/src/command/upgrade.rs
NotGyro: This is implemented within `cargo-edit`, it looks like it initializes as `None` and is filled in later as the manifest file is parsed. Reading `cargo-edit`'s source code at https://github.com/killercup/cargo-edit/blob/master/src/dependency.rs#L269, it looks like `source` will always be `Some()` if the `Dependency` struct is instantiated by `from_toml()` This might just be an API design mistake on their part, or there might be some deeper reason I cannot spot here.

NotGyro: Resolved in d0f202cb

cargo-pgrx/src/command/upgrade.rs
NotGyro: After a conversation with @thomcc  on Discord it seems safer to throw a warning upon trying to upgrade a path or git repo dependency, so that the program doesn't modify Cargo.toml under incorrect assumptions. It's probably better for software development tools to avoid trying to be too clever.

NotGyro: Resolved in [2c9e997](https://github.com/pgcentralfoundation/pgrx/pull/1687/commits/2c9e997bcf521aa26c8a75d12193a70ca7e8a517)

cargo-pgrx/src/command/upgrade.rs
NotGyro: Validating this would require yet another dependency, namely [crates_index](https://docs.rs/crates-index/latest/crates_index/). It might be worth it, though.

cargo-pgrx/src/command/upgrade.rs
```diff
@@ -0,0 +1,235 @@
...
+            }
+        }
+        Ok(())
+    }
+}
+
+impl CommandExecute for Upgrade {
+    #[tracing::instrument(level = "error", skip(self))]
+    fn execute(self) -> eyre::Result<()> {
+        const RELEVANT_PACKAGES: [&str; 3] = ["pgrx", "pgrx-macros", "pgrx-tests"];
```
thomcc: I guess we don't need `pgrx-pg-sys` here since people should get it through `pgrx::pg_sys`. This list seems a little fragile, though.

cargo-pgrx/src/command/upgrade.rs
```diff
@@ -0,0 +1,235 @@
...
+                a package because it is set in the \
+                workspace, please run `cargo pgrx \
+                upgrade` in the workspace directory."
+            );
+        }
+    }
+}
+fn get_dep_source_version(src: &cargo_edit::Source) -> Option<&String> {
+    match src {
+        cargo_edit::Source::Registry(reg) => Some(&reg.version),
```
thomcc, line 67: Should we ignore non-crates.io registries? Admittedly, I doubt it's common (but who knows what weird uses enterprise folks are getting up to with pgrx).

thomcc (commented):
I noted a couple of concerns but for the most part this looks fine. That is, I wouldn't be afraid to run this, and the code doesn't make me want to shriek in terror (and/or immediately open followups). It's actually simpler than I expected, which is nice.

That said, I think you should still wait for @jyn514's review if possible (although this depends on when they're coming back I suppose), as they've already done a lot of review on this and have stronger opinions on the feature than I do.

cargo-pgrx/src/command/upgrade.rs
thomcc: Seems like overkill to ask the registry for this. I don't feel that strongly about it, but I think it's fine to just do what the user asked. (They can run again if they messed up the first time)

That said, if you disagree, feel free to go for it. It wouldn't hurt, I just don't really think it's needed.

cargo-pgrx/src/command/upgrade.rs
workingjubilee: We should still bump pgrx-pg-sys if we find it.

workingjubilee: Same with pgrx-sql-entity-graph, or pgrx-pg-config, or...

cargo-pgrx/src/command/upgrade.rs
workingjubilee: If enterprise users are running their own registry and run `cargo pgrx upgrade`, they probably believe they have their own applicable versions of that crate (or their registry will fetch and cache the crates.io registry), so we should treat them equally.

cargo-pgrx/src/command/upgrade.rs
workingjubilee: So I think we probably _should_ make it smarter, but this might be the wrong direction, so I think we can defer making it smarter until we get some user experience reports.

cargo-pgrx/src/command/upgrade.rs
workingjubilee: This assumption of course fails if enterprise software developers call the tool even when it is not applicable in their environment, and enterprise registry operators do not decide to make software that interoperates correctly (by usefully emulating crates.io or emitting useful errors). But otherwise, we should simply reward everyone for playing along. A custom registry *should* behave correctly.

cargo-pgrx/src/command/upgrade.rs
workingjubilee: I'd like this to land with the ability to point it at a specific package in a workspace, at least. It doesn't have to have a full-on recursive WalkDir logic.

cargo-pgrx/src/command/upgrade.rs
```diff
@@ -0,0 +1,235 @@
...
+    #[clap(long, short)]
+    pub(crate) target_version: Option<String>, //TODO: typed version not string
+
+    /// Path to the manifest file (usually Cargo.toml). Defaults to
+    /// "./Cargo.toml" in the working directory.
+    #[clap(long = "manifest-path", short)]
+    pub(crate) manifest_path: Option<PathBuf>,
+
+    // Flag to permit pre-release builds
+    #[clap(long, short)]
```
workingjubilee, line 40: hmm, what is the short form here? I don't have the logic for short forms ingrained in my brain...

cargo-pgrx/src/command/upgrade.rs
workingjubilee: eh? while reading this conversation, I wondered if I slightly misinterpreted the name of this. it first seemed that this flag was required for using a prerelease at all? is it only for setting the latest prerelease?

maybe should be `--include-prereleases` instead? or just `--prerelease`

workingjubilee (commented):
Ooh, nice. I'm pretty happy with this implementation from what I can see. It will hopefully get poked pretty hard under use and we can improve it from there.

I've seen "workspace that has a Rust library and a pgrx extension that essentially reexports the Rust library" as a common pattern, nevermind the "workspace with multiple extensions" case like the pgrx repo. So this should probably land with the `--package` flag so we have a full (albeit minimal) support for such workspaces.

cargo-pgrx/src/command/upgrade.rs
NotGyro: Oh that's less ambiguous naming, I agree.

cargo-pgrx/src/command/upgrade.rs
NotGyro: It's first-letter, so manifest_path would be `-m`

workingjubilee: I kinda think the prerelease flag shouldn't have a short form then.

cargo-pgrx/src/command/upgrade.rs
NotGyro: @workingjubilee Should this be `--package` or would that be ambiguous / seem like it's referring to the package to update?

cargo-pgrx/src/command/upgrade.rs
NotGyro: This should be dealt with in the latest revision. I've renamed it to "include-prereleases" and disabled the short-form argument (which would be "-i")

cargo-pgrx/src/command/upgrade.rs
NotGyro: Right now I've got an implementation of this, which needs review.

cargo-pgrx/src/command/upgrade.rs
NotGyro: Fair, leaving this alone for now.

cargo-pgrx/src/command/upgrade.rs
NotGyro: These should be present in the latest revision, but it'd be good to double-check this.

cargo-pgrx/src/command/upgrade.rs
workingjubilee: > @workingjubilee Should this be --package or would that be ambiguous / seem like it's referring to the package to update?

...oof, uh, idk.

workingjubilee: I think `--package` is unfortunately probably what people will expect here, even though it's confusing, and we can revise on user experience reports.

cargo-pgrx/src/command/upgrade.rs
```diff
@@ -0,0 +1,331 @@
...
+    /// Select a package within the current workspace within which to upgrade
+    /// pgrx package versions. Defaults to the root manifest in the working
+    /// directory.
+    #[clap(long, short)]
+    pub(crate) package: Option<String>,
+
+    /// Dry-run - if this flag is set, Cargo.toml will not be modified.
+    /// Instead, this command will print the text of the new Cargo.toml
+    /// that would have been generated if it was modified.
+    #[clap(long = "dry-run", short)]
```
workingjubilee: Cargo always uses this for its "dry run" short form, idk why.
```suggestion
    #[clap(long = "dry-run", short = "n")]
```

cargo-pgrx/src/command/upgrade.rs
```diff
@@ -0,0 +1,331 @@
...
+
+/// Upgrade pgrx crate versions in `Cargo.toml`.
+/// Defaults to latest.
+#[derive(clap::Args, Debug)]
+#[clap(author)]
+pub(crate) struct Upgrade {
+    /// Specify a version to upgrade to, rather than defaulting to the latest
+    /// version.
+    #[clap(long, short)]
+    pub(crate) target_version: Option<String>, //TODO: semver not string
```
workingjubilee: we might benefit from this just being
```suggestion
    #[clap(long)]
    pub(crate) to: Option<String>, //TODO: semver not string
```

cargo-pgrx/src/command/upgrade.rs
```diff
@@ -0,0 +1,331 @@
...
+
+    /// Path to the manifest file (usually Cargo.toml). Defaults to
+    /// "./Cargo.toml" in the working directory.
+    #[clap(long = "manifest-path", short)]
```
workingjubilee, line 30: Cargo doesn't support `-m` as a short for `--manifest-path`, but I kinda _don't_ want to tell you to change this because I think it's probably a case where they should. @_@ However, I wanted to make sure you know that.

cargo-pgrx/src/command/upgrade.rs
```diff
@@ -279,27 +279,50 @@ impl CommandExecute for Upgrade {
...
+                    // to do a little bit of finagling here. 
+                    let mut child_path_maybe = None;
+                    for member in &work.members {
+                        if member == &package { 
+                            // Canonicalized, should have a parent path
+                            let root_path = path.parent()
+                                .ok_or(eyre!("Failed to canonicalize path, no \
+                                parent directory found."))?;
+                            let member_path = root_path
+                                .join(PathBuf::from(member))
```
workingjubilee: Converting `member` into a `PathBuf` here shouldn't be necessary because the arg type of `Path::join` is `impl AsRef<Path>`, and this is a String. Right?

cargo-pgrx/src/command/upgrade.rs
```diff
@@ -0,0 +1,331 @@
...
+    fn process_manifest(&self, path: &PathBuf, manifest: &mut LocalManifest) -> eyre::Result<()> {
+        const RELEVANT_PACKAGES: [&str; 8] = [
+            "cargo-pgrx",
+            "pgrx",
+            "pgrx-macros",
+            "pgrx-pg-config",
+            "pgrx-pg-sys",
+            "pgrx-sql-entity-graph",
+            "pgrx-tests",
+            "pgrx-version-updater",
```
workingjubilee: This package actually isn't lockstepped with the rest of our workspace:
```suggestion
```

cargo-pgrx/src/command/upgrade.rs
```diff
@@ -0,0 +1,331 @@
...
+                        workspace, please run `cargo pgrx \
+                        upgrade` in the workspace directory."
+                )
+            }
+        }
+        Ok(())
+    }
+    fn process_manifest(&self, path: &PathBuf, manifest: &mut LocalManifest) -> eyre::Result<()> {
+        const RELEVANT_PACKAGES: [&str; 8] = [
+            "cargo-pgrx",
```
workingjubilee: Hmm. I don't think we need to worry about supporting people depending on cargo-pgrx as a crate.
```suggestion
```

workingjubilee (approved):
Some lingering comments that emerged from reviewing things, but aside from that, seems good!

cargo-pgrx/src/command/upgrade.rs
NotGyro: After some discussion with Jubilee, we went with `--to`

workingjubilee:
Nice! Looks good to me!

PR #1690, opened 2024-05-03
Ignore clippy when clippy is wrong

workingjubilee:


eeeebbbbrrrr:
#![disallow(clippy:wrong)]

PR #1692, opened 2024-05-03
Break `srf_next` into component fn

workingjubilee:
*Most* of the code between `TableIterator::srf_next` and `SetOfIterator::srf_next` is duplicated. Unify them by breaking each into smaller parts, so that the components that are used by both can be the same.


eeeebbbbrrrr (approved):
LGTM.  It looks like it follows the same execution paths as the original code.

PR #1696, opened 2024-05-07
Use UnsafeCell for PgLwLock and PgAtomic

### LINKED ISSUE #1223: PgLwLock breaks postgres reinitialization after backend crash

JelteF:
As explained in detail in #1223, due to the Postgres restart-on-crash logic, `PgAtomic<T>::attach` and `PgLwLock<T>::attach` would be called again after a crash. Using OnceCell prevented the restarted extension from attaching new PgLwLocks, so Postgres failed to restart after a crash. This addresses that by using UnsafeCell instead of OnceCell and marking `attach` as unsafe.

Fixes #1223

### DIFF
--- a/pgrx/src/atomics.rs
+++ b/pgrx/src/atomics.rs
@@ -7,15 +7,16 @@
 //LICENSE All rights reserved.
 //LICENSE
 //LICENSE Use of this source code is governed by the MIT license that can be found in the LICENSE file.
-use once_cell::sync::OnceCell;
+#![deny(unsafe_op_in_unsafe_fn)]
+use std::cell::UnsafeCell;
 
 pub struct PgAtomic<T> {
-    inner: OnceCell<*mut T>,
+    inner: UnsafeCell<*mut T>,
 }
 
 impl<T> PgAtomic<T> {
     pub const fn new() -> Self {
-        Self { inner: OnceCell::new() }
+        Self { inner: UnsafeCell::new(std::ptr::null_mut()) }
     }
 }
 
@@ -23,14 +24,15 @@ impl<T> PgAtomic<T>
 where
     T: atomic_traits::Atomic + Default,
 {
-    pub fn attach(&self, value: *mut T) {
-        self.inner.set(value).expect("This PgAtomic is not empty, can't re-attach");
+    /// SAFETY: Must only be called from inside the Postgres shared memory init hook
+    pub unsafe fn attach(&self, value: *mut T) {
+        unsafe {
+            *self.inner.get() = value;
+        }
     }
 
     pub fn get(&self) -> &T {
-        unsafe {
-            self.inner.get().expect("This PgAtomic as not been initialized").as_ref().unwrap()
-        }
+        unsafe { (*self.inner.get()).as_ref().expect("PgAtomic was not initialized") }
     }
 }
 
--- a/pgrx/src/shmem.rs
+++ b/pgrx/src/shmem.rs
@@ -7,6 +7,7 @@
 //LICENSE All rights reserved.
 //LICENSE
 //LICENSE Use of this source code is governed by the MIT license that can be found in the LICENSE file.
+#![deny(unsafe_op_in_unsafe_fn)]
 use crate::lwlock::*;
 use crate::{pg_sys, PgAtomic};
 use std::hash::Hash;
@@ -65,8 +66,8 @@ macro_rules! pg_shmem_init {
                     if let Some(i) = PREV_SHMEM_STARTUP_HOOK {
                         i();
                     }
+                    $thing.shmem_init();
                 }
-                $thing.shmem_init();
             }
         }
     };
@@ -87,8 +88,8 @@ macro_rules! pg_shmem_init {
                     if let Some(i) = PREV_SHMEM_REQUEST_HOOK {
                         i();
                     }
+                    $thing.pg_init();
                 }
-                $thing.pg_init();
             }
         }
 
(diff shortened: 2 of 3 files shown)


pgrx/src/shmem.rs
```diff
@@ -184,12 +187,11 @@ impl PgSharedMem {
...
             lock.attach(fv_shmem);
             pg_sys::LWLockRelease(addin_shmem_init_lock);
-        }
     }
 
     /// Must be run from the shared memory init hook, use for rust atomics behind `PgAtomic`
-    pub fn shmem_init_atomic<T: atomic_traits::Atomic + Default>(atomic: &PgAtomic<T>) {
-        unsafe {
+    /// SAFETY: Can only be called from Postgres its shared memory init hook
+    pub unsafe fn shmem_init_atomic<T: atomic_traits::Atomic + Default>(atomic: &PgAtomic<T>) {
```
workingjubilee: We can put this unsafe block back, and in so doing not have to reformat everything.
```suggestion
    /// # Safety
    /// Must only be called from Postgres its shared memory init hook
    pub unsafe fn shmem_init_atomic<T: atomic_traits::Atomic + Default>(atomic: &PgAtomic<T>) {
        unsafe {
```

pgrx/src/shmem.rs
```diff
@@ -206,7 +208,6 @@ impl PgSharedMem {
             let atomic = T::default();
             std::ptr::copy(&atomic, fv_shmem, 1);
             pg_sys::LWLockRelease(addin_shmem_init_lock);
```
workingjubilee:
```suggestion
            pg_sys::LWLockRelease(addin_shmem_init_lock);
        }
```

pgrx/src/shmem.rs
```diff
@@ -168,9 +171,9 @@ impl PgSharedMem {
     }
 
     /// Must be run from the shared memory init hook, use for types which are guarded by a `LWLock`
-    pub fn shmem_init_locked<T: Default + PGRXSharedMemory>(lock: &PgLwLock<T>) {
+    /// SAFETY: Can only be called from Postgres its shared memory init hook
+    pub unsafe fn shmem_init_locked<T: Default + PGRXSharedMemory>(lock: &PgLwLock<T>) {
```
workingjubilee: same here.

pgrx/src/shmem.rs
workingjubilee, line 1: Please add
```rust
```
if we're going to touch all the unsafe blocks in this file.

workingjubilee (changes requested):

pgrx/src/shmem.rs
JelteF: All fixed now.

workingjubilee: nice.

pgrx/src/atomics.rs
```diff
@@ -7,30 +7,32 @@
...
     }
 }
 
 impl<T> PgAtomic<T>
 where
     T: atomic_traits::Atomic + Default,
 {
-    pub fn attach(&self, value: *mut T) {
-        self.inner.set(value).expect("This PgAtomic is not empty, can't re-attach");
+    /// SAFETY: Can only be called from Postgres its shared memory init hook
```
workingjubilee: ...was slow to notice this typo. The extra "its"
```suggestion
    /// SAFETY: Must only be called from inside the Postgres shared memory init hook
```

workingjubilee (approved):
Ah, I had originally thought it was best to not change the safety of these fn unnecessarily, which is why I had dawdled to work out a slightly more sophisticated solution, but if this is what you would prefer, then that is fine. It's plausible you might be the heaviest user of this particular API currently, if my guess is correct.

By my evaluation, with the new safety constraint, this API is no more or less sound than it was.

One last nit, typo in the safety comment, and we'll be off to the races.

pgrx/src/atomics.rs
JelteF: Applied everywhere

JelteF:
> Ah, I had originally thought it was best to not change the safety of these fn unnecessarily, which is why I had dawdled to work out a slightly more sophisticated solution, but if this is what you would prefer, then that is fine.

Obviously safe functions are always nicer, but the nice thing is that the `pg_shmem_init!` macro is still marked as safe to call, as well as all the functions to lock/unlock the values. All the newly marked unsafe functions shouldn't really ever be called directly by people.

> It's plausible you might be the heaviest user of this particular API currently, if my guess is correct.

Possibly, and even I don't use it heavily.

> By my evaluation, with the new safety constraint, this API is no more or less sound than it was.

I'm not a huge expert in soundness of rust, but that was my assessment as well: I'm not 100% sure if it was sound before with all possible usages, but I don't think it is any worse than it is now. And now at least Postgres can restart after a crash.

workingjubilee:
> Obviously safe functions are always nicer, but the nice thing is that the `pg_shmem_init!` macro is still marked as safe to call,

This is the thing that I am least confident of its correctness. But even if it does, that's because of a preexisting problem, so it shouldn't bar this PR. I am pretty sure it would at most be a one-time `unsafe` block with a trivial condition that amounts to "uh... don't... don't write really weird and complicated init hooks?" which is easy enough to satisfy.

> as well as all the functions to lock/unlock the values.

Yeah, these should always remain safe, and those are the ones that will actually be in most code.

PR #1699, opened 2024-05-10
Remove seq-macro

workingjubilee:
This adds some local boilerplate but removes an external proc-macro from our build.

### DIFF
--- a/pgrx/src/iter.rs
+++ b/pgrx/src/iter.rs
@@ -10,7 +10,7 @@
 #![allow(clippy::vec_init_then_push)]
 use std::iter::once;
 
-use crate::IntoHeapTuple;
+use crate::{pg_sys, IntoDatum, IntoHeapTuple};
 use pgrx_sql_entity_graph::metadata::{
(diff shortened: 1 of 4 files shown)


eeeebbbbrrrr (approved):
👀 

Fair.

pgrx/src/iter.rs
```diff
@@ -155,31 +155,120 @@ impl<'a, T> Iterator for TableIterator<'a, T> {
...
+                let datums = [$($C.into_datum(),)*];
+                let mut nulls = datums.map(|option| option.is_none());
+                let mut datums = datums.map(|option| option.unwrap_or(pg_sys::Datum::from(0)));
+
+
+                unsafe {
+                    // SAFETY:  Caller has asserted that `tupdesc` is valid, and we just went
+                    // through a little bit of effort to setup properly sized arrays for
+                    // `datums` and `nulls`
+                    pg_sys::heap_form_tuple(tupdesc, datums.as_mut_ptr(), nulls.as_mut_ptr())
```
workingjubilee, line 196: honestly half the reason here for this code is that both of these should be the same thing.

PR #1701, opened 2024-05-14
Implement `RetAbi` and `BoxRet`

### LINKED ISSUE #1484: Remove TableIterator-to-SetOfIterator remapping codegen?
I think the way #1091 was solved leaves some lingering questions.

The error here is emitted from our code:

https://github.com/pgcentralfoundation/pgrx/blob/4fa87bd57d5e673e280c987a75cd8129ff349f43/pgrx/src/srf.rs#L96-L100

But we apparently "need" to convert the TableIterator into a SetOfIterator: https://github.com/pgcentralfoundation/pgrx/blob/5a45b3ae7ac3fe120181f4c50df689c1fb067c78/pgrx-sql-entity-graph/src/pg_extern/mod.rs#L557-L567

I find this to be a somewhat unclear statement, as it's not clear if the obligation is coming from Postgres or from our own code. It is strongly implied to be "from Postgres":

https://github.com/pgcentralfoundation/pgrx/blob/5a45b3ae7ac3fe120181f4c50df689c1fb067c78/pgrx-sql-entity-graph/src/pg_extern/mod.rs#L547-L549

However, what precisely seems to be happening is that we're the ones making the assumption that the return type must be a composite type if we're using a TableIterator. If this understanding is correct, then I think it should be written down as such. It also opens up an alternative solution that can move handling the one-item-tuple case out of the generated code, by using the fact this code is a trait impl and thus we can make generic impls do the lifting. And solving fewer problems during codegen is always preferable.

cc @eeeebbbbrrrr

workingjubilee:
This completely reworks the way we handle returning types from functions, so that we no longer have to rely on a macro expansion behavior that has to already know the types at expansion time (and thus has to parse them somehow, which it cannot realistically do because type aliases exist). Instead, we simply expand into code that asks the types themselves to modify the FunctionCallInfo appropriately and then return a raw Datum to Postgres.

This breaks support for certain returns because it was difficult to do this and also support arbitrary nesting, because Postgres does not support arbitrary nesting. For instance, you can no longer return:
- `SetOfIterator<'a, Result<T, E>>`
- `TableIterator<'a, (Result<T, E>, Result<U, D>)>`
- `Option<TableIterator<'a, Tuple>>`

It's expected that this will improve in the near-ish future.

This also breaks returning values from `#[pg_extern]` functions that were relying on `IntoDatum` implementations being enough. It is not expected this will improve, for the reasons described on the documentation of the new traits, which can be summarized as "`IntoDatum` should never have been used for that bound". This change blocks off several latent correctness problems from affecting pgrx going forward.

Fixes https://github.com/pgcentralfoundation/pgrx/issues/1484

### DIFF
--- a/pgrx-examples/spi_srf/src/lib.rs
+++ b/pgrx-examples/spi_srf/src/lib.rs
@@ -38,9 +38,9 @@ fn calculate_human_years() -> Result<
     TableIterator<
         'static,
         (
-            name!(dog_name, Result<Option<String>, pgrx::spi::Error>),
+            name!(dog_name, Option<String>),
             name!(dog_age, i32),
-            name!(dog_breed, Result<Option<String>, pgrx::spi::Error>),
+            name!(dog_breed, Option<String>),
             name!(human_age, i32),
         ),
     >,
@@ -62,7 +62,7 @@ fn calculate_human_years() -> Result<
             let dog_age = row["dog_age"].value::<i32>()?.expect("dog_age was null");
             let dog_breed = row["dog_breed"].value::<String>();
             let human_age = dog_age * 7;
-            results.push((dog_name, dog_age, dog_breed, human_age));
+            results.push((dog_name?, dog_age, dog_breed?, human_age));
         }
 
         Ok(TableIterator::new(results))
@@ -76,9 +76,9 @@ fn filter_by_breed(
     TableIterator<
         'static,
         (
-            name!(dog_name, Result<Option<String>, pgrx::spi::Error>),
-            name!(dog_age, Result<Option<i32>, pgrx::spi::Error>),
-            name!(dog_breed, Result<Option<String>, pgrx::spi::Error>),
+            name!(dog_name, Option<String>),
+            name!(dog_age, Option<i32>),
+            name!(dog_breed, Option<String>),
         ),
     >,
     spi::Error,
@@ -95,9 +95,11 @@ fn filter_by_breed(
         let tup_table = client.select(query, None, Some(args))?;
 
         let filtered = tup_table
-            .map(|row| (row["dog_name"].value(), row["dog_age"].value(), row["dog_breed"].value()))
-            .collect::<Vec<_>>();
-        Ok(TableIterator::new(filtered))
+            .map(|row| {
+                Ok((row["dog_name"].value()?, row["dog_age"].value()?, row["dog_breed"].value()?))
+            })
+            .collect::<Result<Vec<_>, _>>();
+        filtered.map(|v| TableIterator::new(v))
     })
 }
 
@@ -110,16 +112,16 @@ mod tests {
     #[rustfmt::skip]
     #[pg_test]
     fn test_calculate_human_years() -> Result<(), pgrx::spi::Error> {
-        let mut results: Vec<(Result<Option<String>, _>, i32, Result<Option<String>, _>, i32)> =
+        let mut results =
             Vec::new();
 
-        results.push((Ok(Some("Fido".to_string())), 3, Ok(Some("Labrador".to_string())), 21));
-        results.push((Ok(Some("Spot".to_string())), 5, Ok(Some("Poodle".to_string())), 35));
-        results.push((Ok(Some("Rover".to_string())), 7, Ok(Some("Golden Retriever".to_string())), 49));
-        results.push((Ok(Some("Snoopy".to_string())), 9, Ok(Some("Beagle".to_string())), 63));
-        results.push((Ok(Some("Lassie".to_string())), 11, Ok(Some("Collie".to_string())), 77));
-        results.push((Ok(Some("Scooby".to_string())), 13, Ok(Some("Great Dane".to_string())), 91));
-        results.push((Ok(Some("Moomba".to_string())), 15, Ok(Some("Labrador".to_string())), 105));
+        results.push((Some("Fido".to_string()), 3, Some("Labrador".to_string()), 21));
+        results.push((Some("Spot".to_string()), 5, Some("Poodle".to_string()), 35));
+        results.push((Some("Rover".to_string()), 7, Some("Golden Retriever".to_string()), 49));
+        results.push((Some("Snoopy".to_string()), 9, Some("Beagle".to_string()), 63));
+        results.push((Some("Lassie".to_string()), 11, Some("Collie".to_string()), 77));
+        results.push((Some("Scooby".to_string()), 13, Some("Great Dane".to_string()), 91));
+        results.push((Some("Moomba".to_string()), 15, Some("Labrador".to_string()), 105));
         let func_results = calculate_human_years()?;
 
         for (expected, actual) in results.iter().zip(func_results) {
--- a/pgrx-examples/spi_srf/src/lib.rs
+++ b/pgrx-examples/spi_srf/src/lib.rs
@@ -109,11 +109,9 @@ mod tests {
     use crate::calculate_human_years;
     use pgrx::prelude::*;
 
-    #[rustfmt::skip]
     #[pg_test]
     fn test_calculate_human_years() -> Result<(), pgrx::spi::Error> {
-        let mut results =
-            Vec::new();
+        let mut results = Vec::new();
 
         results.push((Some("Fido".to_string()), 3, Some("Labrador".to_string()), 21));
         results.push((Some("Spot".to_string()), 5, Some("Poodle".to_string()), 35));
--- a/pgrx-tests/src/tests/srf_tests.rs
+++ b/pgrx-tests/src/tests/srf_tests.rs
@@ -298,12 +298,15 @@ mod tests {
     }
 
     #[pg_test(error = "column \"cause_an_error\" does not exist")]
-    pub fn spi_in_setof() -> SetOfIterator<'static, Result<Option<String>, spi::Error>> {
+    pub fn spi_in_setof() -> Result<SetOfIterator<'static, Option<String>>, spi::Error> {
         let oids = vec![1213, 1214, 1232, 1233, 1247, 1249, 1255];
-
-        SetOfIterator::new(oids.into_iter().map(|oid| {
-            Spi::get_one(&format!("SELECT CAUSE_AN_ERROR FROM pg_class WHERE oid = {oid}"))
-        }))
+        let result = oids
+            .into_iter()
+            .map(|oid| {
+                Spi::get_one(&format!("SELECT CAUSE_AN_ERROR FROM pg_class WHERE oid = {oid}"))
+            })
+            .collect::<Result<Vec<Option<_>>, _>>();
+        result.map(SetOfIterator::new)
     }
 
     #[pg_test]
--- a/pgrx-tests/src/tests/srf_tests.rs
+++ b/pgrx-tests/src/tests/srf_tests.rs
@@ -280,13 +280,21 @@ mod tests {
 
     #[pg_test(error = "column \"cause_an_error\" does not exist")]
     pub fn spi_in_iterator(
-    ) -> TableIterator<'static, (name!(id, i32), name!(relname, Result<Option<String>, spi::Error>))>
+    ) -> Result<TableIterator<'static, (name!(id, i32), name!(relname, Option<String>))>, spi::Error>
     {
         let oids = vec![1213, 1214, 1232, 1233, 1247, 1249, 1255];
-
-        TableIterator::new(oids.into_iter().map(|oid| {
-            (oid, Spi::get_one(&format!("SELECT CAUSE_AN_ERROR FROM pg_class WHERE oid = {oid}")))
-        }))
+        let result = oids
+            .into_iter()
+            .map(|oid| {
+                Ok((
+                    oid,
+                    Spi::get_one(&format!(
+                        "SELECT CAUSE_AN_ERROR FROM pg_class WHERE oid = {oid}"
+                    ))?,
+                ))
+            })
+            .collect::<Result<Vec<_>, _>>();
+        result.map(|v| TableIterator::new(v))
     }
 
     #[pg_test(error = "column \"cause_an_error\" does not exist")]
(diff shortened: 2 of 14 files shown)


pgrx/src/callconv.rs
```diff
@@ -134,3 +314,325 @@ fn finish_srf_init<T>(
...
+        Ret::Once(self)
+    }
+
+    unsafe fn box_return(fcinfo: pg_sys::FunctionCallInfo, ret: Ret<Self>) -> pg_sys::Datum {
+        match ret {
+            Ret::Zero => unsafe { pg_return_null(fcinfo) },
+            Ret::Once(value) => pg_sys::Datum::from(value.to_bits()),
+            Ret::Many(_, _) => unreachable!(),
+        }
+    }
```
workingjubilee: This is what I meant about the awkwardness of forcing the scalar type cases into a framework that can support the iterator and composite type cases.

pgrx/src/callconv.rs
```diff
@@ -134,3 +314,325 @@ fn finish_srf_init<T>(
...
+    T: ReturnShipping,
+    <T as ReturnShipping>::CallRet: IntoDatum,
+{
+    match ret {
+        Ret::Zero => unsafe { pg_return_null(fcinfo) },
+        Ret::Once(value) => match value.into_datum() {
+            None => unsafe { pg_return_null(fcinfo) },
+            Some(datum) => datum,
+        },
+        Ret::Many(_, _) => unreachable!(),
```
workingjubilee: Most of these being should-never-happen cases, effectively.

pgrx/src/iter.rs
```diff
@@ -155,6 +158,55 @@ impl<'a, T> Iterator for TableIterator<'a, T> {
...
+    }
+
+    unsafe fn box_return(fcinfo: pg_sys::FunctionCallInfo, ret: Ret<Self>) -> pg_sys::Datum {
+        let value = match ret {
+            Ret::Zero => return empty_srf(fcinfo),
+            Ret::Once(value) => value,
+            Ret::Many(iter, value) => {
+                iter.into_context(fcinfo);
+                value
+            }
```
workingjubilee: Probably not even correct, here. We don't have enough tests for this edge-case.

pgrx/src/iter.rs
```diff
@@ -198,10 +250,38 @@ macro_rules! impl_table_iter {
...
+                        iter.into_context(fcinfo);
+                        value
+                    }
+                };
+
+                unsafe {
+                    let fcx = deref_fcx(fcinfo);
+                    let heap_tuple = value.into_heap_tuple((*fcx).tuple_desc);
+                    pg_sys::HeapTupleHeaderGetDatum((*heap_tuple).t_data)
+                }
```
workingjubilee: This is the most awkward thing that actually makes me think "dealbreaker, maybe?" until I restructure it: the tuple types shouldn't have to have magical knowledge of the TupleDesc that was set up for them by TableIterator.

pgrx/src/iter.rs
```diff
@@ -198,10 +250,38 @@ macro_rules! impl_table_iter {
             }
         }
 
+        unsafe impl<$($C),*> ReturnShipping for ($($C,)*)
+        where
+             $($C: ReturnShipping,)*
+             Self: IntoHeapTuple,
+        {
+            type CallRet = Self;
```
workingjubilee: so many cases of this...

pgrx/src/callconv.rs
workingjubilee: I resolved this by using a blanket impl.

pgrx/src/iter.rs
```diff
@@ -155,6 +158,56 @@ impl<'a, T> Iterator for TableIterator<'a, T> {
...
+
+impl<C: IntoDatum> IntoHeapTuple for (C,) {
+    unsafe fn into_heap_tuple(self, _tupdesc: pg_sys::TupleDesc) -> *mut pg_sys::HeapTupleData {
+        unimplemented!()
+    }
+}
+
+unsafe impl<C> ReturnShipping for (C,)
+where
+    C: ReturnShipping, // so we support TableIterator<'a, (Option<T>,)> as well
```
workingjubilee: This is... not ideal, as it traps us in a position where we might accidentally try to return something like `TableIterator<'a, (TableIterator<'b, TupleA>, TableIterator<'c, TupleB>)>`.

workingjubilee: This would be preferable, as it closes off that route:
```suggestion
    C: RetPackage,
```

workingjubilee: I should note that I'm pretty sure that this is currently impossible anyways even with this PR because the schema generator chokes on it, I just would rather it didn't require "the schema generator chokes on it" in order to stop us from doing something silly.

pgrx-tests/src/tests/srf_tests.rs
```diff
@@ -280,13 +280,21 @@ mod tests {
...
+            .map(|oid| {
+                Ok((
+                    oid,
+                    Spi::get_one(&format!(
+                        "SELECT CAUSE_AN_ERROR FROM pg_class WHERE oid = {oid}"
+                    ))?,
+                ))
+            })
+            .collect::<Result<Vec<_>, _>>();
+        result.map(|v| TableIterator::new(v))
```
workingjubilee, line 297: So, this _should_ be `TableIterator<'_, Result<Tuple, Error>>` so that no eager evaluation has to happen. I think I can make that work with a bit more effort. But the schema generator chokes on **that**! And that is another reason why I want it out of our way.

I do not think people should be required to return `TableIterator<'a, (Result, Result, Result, Result, Result, Result)>` if they want to fallibly return a row, but they currently _are_. And I do not think this should be blocked on that.

Asking people to propagate errors to the top of each row return should be fine.

pgrx-examples/spi_srf/src/lib.rs
```diff
@@ -107,19 +109,17 @@ mod tests {
...
-            Vec::new();
-
-        results.push((Ok(Some("Fido".to_string())), 3, Ok(Some("Labrador".to_string())), 21));
-        results.push((Ok(Some("Spot".to_string())), 5, Ok(Some("Poodle".to_string())), 35));
-        results.push((Ok(Some("Rover".to_string())), 7, Ok(Some("Golden Retriever".to_string())), 49));
-        results.push((Ok(Some("Snoopy".to_string())), 9, Ok(Some("Beagle".to_string())), 63));
-        results.push((Ok(Some("Lassie".to_string())), 11, Ok(Some("Collie".to_string())), 77));
-        results.push((Ok(Some("Scooby".to_string())), 13, Ok(Some("Great Dane".to_string())), 91));
-        results.push((Ok(Some("Moomba".to_string())), 15, Ok(Some("Labrador".to_string())), 105));
+        let mut results = Vec::new();
```
workingjubilee, line 114: An advantage of using a somewhat more sensible return type is that you don't have this sort of thing happen...

workingjubilee (commented):
As the iterator conveniences are basically independent, I'll rebase this on #1703 after that lands.

pgrx/src/iter.rs
workingjubilee: I intend to fix this before we land it but I believe otherwise this is ready for review, I don't want to do much more work on this without feedback.

workingjubilee:
This PR opens a LOT of pure simplifications like 9dd8f5fd055748bf6cd3891acb85d3f731f5d782, it's just going to take forever to find them all.

pgrx-tests/src/tests/srf_tests.rs
eeeebbbbrrrr: > Asking people to propagate errors to the top of each row return should be fine.

I'm not sure this is a nice thing to ask.  When SPI, for example, is involved, and what you're trying to do is return a TableIterator of a bunch of individual values you somehow got from SPI, it's definitely a lot easier to just return the `Result<T>` (or `Result<Option<T>>`) in whatever tuple position.  Especially if there's quite a few values in the tuple.

This isn't a hill I'm willing to die on, but it'd be nice to support, for example, `TableIterator<(Result<S>, T, Result<U>, Result<V>, W, Result<X>, Y, Result<Z>)>`.  Then we'll panic whenever we get to unwrapping that result and the user doesn't have to worry about it.

workingjubilee: @eeeebbbbrrrr Honestly, we'll never have good UX until we expand to supporting actual `struct` arguments, really.

eeeebbbbrrrr: What does "actual `struct` arguments" mean?

pgrx-examples/spi_srf/src/lib.rs
```diff
@@ -95,9 +95,11 @@ fn filter_by_breed(
         let tup_table = client.select(query, None, Some(args))?;
 
         let filtered = tup_table
-            .map(|row| (row["dog_name"].value(), row["dog_age"].value(), row["dog_breed"].value()))
-            .collect::<Vec<_>>();
-        Ok(TableIterator::new(filtered))
+            .map(|row| {
+                Ok((row["dog_name"].value()?, row["dog_age"].value()?, row["dog_breed"].value()?))
```
workingjubilee, line 99: @eeeebbbbrrrr it's certainly not less code but in most cases it just looks like going "hello? I? have? a? question? if that's okay?"

pgrx-tests/src/tests/srf_tests.rs
workingjubilee: TableIterator should be able to return a struct that describes the row in question, but this requires being able to use traits like these to return fairly arbitrary generic types. It doesn't work as long as we rely on parsing.

pgrx-examples/spi_srf/src/lib.rs
eeeebbbbrrrr: It's not terrible, but I'm sure you agree that returning a Result from a closure is awkward at best.  In this case we `.collect::<Result<Vec<_>>>()` so it's not too terrible.  

I prefer the left side of this diff, but like I said elsewhere, I'm not willing to die on this hill.

workingjubilee: yeah, the left side is a bit nicer, but that type signature... that's what I really have it in for. so yeah, if I didn't believe it would open up better possibilities in short order, I wouldn't do it.

eeeebbbbrrrr: > that's what I really have it in for

Yeah, that's why I'm not really gonna fight on this one.  I hate typing out all that mess too.  But once the return signature has been typed out, it makes it a lot easier to just write some code and return some values.

pgrx/src/callconv.rs
```diff
@@ -9,96 +9,280 @@
...
-        match setof_iterator.next() {
-            Some(datum) => {
-                srf_return_next(fcinfo, fcx);
-                datum.into_datum().unwrap_or_else(|| pg_return_null(fcinfo))
-            }
-            None => empty_srf(fcinfo),
+    unsafe fn ret_from_context(fcinfo: pg_sys::FunctionCallInfo) -> Ret<Self> {
+        let fcx = deref_fcx(fcinfo);
+        // SAFETY: fcx.user_fctx was set earlier, immediately before or in a prior call
+        let iter = &mut *(*fcx).user_fctx.cast::<TableIterator<T>>();
```
workingjubilee: uh that is extremely not correct
```suggestion
        let iter = &mut *(*fcx).user_fctx.cast::<SetOfIterator<'_, T>>();
```

workingjubilee:
I moved the implementations for the iterator types to live next to the iterators. It seemed sensical for some reason.

eeeebbbbrrrr:
Ya know. What if we called it…

I’m kidding. This is great. Thanks, @workingjubilee

workingjubilee:
Resolved the "tuples have psychic powers" issue by splitting out their knowledge of how to add a tupdesc into a new fn.

pgrx-tests/src/tests/srf_tests.rs
```diff
@@ -298,12 +298,15 @@ mod tests {
...
-        SetOfIterator::new(oids.into_iter().map(|oid| {
-            Spi::get_one(&format!("SELECT CAUSE_AN_ERROR FROM pg_class WHERE oid = {oid}"))
-        }))
+        let result = oids
+            .into_iter()
+            .map(|oid| {
+                Spi::get_one(&format!("SELECT CAUSE_AN_ERROR FROM pg_class WHERE oid = {oid}"))
+            })
+            .collect::<Result<Vec<Option<_>>, _>>();
+        result.map(SetOfIterator::new)
```
workingjubilee, line 309: Breaking this one annoys me much more than the TableIterator nesting case but I want to make sure that SetOfIterator has exactly as stringent requirements, currently, as `TableIterator<'a, (T,)>`, because it will make things much worse down the line if they meaningfully differ.

pgrx-examples/spi_srf/src/lib.rs
NotGyro: Many question marks is much-preferable to having excessively-verbose nested type signatures, yeah.

NotGyro (approved):
Looks good to me. This really does make code working with the returns from Rust to Postgres much cleaner and more elegant. 

I had some confusion about the SetOfIterator type, but it was cleared up in a company meeting. The single-item case behaves differently than the iterator case for legacy Postgres reasons, which is why SetOfIterator can be one item and/or an iterator.

NotGyro (approved):
Thanks for the documentation! This explains it neatly.

workingjubilee:
This last commit makes the "Ret" type into an associated type and thus makes the iterators eat all the complexity of their return ABI as now only they have to know about how to drive themselves forward. `Result` does not have to think about them, and the implementation details are once more private to iterators and fully hidden behind the trait. This prevents anyone from trying to make any of the **many** bug-inducing nonsensical calls that were technically allowed at first, and probably made the code have to carry around way more panic branches.

workingjubilee:
The last few commits were just cleanup.

PR #1713, opened 2024-05-22
Deprecate `variadic!` support

workingjubilee:
This removes `variadic!` support. The macro itself remains, but it does nothing now, and has a deprecation note. There are no uses of it in this codebase, the ZomboDB codebase, or other pgrx-using codebases I have seen. `VariadicArray` completely surpasses it in utility. I want to remove, ideally, all these ad-hoc type macros, but they may need to be replaced in functionality. This one, however, is already dead.


pgrx/src/fcinfo.rs
```diff
@@ -80,6 +80,7 @@ macro_rules! name {
     };
 }
 
+#[deprecated(since = "0.12.0", note = "try VariadicArray")]
```
eeeebbbbrrrr, line 83: Just delete the macro.  There's enough breaking changes in develop that one more doesn't matter.

eeeebbbbrrrr (approved):

pgrx/src/fcinfo.rs
workingjubilee: All the real parts of the macro, and thus the things that complicate its support internally, are gone. There are other users that don't have their code published, and for them, telling them what else to do may make migration slightly easier.

workingjubilee: Obviously I've broken probably a fair number of unpublished extensions with this version but it seems low-effort.

eeeebbbbrrrr: can it become a compile error instead of a deprecated warning?  Cuz using it now would mean it doesn't dooooooooo what they want, right?

workingjubilee: oh sure.

workingjubilee: actually it already is lol

PR #1720, opened 2024-05-28
FcInfo wrapper

NotGyro:
Implements a thin wrapper around FunctionCallInfo in order to allow us to track the lifetimes of arguments provided to a function callable by Postgres.


pgrx/src/callconv.rs
```diff
@@ -283,3 +283,182 @@ where
...
+            let mut flinfo = PgBox::from_pg(fcinfo.flinfo);
+            if flinfo.fn_extra.is_null() {
+                flinfo.fn_extra = PgMemoryContexts::For(flinfo.fn_mcxt)
+                    .leak_and_drop_on_delete(default())
+                    as crate::void_mut_ptr;
+            }
+
+            PgBox::from_pg(flinfo.fn_extra as *mut ReturnType)
+        }
+    }
```
workingjubilee: For this, I basically consider the PgBox type soft-deprecated (there's not a good alternative atm or I would actually set the deprecation notice up), so I would rather we not use it here. It would still be safe, however, if it simply returned a raw pointer, as that would be unsafe to use.

pgrx/src/callconv.rs
```diff
@@ -283,3 +283,182 @@ where
...
+        fcinfo.fncollation
+    }
+
+    /// Retrieve the type (as an Oid) of argument number `num`.
+    /// In other words, the type of `self.raw_args()[num]`
+    #[inline]
+    pub fn pg_getarg_type(&self, num: usize) -> pg_sys::Oid {
+        // Safety: User must supply a valid fcinfo to assume_valid() in order
+        // to construct a FcInfo. If that constraint is maintained, this should
+        // be safe.
```
workingjubilee: You can just say something like
```suggestion
        // SAFETY: see FcInfo::assume_valid
```

pgrx/src/callconv.rs
```diff
@@ -283,3 +283,182 @@ where
...
+    ///
+    /// ```rust,no_run
+    /// use pgrx::pg_return_null;
+    /// use pgrx::prelude::*;
+    /// fn foo(fcinfo: pg_sys::FunctionCallInfo) -> pg_sys::Datum {
+    ///     return unsafe { pg_return_null(fcinfo) };
+    /// }
+    /// ```
+    #[inline]
+    pub unsafe fn pg_return_null(&mut self) -> pg_sys::Datum {
```
workingjubilee: Needs to document its Safety constraint

pgrx/src/callconv.rs
```diff
@@ -283,3 +283,182 @@ where
...
+        let fcinfo = unsafe { self.0.as_mut() }.unwrap();
+        fcinfo.isnull = true;
+        pg_sys::Datum::from(0)
+    }
+
+    /// Get the collation the function call should use
+    #[inline]
+    pub unsafe fn pg_get_collation(&self) -> pg_sys::Oid {
```
workingjubilee: Why is this one unsafe?

pgrx/src/callconv.rs
```diff
@@ -283,3 +283,182 @@ where
...
+        }
+    }
+
+    #[inline]
+    pub fn srf_is_first_call(&self) -> bool {
+        // Safety: User must supply a valid fcinfo to assume_valid() in order
+        // to construct a FcInfo. If that constraint is maintained, this should
+        // be safe.
+        unsafe { (*(*self.0).flinfo).fn_extra.is_null() }
+    }
```
workingjubilee: should be something like `is_initialized`, since `first_call` is misleading: this will return false for the rest of the code after it is set.

tbh we could just offer this as something like `init_multicall_context` and then have it return `true` if initialization was actually done (otherwise it is `false`).

pgrx/src/callconv.rs
```diff
@@ -283,3 +283,182 @@ where
...
+                pg_sys::ExprDoneCond_ExprEndResult;
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct FcContext<'fcx>(
+    *mut pgrx_pg_sys::FuncCallContext,
+    std::marker::PhantomData<&'fcx mut pgrx_pg_sys::FuncCallContext>,
+);
```
workingjubilee: It's not clear to me that we can't just offer `FcContext` as `&mut pg_sys::FuncCallContext` as unlike the fcinfo, the callee "owns" this.

pgrx/src/callconv.rs
```diff
@@ -283,3 +283,182 @@ where
...
+        unsafe { FcContext::assume_valid(pg_sys::per_MultiFuncCall(self.0)) }
+    }
+    /// Equivalent to "per_MultiFuncCall" with no FFI cost, and a lifetime
+    /// constraint.
+    ///
+    /// Safety: This is only equivalent to the current (as of Postgres 16.3)
+    /// implementation of [`pg_sys::per_MultiFuncCall`], and future changes
+    /// to the complexity of the calling convention may break this method.
+    pub(crate) unsafe fn deref_fcx(&mut self) -> FcContext<'fcx> {
+        unsafe { FcContext::assume_valid((*(*self.0).flinfo).fn_extra.cast()) }
```
workingjubilee: They cannot really change this without breaking basically all SRFs ever. They only got away with the FunctionCallInfoBaseData change because the primary entity that constructs FunctionCallInfoDatas to call functions with is Postgres itself, and everyone else used it as the pointer typedef of FunctionCallInfo.

workingjubilee:
it seems things are coming along nicely, I'm just noting you're allowed to interpret a bit more into things~

pgrx/src/callconv.rs
NotGyro: I removed that safety warning and replaced it with one that simply states `self.0.flinfo.fn_extra` must be non-null

pgrx/src/callconv.rs
NotGyro: Rewrote it as that.

pgrx/src/callconv.rs
NotGyro: Does `srf_is_initialized()` work? 
```rust
    #[inline]
    pub fn srf_is_initialized(&self) -> bool {
        // Safety: User must supply a valid fcinfo to assume_valid() in order
        // to construct a FcInfo. If that constraint is maintained, this should
        // be safe.
        unsafe { !(*(*self.0).flinfo).fn_extra.is_null() }
    }
```

pgrx/src/callconv.rs
NotGyro: I've documented it now, as per our call. Thanks for clearing up my suspicions re: this causing memory leaks if misused!

pgrx/src/callconv.rs
NotGyro: As per our call, it doesn't need to be unsafe. Thanks for clearing it up

pgrx/src/callconv.rs
NotGyro: I've rewritten it to return a simpler pointer type, just a NonNull<T>.

pgrx/src/callconv.rs
```diff
@@ -283,3 +284,170 @@ where
...
+
+    /// Retrieve the type (as an Oid) of argument number `num`.
+    /// In other words, the type of `self.raw_args()[num]`
+    #[inline]
+    pub fn get_arg_type(&self, num: usize) -> pg_sys::Oid {
+        // SAFETY: see FcInfo::assume_valid
+        unsafe {
+            pg_sys::get_fn_expr_argtype(self.0.as_ref().unwrap().flinfo, num as std::os::raw::c_int)
+        }
+    }
```
workingjubilee: Should have an Option signature for the out-of-bound case.

pgrx/src/callconv.rs
```diff
@@ -283,3 +284,170 @@ where
...
+    /// fn foo<'a>(mut fcinfo: FcInfo<'a>) -> pg_sys::Datum {
+    ///     return unsafe { fcinfo.set_return_null() };
+    /// }
+    /// ```
+    #[inline]
+    pub unsafe fn set_return_null(&mut self) -> pg_sys::Datum {
+        let fcinfo = unsafe { self.0.as_mut() }.unwrap();
+        fcinfo.isnull = true;
+        pg_sys::Datum::from(0)
+    }
```
workingjubilee: Should return a `pgrx::datum::Datum<'a>`

pgrx/src/callconv.rs
```diff
@@ -283,3 +284,170 @@ where
...
+
+    /// Get the collation the function call should use
+    #[inline]
+    pub fn get_collation(&self) -> pg_sys::Oid {
+        // SAFETY: see FcInfo::assume_valid
+        let fcinfo = unsafe { self.0.as_mut() }.unwrap();
+        fcinfo.fncollation
+    }
```
workingjubilee: Option.

pgrx/src/callconv.rs
```diff
@@ -283,3 +284,170 @@ where
...
+
+    /// Safety: Assumes `self.0.flinfo.fn_extra` is non-null
+    /// i.e. [`FcInfo::srf_is_initialized()`] would be `true`.
+    #[inline]
+    pub unsafe fn srf_return_done(&mut self) {
+        unsafe {
+            pg_sys::end_MultiFuncCall(self.0, self.deref_fcx());
+            (*((*self.0).resultinfo as *mut pg_sys::ReturnSetInfo)).isDone =
+                pg_sys::ExprDoneCond_ExprEndResult;
```
workingjubilee: can extract the accessor for the resultinfo here as an unsafe fn

pgrx/src/callconv.rs
workingjubilee: ideally there'd be an args iterator that you can use to optionally call `.get_arg_type` for that index. can be deferred to a followup though.

workingjubilee:
This does need an `as_mut_ptr` function for continued FFI uses.

NotGyro:
> This does need an `as_mut_ptr` function for continued FFI uses.

Could this be a crate-level internal method?

workingjubilee:
Hm, sadly I don't think so.

workingjubilee:
We can defer adding it for a bit but I don't think we can do it forever.

pgrx/src/callconv.rs
```diff
@@ -283,3 +286,190 @@ where
...
+            // at least nargs elements of NullableDatum.
+            std::slice::from_raw_parts(args_ptr, arg_len as _)
+        }
+    }
+
+    /// Retrieves the internal [`pg_sys::FunctionCallInfo`] wrapped by this type. 
+    /// A FunctionCallInfo is a mutable pointer to a FunctionCallInfoBaseData already, and so
+    /// that type is sufficient.
+    #[inline]
+    pub unsafe fn as_mut_ptr(&self) -> FunctionCallInfo {
```
workingjubilee:
```suggestion
    pub unsafe fn as_mut_ptr(&self) -> pg_sys::FunctionCallInfo {
```

pgrx/src/callconv.rs
```diff
@@ -10,12 +10,15 @@
 #![doc(hidden)]
 #![deny(unsafe_op_in_unsafe_fn)]
 //! Helper implementations for returning sets and tables from `#[pg_extern]`-style functions
+use pgrx_pg_sys::FunctionCallInfo;
```
workingjubilee:
```suggestion
```

pgrx/src/callconv.rs
```diff
@@ -283,3 +286,190 @@ where
...
+    #[inline]
+    pub unsafe fn srf_return_done(&mut self) {
+        unsafe {
+            pg_sys::end_MultiFuncCall(self.0, self.deref_fcx());
+            (*self.get_result_info()).isDone =
+                pg_sys::ExprDoneCond_ExprEndResult;
+        }
+    }
+
+    pub unsafe fn get_result_info(&self) -> *mut pg_sys::ReturnSetInfo {
```
workingjubilee:
```suggestion
    /// # Safety
    /// Do not corrupt the `pg_sys::ReturnSetInfo` struct's data.
    pub unsafe fn get_result_info(&mut self) -> &mut pg_sys::ReturnSetInfo { 
```

Alternately, expose safe accessors for just the last four fields (they are "owned" by the callee):

https://github.com/postgres/postgres/blob/4ebe51a5fbc3e08861db85fe3868c31c6bc8ed59/src/include/nodes/execnodes.h#L335-L341

workingjubilee: or just the one we set.

NotGyro: Which ones do we mutate? Just isDone?

pgrx/src/callconv.rs
NotGyro: This should be resolved, except the iterator thing which I will make into another issue.

pgrx/src/callconv.rs
NotGyro: Resolved

pgrx/src/callconv.rs
NotGyro: Resolved this.

pgrx/src/callconv.rs
NotGyro: I cut this out in a recent commit.

pgrx/src/callconv.rs
workingjubilee: seems so?

NotGyro: Alright, implemented in 9b420292aae38b1beaa324f067378cce89ee6b90

pgrx/src/callconv.rs
NotGyro: Resolved

workingjubilee (approved):
This is good! I think we can simplify the ReturnSetInfoWrapper but, later

PR #1724, opened 2024-05-29
Allow real borrow-checking of `#[pg_extern] fn` lifetimes

workingjubilee:
This doesn't finish fixing things, but it introduces a skeleton for actually trying to impose borrow-checking further on the rest of pgrx's code. This small refactor is effectively a chokepoint that all future code has to build from.


NotGyro (approved):
Looks good! This is a completion of a lot of the work required to get lifetimes for an FcInfo wrapper plumbed through to everywhere they need to go, and I'm seeing some safety/soundness improvements in there as well.

Issue #1582, opened 2024-02-22
Document `cargo pgrx package` including `./sql`


workingjubilee:
This is an undocumented feature that is useful for including extension upgrade scripts (`ALTER EXTENSION`), which currently must be written by hand.

workingjubilee:
Apparently it works for all of `cargo pgrx {run,test,package}`.

workingjubilee:
Assigning it to the person who knows it best~

workingjubilee:
maybe we should write a comment in a file in the `./sql` directory like
```sql
-- cargo pgrx will copy .sql files from this directory when installing this extension
```

eeeebbbbrrrr:
> `-- cargo pgrx will copy .sql files from this directory when installing this extension`

It only copies files in the format of `extname--oldver--newver.sql`.  All other files are ignored.  See PR #1725

PR #1725, opened 2024-05-30
Document what happens with the `./sql` directory

eeeebbbbrrrr:
Address #1582 by documenting how `cargo pgrx run/test/install/package` handle extension upgrade scripts.


workingjubilee:
Oh!
That's more specific than I imagined. Thanks!

PR #1726, opened 2024-05-31
add bindings of fmgroids.h, OIDs of built-in functions

usamoi:
`fmgroids.h` is helpful to perform an index scan.

For example, https://github.com/apache/age/blob/ef9e0bd67bb8c342c4f6dd04761e2c4b3a3a5f80/src/backend/catalog/ag_graph.c#L74. It uses `F_NAMEEQ` to perform a search for catalog defined by extension.


workingjubilee (approved):
Sure!

PR #1729, opened 2024-05-31
document pg_guard to the best of my abilities

### LINKED ISSUE #1406: Give directions on correct usage of `pg_guard` and similar
For a few fundamentally `unsafe` declarations, the ideal size should be as small as possible. There will be a similar function on MemCx. In the guide, we will need to explain code organization on a high level, and saying "really, try to make these as small as possible. `pg_guard` is for `pgrx-pg-sys` and certain callback-driven APIs, use `pg_extern` for anything you want to expose through an extension" is useful.

eeeebbbbrrrr:
Closes #1406

### DIFF
--- a/docs/src/ffi-error-handling.md
+++ b/docs/src/ffi-error-handling.md
@@ -8,7 +8,8 @@ There are many other concerns across this boundary such as function call ABIs an
 # High-level Postgres Error Handling Overview
 
 Most Postgres internal functions (those accessible via the `pg_sys` module) are capable of raising an `ERROR`.  This "error"
-comes into existence when, internally, Postgres calls `errstart()`, which does the work to instantiate the error.
+comes into existence when, internally, Postgres code calls the `ereport()` (or `elog()`) macro.  `ereport()` does the work
+to instantiate the error by first calling the `errstart()` function.
 
 Code execution then finds its way to `errfinish()` where, finally, `siglongjmp()` is called to instantly move the stack 
 back to the frame where Postgres began the current transaction (where it previously created a `sigsetjmp()` point).  
@@ -91,12 +92,21 @@ function.  This function sets up pgrx' own `sigsetjmp` point, lies to Postgres' 
 going to jump to in case of an ERROR, calls the function via the closure argument, then restores Postgres' exception handling
 stack.
 
+"Lies to Postgres" is a little curious here, but Postgres doesn't expect that whenever it raises an ERROR it'll be jumping
+into a Rust-created stack frame.  In fact, it doesn't have *any* expectations about where it's jumping other than, eventually,
+the raised ERROR will either rollback the current transaction, be re-thrown, or in some cases, simply ignored.
+
 As it relates to this document, the specific workings of this process is more of an implementation detail, but the gist of
 the process is that we set up our own `sigsetjmp` point so that we can trap, in Rust, any ERROR Postgres might raise while
 calling the internal Postgres function.  This then allows us to convert that error into a Rust panic and have it propagated
-through the call stack so that Rust's stack properly unwinds and type destructors are called.  
+through the call stack so that Rust's stack properly unwinds and type destructors are called.
+
+We make sure to limit the amount of `sigsetjmp`-protected code to only be the internal Postgres function being called.
+Doing so ensures we're doing the minimal amount of work necessary to properly protect from a Postgres ERROR and not also
+accidentally defeating Rust's stack unwinding.
 
 While Rust doesn't guarantee that `drop()` will get called for any instantiated type, we do our best to encourage it.
+Of course, using `std::mem::forget()` on an instantiated type will never have its drop implementation called.
 
 Ultimately, at the top of the Rust callstack, the panic raised from a Postgres ERROR is then converted back into a normal
 Postgres ERROR using its internal facility for raising errors.
--- a/docs/src/ffi-error-handling.md
+++ b/docs/src/ffi-error-handling.md
@@ -8,7 +8,7 @@ There are many other concerns across this boundary such as function call ABIs an
 # High-level Postgres Error Handling Overview
 
 Most Postgres internal functions (those accessible via the `pg_sys` module) are capable of raising an `ERROR`.  This "error"
-comes into existence when, internally, Postgres calls `errstart()`, which does the work instantiate the error.
+comes into existence when, internally, Postgres calls `errstart()`, which does the work to instantiate the error.
 
 Code execution then finds its way to `errfinish()` where, finally, `siglongjmp()` is called to instantly move the stack 
 back to the frame where Postgres began the current transaction (where it previously created a `sigsetjmp()` point).  
--- a/docs/src/ffi-error-handling.md
+++ b/docs/src/ffi-error-handling.md
@@ -42,7 +42,7 @@ pgrx uses two different approaches to protect these FFI boundaries.  While both 
 Rust from Postgres `setlongjmp` ERRORs and the other protects Postgres from Rust `panic()!`s.  To make things confusing
 they're both called `#[pg_guard]`.
 
-Essentially, pgrx needs to guard two styles of `extern "C"` functions.  One style is the `extern "C" {}` block that 
+Essentially, pgrx needs to guard two styles of `extern "C"` functions.  One style is the [`extern "C" {}` block][extern-blocks] that 
 declares a function lives "somewhere else" (in our case, the Postgres process in which the pgrx extension is loaded).
 The other style is `extern "C" fn foo() { ... }` functions that are written in Rust and might be passed to Postgres (for
 it to later call) via a standard function pointer.
--- a/docs/src/ffi-error-handling.md
+++ b/docs/src/ffi-error-handling.md
@@ -49,7 +49,7 @@ it to later call) via a standard function pointer.
 
 ## Guarding Postgres Internal Functions
 
-pgrx uses the `bindgen` tool to generate "bindings" for exported Postgres symbols.  Postgres' source header files (`*.h`)
+pgrx uses the [`bindgen`] tool to generate "bindings" for exported Postgres symbols.  Postgres' source header files (`*.h`)
 are read, parsed, and transformed, as much as bindgen knows how, into Rust declarations.  In the case of exported functions,
 bindgen generates blocks similar to:
 
--- a/docs/src/ffi-error-handling.md
+++ b/docs/src/ffi-error-handling.md
@@ -1,0 +1,249 @@
+Postgres is written in C.  pgrx is written in Rust.  Between them is a boundary where each blindly believes the other
+behaves in an expected way.  Our primary concern with this boundary is error handling.
+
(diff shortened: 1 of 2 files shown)


docs/src/ffi-error-handling.md
```diff
@@ -0,0 +1,249 @@
...
+behaves in an expected way.  Our primary concern with this boundary is error handling.
+
+There are many other concerns across this boundary such as function call ABIs and pointer ownership, but these are generally
+"obvious" concerns to anyone that's done any FFI development and won't be discussed here in detail.  
+
+
+# High-level Postgres Error Handling Overview
+
+Most Postgres internal functions (those accessible via the `pg_sys` module) are capable of raising an `ERROR`.  This "error"
+comes into existence when, internally, Postgres calls `errstart()`, which does the work instantiate the error.
```
workingjubilee:
```suggestion
comes into existence when, internally, Postgres calls `errstart()`, which does the work to instantiate the error.
```

docs/src/ffi-error-handling.md
```diff
@@ -0,0 +1,249 @@
...
+
+Code execution then finds its way to `errfinish()` where, finally, `siglongjmp()` is called to instantly move the stack 
+back to the frame where Postgres began the current transaction (where it previously created a `sigsetjmp()` point).
```
workingjubilee, line 15: Hmm. This should probably clarify that it moves the state to that exact point-in-time of the sigsetjmp, "right after" it was called.

docs/src/ffi-error-handling.md
```diff
@@ -0,0 +1,249 @@
...
+# High-level Rust Error Handling Overview
+
+Rust, on the other hand, aborts the current process whenever `panic!()` is called.  This is, quite clearly, incompatible
+with Postgres' error handling approach.  
+
+It's technically incompatible, it's spiritually incompatible, and it robs Postgres of the opportunity to cleanly rollback 
+the current transaction (Postgres is supposed to be tolerant of such situations, but who wants to test Postgres' 
+recoverability in production?).
+
+Conversely, Postgres' `sigsetjmp/siglongjmp` approach is as egregiously incompatible with Rust.  `siglongjmp` will blindly
```
workingjubilee:
```suggestion
Conversely, Postgres' `sigsetjmp`/`siglongjmp` approach is as egregiously incompatible with Rust.  `siglongjmp` will blindly
```

docs/src/ffi-error-handling.md
```diff
@@ -0,0 +1,249 @@
...
+extern "C" fn foo() -> bool {
+    pgrx::pg_sys::submodules::panic::pgrx_extern_c_guard(move || {
+        // ... user-written Rust code here ...
+        return true;
+    })
+}
+```
+
+Behind the scenes, `pgrx_extern_c_guard(|| ...)` executes the closure argument inside a rust `std::panic::catch_unwind(|| ...)`
+block.  Doing so allows pgrx to capture any Rust `panic!()` and contain it its stack unwinding to the `catch_unwind()` block
```
workingjubilee: It's not _really_ a block, it's a closure, but saying closureclosureclosure risks semantic saturation, so let's just say...
```suggestion
block.  Doing so allows pgrx to capture any Rust `panic!()` and contain its stack unwinding to within the [`catch_unwind`]
```

docs/src/ffi-error-handling.md
```diff
@@ -0,0 +1,249 @@
...
+the current transaction. Finally, Postgres is again ready and waiting to begin a new transaction.
+
+This is an elegant solution to error handling as it allows Postgres to cleanly rollback the current transaction, free
+used memory, release locks, and whatever else might be necessary.
+
+
+# High-level Rust Error Handling Overview
+
+Rust, on the other hand, aborts the current process whenever `panic!()` is called.  This is, quite clearly, incompatible
+with Postgres' error handling approach.
```
workingjubilee: I think we should mention the dread stack unwinding up-front, as _merely_ aborting would be easier to handle. Something like this?
```suggestion
Rust, on the other hand, will call its panic hook and then run its panic handler when `panic!()` is called.
The panic handler itself will either *unwind* the stack or *abort* the current process.
This is, quite clearly, incompatible with Postgres' error handling approach:
unwinding destroys `sigsetjmp` checkpoints, and aborting shuts down the entire database!
```

docs/src/ffi-error-handling.md
```diff
@@ -0,0 +1,249 @@
...
+function.  This function sets up pgrx' own `sigsetjmp` point, lies to Postgres' exception handling stack about where it's 
+going to jump to in case of an ERROR, calls the function via the closure argument, then restores Postgres' exception handling
+stack.
+
+As it relates to this document, the specific workings of this process is more of an implementation detail, but the gist of
+the process is that we set up our own `sigsetjmp` point so that we can trap, in Rust, any ERROR Postgres might raise while
+calling the internal Postgres function.  This then allows us to convert that error into a Rust panic and have it propagated
+through the call stack so that Rust's stack properly unwinds and type destructors are called.  
+
+While Rust doesn't guarantee that `drop()` will get called for any instantiated type, we do our best to encourage it.
```
workingjubilee, line 108: This should refer people to [`mem::forget`](https://doc.rust-lang.org/std/mem/fn.forget.html), which documents the implications of this.

docs/src/ffi-error-handling.md
```diff
@@ -0,0 +1,249 @@
...
+    }
+    
+    unsafe {
+        crate::ffi::pg_guard_ffi_boundary(|| palloc(size))
+    }
+}
+```
+
+Essentially, in this usage, `#[pg_guard]` generates standalone wrapper functions that delegate to pgrx' `pg_guard_ffi_boundary(|| ...)`
+function.  This function sets up pgrx' own `sigsetjmp` point, lies to Postgres' exception handling stack about where it's
```
workingjubilee, line 91: lies, huh?

docs/src/ffi-error-handling.md
```diff
@@ -0,0 +1,249 @@
...
+            // `_foo` should drop() here
+        }
+    })
+}
+```
+
+Combined, we're ensuring that if any of the Postgres functions (`relation_open`/`relation_close`) raise an ERROR (say, due
+to an invalid Oid value), `pg_guard_ffi_boundary` will catch that and convert into a Rust panic.  Then ultimately, the top-level
+`pgrx_extern_c_guard` call will convert it back into a Postgres ERROR once the Rust stack has properly unwound and drop
+impls have been called.
```
workingjubilee: Let's get some hyperlinks in here~
```suggestion
impls have been called.

[`mem::forget`]: https://doc.rust-lang.org/std/mem/fn.forget.html
[`catch_unwind`]: https://doc.rust-lang.org/std/panic/fn.catch_unwind.html
[`bindgen`]: https://rust-lang.github.io/rust-bindgen/
[extern-blocks]: https://doc.rust-lang.org/reference/items/external-blocks.html
[extern-fn]: https://doc.rust-lang.org/reference/items/functions.html#extern-function-qualifier
```

docs/src/ffi-error-handling.md
```diff
@@ -0,0 +1,249 @@
...
+any opportunity to participate in error handling.
+
+
+# A Wolf, a goat, and some cabbage
+
+pgrx uses two different approaches to protect these FFI boundaries.  While both are implemented in Rust, one protects
+Rust from Postgres `setlongjmp` ERRORs and the other protects Postgres from Rust `panic()!`s.  To make things confusing
+they're both called `#[pg_guard]`.
+
+Essentially, pgrx needs to guard two styles of `extern "C"` functions.  One style is the `extern "C" {}` block that
```
workingjubilee:
```suggestion
Essentially, pgrx needs to guard two styles of `extern "C"` functions.  One style is the [`extern "C" {}` block][extern-blocks] that 
```

docs/src/ffi-error-handling.md
```diff
@@ -0,0 +1,249 @@
...
+
+# A Wolf, a goat, and some cabbage
+
+pgrx uses two different approaches to protect these FFI boundaries.  While both are implemented in Rust, one protects
+Rust from Postgres `setlongjmp` ERRORs and the other protects Postgres from Rust `panic()!`s.  To make things confusing
+they're both called `#[pg_guard]`.
+
+Essentially, pgrx needs to guard two styles of `extern "C"` functions.  One style is the `extern "C" {}` block that 
+declares a function lives "somewhere else" (in our case, the Postgres process in which the pgrx extension is loaded).
+The other style is `extern "C" fn foo() { ... }` functions that are written in Rust and might be passed to Postgres (for
```
workingjubilee:
```suggestion
The other style is [`extern "C" fn foo() { ... }` functions][extern-fn] that are written in Rust and might be passed to Postgres (for
```

docs/src/ffi-error-handling.md
```diff
@@ -0,0 +1,249 @@
...
+it to later call) via a standard function pointer.
+
+## Guarding Postgres Internal Functions
+
+pgrx uses the `bindgen` tool to generate "bindings" for exported Postgres symbols.  Postgres' source header files (`*.h`)
```
workingjubilee:
```suggestion
pgrx uses the [`bindgen`] tool to generate "bindings" for exported Postgres symbols.  Postgres' source header files (`*.h`)
```

docs/src/ffi-error-handling.md
```diff
@@ -0,0 +1,249 @@
+Postgres is written in C.  pgrx is written in Rust.  Between them is a boundary where each blindly believes the other
+behaves in an expected way.  Our primary concern with this boundary is error handling.
+
+There are many other concerns across this boundary such as function call ABIs and pointer ownership, but these are generally
+"obvious" concerns to anyone that's done any FFI development and won't be discussed here in detail.  
+
+
+# High-level Postgres Error Handling Overview
```
workingjubilee, line 8: We should probably mention that this is describing the workings behind `ereport`, as C programmers are more likely to recognize that name.

docs/src/ffi-error-handling.md
eeeebbbbrrrr: ehh.  I'll expound on that, I guess.  I mentally consider it a lie because **we** are updating Postgres' exception stack pointer thingiemabob to point to **our** code.  Postgres doesn't expect to jump into some Rust stack frame.  Fortunately, it doesn't need to know.

docs/src/ffi-error-handling.md
eeeebbbbrrrr: I think I understand what you mean, but I'm not *quite* sure of the words you'd want to explain it.  I'll let you add this point.  ;)

eeeebbbbrrrr:
@workingjubilee I think this is good enough.  

I left https://github.com/pgcentralfoundation/pgrx/pull/1729#discussion_r1623643435 for you.  As I said, I'm not quite sure how you'd want that to be explained.

workingjubilee (approved):
Cool. Thanks! I'll take a look at poking the last thing I mentioned later.

PR #1731, opened 2024-06-06
Make unboxing args sound with `ArgAbi`

workingjubilee:
This makes moving arguments into pgrx's `pg_extern` functions sound, by using a new trait, `ArgAbi`, to describe this behavior. The supporting `Arg` and `Args` types allow effectively treating the arguments of the function like an iterator, but the behavior is much more `unsafe`.

## The victories
- A `#[pg_extern] fn` must be sound in its arguments and return types.
- A large amount of our code generation for argument unboxing is now driven **entirely by type inference**, which simplifies the code analysis we must do.

## The collateral damage
- This kills a ton of our aggregate examples because they were relying on `'static` lifetimes and `#[pg_aggregate]` does not yet understand how to pass through lifetimes correctly.
- Arguments borrowed for `'static` now fail to compile, even in a hypothetical case it might be okay. This is because pgrx attempts to translate an argument from a lifetime-bound `FcInfo` into the `'static` lifetime, which is correctly rejected by Rust's type system.
- Returning borrows with the `'static` lifetime also fails, for similar reasons. This affects even hypothetical cases it would be okay (e.g. if the return was *actually* a `&'static str`). Types that are inherently `'static` like `String` are still fine.

## Future work
- We're going to bump into a lot of bad ergonomics very shortly as people test the new bounds, and will have to hammer out new API to make the desired patterns easy to write.
- This does not retire `FromDatum` and thus doesn't hammer the nail in the coffin on the unsoundness in the numerous SPI functions that use `FromDatum`.
- I have not yet begun to cull all the now-unnecessary code that tries to parse out types, but given that arguments are now driven from inference? There's a *lot* of it.

### DIFF
--- a/pgrx-examples/composite_type/src/lib.rs
+++ b/pgrx-examples/composite_type/src/lib.rs
@@ -145,6 +145,7 @@ fn make_friendship(
 }
 
 /*
+FIXME: make this example no longer DOA
 Create sum the scritches received by dogs, roughly the equivalent of:
 
 ```sql
@@ -161,64 +162,7 @@ CREATE AGGREGATE sum_scritches ("value" Dog) (
 )
 ```
 */
-struct SumScritches;
-
-#[pg_aggregate]
-impl Aggregate for SumScritches {
-    type State = i32;
-    const INITIAL_CONDITION: Option<&'static str> = Some("0");
-    type Args = pgrx::name!(value, pgrx::composite_type!('static, "Dog"));
-
-    fn state(
-        current: Self::State,
-        arg: Self::Args,
-        _fcinfo: pg_sys::FunctionCallInfo,
-    ) -> Self::State {
-        let arg_scritches: i32 = arg
-            .get_by_name("scritches")
-            .unwrap() // Unwrap the result of the conversion
-            .unwrap_or_default(); // The number of scritches, or 0 if there was none set
-        current + arg_scritches
-    }
-}
-
-/*
-Create sum the scritches received by dogs, roughly the equivalent of:
-
-```sql
-CREATE FUNCTION scritch_collector_state(state Dog, new integer)
-    RETURNS Dog
-    LANGUAGE SQL
-    STRICT
-    RETURN ROW(state.name, state.scritches + new)::Dog;
-
-CREATE AGGREGATE scritch_collector ("value" integer) (
-    SFUNC = "sum_scritches_state",
-    STYPE = Dog,
-)
-```
-*/
-struct ScritchCollector;
-
-#[pg_aggregate]
-impl Aggregate for ScritchCollector {
-    type State = Option<pgrx::composite_type!('static, "Dog")>;
-    type Args = i32;
-
-    fn state(
-        current: Self::State,
-        arg: Self::Args,
-        _fcinfo: pg_sys::FunctionCallInfo,
-    ) -> Self::State {
-        let mut current = match current {
-            Some(v) => v,
-            None => PgHeapTuple::new_composite_type(DOG_COMPOSITE_TYPE).unwrap(),
-        };
-        let current_scritches: i32 = current.get_by_name("scritches").unwrap().unwrap_or_default();
-        current.set_by_name("scritches", current_scritches + arg).unwrap();
-        Some(current)
-    }
-}
+// struct SumScritches;
 
 /*
 Create an operator allowing dogs to accept scritches directly.
--- a/pgrx-examples/composite_type/src/lib.rs
+++ b/pgrx-examples/composite_type/src/lib.rs
@@ -243,14 +243,6 @@ mod tests {
     }
 
     #[pg_test]
-    fn test_scritch_collector() {
-        let retval = Spi::get_one::<i32>(
-            "SELECT (scritchcollector(value)).scritches FROM UNNEST(ARRAY [1,2,3]) as value;",
-        );
-        assert_eq!(retval, Ok(Some(6)));
-    }
-
-    #[pg_test]
     fn test_dog_add_operator() {
         let retval = Spi::get_one::<i32>("SELECT (ROW('Nami', 0)::Dog + 1).scritches;");
         assert_eq!(retval, Ok(Some(1)));
--- a/pgrx-examples/strings/src/lib.rs
+++ b/pgrx-examples/strings/src/lib.rs
@@ -34,11 +34,6 @@ fn append(mut input: String, extra: &str) -> String {
 }
 
 #[pg_extern]
-fn split(input: &'static str, pattern: &str) -> Vec<&'static str> {
-    input.split_terminator(pattern).collect()
-}
-
-#[pg_extern]
 fn split_set<'a>(input: &'a str, pattern: &'a str) -> SetOfIterator<'a, &'a str> {
     SetOfIterator::new(input.split_terminator(pattern))
 }
--- a/pgrx-examples/strings/src/lib.rs
+++ b/pgrx-examples/strings/src/lib.rs
@@ -34,6 +34,11 @@ fn append(mut input: String, extra: &str) -> String {
 }
 
 #[pg_extern]
+fn split<'a>(input: &'a str, pattern: &str) -> Vec<&'a str> {
+    input.split_terminator(pattern).collect()
+}
+
+#[pg_extern]
 fn split_set<'a>(input: &'a str, pattern: &'a str) -> SetOfIterator<'a, &'a str> {
     SetOfIterator::new(input.split_terminator(pattern))
 }
--- a/pgrx-tests/src/tests/lifetime_tests.rs
+++ b/pgrx-tests/src/tests/lifetime_tests.rs
@@ -39,9 +39,9 @@ fn returns_option_ref_with_lifetime() -> Option<&'static str> {
 }
 
 #[pg_extern]
-fn returns_tuple_with_lifetime(
-    value: &'static str,
-) -> TableIterator<(name!(a, &'static str), name!(b, Option<&'static str>))> {
+fn returns_tuple_with_lifetime<'a>(
+    value: &'a str,
+) -> TableIterator<(name!(a, &'a str), name!(b, Option<&'a str>))> {
     TableIterator::once((value, Some(value)))
 }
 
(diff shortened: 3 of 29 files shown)


workingjubilee:
oh, right, tests.

workingjubilee:
Most of the current failures are from this, which will require the most thought. https://blog.rust-lang.org/2022/10/28/gats-stabilization.html#implied-static-requirement-from-higher-ranked-trait-bounds

pgrx/src/callconv.rs
```diff
@@ -488,12 +690,75 @@ impl<'fcx> FcInfo<'fcx> {
...
         }
     }
+
+    #[inline]
+    pub fn arguments<'arg>(&'arg self) -> Arguments<'arg, 'fcx> {
+        Arguments { iter: self.raw_args().iter().enumerate(), fcinfo: self }
+    }
+}
+
+// TODO: rebadge this as AnyElement
```
NotGyro, line 755: Nit - could this be within the scope of this pull request? "Argument" does describe it, though, I don't mind this.

pgrx/src/callconv.rs
```diff
@@ -488,12 +690,75 @@ impl<'fcx> FcInfo<'fcx> {
...
+pub struct Argument<'a, 'fcx>(&'a FcInfo<'fcx>, usize, &'a pg_sys::NullableDatum);
+
+impl<'a, 'fcx> Argument<'a, 'fcx> {
+    /// Note: some overhead
```
NotGyro: Might be worth documenting (vague shape of) the overhead here

NotGyro (approved):
I've read through this, I'm very impressed with how much cleaner / more sensible the lifetimes are with these changes. Looks good to me apart from a few documentation nitpicks.

pgrx/src/callconv.rs
workingjubilee: Oh, so AnyElement has a special type definition relative to Postgres. Our AnyElement is here: https://github.com/pgcentralfoundation/pgrx/blob/7ea24807bd5382f74cf5b4ea547452bca37203ee/pgrx/src/datum/anyelement.rs#L22-L26

And it's modeling this type: https://www.postgresql.org/docs/current/datatype-pseudo.html

But there may have to be a bit of additional work to make a wrapper around Argument usable as AnyElement. It's based on the use-cases.

pgrx/src/callconv.rs
workingjubilee: Oh, it's an FFI call, so.

pgrx-examples/composite_type/src/lib.rs
```diff
@@ -145,6 +145,7 @@ fn make_friendship(
 }
 
 /*
+FIXME: make this example no longer DOA
```
eeeebbbbrrrr, line 148: What's the plan for this?  Aggregate support is a pretty important feature

pgrx-examples/strings/src/lib.rs
```diff
@@ -33,11 +33,6 @@ fn append(mut input: String, extra: &str) -> String {
     input
 }
 
-#[pg_extern]
-fn split(input: &'static str, pattern: &str) -> Vec<&'static str> {
-    input.split_terminator(pattern).collect()
-}
-
```
eeeebbbbrrrr: Why was this removed?  And what *should* the function signature look like?

pgrx-tests/src/tests/lifetime_tests.rs
```diff
@@ -39,9 +39,9 @@ fn returns_option_ref_with_lifetime() -> Option<&'static str> {
 }
 
 #[pg_extern]
-fn returns_tuple_with_lifetime(
-    value: &'static str,
-) -> TableIterator<(name!(a, &'static str), name!(b, Option<&'static str>))> {
+fn returns_tuple_with_lifetime<'a>(
+    value: &'a str,
+) -> TableIterator<(name!(a, &'a str), name!(b, Option<&'a str>))> {
```
eeeebbbbrrrr, line 44: +1

eeeebbbbrrrr (approved):
My only question is:  What's the plan for Aggregates?

Also, do you have a simple example of what a `#[pg_extern]` function expands to now?  I feel like I just recently wrote some docs that *kinda* outlines that, and we should probably update those too.

eeeebbbbrrrr:
This has clearly been a ton of work!  I'm excited!!

pgrx-examples/strings/src/lib.rs
NotGyro: I'm guessing it's the `'static` that's objectionable here.

workingjubilee: Correct.

workingjubilee:
@eeeebbbbrrrr

The wrapper function will always expand like this:
```rust
#[no_mangle]
#[doc(hidden)]
pub unsafe extern "C" fn substring_wrapper(
    fcinfo: ::pgrx::pg_sys::FunctionCallInfo,
) -> ::pgrx::pg_sys::Datum {
    fn _internal_wrapper<'fcx>(
        fcinfo: &mut ::pgrx::callconv::FcInfo<'fcx>,
    ) -> ::pgrx::datum::Datum<'fcx> {
        #[allow(unused_unsafe)]
        unsafe {
            let call_flow = <&str as ::pgrx::callconv::RetAbi>::check_and_prepare(fcinfo);
            let result = match call_flow {
                ::pgrx::callconv::CallCx::WrappedFn(mcx) => {
                    let mut mcx = ::pgrx::PgMemoryContexts::For(mcx);
                    let _args = &mut fcinfo.arguments();
                    let call_result = mcx.switch_to(|_| {
                        let input_ = _args.unbox_next_unchecked().unwrap_or_else(|| {
                            ::core::panicking::panic_fmt(format_args!(
                                "unboxing {0} argument failed",
                                "input_"
                            ));
                        });
                        let start_ = _args.unbox_next_unchecked().unwrap_or_else(|| {
                            ::core::panicking::panic_fmt(format_args!(
                                "unboxing {0} argument failed",
                                "start_"
                            ));
                        });
                        let end_ = _args.unbox_next_unchecked().unwrap_or_else(|| {
                            ::core::panicking::panic_fmt(format_args!(
                                "unboxing {0} argument failed",
                                "end_"
                            ));
                        });
                        substring(input_, start_, end_)
                    });
                    ::pgrx::callconv::RetAbi::to_ret(call_result)
                }
                ::pgrx::callconv::CallCx::RestoreCx => {
                    <&str as ::pgrx::callconv::RetAbi>::ret_from_fcx(fcinfo)
                }
            };
            ::core::mem::transmute(unsafe {
                <&str as ::pgrx::callconv::RetAbi>::box_ret_in(fcinfo, result)
            })
        }
    }
    let datum = unsafe {
        ::pgrx::pg_sys::submodules::panic::pgrx_extern_c_guard(|| {
            let mut fcinfo = ::pgrx::callconv::FcInfo::from_ptr(fcinfo);
            _internal_wrapper(&mut fcinfo)
        })
    };
    datum.sans_lifetime()
}
```

There is no meaningful variation except based on number of arguments.

pgrx-examples/strings/src/lib.rs
eeeebbbbrrrr:
```rust
#[pg_extern]
fn split<'a>(input: &'a str, pattern: &str) -> Vec<&'a str> {
    input.split_terminator(pattern).collect()
}
```

Is that the new form?  Should we just update that example rather than removing it?

eeeebbbbrrrr:
```rust
...
::core::mem::transmute(unsafe {
                <&str as ::pgrx::callconv::RetAbi>::box_ret_in(fcinfo, result)
            })
...
```            

What is the thing being transmuted here?  Is it the lifetime of the return value?

workingjubilee:
Correct.

I wonder if I can erase that...

workingjubilee:
Yeah, that lifetime transmutation seems to be completely unnecessary now.

workingjubilee:
lol, well, CI isn't going to prove it anytime soon.

GitHub is zzz

pgrx-examples/strings/src/lib.rs
workingjubilee: Sure. It seemed slightly redundant with the next example, is probably why I deleted it.

pgrx-examples/composite_type/src/lib.rs
workingjubilee: The problem isn't with Aggregate, per se, it's with certain types.

These are aggregates that are supposed to manage composite types. But we know composite types are allocated in Postgres, thus are effectively lifetime-bound. But `#[pg_aggregate]` and Aggregate don't know how to convey the lifetimes through. Thus even annotating it with something like this does not actually fix the problem:
```rust
#[pg_aggregate]
impl<'a> Aggregate for SumScritches<'a> {
    type State = i32;
    const INITIAL_CONDITION: Option<&'static str> = Some("0");
    type Args = pgrx::name!(value, pgrx::composite_type!('a, "Dog"));

    fn state(
        current: Self::State,
        arg: Self::Args,
        _fcinfo: pg_sys::FunctionCallInfo,
    ) -> Self::State {
        todo!()
    }
}
```

Thus in order to make aggregates work for complex cases we must break the world for them as well, making all their functions use the correct, lifetime-bound types. Hopefully this will let us simplify the code expansion for them as well.

workingjubilee:
I wasn't quite happy with the remainder of the point of InOutFuncs being part of the collateral damage, so I adjusted it to better support the quirky behavior of Postgres on that count. Regardless of what I think about the API design, the fact that its test failed seemed bad for correctness, especially given how easy it was to incur. Now, if you ask for, say, a `Nullable<Option<&'a CStr>>`? Then everything should behave correctly, and you should get exactly what you would expect. This sequence isn't actually in the code due to the sheer levels of argument forwarding, but the result is this:

```rust
if nullable_datum.isnull {
    Nullable::None
} else if nullable_datum.value == pg_sys::Datum::null() {
    Nullable::Valid(None)
} else {
    Nullable::Valid(Some(the_conversion(nullable_datum.value)))
}
```

workingjubilee:
argh.

jamessewell:
So this lays the ground work for pg_agg, but that work isn't happening atm is it?

workingjubilee:
correct.

PR #1732, opened 2024-06-06
Remove `UnwindSafe` bound from `pgrx_extern_c_guard`

workingjubilee:
Our invariants are so complicated and need such special handling that once they are take care of, `UnwindSafe` adds nothing. The particular example that motivated this: `&mut T` isn't `UnwindSafe` because it's *easy* to witness a broken invariant, not that you will.

See https://github.com/rust-lang/rfcs/pull/3260 for further rationale.

### DIFF
--- a/pgrx-pg-sys/src/submodules/panic.rs
+++ b/pgrx-pg-sys/src/submodules/panic.rs
@@ -16,7 +16,7 @@ use std::cell::Cell;
 use std::fmt::{Display, Formatter};
 use std::hint::unreachable_unchecked;
 use std::panic::{
-    catch_unwind, panic_any, resume_unwind, Location, PanicInfo, RefUnwindSafe, UnwindSafe,
+    catch_unwind, panic_any, resume_unwind, AssertUnwindSafe, Location, PanicInfo, UnwindSafe,
 };
 
 use crate::elog::PgLogLevel;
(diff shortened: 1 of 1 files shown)


pgrx-pg-sys/src/submodules/panic.rs
```diff
@@ -381,9 +381,9 @@ enum GuardAction<R> {
 // what we really want is a bound of R: !Drop, but negative bounds don't exist yet
 pub unsafe fn pgrx_extern_c_guard<Func, R>(f: Func) -> R
 where
-    Func: FnOnce() -> R + UnwindSafe + RefUnwindSafe,
+    Func: FnOnce() -> R,
 {
-    match unsafe { run_guarded(f) } {
+    match unsafe { run_guarded(AssertUnwindSafe(f)) } {
```
eeeebbbbrrrr, line 386: Interesting.

eeeebbbbrrrr (commented):
while I understand the what of this, I don't quite understand the why.

Then again, when I added the `+ RefUnwindSafe` bound, I didn't quite understand the why at that time either.  
EDIT:  As I recall, that was me appeasing the compiler.  Which I've since learned isn't always the correct approach!

My understanding tho is that `AssertUnwindSafe` is fine to basically YOLO?

workingjubilee:
Hm.

...I actually just thought of a trivial change that makes the ArgAbi code compile without this, so now I feel slightly silly. :V

...but that's kind of the problem, in a way. It's basically meaningless.

workingjubilee:
@eeeebbbbrrrr In a sense, we do have a need for "unwind safety" in pgrx. However, our invariants are so horribly complicated and need so much special handling that essentially, once they are take care of, `UnwindSafe` adds nothing. And `&mut T` isn't `UnwindSafe` because it's _easy_ to witness a broken invariant, not that you _will_.

eeeebbbbrrrr:
> `UnwindSafe` adds nothing.

fair.

eeeebbbbrrrr (approved):

PR #1733, opened 2024-06-07
Fix lifetimes for IO functions

workingjubilee:
While sorting out lifetime problems to make #1731 landable, I noticed JsonInOutFuncs is completely replaceable with two reexports. So I replaced it. This also made the changes to sort out the rest of the lifetimes obvious, and now we return CString always from these functions.


workingjubilee:
This plus another small fix reduces the error output remaining for #1731 from 465 to 305 lines.

workingjubilee:
235, actually.

workingjubilee:
Another fix and now down to 199, and now I just need to rethink how I'm doing the ArgAbi impl for Vec.

PR #1737, opened 2024-06-19
impl Allocator for MemCx

NotGyro:
Implements the nightly-only [`std::alloc::Allocator`] trait on `MemCx`, and adds a `nightly` feature flag to conditionally-compile this implementation (so that builds on `stable` will not break, but builds on `nightly` will have the new `Allocator` implementation).

Tests for this allocator have also been written.

Additionally, I have expanded documentation on existing `MemCx` methods.

# Future work

We may want to determine if / how to deal with large allocations (larger than 1 GB in a single allocation). `MCXT_ALLOC_HUGE` will be useful here.

[`std::alloc::Allocator`]: https://doc.rust-lang.org/std/alloc/trait.Allocator.html


pgrx/src/memcx.rs
```diff
@@ -16,27 +18,35 @@ pub struct MemCx<'mcx> {
...
     }
 
-    /// You probably shouldn't be using this.
-    pub(crate) unsafe fn exec_in<T>(&self, f: impl FnOnce() -> T) -> T {
+    /// Stores the current global memory context, switches to *this* memory context,
+    /// and executes the closure `f`.
+    /// Once `f` completes, the previous global memory context is restored.
+    pub unsafe fn exec_in<T>(&self, f: impl FnOnce() -> T) -> T {
         let remembered = pg_sys::MemoryContextSwitchTo(self.ptr.as_ptr());
         let res = f();
```
eeeebbbbrrrr, line 48: if `f()` panics then we're left in `self`'s memory context and not whatever we `remembered`.  I'm not sure this is what we'd want here.  

It might make sense to use `PgTryBuilder()` to run `f()`, which would give us a "finally" block where we can make sure to restore the memory context to `remembered`.

pgrx/src/memcx.rs
```diff
@@ -45,3 +55,99 @@ where
...
+        layout: std::alloc::Layout,
+    ) -> Result<NonNull<[u8]>, std::alloc::AllocError> {
+        // Overriding default function here to use Postgres' zeroing implementation.
+        // Postgres 16 and newer permit any arbitrary power-of-2 alignment
+        unsafe {
+            // Bitflags for MemoryContextAllocAligned:
+            // #define MCXT_ALLOC_HUGE    0x01 /* allow huge allocation (> 1 GB) */
+            // #define MCXT_ALLOC_NO_OOM  0x02 /* no failure if out-of-memory */
+            // #define MCXT_ALLOC_ZERO    0x04 /* zero allocated memory */
+            let ptr = pgrx_pg_sys::MemoryContextAllocAligned(
```
eeeebbbbrrrr: Did you actually want to use `MemoryContextAllocZeroAligned` here?

pgrx/src/memcx.rs
```diff
@@ -45,3 +55,99 @@ where
...
+        // On versions before Postgres 16, alignment is always 8 byte / 64 bit.
+        let layout = layout_for_pre16(layout)?;
+        unsafe {
+            let ptr = pgrx_pg_sys::MemoryContextAllocZero(self.ptr.as_ptr(), layout.size());
+            let slice: &mut [u8] = slice::from_raw_parts_mut(ptr.cast(), layout.size());
+            Ok(NonNull::new_unchecked(slice))
+        }
+    }
+
+    #[cfg(not(any(feature = "pg12", feature = "pg13", feature = "pg14", feature = "pg15")))]
```
eeeebbbbrrrr: I don't know if @workingjubilee has a preference, but writing this as just `#[cfg(feature = "pg16")]` would be a lot easer to reason about.

pgrx/src/memcx.rs
```diff
@@ -45,3 +55,99 @@ where
...
+    fn allocate(
+        &self,
+        layout: std::alloc::Layout,
+    ) -> Result<NonNull<[u8]>, std::alloc::AllocError> {
```
eeeebbbbrrrr: One thing I don't quite understand.  The returned byte array isn't associated with any lifetime.  Not the lifetime of `&self` nor even `'mcx`.

The doc comments on `std::alloc::Allocator::allocate()` don't make a lot of sense to me, so how does the compiler otherwise ensure that the function's result can't be used after free?

workingjubilee:
@NotGyro Can you create a set of nightly-only compile-fail tests, idk, in `pgrx-tests/tests/nightly-fail/` or something, gated on the pgrx `nightly` feature?

pgrx/src/memcx.rs
workingjubilee: This is not meant to be commonly used.

An additional "this is UB if anything panics" constraint is acceptable.

pgrx/src/memcx.rs
workingjubilee: For `Vec<T, A>` the lifetime of `A` is what we're using to make it so that allocations cannot outlive their memory context.

pgrx/src/memcx.rs
```diff
@@ -16,27 +18,35 @@ pub struct MemCx<'mcx> {
...
     pub(crate) unsafe fn from_ptr(raw: pg_sys::MemoryContext) -> MemCx<'mcx> {
         let ptr = NonNull::new(raw).expect("memory context must be non-null");
         MemCx { ptr, _marker: PhantomData }
     }
 
-    /// You probably shouldn't be using this.
-    pub(crate) unsafe fn alloc_bytes(&self, size: usize) -> *mut u8 {
+    /// Allocate a raw byte buffer `size` bytes in length
+    /// and returns a pointer to the new allocation.
+    pub unsafe fn alloc_bytes(&self, size: usize) -> *mut u8 {
```
workingjubilee: Needs the safety constraint documented.

pgrx/src/memcx.rs
```diff
@@ -16,27 +18,35 @@ pub struct MemCx<'mcx> {
 }
 
 impl<'mcx> MemCx<'mcx> {
-    /// You probably shouldn't be using this.
+    /// Wrap the provided [`pg_sys::MemoryContext`]
+    ///
+    /// # Safety:
```
workingjubilee: This line gets formatted as a title.
```suggestion
    /// # Safety
```

pgrx/src/memcx.rs
```diff
@@ -16,27 +18,35 @@ pub struct MemCx<'mcx> {
...
+    pub unsafe fn exec_in<T>(&self, f: impl FnOnce() -> T) -> T {
         let remembered = pg_sys::MemoryContextSwitchTo(self.ptr.as_ptr());
         let res = f();
         pg_sys::MemoryContextSwitchTo(remembered);
         res
     }
 }
 
-/// Acquire the current context and operate inside it.
+/// Acquire the current global context and operate inside it.
```
workingjubilee: Not sure what this means? Memory contexts are usually not conceptually global, but rather bound to increasingly local scopes. The current context is often one which will end immediately after the outer function call is over.

pgrx/src/memcx.rs
```diff
@@ -45,3 +55,99 @@ where
...
+        layout: std::alloc::Layout,
+    ) -> Result<NonNull<[u8]>, std::alloc::AllocError> {
+        // On versions before Postgres 16, alignment is always 8 byte / 64 bit.
+        let layout = layout_for_pre16(layout)?;
+        unsafe {
+            let ptr = pgrx_pg_sys::MemoryContextAlloc(self.ptr.as_ptr(), layout.size());
+            let slice: &mut [u8] = slice::from_raw_parts_mut(ptr.cast(), layout.size());
+            Ok(NonNull::new_unchecked(slice))
+        }
+    }
```
workingjubilee: This is not correct. You cannot achieve the goal you are trying to accomplish by tinkering with `alloc::Layout`. The `Layout` is a *command* to you. You obey or you fail, you must not modify the `Layout`. Please just return an AllocError always on these versions for now.

workingjubilee (changes requested):
You need a test for this working with a highly-aligned type which will show that allocations fail if you try to allocate something with e.g. `repr(align(4096))`.

pgrx/src/memcx.rs
eeeebbbbrrrr: > An additional "this is UB if anything panics" constraint is acceptable.

Fine by me.  ;)

pgrx/src/memcx.rs
NotGyro: `MemoryContextAllocZeroAligned` is... *weird*. It existed before Postgres 16 introduced the arbitrary-alignment allocation API, and it does not take an alignment argument. The doc-string says 

```
/*
* MemoryContextAllocZeroAligned
*      MemoryContextAllocZero where length is suitable for MemSetLoop
*
*  This might seem overly specialized, but it's not because newNode()
*  is so often called with compile-time-constant sizes.
*/
```

[and it was removed from the Postgres 17 beta in a recent commit.](https://github.com/postgres/postgres/commit/3c080fb4fad3e1c1e34f74a7b84a443137adc9f2) 

I suspect it is misleadingly-named, and is instead "aligned" in the sense of "aligned for a very specific internal use-case"

eeeebbbbrrrr: As this is, it's missing setting the allocated chunk to zeros, isn't it?

pgrx/src/memcx.rs
NotGyro: Yeah it does state 
```
    /// The returned block of memory remains valid as long as it is [*currently allocated*] and the shorter of:
    ///   - the borrow-checker lifetime of the allocator type itself.
    ///   - as long as at the allocator and all its clones has not been dropped.
```

but the problem is that these methods do not take or return lifetimes, so it might just be compiler magic we have to trust? @workingjubilee would know more here, I think.

pgrx/src/memcx.rs
NotGyro: Should be corrected in https://github.com/pgcentralfoundation/pgrx/pull/1737/commits/f149072c318f7e39c212daf3b59e0f723d1dfd46

pgrx/src/memcx.rs
NotGyro: Should be corrected in https://github.com/pgcentralfoundation/pgrx/pull/1737/commits/f149072c318f7e39c212daf3b59e0f723d1dfd46

pgrx/src/memcx.rs
NotGyro: Corrected in https://github.com/pgcentralfoundation/pgrx/pull/1737/commits/f149072c318f7e39c212daf3b59e0f723d1dfd46

pgrx/src/memcx.rs
NotGyro: Still pondering if this should be safe or not...

pgrx/src/memcx.rs
NotGyro: Corrected in https://github.com/pgcentralfoundation/pgrx/pull/1737/commits/f149072c318f7e39c212daf3b59e0f723d1dfd46

pgrx/src/memcx.rs
NotGyro: Per Discord conversation on correctness, documented this as a safety constraint instead in https://github.com/pgcentralfoundation/pgrx/pull/1737/commits/f149072c318f7e39c212daf3b59e0f723d1dfd46

pgrx/src/memcx.rs
NotGyro: > As this is, it's missing setting the allocated chunk to zeros, isn't it?

[Per Postgres documentation](https://doxygen.postgresql.org/palloc_8h.html), passing in bitflag 4 should zero the memory.

See: https://github.com/postgres/postgres/blob/master/src/backend/utils/mmgr/mcxt.c#L1256
https://github.com/postgres/postgres/blob/master/src/include/utils/palloc.h#L66

pgrx/src/memcx.rs
NotGyro: > For `Vec<T, A>` the lifetime of `A` is what we're using to make it so that allocations cannot outlive their memory context.

This sounds right.

NotGyro:
> @NotGyro Can you create a set of nightly-only compile-fail tests, idk, in `pgrx-tests/tests/nightly-fail/` or something, gated on the pgrx `nightly` feature?

Done. This has made the number of files modified in the PR enormous, but that's somewhat unavoidable. I'd say just focus on the non-compilefail/todo files when reviewing.

NotGyro:
> You need a test for this working with a highly-aligned type which will show that allocations fail if you try to allocate something with e.g. `repr(align(4096))`.

Should be taken care of in 69bb9d48269bd10acf2b11fcc60ea2058a5bc09f

pgrx/src/memcx.rs
NotGyro: Per our conversation, this appears to be safe. I've made it a safe method.

workingjubilee:
> > @NotGyro Can you create a set of nightly-only compile-fail tests, idk, in `pgrx-tests/tests/nightly-fail/` or something, gated on the pgrx `nightly` feature?
> 
> Done. This has made the number of files modified in the PR enormous, but that's somewhat unavoidable. I'd say just focus on the non-compilefail/todo files when reviewing.

Oh, I mostly meant "so we can add compile-fail tests to verify a `Vec<T, MemCx<'mcx>>` cannot outlive its associated `MemCx<'mcx>`".

NotGyro:
> > > @NotGyro Can you create a set of nightly-only compile-fail tests, idk, in `pgrx-tests/tests/nightly-fail/` or something, gated on the pgrx `nightly` feature?
> > 
> > 
> > Done. This has made the number of files modified in the PR enormous, but that's somewhat unavoidable. I'd say just focus on the non-compilefail/todo files when reviewing.
> 
> Oh, I mostly meant "so we can add compile-fail tests to verify a `Vec<T, MemCx<'mcx>>` cannot outlive its associated `MemCx<'mcx>`".

One of these has been implemented in 6b5c34dd99ea887834d83a6ed435fdcbe88671e8. I could also set up another test with a MemCx constructed from a nonsense pointer (since the compiler won't catch this in unsafe code, but it would catch if the lifetime referencing the MemCx outlives it) if we need one that doesn't use a closure, but this behavior should be consistent between normal scopes and closures.

pgrx-tests/tests/nightly/compile-fail/allocation-doesnt-outlive-memcx.rs
```diff
@@ -0,0 +1,18 @@
...
+    let mut vector_used = Arc::new(None);
+
+    pgrx::memcx::current_context(|mcx: &MemCx| {
+        let mut inner_vec = Vec::new_in(&mcx);
+        inner_vec.push(451);
+        inner_vec.push(242);
+        *Arc::<std::option::Option<_>>::get_mut(&mut vector_used).unwrap() = Some(inner_vec);
+    });
+    let _vector = Arc::<_>::get_mut(&mut vector_used).take().unwrap();
+}
```
workingjubilee: nit:
```suggestion
}

```

pgrx-tests/tests/nightly/compile-fail/allocation-doesnt-outlive-memcx.rs
workingjubilee, line 1: Nice!

pgrx-tests/tests/nightly/compile-fail/arrays-dont-leak.rs
workingjubilee, line 1: I don't think we need this duplicate set of tests for all the others on nightly, I'm going to instead look into migrating to `ui_test` which will allow specifically documenting our expectations and using [normalization regexes](https://github.com/oli-obk/ui_test/blob/ed37e39d9c37f164df54ab947719c25a5e136b50/README.md?plain=1#L49-L50) to handle other cases.

pgrx/src/memcx.rs
```diff
@@ -45,3 +62,54 @@ where
 
     f(&memcx)
 }
+
+#[cfg(all(feature = "nightly", feature = "pg16"))]
+unsafe impl<'mcx> std::alloc::Allocator for &MemCx<'mcx> {
```
workingjubilee: I think it would be slightly nicer to do
```suggestion
#[cfg(all(feature = "nightly", feature = "pg16"))]
mod nightly {
    use std::slice;
    use super::*;

    unsafe impl<'mcx> std::alloc::Allocator for &MemCx<'mcx> {
```

pgrx/src/memcx.rs
```diff
@@ -45,3 +62,54 @@ where
...
+                self.ptr.as_ptr(),
+                layout.size(),
+                layout.align(),
+                4,
+            );
+            let slice: &mut [u8] = slice::from_raw_parts_mut(ptr.cast(), layout.size());
+            Ok(NonNull::new_unchecked(slice))
+        }
+    }
+}
```
workingjubilee, line 118:
```suggestion
            }
        }
    }
}
```

pgrx/src/memcx.rs
```diff
@@ -8,6 +8,8 @@
 // And it's nice-ish to have shorter lifetime names and have 'mcx consistently mean the lifetime.
 use crate::pg_sys;
 use core::{marker::PhantomData, ptr::NonNull};
+#[cfg(feature = "nightly")]
+use std::slice;
```
workingjubilee: Then we can
```suggestion
```

pgrx/src/memcx.rs
```diff
@@ -45,3 +62,54 @@ where
...
+            );
+            let slice: &mut [u8] = slice::from_raw_parts_mut(ptr.cast(), layout.size());
+            Ok(NonNull::new_unchecked(slice))
+        }
+    }
+
+    unsafe fn deallocate(&self, ptr: NonNull<u8>, _layout: std::alloc::Layout) {
+        // TODO: Find faster free for use when MemoryContext is known.
+        // This is the global function that looks up the relevant Memory Context by address range.
+        pgrx_pg_sys::pfree(ptr.as_ptr().cast())
```
workingjubilee: They do this via a macro that accesses a `static const` and calls a `static inline` function, so we have no realistic way of doing better without PRing Postgres, basically.

nit:
```suggestion
        pg_sys::pfree(ptr.as_ptr().cast())
```

pgrx/src/memcx.rs
```diff
@@ -45,3 +62,54 @@ where
...
+    fn allocate(
+        &self,
+        layout: std::alloc::Layout,
+    ) -> Result<NonNull<[u8]>, std::alloc::AllocError> {
+        unsafe {
+            // Bitflags for MemoryContextAllocAligned:
+            // #define MCXT_ALLOC_HUGE    0x01 /* allow huge allocation (> 1 GB) */
+            // #define MCXT_ALLOC_NO_OOM  0x02 /* no failure if out-of-memory */
+            // #define MCXT_ALLOC_ZERO    0x04 /* zero allocated memory */
+            let ptr = pgrx_pg_sys::MemoryContextAllocAligned(
```
workingjubilee: nit:
```suggestion
            let ptr = pg_sys::MemoryContextAllocAligned(
```

pgrx/src/memcx.rs
```diff
@@ -45,3 +62,54 @@ where
...
+        layout: std::alloc::Layout,
+    ) -> Result<NonNull<[u8]>, std::alloc::AllocError> {
+        // Overriding default function here to use Postgres' zeroing implementation.
+        // Postgres 16 and newer permit any arbitrary power-of-2 alignment
+        unsafe {
+            // Bitflags for MemoryContextAllocAligned:
+            // #define MCXT_ALLOC_HUGE    0x01 /* allow huge allocation (> 1 GB) */
+            // #define MCXT_ALLOC_NO_OOM  0x02 /* no failure if out-of-memory */
+            // #define MCXT_ALLOC_ZERO    0x04 /* zero allocated memory */
+            let ptr = pgrx_pg_sys::MemoryContextAllocAligned(
```
workingjubilee: nit:
```suggestion
            let ptr = pg_sys::MemoryContextAllocAligned(
```

pgrx-tests/src/tests/memcxt_tests.rs
```diff
@@ -97,4 +98,127 @@ mod tests {
...
+                let foo = Foo::new(i, mcx);
+                structures.push(foo);
+            }
+            for foo in structures.iter() {
+                foo.validate_content();
+                let addr = (foo as *const Foo) as usize;
+                assert!((addr % 4096) == 0);
+            }
+        });
+    }
```
workingjubilee, line 220: Nice, thank you!

pgrx-tests/src/tests/memcxt_tests.rs
```diff
@@ -97,4 +98,127 @@ mod tests {
         drop(ctx);
         assert_eq!(unsafe { pg_sys::CurrentMemoryContext }, ctx_parent);
     }
+
+    #[cfg(feature = "nightly")]
+    #[pg_test]
+    #[no_mangle]
```
workingjubilee: hm. why did you add `#[no_mangle]` to these?

workingjubilee (approved):
Nice! Just a small cleanup detail and we're done, I think!

workingjubilee:
>
> One of these has been implemented in [6b5c34d](https://github.com/pgcentralfoundation/pgrx/commit/6b5c34dd99ea887834d83a6ed435fdcbe88671e8). I could also set up another test with a MemCx constructed from a nonsense pointer (since the compiler won't catch this in unsafe code, but it would catch if the lifetime referencing the MemCx outlives it) if we need one that doesn't use a closure, but this behavior should be consistent between normal scopes and closures.

Yeah, I don't think we need to test a nonsensical `unsafe` configuration.

pgrx/src/memcx.rs
workingjubilee: This would also make it easier to add other std imports for just this segment of code.

pgrx-tests/src/tests/memcxt_tests.rs
NotGyro: Probably doesn't help, I just wanted to make sure this doesn't optimize the behavior we actually want to test out of existence because the test is too structurally simple or w/e. And so it is probably not necessary.

pgrx-tests/tests/nightly/compile-fail/arrays-dont-leak.rs
NotGyro: Done in 9fcca5c3

PR #1740, opened 2024-06-24
Address `clippy::must_use_candidate` in template

nyurik:
This addresses `clippy::must_use_candidate` and applies a minor optimization: it removes a Vec realloc by simply taking the existing one, instead of cloning and clearing it.

### DIFF
--- a/cargo-pgrx/src/templates/lib_rs
+++ b/cargo-pgrx/src/templates/lib_rs
@@ -23,10 +23,12 @@ mod tests {{
 /// It must be visible at the root of your extension crate.
 #[cfg(test)]
 pub mod pg_test {{
+    #[allow(clippy::needless_pass_by_value)]
     pub fn setup(_options: Vec<&str>) {{
         // perform one-off initialization when the pg_test framework starts
     }}
 
+    #[must_use]
     pub fn postgresql_conf_options() -> Vec<&'static str> {{
         // return any postgresql.conf settings that are required for your tests
         vec![]
--- a/cargo-pgrx/src/templates/lib_rs
+++ b/cargo-pgrx/src/templates/lib_rs
@@ -23,7 +23,6 @@ mod tests {{
 /// It must be visible at the root of your extension crate.
 #[cfg(test)]
 pub mod pg_test {{
-    #[allow(clippy::needless_pass_by_value)]
     pub fn setup(_options: Vec<&str>) {{
         // perform one-off initialization when the pg_test framework starts
     }}
--- a/pgrx-macros/src/rewriter.rs
+++ b/pgrx-macros/src/rewriter.rs
@@ -10,6 +10,7 @@
 extern crate proc_macro;
 
 use quote::{format_ident, quote, quote_spanned};
+use std::mem::take;
 use std::ops::Deref;
 use std::str::FromStr;
 use syn::punctuated::Punctuated;
@@ -35,8 +36,7 @@ pub fn item_fn_without_rewrite(mut func: ItemFn) -> syn::Result<proc_macro2::Tok
     let input_func_name = func.sig.ident.to_string();
     let sig = func.sig.clone();
     let vis = func.vis.clone();
-    let attrs = func.attrs.clone();
-
+    let mut attrs = take(&mut func.attrs);
     let generics = func.sig.generics.clone();
 
     if attrs.iter().any(|attr| attr.path().is_ident("no_mangle"))
@@ -49,10 +49,12 @@ pub fn item_fn_without_rewrite(mut func: ItemFn) -> syn::Result<proc_macro2::Tok
         panic!("#[pg_guard] for function with generic parameters must not be combined with #[no_mangle]");
     }
 
+    attrs.push(syn::parse_quote!(#[automatically_derived]));
+    attrs.push(syn::parse_quote!(#[allow(clippy::used_underscore_binding)]));
+
     // but for the inner function (the one we're wrapping) we don't need any kind of
     // abi classification
     func.sig.abi = None;
-    func.attrs.clear();
 
     // nor do we need a visibility beyond "private"
     func.vis = Visibility::Inherited;
--- a/pgrx-macros/src/rewriter.rs
+++ b/pgrx-macros/src/rewriter.rs
@@ -10,7 +10,7 @@
 extern crate proc_macro;
 
 use quote::{format_ident, quote, quote_spanned};
-use std::mem::take;
+use std::mem;
 use std::ops::Deref;
 use std::str::FromStr;
 use syn::punctuated::Punctuated;
@@ -36,7 +36,7 @@ pub fn item_fn_without_rewrite(mut func: ItemFn) -> syn::Result<proc_macro2::Tok
     let input_func_name = func.sig.ident.to_string();
     let sig = func.sig.clone();
     let vis = func.vis.clone();
-    let mut attrs = take(&mut func.attrs);
+    let mut attrs = mem::take(&mut func.attrs);
     let generics = func.sig.generics.clone();
 
     if attrs.iter().any(|attr| attr.path().is_ident("no_mangle"))
@@ -48,9 +48,6 @@ pub fn item_fn_without_rewrite(mut func: ItemFn) -> syn::Result<proc_macro2::Tok
     {
         panic!("#[pg_guard] for function with generic parameters must not be combined with #[no_mangle]");
     }
-
-    attrs.push(syn::parse_quote!(#[automatically_derived]));
-    attrs.push(syn::parse_quote!(#[allow(clippy::used_underscore_binding)]));
 
     // but for the inner function (the one we're wrapping) we don't need any kind of
     // abi classification


cargo-pgrx/src/templates/lib_rs
```diff
@@ -23,10 +23,12 @@ mod tests {{
 /// It must be visible at the root of your extension crate.
 #[cfg(test)]
 pub mod pg_test {{
+    #[allow(clippy::needless_pass_by_value)]
     pub fn setup(_options: Vec<&str>) {{
```
workingjubilee: I would rather just fix the actual signature of `setup` for pgrx.

pgrx-macros/src/rewriter.rs
```diff
@@ -35,8 +36,7 @@ pub fn item_fn_without_rewrite(mut func: ItemFn) -> syn::Result<proc_macro2::Tok
     let input_func_name = func.sig.ident.to_string();
     let sig = func.sig.clone();
     let vis = func.vis.clone();
-    let attrs = func.attrs.clone();
-
+    let mut attrs = take(&mut func.attrs);
```
workingjubilee: I don't know why this is passing tests, seeing as we use the entire func, including its attrs, down here: https://github.com/pgcentralfoundation/pgrx/blob/7ea24807bd5382f74cf5b4ea547452bca37203ee/pgrx-macros/src/rewriter.rs#L108

workingjubilee: I suppose because we apply attrs to the outer function as well, but I don't feel confident asserting that it's fine to simply lift the attrs up in all cases.

pgrx-macros/src/rewriter.rs
```diff
@@ -49,10 +49,12 @@ pub fn item_fn_without_rewrite(mut func: ItemFn) -> syn::Result<proc_macro2::Tok
         panic!("#[pg_guard] for function with generic parameters must not be combined with #[no_mangle]");
     }
 
+    attrs.push(syn::parse_quote!(#[automatically_derived]));
```
workingjubilee: What, concretely, does this help with?

pgrx-macros/src/rewriter.rs
```diff
@@ -49,10 +49,12 @@ pub fn item_fn_without_rewrite(mut func: ItemFn) -> syn::Result<proc_macro2::Tok
...
+    attrs.push(syn::parse_quote!(#[allow(clippy::used_underscore_binding)]));
```
workingjubilee: I accepted the other PR that added an `allow(clippy::lint)` because I am working on a branch that changes the code in a way that makes it non-applicable. A similar logic applies here: I would prefer to actually solve this. But I don't know if my upcoming branch can solve it.

pgrx-macros/src/rewriter.rs
workingjubilee: **oh**, we were already clearing the attrs, I see.

pgrx-macros/src/rewriter.rs
nyurik: @workingjubilee https://stackoverflow.com/questions/51481551/what-does-automatically-derived-mean -- see also serde [implementation](https://github.com/search?q=repo%3Aserde-rs%2Fserde+automatically_derived&type=code)

pgrx-macros/src/rewriter.rs
nyurik: thx!  I think the easiest is to run `cargo clippy --workspace --bins --tests --lib --benches --examples -- -D warnings -W clippy::pedantic` on the generated code as part of the CI tests.

You may also add this section to the auto-generated `Cargo.toml` and run the above command without the `-W clippy::pedantic`

```toml
[lints.rust]
unused_qualifications = "warn"

[lints.clippy]
pedantic = { level = "warn", priority = -1 }
# list any other clippy lints you want to allow, e.g.
# missing_errors_doc = "allow"
```

cargo-pgrx/src/templates/lib_rs
nyurik: i wasn't sure why you passed the ownership to the setup... but i guess this might be a relatively rare case, so might as well just pass `&[&str]`?

nyurik: P.S. Note that every single user of this code would need to update their signature - so it might not be worth the disruption?

workingjubilee: that is why we added a command to automatically upgrade code to pgrx.

pgrx-macros/src/rewriter.rs
workingjubilee: I will not.

cargo-pgrx/src/templates/lib_rs
workingjubilee: I also have merged quite a few PRs that will probably break a lot of extensions lately, and am going to merge another.

workingjubilee: something about actually spending my time on fixing critical soundness issues instead of fussing over lints.

pgrx-macros/src/rewriter.rs
workingjubilee: I know what it means. I have looked into the compiler. My question is why are you suggesting I add this, specifically. What does it actually change? I don't care if serde does it. Just because the build tool's name is cargo does not mean we should actually cargo cult things.

pgrx-macros/src/rewriter.rs
nyurik: ?

pgrx-macros/src/rewriter.rs
nyurik: I agree about curgo culting.  My issue is that I would like MY code to be clean, and I rely on Clippy to catch issues for me. Clippy has done an awesome job **for me** in this regard.  The problem is that any generated code is not as easy to annotate with attributes. So instead, I am forced to place just the pgrx generated code into separate modules/files and add module-level attributes like `#![allow(clippy::used_underscore_binding)]` so that Clippy does not complain about it -- making code management harder. You may say this is a limitation of Clippy (although I could see why they choose to do it), but at the end of the day, if no other auto-generated code cause Clippy warnings, and PGRX does, it affects your users, and creates un-needed frictions. It is far easier to simply include extra invisible attributes to the generated code than to cause each one of your user to deal with it. Just like all other macro-generating crates do.

workingjubilee: Report these bugs to clippy **first** before coming to me with them.

workingjubilee: And stop asking me to add these annotations without first trying to actually fix the issue that gets linted on.

workingjubilee: If you really can be bothered with neither of those, then there is no use for filing an allowance for every single lint individually. Simply add `allow(clippy::all)`.

workingjubilee: You're talking about friction because of a few lints, while I'm spending my work days trying to fix high-priority soundness issues that also reduce the performance of the real generated artifacts. In other words, they undermine the very point of the pgrx library. Clippy doesn't exactly detect that. I don't really care if the generated code is pretty or not or generates a lot of warnings while these deep, structural issues remain. I think you *should* be alarmed at how bad the code it generates is. I think people *should* experience friction while using it, until the actual problems get actually solved, instead of merely papered over. I don't like painting a pretty picture and hanging it up over a *rotting inside*. It feels like lying.

And meanwhile clippy has over 400 false positive issues open. https://github.com/rust-lang/rust-clippy/issues?q=is%3Aissue+is%3Aopen+label%3AI-false-positive

workingjubilee: The purpose of lints is to highlight issues so they get actually fixed. Every single `allow` is itself, something that should be linted on: pointless code doing nothing, or worse, concealing actual structural problems.

workingjubilee: This attribute, if added, would probably increase the number of warnings generated by rustc.

https://github.com/rust-lang/rust/blob/4bc39f028d14c24b04dd17dc425432c6ec354536/tests/ui/feature-gates/issue-43106-gating-of-builtin-attrs.rs#L266-L273

workingjubilee: @nyurik This is why I asked that question. I want to understand what actually happens. I want to know how rustc will use this. And apparently the answer is mostly in ways that are liable to impair diagnostics:

https://github.com/rust-lang/rust/blob/4bc39f028d14c24b04dd17dc425432c6ec354536/compiler/rustc_trait_selection/src/traits/error_reporting/type_err_ctxt_ext.rs#L2104-L2114

And that is only when it doesn't generate warnings outright. So no, I don't really want to add `#[automatically_derived]` to the code that pgrx emits on top of a function for the supposed purpose of generating _less_ warnings, when it doesn't do that?

pgrx-macros/src/rewriter.rs
workingjubilee: I have pushed a commit to #1731 which will silence this warning for all pgrx functions when that PR lands, because it alters the codegen strategy significantly (and so a PR fixing this would conflict with it).

pgrx-macros/src/rewriter.rs
workingjubilee:
```suggestion
```

pgrx-macros/src/rewriter.rs
workingjubilee:
```suggestion
```

cargo-pgrx/src/templates/lib_rs
```diff
@@ -23,10 +23,12 @@ mod tests {{
 /// It must be visible at the root of your extension crate.
 #[cfg(test)]
 pub mod pg_test {{
+    #[allow(clippy::needless_pass_by_value)]
```
workingjubilee: The fix for this should happen in another PR.
```suggestion
```

pgrx-macros/src/rewriter.rs
```diff
@@ -10,6 +10,7 @@
 extern crate proc_macro;
 
 use quote::{format_ident, quote, quote_spanned};
+use std::mem::take;
```
workingjubilee:
```suggestion
use std::mem;
```

pgrx-macros/src/rewriter.rs
workingjubilee:
```suggestion
    let mut attrs = mem::take(&mut func.attrs);
```

workingjubilee (dismissed):

PR #1748, opened 2024-07-04
C enums as module consts + deprecation

### LINKED ISSUE #1370: Use `.default_enum_style(EnumVariation::ModuleConsts)`

workingjubilee:
I tried to do without the deprecation warning code but I found it simply so annoying that rustc couldn't find the right names (its Levenshtein-derived algorithm doesn't reason about pathnames quite that way) that I figured it would be easiest to simply add the relevant codegen. That made it possible to distinguish between disappearing types and now-outdated names.

Closes #1370.

### DIFF
--- a/pgrx-pg-sys/build.rs
+++ b/pgrx-pg-sys/build.rs
@@ -106,6 +106,7 @@ impl bindgen::callbacks::ParseCallbacks for BindingOverride {
     // FIXME: implement a... C compiler?
     fn func_macro(&self, _name: &str, _value: &[&[u8]]) {}
 
+    /// Intentionally doesn't do anything, just updates internal state.
     fn enum_variant_behavior(
         &self,
         enum_name: Option<&str>,
@@ -113,6 +114,7 @@ impl bindgen::callbacks::ParseCallbacks for BindingOverride {
         variant_value: bindgen::callbacks::EnumVariantValue,
     ) -> Option<bindgen::callbacks::EnumVariantCustomBehavior> {
         enum_name.inspect(|name| match name.strip_prefix("enum").unwrap_or(name).trim() {
+            // specifically overridden enum
             "NodeTag" => return,
             name if name.contains("unnamed at") => return,
             // to prevent problems with BuiltinOid
@@ -781,8 +783,9 @@ fn run_bindgen(
         .generate()
         .wrap_err_with(|| format!("Unable to generate bindings for pg{major_version}"))?;
     let mut binding_str = bindings.to_string();
-    drop(bindings);
-    let enum_names = Rc::into_inner(enum_names).unwrap().into_inner();
+    drop(bindings); // So the Rc::into_inner can unwrap
+    // FIXME: do this for the Node graph instead of reparsing?
+    let enum_names: EnumMap = Rc::into_inner(enum_names).unwrap().into_inner();
     binding_str.extend(enum_names.into_iter().flat_map(|(name, variants)| {
         const MIN_I32: i64 = i32::MIN as _;
         const MAX_I32: i64 = i32::MAX as _;
@@ -799,7 +802,7 @@ fn run_bindgen(
                 r#"
 #[deprecated(since = "0.12.0", note = "you want pg_sys::{module}::{variant}")]
 pub const {module}_{variant}: {ty} = {value};"#,
-                module = &*name,
+                module = &*name, // imprecise closure capture
             )
         })
     }));
--- a/pgrx-pg-sys/build.rs
+++ b/pgrx-pg-sys/build.rs
@@ -784,6 +784,7 @@ fn run_bindgen(
         .wrap_err_with(|| format!("Unable to generate bindings for pg{major_version}"))?;
     let mut binding_str = bindings.to_string();
     drop(bindings); // So the Rc::into_inner can unwrap
+
     // FIXME: do this for the Node graph instead of reparsing?
     let enum_names: EnumMap = Rc::into_inner(enum_names).unwrap().into_inner();
     binding_str.extend(enum_names.into_iter().flat_map(|(name, variants)| {
--- a/pgrx-pg-sys/build.rs
+++ b/pgrx-pg-sys/build.rs
@@ -30,44 +30,44 @@ mod build {
 }
 
 #[derive(Debug)]
-struct PgrxOverrides(HashSet<String>);
+struct BindingOverride {
+    ignore_macros: HashSet<&'static str>,
+}
 
-impl PgrxOverrides {
-    fn default() -> Self {
+impl BindingOverride {
+    fn new() -> Self {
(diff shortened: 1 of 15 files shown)


eeeebbbbrrrr:
This is cool. Has this ModuleConsts form always been there or is it relatively new to bindgen?  Neat

workingjubilee:
It's been there since before pgrx's initial commit.

pgrx-pg-sys/build.rs
```diff
@@ -90,6 +97,43 @@ impl bindgen::callbacks::ParseCallbacks for PgrxOverrides {
...
+
+    // FIXME: alter types on some int macros to the actually-used types so we can stop as-casting them
+    fn int_macro(&self, _name: &str, _value: i64) -> Option<bindgen::callbacks::IntKind> {
+        None
+    }
+
+    // FIXME: implement a... C compiler?
+    fn func_macro(&self, _name: &str, _value: &[&[u8]]) {}
+
+    fn enum_variant_behavior(
```
workingjubilee, line 110:
```suggestion
    /// Intentionally doesn't do anything, just updates internal state.
    fn enum_variant_behavior(
```

pgrx-pg-sys/build.rs
```diff
@@ -733,8 +780,31 @@ fn run_bindgen(
         .default_non_copy_union_style(NonCopyUnionStyle::ManuallyDrop)
         .generate()
         .wrap_err_with(|| format!("Unable to generate bindings for pg{major_version}"))?;
+    let mut binding_str = bindings.to_string();
+    drop(bindings);
```
workingjubilee:
```suggestion
    drop(bindings); // So the Rc::into_inner can unwrap
```

pgrx-pg-sys/build.rs
```diff
@@ -90,6 +97,43 @@ impl bindgen::callbacks::ParseCallbacks for PgrxOverrides {
...
+    fn func_macro(&self, _name: &str, _value: &[&[u8]]) {}
+
+    fn enum_variant_behavior(
+        &self,
+        enum_name: Option<&str>,
+        variant_name: &str,
+        variant_value: bindgen::callbacks::EnumVariantValue,
+    ) -> Option<bindgen::callbacks::EnumVariantCustomBehavior> {
+        enum_name.inspect(|name| match name.strip_prefix("enum").unwrap_or(name).trim() {
+            "NodeTag" => return,
```
workingjubilee, line 118:
```suggestion
            // specifically overridden enum
            "NodeTag" => return,
```

pgrx-pg-sys/build.rs
```diff
@@ -733,8 +780,31 @@ fn run_bindgen(
         .default_non_copy_union_style(NonCopyUnionStyle::ManuallyDrop)
         .generate()
         .wrap_err_with(|| format!("Unable to generate bindings for pg{major_version}"))?;
+    let mut binding_str = bindings.to_string();
+    drop(bindings);
+    let enum_names = Rc::into_inner(enum_names).unwrap().into_inner();
```
workingjubilee:
```suggestion
    // FIXME: do this for the Node graph instead of reparsing?
    let enum_names: EnumMap = Rc::into_inner(enum_names).unwrap().into_inner();
```

pgrx-pg-sys/build.rs
```diff
@@ -733,8 +780,31 @@ fn run_bindgen(
...
+                EnumVariantValue::Signed(v @ MIN_I32..=MAX_I32) => ("i32", v.to_string()),
+                EnumVariantValue::Signed(v) => ("i64", v.to_string()),
+                EnumVariantValue::Unsigned(v @ 0..=MAX_U32) => ("u32", v.to_string()),
+                EnumVariantValue::Unsigned(v) => ("u64", v.to_string()),
+            };
+            format!(
+                r#"
+#[deprecated(since = "0.12.0", note = "you want pg_sys::{module}::{variant}")]
+pub const {module}_{variant}: {ty} = {value};"#,
+                module = &*name,
```
workingjubilee:
```suggestion
                module = &*name, // imprecise closure capture
```

pgrx/src/callbacks.rs
```diff
@@ -57,18 +57,17 @@ pub enum PgXactCallbackEvent {
...
-            pg_sys::XactEvent_XACT_EVENT_PRE_COMMIT => PgXactCallbackEvent::PreCommit,
-            pg_sys::XactEvent_XACT_EVENT_PRE_PREPARE => PgXactCallbackEvent::PrePrepare,
+            XACT_EVENT_ABORT => PgXactCallbackEvent::Abort,
+            XACT_EVENT_COMMIT => PgXactCallbackEvent::Commit,
+            XACT_EVENT_PARALLEL_ABORT => PgXactCallbackEvent::ParallelAbort,
+            XACT_EVENT_PARALLEL_COMMIT => PgXactCallbackEvent::ParallelCommit,
+            XACT_EVENT_PARALLEL_PRE_COMMIT => PgXactCallbackEvent::ParallelPreCommit,
+            XACT_EVENT_PREPARE => PgXactCallbackEvent::Prepare,
+            XACT_EVENT_PRE_COMMIT => PgXactCallbackEvent::PreCommit,
+            XACT_EVENT_PRE_PREPARE => PgXactCallbackEvent::PrePrepare,
```
workingjubilee, line 70: Making these matches a bit less awful to read was basically my primary goal here.

pgrx-pg-sys/build.rs
```diff
@@ -733,8 +782,32 @@ fn run_bindgen(
         .default_non_copy_union_style(NonCopyUnionStyle::ManuallyDrop)
         .generate()
         .wrap_err_with(|| format!("Unable to generate bindings for pg{major_version}"))?;
+    let mut binding_str = bindings.to_string();
+    drop(bindings); // So the Rc::into_inner can unwrap
+    // FIXME: do this for the Node graph instead of reparsing?
```
workingjubilee, line 788:
```suggestion

    // FIXME: do this for the Node graph instead of reparsing?
```

PR #1750, opened 2024-07-05
fix(cargo-pgrx): fix artifact naming on newer rust versions

t3hmrman:
On newer versions of Rust (1.79+), artifacts that are produced by `cargo` are reliably snake_cased rather than kebab-cased, so older code checking for artifact names didn't work.

Came across this when trying to run *in-repo* tests with an *in-repo* build of `cargo-pgrx` on 1.79 which is now stable.

### DIFF
--- a/cargo-pgrx/src/command/install.rs
+++ b/cargo-pgrx/src/command/install.rs
@@ -415,6 +415,23 @@ fn copy_sql_files(
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


t3hmrman:
Sorry -- my forks version of `develop` was very old, so ran into this, should have pulled in the fix (from 3 months ago)!

Issue #1749, opened 2024-07-04
register_hook(&mut HOOKS) : mutable reference to mutable static


daamien:
Not a actually bug, but a strange warning....

I registered my hooks with the following line 

```rust
        static mut HOOKS: AnonHook = AnonHook { };
        pgrx::hooks::register_hook(&mut HOOKS)
```

which I copy-pasted blindly from  the hook_tests here:

https://github.com/pgcentralfoundation/pgrx/blob/2c1d3887d16744e5230efe506d98a51836cf8228/pgrx-tests/src/tests/hooks_tests.rs#L126C1-L127C46

Cargo 1.77 issues the following warning

```
473 |     pgrx::hooks::register_hook(&mut HOOKS);
    |                                ^^^^^^^^^^ mutable reference to mutable static
    |
    = note: for more information, see issue #114447 <https://github.com/rust-lang/rust/issues/114447>
    = note: this will be a hard error in the 2024 edition
    = note: this mutable reference has lifetime `'static`, but if the static gets accessed (read or written) by any other means, or any other reference is created, then any further use of this mutable reference is Undefined Behavior
    = note: `#[warn(static_mut_refs)]` on by default
help: use `addr_of_mut!` instead to create a raw pointer
    |
473 |     pgrx::hooks::register_hook(addr_of_mut!(HOOKS));
```

Trying to rewrite that register_hook line to remove the warning, I ended up with the syntax below that seems to please cargo but looks fishy

```rust
        let h = &mut *std::ptr::addr_of_mut!(HOOKS) as &mut dyn PgHooks;                                                                         
        pgrx::hooks::register_hook(h);
```

I tried to understand the issue 114447, but it is way out of my league.... 

It seems obvious that we won't create another reference to the HOOKS variable, but is that enough to simply ignore this warning ?

workingjubilee:
We are going to move away from using `static mut` in pgrx. The use here is actually quite inappropriate and almost certainly UB, because we later do materialize references to it inside the various hooks:

https://github.com/pgcentralfoundation/pgrx/blob/2c1d3887d16744e5230efe506d98a51836cf8228/pgrx/src/hooks.rs#L601-L608

So I don't think

>It seems obvious that we won't create another reference to the HOOKS variable

is quite correct.

Fixing this requires a redesign of the API.

workingjubilee:
> Trying to rewrite that register_hook line to remove the warning, I ended up with the syntax below that seems to please cargo but looks fishy
[...]

PR #1752, opened 2024-07-05
Deprecate `pgrx::hooks` without replacement

workingjubilee:
Even if we could fix this with internal refactors, I think it's a sufficiently bad pattern, and sufficiently hard to reason about the soundness of it, that we should just tell people to drop this API.

### DIFF
--- a/pgrx/src/hooks.rs
+++ b/pgrx/src/hooks.rs
@@ -9,6 +9,10 @@
 //LICENSE Use of this source code is governed by the MIT license that can be found in the LICENSE file.
 //! A trait and registration system for hooking Postgres internal operations such as its planner and executor
 #![allow(clippy::unit_arg)]
+#![deprecated(
+    since = "0.12.1",
+    note = "currently always UB, use FFI + pointers to `static UnsafeCell`"
+)]
 use crate as pgrx; // for #[pg_guard] support from within ourself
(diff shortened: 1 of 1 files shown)


daamien:
As far as I understand, this is going to break any PGRX extension that defines PostgresQL hooks ? Seems like a radical choice, given that there may be some of these hook-based extensions currently in production...  What would be the best approach for them ? Staying in version 0.11 indefinetely with no support for Postgres 17+  or maybe something like `#[allow(pgrx::hooks)]` ?

workingjubilee:
@daamien you can just `allow(deprecated)`. warnings don't prevent anything from compiling.

workingjubilee:
My only hesitation for committing this at the moment is currently I'm also using deprecations to communicate about other things that can be effectively replaced, so I am reluctant to fill the deprecation notices with even more.

workingjubilee:
I've decided I'm only going to merge this deprecation notice after I cut 0.12.0 proper, as otherwise a bunch of useful deprecation notices will be shadowed by one kinda useless one (I feel the request is a bit unfair, even if it's true it's more sound code).

yrashk:
To keep some decision-making trail, is it fair to note that #1749 would be a good reference for the problems discussed here?

PR #1754, opened 2024-07-05
Prepare pgrx 0.12.0-beta.0

workingjubilee:


workingjubilee:
Will It Blend run: https://github.com/pgcentralfoundation/pgrx/actions/runs/9813695603

It didn't take up the fix in #1755 "because GitHub".

PR #1756, opened 2024-07-06
pgrx-pg-sys missing build directory in Cargo.toml

danbluhmhansen:
0.12.0-beta.0 errors in pgrx-pg-sys/build.rs due to the toml's `include` directive not including the build script's module.


workingjubilee:
Weird and annoying.

workingjubilee:
Oh right, we have to no-verify that crate...

PR #1757, opened 2024-07-06
Missing ArgAbi and BoxRet trait implementations for datums

danbluhmhansen:
After updating to version `0.12.0-beta.0` I can no longer use `JsonB` as a `pg_extern` function argument, and I can no longer use `JsonB`, `AnyArray`, or `AnyElement` as state type in `pg_aggregate`. Unfortunately, pgrx lacked any tests for these types, so they were not checked for compilation when adding ArgAbi and RetAbi.

### DIFF
--- a/pgrx-tests/src/tests/aggregate_tests.rs
+++ b/pgrx-tests/src/tests/aggregate_tests.rs
@@ -143,6 +143,74 @@ impl Aggregate for DemoCustomState {
     }
 }
 
+struct FirstJson;
+
+#[pg_aggregate]
+impl Aggregate for FirstJson {
+    type State = pgrx::Json;
+    type Args = pgrx::name!(value, pgrx::Json);
+
+    #[pgrx(parallel_safe, immutable, strict)]
+    fn state(
+        current: Self::State,
+        _arg: Self::Args,
+        _fcinfo: pg_sys::FunctionCallInfo,
+    ) -> Self::State {
+        current
+    }
+}
+
+struct FirstJsonB;
+
+#[pg_aggregate]
+impl Aggregate for FirstJsonB {
+    type State = pgrx::JsonB;
+    type Args = pgrx::name!(value, pgrx::JsonB);
+
+    #[pgrx(parallel_safe, immutable, strict)]
+    fn state(
+        current: Self::State,
+        _arg: Self::Args,
+        _fcinfo: pg_sys::FunctionCallInfo,
+    ) -> Self::State {
+        current
+    }
+}
+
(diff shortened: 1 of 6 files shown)


eeeebbbbrrrr:
Interesting. Surprised we don’t have tests to catch these. 

Thanks @danbluhmhansen. I’ll approve CI and we’ll see about merging next week, I imagine.

workingjubilee (changes requested):
This should land with tests that prevent this omission in the future. Just simple "accepts the type, returns the type" somewhere in pgrx-tests is all.

workingjubilee:
I have a suspicion that these BoxRet impls for AnyArray and AnyElement will prove... optimistic, but if we can at least make it through a function call, then at least it kinda works.

workingjubilee:
Preferably, the Any* types should at least have the mercy to have their impl check that you created an AnyType with the same OID returned by `get_fn_expr_rettype`. Perhaps only beneath `cfg(debug_assertions)`? That can wait for a separate PR if you are not interested in writing it, however.

danbluhmhansen:
> This should land with tests that prevent this omission in the future. Just simple "accepts the type, returns the type" somewhere in pgrx-tests is all.

Alright I have added tests that should catch possible future regressions.

- json argument type tests to [json_tests.rs](https://github.com/pgcentralfoundation/pgrx/pull/1757/files#diff-e492475e4931af3735736bdd8efeaa2b5eade38d90291679621c5e2ffb63d619)
- anyelement argument type tests to new [anyelement_tests.rs](https://github.com/pgcentralfoundation/pgrx/pull/1757/files#diff-1f86317ee25d15e38f376db54d456885ce5adda4666d88912ce1c416dfeca994)
- anynumeric argument type tests to new [anynumeric_tests.rs](https://github.com/pgcentralfoundation/pgrx/pull/1757/files#diff-185d5b61c239f04a13f4fc4ac33f310fa407c14a45437c8012c8904b6d7eb0ec)
- aggregate state type tests for types `json`, `jsonb`, `anyarray`, `anyelement` to [aggregate_tests.rs](https://github.com/pgcentralfoundation/pgrx/pull/1757/files#diff-f26401de6f40e8eb824c99b1aa7bb3acdc50edaaa14ec0f6fb350f449fcf1945)
  - I couldn't figure out how to write an actual pg_test for the anyarray aggregate state so I left that out, besides any regression should be caught by having the pg_aggregate implementation, regardless of the tests.

workingjubilee (approved):
Thank you.

PR #1772, opened 2024-07-10
Don't hide docs for the callconv module

syvb:
There are types in `callconv` that are needed by user code (e.g. when manually implementing a Postgres type used in function signatures). It was really confusing when I saw errors referencing types that I couldn't find in the documentation.

### DIFF
--- a/pgrx/src/callconv.rs
+++ b/pgrx/src/callconv.rs
@@ -7,7 +7,6 @@
 //LICENSE All rights reserved.
 //LICENSE
 //LICENSE Use of this source code is governed by the MIT license that can be found in the LICENSE file.
-#![doc(hidden)]
 #![deny(unsafe_op_in_unsafe_fn)]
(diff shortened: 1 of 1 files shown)


workingjubilee:
Ah, you want to review the docs for... ArgAbi, RetAbi, FcInfo, Arg, and Args, yeah?

workingjubilee:
I guess BoxRet too.

In fact I guess I somewhat would prefer if people only needed BoxRet instead of RetAbi.

syvb:
Yes, I'm trying to update an extension to the latest version and I got errors referencing those types. BoxRet and ArgAbi are the only things I've needed to use so far.

workingjubilee:
I am a bit persnickety about how I want to go about things, so I am gonna resolve this in https://github.com/pgcentralfoundation/pgrx/pull/1776

Please feel free to review it if you like.

PR #1776, opened 2024-07-13
Refactor BoxRet to be easier to impl before showing it

workingjubilee:
People need to implement some of the code in this module, but I want to make sure it is easier to actually do so. To that end, refactor BoxRet to have a more user-friendly interface, ish, and then expose it.

While doing so, adjust some safety requirements here and there.

### DIFF
--- a/pgrx/src/callconv.rs
+++ b/pgrx/src/callconv.rs
@@ -348,9 +348,7 @@ pub unsafe trait BoxRet: Sized {
 
     /// # Safety
     /// You have to actually return the resulting Datum to the function.
-    unsafe fn box_into<'fcx>(self, fcinfo: &mut FcInfo<'fcx>) -> Datum<'fcx> {
-        unsafe { fcinfo.return_raw_datum(self.box_in_fcinfo(fcinfo.0)) }
-    }
+    unsafe fn box_into<'fcx>(self, fcinfo: &mut FcInfo<'fcx>) -> Datum<'fcx>;
 }
 
 unsafe impl<T> RetAbi for T
--- a/pgrx/src/callconv.rs
+++ b/pgrx/src/callconv.rs
@@ -14,9 +14,9 @@ use crate::datum::Datum;
 use crate::heap_tuple::PgHeapTuple;
 use crate::nullable::Nullable;
 use crate::{
-    pg_return_null, pg_sys, AnyArray, AnyElement, AnyNumeric, Date, FromDatum, Inet, Internal,
-    Interval, IntoDatum, Json, JsonB, PgBox, PgMemoryContexts, PgVarlena, Time, TimeWithTimeZone,
-    Timestamp, TimestampWithTimeZone, UnboxDatum, Uuid,
+    pg_sys, AnyArray, AnyElement, AnyNumeric, Date, FromDatum, Inet, Internal, Interval, IntoDatum,
+    Json, JsonB, PgBox, PgMemoryContexts, PgVarlena, Time, TimeWithTimeZone, Timestamp,
+    TimestampWithTimeZone, UnboxDatum, Uuid,
 };
 use core::marker::PhantomData;
 use core::{iter, mem, ptr, slice};
@@ -395,12 +395,10 @@ unsafe impl<T> BoxRet for Option<T>
 where
     T: BoxRet,
 {
-    unsafe fn box_in_fcinfo(self, fcinfo: pg_sys::FunctionCallInfo) -> pg_sys::Datum {
-        unsafe {
-            match self {
-                None => pg_return_null(fcinfo),
-                Some(value) => value.box_in_fcinfo(fcinfo),
-            }
+    unsafe fn box_into<'fcx>(self, fcinfo: &mut FcInfo<'fcx>) -> Datum<'fcx> {
+        match self {
+            Some(datum) => unsafe { datum.box_into(fcinfo) },
+            None => fcinfo.return_null(),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


pgrx/src/callconv.rs
```diff
@@ -605,30 +651,57 @@ impl<'fcx> FcInfo<'fcx> {
...
+    /// fn foo<'a>(mut fcinfo: FcInfo<'a>) -> Datum<'a> {
+    ///     unsafe { fcinfo.return_null() }
     /// }
     /// ```
     #[inline]
-    pub unsafe fn set_return_null(&mut self) -> crate::datum::Datum<'fcx> {
-        let fcinfo = unsafe { self.0.as_mut() }.unwrap();
-        fcinfo.isnull = true;
-        crate::datum::Datum::null()
+    pub fn return_null(&mut self) -> Datum<'fcx> {
```
workingjubilee, line 676: It is safe to return null because that means that Postgres must not examine the resulting Datum. If Postgres ignores the SQL null to examine the null Datum and dereferences it, that is Postgres being unsound. If Postgres interprets the bytes as a type that has no valid 0-initialization... how would it do that, for one...? ...but for two, that is also Postgres being unsound.

pgrx/src/callconv.rs
```diff
@@ -605,30 +651,57 @@ impl<'fcx> FcInfo<'fcx> {
...
+    ///
+    /// # Safety
+    /// - If the function call expects a by-reference return, and the `pg_sys::Datum`
+    ///   does not point to a valid object of the correct type, this is *undefined behavior*.
+    /// - If the function call expects a by-reference return, and the `pg_sys::Datum` points
+    ///   to an object that will not last for the caller's lifetime, this is *undefined behavior*.
+    /// - If the function call expects a by-value return, and the `pg_sys::Datum` is not a valid
+    ///   instance of that type when interpreted as such, this is **undefined behavior**.
+    /// - If the return must be null, then calling this function is **undefined behavior**.
+    #[inline]
```
workingjubilee:
```suggestion
    /// - If the return must be the SQL null, then calling this function is **undefined behavior**.
    ///
    /// Functionally, this can be thought of as a convenience for a [`mem::transmute`]
    /// from `pg_sys::Datum` to `pgrx::datum::Datum<'fcx>`. This gives it similar constraints.
    #[inline]
```

pgrx/src/callconv.rs
```diff
@@ -605,30 +651,57 @@ impl<'fcx> FcInfo<'fcx> {
...
 
-    /// Modifies the contained `fcinfo` struct to flag its return value as null.
+    /// Accessor for the "is null" flag
     ///
-    /// Returns a null-pointer Datum for use with the calling function's return value.
+    /// # Safety
+    /// If this flag is set to "false", then the resulting return must be a valid [`Datum`] for
+    /// the function call's result type.
+    pub unsafe fn set_return_is_null(&mut self) -> &mut bool {
+        unsafe { &mut (*self.0).isnull }
```
workingjubilee, line 653: generally using `as_mut` is less-than-ideal because it takes the entire "place"

pgrx/src/callconv.rs
```diff
@@ -334,9 +334,23 @@ pub unsafe trait RetAbi: Sized {
...
+/// In order to handle all of the nuances of the calling convention for returning result sets from
+/// Postgres functions, pgrx uses a very complicated trait, RetAbi. In practice, however, most
+/// types do not need to think about its many sharp-edged cases. Instead, they should implement
+/// this simplified trait, BoxRet. A blanket impl of RetAbi for BoxRet takes care of the rest.
 pub unsafe trait BoxRet: Sized {
-    unsafe fn box_in_fcinfo(self, fcinfo: pg_sys::FunctionCallInfo) -> pg_sys::Datum;
+    #[doc(hidden)]
+    unsafe fn box_in_fcinfo(self, fcinfo: pg_sys::FunctionCallInfo) -> pg_sys::Datum {
+        unsafe { Self::box_into(self, &mut FcInfo::from_ptr(fcinfo)).sans_lifetime() }
+    }
```
workingjubilee: Going to actually completely erase this function before this gets committed on Monday, rather than having this recursive impl, I just was looking for a stopping point today.

pgrx/src/callconv.rs
```diff
@@ -334,9 +334,23 @@ pub unsafe trait RetAbi: Sized {
...
+
+    /// # Safety
+    /// You have to actually return the resulting Datum to the function.
+    unsafe fn box_into<'fcx>(self, fcinfo: &mut FcInfo<'fcx>) -> Datum<'fcx> {
+        unsafe { fcinfo.return_raw_datum(self.box_in_fcinfo(fcinfo.0)) }
+    }
```
workingjubilee: This safety constraint isn't a big concern because generally "user code" should not be calling this function directly, but rather implementing it so that it can _be_ called.

pgrx/src/callconv.rs
```diff
@@ -605,30 +651,57 @@ impl<'fcx> FcInfo<'fcx> {
...
+    ///     unsafe { fcinfo.return_null() }
     /// }
     /// ```
     #[inline]
-    pub unsafe fn set_return_null(&mut self) -> crate::datum::Datum<'fcx> {
-        let fcinfo = unsafe { self.0.as_mut() }.unwrap();
-        fcinfo.isnull = true;
-        crate::datum::Datum::null()
+    pub fn return_null(&mut self) -> Datum<'fcx> {
+        unsafe { *self.set_return_is_null() = true };
```
workingjubilee, line 678:
```suggestion
        // SAFETY: returning null is always safe
        unsafe { *self.set_return_is_null() = true };
```

pgrx/src/callconv.rs
```diff
@@ -334,9 +334,23 @@ pub unsafe trait RetAbi: Sized {
...
 pub unsafe trait BoxRet: Sized {
-    unsafe fn box_in_fcinfo(self, fcinfo: pg_sys::FunctionCallInfo) -> pg_sys::Datum;
+    #[doc(hidden)]
+    unsafe fn box_in_fcinfo(self, fcinfo: pg_sys::FunctionCallInfo) -> pg_sys::Datum {
+        unsafe { Self::box_into(self, &mut FcInfo::from_ptr(fcinfo)).sans_lifetime() }
+    }
+
+    /// # Safety
+    /// You have to actually return the resulting Datum to the function.
+    unsafe fn box_into<'fcx>(self, fcinfo: &mut FcInfo<'fcx>) -> Datum<'fcx> {
```
workingjubilee: maybe I should make the datum returned into a field so that this doesn't have to be unsafe...

workingjubilee: it would have to be a struct that only allows mutating all "fields" callees can mutate, so that they can update all of them in one function call...

PR #1780, opened 2024-07-16
docs: remove an extra backtick

SteveLauC:
Removed an extra backtick "`".

 Noticed it while reading the doc, so let's remove it.


workingjubilee:
Thanks!

Issue #1738, opened 2024-06-20
Support for `ClientAuthentication_hook`


draescherl:
Hi, thanks for the amazing work.

I was wondering if the [`ClientAuthentication_hook`](https://pgpedia.info/c/ClientAuthentication_hook.html) was supported? 

I couldn't find a reference to it in the docs. Looking through issues and pull requests, all I could find are the following issue https://github.com/pgcentralfoundation/pgrx/issues/1073 and related PR https://github.com/pgcentralfoundation/pgrx/pull/1077, which seem to indicate they aren't available in PGRX yet.

Has this changed since then? If not, is it because supporting it is non-trivial? 

Thanks in advance!

eeeebbbbrrrr:
We don’t have safe wrappers around it but it’s super easy to add the hook yourself with a bit of unsafe {}

draescherl:
Hi, thanks for the quick reply. I'd be very curious to try adding the hook myself, however my experience with Rust is quite limited and I'm not sure how to go about it. Would you have a commit I could use as an example?

daamien:
FTR, this hook is now available in PGRX 0.17

PR #1785, opened 2024-07-18
WIP: add support for ClientAuthentication_hook

draescherl:
Hi, after https://github.com/pgcentralfoundation/pgrx/issues/1738 a colleague and I tried our hands at implementing the `ClientAuthentication_hook`.

We took inspiration from https://github.com/pgcentralfoundation/pgrx/commit/0703e630e9c26202fb2b267f02619430860fa954 but the tests don't pass and we can't figure out why. Do the tests not go through an auth phase? Any advice?

Thanks!

### DIFF
--- a/pgrx-pg-sys/include/pg13.h
+++ b/pgrx-pg-sys/include/pg13.h
@@ -15,6 +15,7 @@
 #include "miscadmin.h"
 #include "pgstat.h"
 
+#include "libpq/auth.h"
 #include "access/amapi.h"
 #include "access/genam.h"
 #include "access/generic_xlog.h"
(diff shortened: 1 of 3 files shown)


eeeebbbbrrrr:
Doesn't that hook require the extension be loaded by the postmaster, via `shared_preload_libraries`?  Or maybe at least via `session_preload_libraries`?

That probably explains why the test framework isn't behaving the way you expect -- the test extension .so isn't actually loaded yet.

Not sure I have a suggestion yet, but that'd be my thought...

workingjubilee:
I think everything that requires `shared_preload_libraries` or `session_preload_libraries` should probably be behind a feature flag in pgrx itself, to make it obvious when people need to configure that.

daamien:
FTR, this hook is now available in PGRX 0.17

eeeebbbbrrrr:
This has been a Draft for quite awhile now.  I'm closing this simply to keep our PR list tidy.  If you decide to come back to this against the current `develop` branch, please open a new PR.

PR #1787, opened 2024-07-22
Include pg_seclabel catalog

daamien:
In complement to #1189, add the `catalog/pg_seclabel.h` header in order to access the `FormData_pg_seclabel` struct and provide a mecanism to list the security labels.


workingjubilee (approved):
Sure, thanks!

PR #1790, opened 2024-07-23
Add bindings for 'jit/jit.h' and 'utils/resowner_private.h' header files

jeltz:
These headers are required when writing a JIT provider for PostgreSQL. Most things in jit.h are required and from resowner_private.h the ResourceOwner*JIT() functions are necessary. Additionally jit.h contains the struct used when measuring how much time is spent doing just-in-time compilation which is useful when working with EXPLAIN ANALYZE outputs.

The resowner_private.h header will be removed in PostgreSQL 17.


pgrx-pg-sys/include/pg12.h
```diff
@@ -167,6 +168,7 @@
 #include "utils/regproc.h"
 #include "utils/rel.h"
 #include "utils/relcache.h"
+#include "utils/resowner_private.h"
```
workingjubilee, line 172: why is this one private?

workingjubilee:
What will happen to the items that are in resowner_private.h then in Postgres 17?

jeltz:
> What will happen to the items that are in resowner_private.h then in Postgres 17?

The whole resowner API was refactored so in PG 17 and all of it is in utils/resowner.h.

workingjubilee:
Are there any items from `resowner_private.h` that don't survive the refactor?

pgrx-pg-sys/include/pg12.h
jeltz: I have no idea. It claims to be private but it contains functions which are necessary for anyone implementing a JIT provider. It might be an error that these functions ended up in this header file.

jeltz: And the contents of it does not look very private.

jeltz:
> Are there any items from `resowner_private.h` that don't survive the refactor?

All of them. They are all replaced by the new resowner API.

https://github.com/postgres/postgres/commit/b8bff07daa85c837a2747b4d35cd5a27e73fb7b2

pgrx-pg-sys/include/pg12.h
jeltz: This commit split the header file: https://github.com/postgres/postgres/commit/45326c5a11530cff5db99c6d4b393439cc901f26

workingjubilee:
huh.

pgrx-pg-sys/include/pg12.h
workingjubilee: Let's include resowner.h then also, so we can just pull a line on the pg17 upgrade.

jeltz: I have done so now.

workingjubilee (approved):
Thanks!

PR #1791, opened 2024-07-27
Remove ArgAbi for `Numeric<P, S>`

workingjubilee:
This implementation is actually false as Postgres does not check typmods before calling a function, thus we can't use them to define arguments.

### DIFF
--- a/pgrx-examples/numeric/src/lib.rs
+++ b/pgrx-examples/numeric/src/lib.rs
@@ -13,7 +13,7 @@ use pgrx::prelude::*;
 pg_module_magic!();
 
 #[pg_extern]
-fn add_numeric(a: Numeric<1000, 33>, b: Numeric<1000, 33>) -> Numeric<1000, 33> {
+fn add_numeric(a: AnyNumeric, b: AnyNumeric) -> Numeric<1000, 33> {
     (a + b).rescale().unwrap()
 }
 
--- a/pgrx-tests/src/tests/roundtrip_tests.rs
+++ b/pgrx-tests/src/tests/roundtrip_tests.rs
@@ -111,7 +111,6 @@ mod tests {
     roundtrip!(rt_refstr, test_rt_refstr, &'a str, "foo");
     roundtrip!(rt_bool, test_rt_bool, bool, true);
     roundtrip!(rt_f32, test_rt_f32, f32, f32::MAX);
-    roundtrip!(rt_numeric, test_rt_numeric, Numeric<100,0>, Numeric::from_str("31241234123412341234").unwrap());
     roundtrip!(
         rt_anynumeric,
         test_rt_anynumeric,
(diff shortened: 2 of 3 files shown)


eeeebbbbrrrr:
Um

eeeebbbbrrrr:
We still need to be able to generate a schema definition for how the user wants it.

workingjubilee:
> We still need to be able to generate a schema definition for how the user wants it.

They can use AnyNumeric.

workingjubilee:
Using it as an argument will not be checked by Postgres. We can check and return Numerics of a defined precision and scale, but we should not allow values wandering around in extensions claiming to be a type and then not actually fulfilling its definition.

eeeebbbbrrrr:
But that’ll write the arg type as NUMERIC, not NUMERIC(X, Y). 

Postgres might not check the typemod but it does get you a value formed with the proper typemod from a SQL literal. 

I think this PR is ill-advised.

workingjubilee:
to hell with you, clippy.

workingjubilee:
Yet there is no requirement that this is from a literal.

And whenever I suggest we support checked-arguments idioms, I seem to remember you suggest they are bad ideas.

eeeebbbbrrrr:
Part of the importance of the schema generator is to expose metadata about a function that client libraries can introspect and act upon on whatever ways they want. 

If what you’re saying is that we ought to check the typemod here in our wrapper, or something, then maybe we can figure that out. 

But I’m not liking the idea of removing the ability for a function to accurately express what it expects its arguments to be.

workingjubilee:
and when writing the Rust code implementing an extension, allowing someone to accept such an argument will naturally lead to the reasonable assumption that the argument meets its constraints.

eeeebbbbrrrr:
When called from sql it will.  That’s what #[pg_extern] is all about. 

Do we just need to require (or rewrite?) all pg_extern functions to be unsafe?  Haha. That’s not the worst idea

workingjubilee:
all `extern "C" fn` implicitly are.

will Postgres check the numeric meets its constraints if it is a nonliteral?

eeeebbbbrrrr:
Like I’d the arg datum comes from a table or something?

I think it’ll go through a cast of sorts to coerce it into the proper scale?  Maybe it won’t. Hmm. I’d have to check to say for sure. Not at my computer. 

Re: extern C functions being unsafe. 

Yes, but pg_extern-annotated functions, as written by the user, aren’t extern C.

workingjubilee:
The wrapper is. If the function is called from Rust its arguments will be typechecked, at least, and that does not depend on whether it is a literal or not.

workingjubilee:
> Like I’d the arg datum comes from a table or something?

Yes.

> I think it’ll go through a cast of sorts to coerce it into the proper scale? Maybe it won’t. Hmm. I’d have to check to say for sure. Not at my computer.

Then we shall resume later.

In any case, if it uses the many APIs we expose for calling other Postgres functions, those should also all be unsafe.

The only one we can't sensibly guard is the basic function of SPI.

eeeebbbbrrrr:
Right.  pg_extern functions are meant to be called from sql.  In my mind, they need to be honest to that, first. 

And rust’s type system will prevent a user internally calling the function with Numeric<10, 20> when it’s declared Numeric<1000, 1>.

eeeebbbbrrrr:
Numeric is a difficult type to work with, that’s for sure. 

Related, I’m surprised nobody has brought up varchar(100) yet. That uses a typemod too.

workingjubilee:
I think the easiest way to simultaneously make typmods and "things that are really just `TEXT`" make sense in our system is have a way to specify "the actual argument is this, but the SQL generated for it should be from this other thing", and then make writing that relatively lightweight.

PR #1796, opened 2024-08-09
Test that pgrx can build after publishing

workingjubilee:
Test that pgrx will always build after we publish it. It should never again take an excessive amount of time to release pgrx, because we will always be confident we are ready to release. This does not finish making the release "turnkey", but it does take care of every obstacle to such.

Note that there is a technicality: we are "only" testing that our _package_ builds. We are not actually pulling a published-to-cargo release and testing it. We are not running the `cargo publish --dry-run` command, either, because there is no `cargo publish --workspace`. Instead, we are packaging the workspace and rebuilding **that package**. This unfortunately demands that we factor out all the packages in the workspace that are not going to be published.

Other details of the refactoring are informed by these oddities:
- `package.publish = false` doesn't play well with the extensions to `cargo package --workspace`: https://github.com/rust-lang/cargo/issues/14356
- `workspace.exclude = []` does not properly support globs: https://github.com/rust-lang/cargo/issues/6009
- We *need* to reuse the same `CARGO_TARGET_DIR` for building and testing our many examples or we run out of storage space on the runners.


workingjubilee:
Will deal with the CI failures when I wake up.

workingjubilee:
Just need to refactor rerunning the individual crate tests.

workingjubilee:
We almost certainly are now doing redundant build/test cycling but that's fine because I'm going to work on cutting all that down LATER.

workingjubilee:
finally.

PR #1802, opened 2024-08-13
Split out pgrx-bindgen crate

workingjubilee:
One last essentially-administrative change, and part of the reason that I did all the mucking-about in #1796. We have a lot of code in pgrx-pg-sys's build.rs, and it's high time that we actually split out the real crate that it is. Currently this *just* moves that around, rather than doing any refactoring, to make the git blame trackable.

The packaging test generates a sufficiently large cache that it cannot be done for every version of pgrx in the combinatoric fashion that we usually do, thus, we don't anymore! We simply run it for one version.

### DIFF
--- a/.github/workflows/package-test.yaml
+++ b/.github/workflows/package-test.yaml
@@ -1,0 +1,45 @@
+name: verify package can build
+
+on:
+  push:
+    branches: [develop]
+  pull_request:
+    branches: [develop]
+
+env:
+  CARGO_TERM_COLOR: always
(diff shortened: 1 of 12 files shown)


workingjubilee:
The packaging test generates a sufficiently large cache that it cannot be done for every version of pgrx in the combinatoric fashion that we usually do, thus, we don't anymore! We simply run it for one version.

workingjubilee:
Which predictably drops several minutes off CI! Nice. Let's go!

PR #1803, opened 2024-08-13
Update versions even if no specifier

workingjubilee:
Found a bug after doing some refactoring to improve how this tool works in general, making it more resilient to whatever the hell we do to our repo. This should allow us to have packages with different names than the paths to them, allowing arbitrary nesting and relocation, and also applies a less ad-hoc rule regarding "what packages are we updating".

Unblocks #1802.

### DIFF
--- a/tools/version-updater/Cargo.lock
+++ b/tools/version-updater/Cargo.lock
@@ -161,6 +161,7 @@ dependencies = [
  "clap",
  "owo-colors",
  "rustc-hash",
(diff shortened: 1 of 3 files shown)


BradyBonnette (approved):
I approve mostly because I was on a call with you while you were fixing it :D

workingjubilee:
That PR is now failing on another problem but hey! This one works as-intended.

PR #1804, opened 2024-08-14
Make bindgen generate CStrs and such

workingjubilee:
While cutting a release, I noticed our bindgen options were wonky. I fixed a few blocklistings and adjusted the bindgen settings. Most of this should be invisible. There's only one case affected by nested struct naming, and it's not even present in pg16 and later. But the most notable change, and one that I expect might force people to have to update their code: now const strings from C are CStrs!


eeeebbbbrrrr:
> now const strings from C are CStrs!

That's very nice.

workingjubilee:
I should default to checking pg16 instead of pg13~14.

PR #1808, opened 2024-08-14
Correct release instructions for current repo tools

workingjubilee:
- Remove a reference to a script that doesn't exist anymore.
- Remind people to pull off develop so they publish the same commit.


eeeebbbbrrrr (approved):
Thank you!

workingjubilee:
Added a pair of cargo commands that we found were actually secretly necessary.

PR #1823, opened 2024-08-21
Implement ArgAbi for PgRelation

thomcc:
This is missing in 0.12.0.

It was originally thought that perhaps extensions didn't need this, as pgrx makes a lot of opinionated choices with PgRelation that aren't always applicable. In actuality there are extensions where this proves to be a significant amount of the 0.12 update friction.

### DIFF
--- a/pgrx/src/callconv.rs
+++ b/pgrx/src/callconv.rs
@@ -21,6 +21,7 @@ use crate::heap_tuple::PgHeapTuple;
 use crate::nullable::Nullable;
 use crate::pg_sys;
 use crate::pgbox::*;
+use crate::rel::PgRelation;
 use crate::{PgBox, PgMemoryContexts};
 
 use core::marker::PhantomData;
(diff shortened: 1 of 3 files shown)


workingjubilee:
Hi, can you add a test for this? (it would be literally just a pass/return fn)

thomcc:
Sure.

pgrx/src/callconv.rs
```diff
@@ -262,7 +262,7 @@ macro_rules! argue_from_datum {
 argue_from_datum! { 'fcx; i8, i16, i32, i64, f32, f64, bool, char, String, Vec<u8> }
 argue_from_datum! { 'fcx; Date, Interval, Time, TimeWithTimeZone, Timestamp, TimestampWithTimeZone }
 argue_from_datum! { 'fcx; AnyArray, AnyElement, AnyNumeric }
-argue_from_datum! { 'fcx; Inet, Internal, Json, JsonB, Uuid }
+argue_from_datum! { 'fcx; Inet, Internal, Json, JsonB, Uuid, crate::rel::PgRelation }
```
workingjubilee: Also, I would prefer this was actually imported, because I had to spend _considerable_ amounts of time refactoring messy imports recently, and I have developed a Strong Preference around actual imports.

specifically `module::Type` is okay by me but I would prefer to avoid from the crate root except in the top level imports (this makes it easier to move modules around or etc).

thomcc: Done.

thomcc:
My test only covers passing it as an argument since I don't know how returning them works really 😅.

workingjubilee:
Oh, yk, it probably doesn't.

PR #1827, opened 2024-08-24
Add UnboxDatum for Range<T>

mhov:
I was getting compiler complaints from this input type `my_func(arr: Array<Range<Date>>)`  and calling `.get_by_name::<Range<Date>>`. Implement the required trait.


workingjubilee:
Thanks!

PR #1831, opened 2024-08-27
Moved Sized from FromDatum methods to trait

YohDeadfall:
Since all methods of `FromDatum` require `Self` to be `Sized` it makes sense to move the bound to the trait itself.


workingjubilee:
Sure! Thanks!

PR #1836, opened 2024-09-02
Used SPI result type for cursor API

YohDeadfall:
A tiny cleanup of the cursor code. `SpiResult<R>` is no different than `Result<R, SpiError>`.


workingjubilee:
Thanks!

PR #1837, opened 2024-09-02
Fixed out of bounds reads for open_cursor

YohDeadfall:
The problem arises when the prepared statement has less arguments than
the used plan since there's no parameter to pass the number of arguments
like it's for `SPI_execute_with_args`. Meanwhile the `execute` method
does a check.

Introduce a variant, `try_open_cursor`, that returns an SpiResult instead.

### DIFF
--- a/pgrx-tests/src/tests/spi_tests.rs
+++ b/pgrx-tests/src/tests/spi_tests.rs
@@ -203,7 +203,6 @@ mod tests {
     }
 
     #[pg_test]
-    #[allow(deprecated)]
     fn test_cursor() -> Result<(), spi::Error> {
         Spi::connect(|mut client| {
             client.update("CREATE TABLE tests.cursor_table (id int)", None, None)?;
@@ -224,7 +223,6 @@ mod tests {
     }
 
     #[pg_test]
-    #[allow(deprecated)]
     fn test_cursor_prepared_statement() -> Result<(), pgrx::spi::Error> {
         Spi::connect(|mut client| {
             client.update("CREATE TABLE tests.cursor_table (id int)", None, None)?;
@@ -263,7 +261,6 @@ mod tests {
         test_cursor_prepared_statement_panics_impl(Some([None, None].to_vec()))
     }
 
-    #[allow(deprecated)]
     fn test_cursor_prepared_statement_panics_impl(
         args: Option<Vec<Option<pg_sys::Datum>>>,
     ) -> Result<(), pgrx::spi::Error> {
@@ -285,7 +282,6 @@ mod tests {
     }
 
     #[pg_test]
-    #[allow(deprecated)]
     fn test_cursor_by_name() -> Result<(), pgrx::spi::Error> {
         let cursor_name = Spi::connect(|mut client| {
             client.update("CREATE TABLE tests.cursor_table (id int)", None, None)?;
@@ -317,7 +313,6 @@ mod tests {
     }
 
     #[pg_test(error = "syntax error at or near \"THIS\"")]
-    #[allow(deprecated)]
     fn test_cursor_failure() {
         Spi::connect(|client| {
             client.open_cursor("THIS IS NOT SQL", None);
@@ -325,7 +320,6 @@ mod tests {
     }
 
     #[pg_test(error = "cursor: CursorNotFound(\"NOT A CURSOR\")")]
-    #[allow(deprecated)]
     fn test_cursor_not_found() {
         Spi::connect(|client| client.find_cursor("NOT A CURSOR").map(|_| ())).expect("cursor");
     }
@@ -358,7 +352,6 @@ mod tests {
     }
 
     #[pg_test]
-    #[allow(deprecated)]
     fn test_spi_non_mut() -> Result<(), pgrx::spi::Error> {
         // Ensures update and cursor APIs do not need mutable reference to SpiClient
         Spi::connect(|mut client| {
--- a/pgrx-tests/src/tests/spi_tests.rs
+++ b/pgrx-tests/src/tests/spi_tests.rs
@@ -244,6 +244,44 @@ mod tests {
     }
 
     #[pg_test]
+    #[should_panic]
+    fn test_cursor_prepared_statement_panics_none_args() -> Result<(), pgrx::spi::Error> {
+        test_cursor_prepared_statement_panics_impl(None)
+    }
+
+    #[pg_test]
+    #[should_panic]
+    fn test_cursor_prepared_statement_panics_less_args() -> Result<(), pgrx::spi::Error> {
+        test_cursor_prepared_statement_panics_impl(Some([].to_vec()))
+    }
+
+    #[pg_test]
+    #[should_panic]
+    fn test_cursor_prepared_statement_panics_more_args() -> Result<(), pgrx::spi::Error> {
+        test_cursor_prepared_statement_panics_impl(Some([None, None].to_vec()))
+    }
+
+    fn test_cursor_prepared_statement_panics_impl(
+        args: Option<Vec<Option<pg_sys::Datum>>>,
+    ) -> Result<(), pgrx::spi::Error> {
+        Spi::connect(|mut client| {
+            client.update("CREATE TABLE tests.cursor_table (id int)", None, None)?;
+            client.update(
+                "INSERT INTO tests.cursor_table (id) \
+            SELECT i FROM generate_series(1, 10) AS t(i)",
+                None,
+                None,
+            )?;
+            let prepared = client.prepare(
+                "SELECT * FROM tests.cursor_table WHERE id = $1",
+                Some([PgBuiltInOids::INT4OID.oid()].to_vec()),
+            )?;
+            client.open_cursor(&prepared, args);
+            unreachable!();
+        })
+    }
+
+    #[pg_test]
     fn test_cursor_by_name() -> Result<(), pgrx::spi::Error> {
         let cursor_name = Spi::connect(|mut client| {
             client.update("CREATE TABLE tests.cursor_table (id int)", None, None)?;
--- a/pgrx-tests/src/tests/spi_tests.rs
+++ b/pgrx-tests/src/tests/spi_tests.rs
@@ -203,6 +203,7 @@ mod tests {
     }
 
     #[pg_test]
+    #[allow(deprecated)]
     fn test_cursor() -> Result<(), spi::Error> {
         Spi::connect(|mut client| {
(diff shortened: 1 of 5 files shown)


workingjubilee:
can you add a FIXME for it? thanks.

workingjubilee:
The main reason it doesn't return a Result is that all of this was a relatively under-designed API that has been scheduled for total demolition and replacement, so comparatively minor but still breaking changes like "maybe we should have returned a Result here..." have been on-hold.

pgrx/src/spi/query.rs
```diff
@@ -262,9 +270,7 @@ impl<'conn: 'stmt, 'stmt> Query<'conn> for &'stmt PreparedStatement<'conn> {
     }
 
     fn open_cursor(self, _client: &SpiClient<'conn>, args: Self::Arguments) -> SpiCursor<'conn> {
-        let args = args.unwrap_or_default();
-
-        let (mut datums, nulls): (Vec<_>, Vec<_>) = args.into_iter().map(prepare_datum).unzip();
+        let (mut datums, nulls) = self.args_to_datums(args).unwrap();
```
workingjubilee: specifically adding the FIXME right here.

Also alternatively: add `fn try_open_cursor`, then tag *this* one with `#[deprecated(since = "0.12.2", note = "lol ub")]`, and then replace the inner impl of this with just `self.try_open_cursor(_client, args).unwrap()`.

workingjubilee (changes requested):
that said, adding new API is optional, but we do need a `#[should_panic]` test on this case for `open_cursor`, or a "returns Err" test on the `try_open_cursor` fn, either one.

pgrx/src/spi/query.rs
YohDeadfall: Will write the new method too then, that's not an issue for me. And actually, I have other plans for improvements there and around (:

workingjubilee:
It feels a little early to bump to 0.13, not that I'm afraid of such, but yk "a month" seems like a good span.

YohDeadfall:
I addressed your changes, but what I honestly don't like is `#[allow(deprecated)]` attributes in tests, but if with the next breaking version it will be removed that seems fine.

If you share your plans on the SPI part, then I can contribute by reshaping it.

workingjubilee:
> If you share your plans on the SPI part, then I can contribute by reshaping it.

The most fundamental issue is that the API does not enforce lifetime bounds properly. Solving this is hard, but it is why 0.12 landed a huge breaking change for most pgrx-using code: solving the problem with the `pgrx::spi` API and then not fixing the entire rest of everything was basically closing the barn door well after every cow had not only run out but acquired demonic powers, picked up a hellforged weapon, and turned into a roaming infernal bovine.

It also has some overly complicated parts, like the overly-cutesy Query trait, which I would wish to move away from.

pgrx-tests/src/tests/spi_tests.rs
```diff
@@ -244,6 +246,46 @@ mod tests {
...
+                None,
+                None,
+            )?;
+            let prepared = client.prepare(
+                "SELECT * FROM tests.cursor_table WHERE id = $1",
+                Some([PgBuiltInOids::INT4OID.oid()].to_vec()),
+            )?;
+            client.open_cursor(&prepared, args);
+            unreachable!();
+        })
```
workingjubilee, line 281: huh, does this return an error because we panic during the connection? interesting.

pgrx/src/spi/client.rs
```diff
@@ -117,22 +117,63 @@ impl<'conn> SpiClient<'conn> {
...
     /// Rows may be then fetched using [`SpiCursor::fetch`].
     ///
     /// See [`SpiCursor`] docs for usage details.
+    ///
+    /// See [`try_open_cursor_mut`][Self::try_open_cursor_mut] which will return an [`SpiError`] rather than panicking.
+    ///
+    /// # Panics
+    ///
+    /// Panics if a cursor wasn't opened.
+    #[deprecated(since = "0.12.2", note = "undefined behavior")]
```
workingjubilee: hm. on second thought, not entirely sure if we want to deprecate this just because it panics... otoh most people don't like panics...

workingjubilee (approved):
thank you, this looks good.

pgrx-tests/src/tests/spi_tests.rs
YohDeadfall: Right, it panics and the tests are marked as `#[should_panic]`. That allows us to verify that there's no undefined behavior.

pgrx/src/spi/client.rs
YohDeadfall: So, deprecate or leave it as it is for now? Personally I hate panicking, but there's a lot of upcoming work with breaking changes anyway, I guess, so forcing users to fix their code on each release isn't wise.

pgrx-tests/src/tests/spi_tests.rs
workingjubilee: Interesting.

pgrx/src/spi/client.rs
workingjubilee: Yeah, I've reconsidered and I think we should remove the deprecation note, I don't want to churn people pointlessly. I'll try to mention it in the release notes.

workingjubilee: or @eeeebbbbrrrr will. :^)

unless he feels that we should in fact deprecate this because it started panicking (but only when the alternative was UB).

pgrx-tests/src/tests/spi_tests.rs
```diff
@@ -203,6 +203,7 @@ mod tests {
     }
 
     #[pg_test]
+    #[allow(deprecated)]
```
workingjubilee:
```suggestion
```

pgrx-tests/src/tests/spi_tests.rs
```diff
@@ -223,6 +224,7 @@ mod tests {
...
```
workingjubilee:
```suggestion
```

pgrx-tests/src/tests/spi_tests.rs
```diff
@@ -244,6 +246,46 @@ mod tests {
...
+        test_cursor_prepared_statement_panics_impl(Some([].to_vec()))
+    }
+
+    #[pg_test]
+    #[should_panic]
+    fn test_cursor_prepared_statement_panics_more_args() -> Result<(), pgrx::spi::Error> {
+        test_cursor_prepared_statement_panics_impl(Some([None, None].to_vec()))
+    }
+
+    #[allow(deprecated)]
```
workingjubilee:
```suggestion
```

pgrx-tests/src/tests/spi_tests.rs
```diff
@@ -244,6 +246,46 @@ mod tests {
...
+                "SELECT * FROM tests.cursor_table WHERE id = $1",
+                Some([PgBuiltInOids::INT4OID.oid()].to_vec()),
+            )?;
+            client.open_cursor(&prepared, args);
+            unreachable!();
+        })
+    }
+
+    #[pg_test]
+    #[allow(deprecated)]
```
workingjubilee:
```suggestion
```

pgrx-tests/src/tests/spi_tests.rs
```diff
@@ -275,13 +317,15 @@ mod tests {
     }
 
     #[pg_test(error = "syntax error at or near \"THIS\"")]
+    #[allow(deprecated)]
```
workingjubilee:
```suggestion
```

pgrx-tests/src/tests/spi_tests.rs
```diff
@@ -275,13 +317,15 @@ mod tests {
...
     fn test_cursor_failure() {
         Spi::connect(|client| {
             client.open_cursor("THIS IS NOT SQL", None);
         })
     }
 
     #[pg_test(error = "cursor: CursorNotFound(\"NOT A CURSOR\")")]
+    #[allow(deprecated)]
```
workingjubilee:
```suggestion
```

pgrx-tests/src/tests/spi_tests.rs
```diff
@@ -314,6 +358,7 @@ mod tests {
     }
 
     #[pg_test]
+    #[allow(deprecated)]
```
workingjubilee:
```suggestion
```

pgrx-tests/src/tests/spi_tests.rs
```diff
@@ -532,6 +577,7 @@ mod tests {
...
```
workingjubilee, line 573:
```suggestion
```

pgrx/src/spi/client.rs
```diff
@@ -117,22 +117,63 @@ impl<'conn> SpiClient<'conn> {
     /// Rows may be then fetched using [`SpiCursor::fetch`].
     ///
     /// See [`SpiCursor`] docs for usage details.
+    ///
+    /// See [`try_open_cursor`][Self::try_open_cursor] which will return an [`SpiError`] rather than panicking.
+    ///
+    /// # Panics
+    ///
+    /// Panics if a cursor wasn't opened.
+    #[deprecated(since = "0.12.2", note = "undefined behavior")]
```
workingjubilee:
```suggestion
```

pgrx/src/spi/client.rs
workingjubilee:
```suggestion
```

pgrx/src/spi/query.rs
```diff
@@ -13,19 +13,33 @@ use crate::pg_sys::{self, PgOid};
...
     ) -> SpiResult<SpiTupleTable<'conn>>;
 
-    /// Open a cursor for the query
-    fn open_cursor(self, client: &SpiClient<'conn>, args: Self::Arguments) -> SpiCursor<'conn>;
+    /// Open a cursor for the query.
+    ///
+    /// # Panics
+    ///
+    /// Panics if a cursor wasn't opened.
+    #[deprecated(since = "0.12.2", note = "undefined behavior")]
```
workingjubilee:
```suggestion
```

workingjubilee (approved):
stripping the deprecated stuff mostly, sorry about changing my mind back and forth.

workingjubilee:
Thanks!

pgrx-tests/src/tests/spi_tests.rs
YohDeadfall: Okay, it's a bit incorrect without the reason I haven't provided for tests because `unreachable!!` panics too and gives us a false negative. See #1843 for a fix.

Issue #1835, opened 2024-09-01
Compilation speed


ccleve:
If you run "`cargo pgrx run --release`", it takes a long time to compile (as it should).

If you run it again, immediately, without making any code changes, it takes a long time. It hangs on this line:

`Building [=======================> ] 231/232: pgrx_embed_relevantdb(bin)  `

I don't think it used to do that prior to 0.12.

I'm using 0.12.1.

eeeebbbbrrrr:
Hey!  Ya know, I've noticed this exact thing now that I've upgraded a project to 0.12.  It's on my list to look at first thing next week.  Off the top of my head, I am unsure what's going on.

workingjubilee:
I'm guessing it's because we kiiinda "modify" the source in a way that cargo interprets as having diffed it...?

usamoi:
It hangs if LTO is enabled. I usually build an optimized artifact with LTO disabled. I enable LTO only when I really need to release a new version.

eeeebbbbrrrr:
I'm taking a look at it now

eeeebbbbrrrr:
One quick experiment that looks good is for cargo-pgrx to **not** pass the `--release` or `--profile XXX` flags down when compiling the pgrx_embed binary -- instead always compile it in debug mode.

Immediately after a `cargo clean`, that causes two full compilations of the code if using `--release/--profile`, but after that, the built of the pgrx_embed binary is nearly instant.

Thoughts?

eeeebbbbrrrr:
I put up PR #1838 that makes this a lot better.  @ccleve / @usamoi, feel free to install cargo-pgrx from that PR if you want and see if it makes ya happier.

workingjubilee:
setting LTO to fat will still be overkill, even for release binaries.

ccleve:
Not passing the --release flag is a good solution.

eeeebbbbrrrr:
@workingjubilee looking at this in some detail this morning brought up the memories of the 4 other ways pgrx has historically solved the schema generation problem.

I almost wonder if there's one more iteration here where we don't need the pg_embed.exe at all?

Why wouldn't it be possible to `dlopen` the extension.so with `RTLD_LAZY` and just directly call the function symbols we need to run?

From the manpage:

```
       RTLD_LAZY
              Perform lazy binding.  Resolve symbols only as the code
              that references them is executed.  If the symbol is never
              referenced, then it is never resolved.  (Lazy binding is
[...]

PR #1838, opened 2024-09-03
always compile `pgrx_embed_*` without opts

eeeebbbbrrrr:
This addresses issue #1835 by always compiling the `pgrx_embed_xxx` binary in debug mode.

Essentially, `cargo pgrx` now ignores the `--release` or `--profile XXX` arguments when compiling the `pgrx_embed_xxx` binary.

(I am not sure how we can avoid the re-compile completely.  My thought was somehow detecting that the contents of the `PGRX_EMBED` file (which is a temp file generated on every build) haven't actually changed, but I don't know what we'd compare it to?)

Anyways, this PR solves my annoyance and I think it should be sufficient for the OP too.

### DIFF
--- a/cargo-pgrx/src/command/schema.rs
+++ b/cargo-pgrx/src/command/schema.rs
@@ -147,6 +147,7 @@ pub(crate) fn generate_schema(
     let lib_filename = manifest.lib_filename()?;
 
     if !skip_build {
+        // NB:  The only path where this happens is via the command line using `cargo pgrx schema`
         first_build(
             user_manifest_path.as_ref(),
             profile,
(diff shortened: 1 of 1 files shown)


workingjubilee:
> (I am not sure how we can avoid the re-compile completely. My thought was somehow detecting that the contents of the PGRX_EMBED file (which is a temp file generated on every build) haven't actually changed, but I don't know what we'd compare it to?)

I believe that it would be possible for cargo to manage this via https://github.com/rust-lang/cargo/issues/6529 but they have so far chosen not to implement that.

workingjubilee:
something something filetimes drool content-addressable storage rules
but regardless, this seems fine to me.

PR #1841, opened 2024-09-06
Include `access/visibilitymap.h` and `utils/tuplestore.h`

eeeebbbbrrrr:
The ability to inspect the visibility map is useful.  So is being able to work with the result of SRFs that return a tuplestore.


eeeebbbbrrrr:
I'm Eric Ridge and I approve this message

PR #1845, opened 2024-09-09
Add an example of a WAL decoder

daamien:
Hi !

I recently wrote this example as an exercice to learn how to build a WAL decoder with PGRX.... I built this as « the thing I would have loved to have before » when I started digging on this topic. Feel free to modify it heavily or discard it completely if it is too specific or if the coding style is not up to the PGRX standards. I tried my best to write idiomic Rust but there may be some naïve/unsound parts...

This example shows how to build a basic Change Data Capture (CDC) mechanism using the Postgres Logical Decoding capabilities. The changes occurring on the database are serialized into JSON and pushed to a queue (a "logical replication slot") where they can be consumed by remote clients.

A Postgres CDC extension can be a used in various purposes:

* Ad hoc replication systems (e.g. Postgres => SQL Server )
* External commit-log for a distributed system (such as Kafka)
* Advanced Monitoring ( e.g. Prometheus/Loki )

This example tries to find the right tradeoff between simplicity and usefulness. Currently it has strong limitations but they should be easy to overcome.

Rust really shines in this example compared to similar implementations in C (see the wal2json extension). The Serde crate provides JSON serialization out of the box, whereas all the C implementations are forced to write their own JSON formatter.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@
 #LICENSE All rights reserved.
 #LICENSE
 #LICENSE Use of this source code is governed by the MIT license that can be found in the LICENSE file.
- 
+
 [workspace]
 resolver = "2"
 members = [
@@ -32,6 +32,7 @@ exclude = [
     "pgrx-examples/custom_types",
     "pgrx-examples/custom_sql",
     "pgrx-examples/datetime",
+    "pgrx-examples/wal_decoder",
     "pgrx-examples/errors",
     "pgrx-examples/nostd",
     "pgrx-examples/numeric",
--- a/pgrx-examples/wal_decoder/.gitignore
+++ b/pgrx-examples/wal_decoder/.gitignore
@@ -1,0 +1,6 @@
+.DS_Store
+.idea/
+/target
+*.iml
+**/*.rs.bk
+Cargo.lock
--- a/pgrx-examples/wal_decoder/Cargo.toml
+++ b/pgrx-examples/wal_decoder/Cargo.toml
@@ -1,0 +1,38 @@
+[package]
+name = "wal_decoder"
+version = "0.0.0"
+edition = "2021"
+
+[lib]
+crate-type = ["cdylib", "lib"]
+
+[[bin]]
+name = "pgrx_embed_wal_decoder"
+path = "./src/bin/pgrx_embed.rs"
+
+[features]
+default = ["pg13"]
+pg12 = ["pgrx/pg12", "pgrx-tests/pg12" ]
+pg13 = ["pgrx/pg13", "pgrx-tests/pg13" ]
+pg14 = ["pgrx/pg14", "pgrx-tests/pg14" ]
+pg15 = ["pgrx/pg15", "pgrx-tests/pg15" ]
+pg16 = ["pgrx/pg16", "pgrx-tests/pg16" ]
+pg17 = ["pgrx/pg17", "pgrx-tests/pg17" ]
+pg_test = []
+
+[dependencies]
+pgrx = { path = "../../pgrx", default-features = false }
+serde = "1.0.209"
+serde_json = "1.0.128"
+
+[dev-dependencies]
+pgrx-tests = { path = "../../pgrx-tests" }
+
+[profile.dev]
+panic = "unwind"
+
+[profile.release]
+panic = "unwind"
+opt-level = 3
+lto = "fat"
+codegen-units = 1
--- a/pgrx-examples/wal_decoder/README.md
+++ b/pgrx-examples/wal_decoder/README.md
@@ -1,0 +1,118 @@
+
+# A simple Change Data Capture ([CDC]) extension.
+
+This extension will extract the [DML] changes from Postgres WAL
+using [Logical Decoding] and export them in JSON format using [serde].
+
+## Principles
+
+* Postgres triggers various callbacks at the different stages of a transaction
+* The decoder defines some of these callbacks: begin, change, commit, etc.
+* The callbacks extract the changes made during the transaction
+* They build Rust structs (Action, Tuple) to represent those changes
+* The structs are then serialized into JSON
+* The JSON output is sent into a logical replication slot (i.e. a queue)
+* The output can be consumed in various ways by a remote client
+
+## Requirements
+
+In order to use this extension with a cargo-pgrx managed instance, you'll
+need to add the configuration below in "$PGRX_HOME/data-$PGVER/postgresql.conf".
+
+``` ini
+shared_preload_libraries = 'wal_decoder'
+wal_level = logical
+```
+
+## Example
+
+1- Create a table and publish it
+
+``` sql
+CREATE TABLE person (name TEXT, age INT);
+ALTER TABLE person REPLICA IDENTITY FULL;
+CREATE PUBLICATION gotham_pub FOR TABLE person;
+```
+
+2- Create a replication slot fed by the decoder
+
+``` sql
+SELECT pg_create_logical_replication_slot('gotham_slot', 'wal_decoder');
+```
+
+3- Consume the changes from the replication slot
+
+``` sql
+INSERT INTO person
+VALUES ('Bruce Wayne',42),('Clark Kent',33);
+```
+
+``` sql
+SELECT * FROM pg_logical_slot_get_changes('gotham_slot', NULL, NULL);
+
+    lsn    | xid |                                     data
+-----------+-----+------------------------------------------------------------------------------
+ 0/16A87C8 | 581 | {"typ":"BEGIN"}
+ 0/16A87C8 | 581 | {"typ":"INSERT","rel":"public.person","new":{"name":"Bruce Wayne","age":42}}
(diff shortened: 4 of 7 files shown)


eeeebbbbrrrr:
Could I trouble you to put a README.md together for this crate that details what it does and how to use it?

daamien:
> Could I trouble you to put a README.md together for this crate that details what it does and how to use it?

Sure, I wrote the explanations in the preamble comment of the `src/lib.rs` but I can move it into README file

pgrx-examples/decoder/src/lib.rs
```diff
@@ -240,10 +240,8 @@ unsafe extern "C" fn pg_decode_begin_txn(
     ctx_ptr: *mut pg_sys::LogicalDecodingContext,
     _txn_ptr: *mut pg_sys::ReorderBufferTXN
 ){
-    let ctx = unsafe { PgBox::from_pg(ctx_ptr) } ;
-    let mut state = unsafe {
-        PgBox::from_pg(ctx.output_plugin_private as *mut DecodingState)
-    };
+    let ctx = PgBox::from_pg(ctx_ptr);
+    let mut state = PgBox::from_pg(ctx.output_plugin_private as *mut DecodingState);
```
workingjubilee: You're gonna want to revert this commit, because this should have `deny(unsafe_op_in_unsafe_fn)`.

pgrx-examples/decoder/src/lib.rs
```diff
@@ -0,0 +1,322 @@
+use pgrx::prelude::*;
```
workingjubilee:
```suggestion
#![deny(unsafe_op_in_unsafe_fn)]
use pgrx::prelude::*;
```

pgrx-examples/decoder/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+mod tests {
+    use pgrx::prelude::*;
+
+    #[pg_test]
+    #[ignore]
+    fn test_decoder() {
+        //
+        // Not sure how to handle the tests of this extension....
+        // This function is running *inside* a transaction
+        // and the WAL decoder will output the changes only when they ar
```
workingjubilee:
```suggestion
        // and the WAL decoder will output the changes only when they are
```

pgrx-examples/decoder/src/bin/pgrx_embed.rs
```diff
@@ -0,0 +1 @@
+::pgrx::pgrx_embed!();
```
workingjubilee: nit but I just don't like GitHub whining at me about it.
```suggestion
::pgrx::pgrx_embed!();

```

pgrx-examples/decoder/README.md
```diff
@@ -0,0 +1,119 @@
...
+* They build Rust structs (Action, Tuple) to represent those changes
+* The structs are then serialized into JSON
+* The JSON output is sent into a replication slot
+* The ouput be can consumed in various ways by a remote client.
+
+
+## Requirements
+
+In order to use this extension with a cargo-pgrx managed database, you'll
+need to add the configuration below in "$PGRX_HOME/data-$PGVER/postgresql.conf".
```
workingjubilee: isn't this "any" database, to the appropriate postgresql.conf?

pgrx-examples/decoder/README.md
```diff
@@ -0,0 +1,119 @@
+
+# A simple Change Data Capture ([CDC]) extension.
+
+This extension will extract the [DML] changes from Postgres WAL
+using [Logical Decoding] and export them in JSON format using [serde].
+
+## Principles
+
+* Postgres triggers various callbacks at the different stages of a transaction
+* The decoder defines some of these callbacks : begin, change, commit, etc.
```
workingjubilee:
```suggestion
* The decoder defines some of these callbacks: begin, change, commit, etc.
```

pgrx-examples/decoder/README.md
```diff
@@ -0,0 +1,119 @@
...
+* The callbacks extract the changes made during the transaction
+* They build Rust structs (Action, Tuple) to represent those changes
+* The structs are then serialized into JSON
+* The JSON output is sent into a replication slot
+* The ouput be can consumed in various ways by a remote client.
```
workingjubilee:
```suggestion
* The output can be consumed in various ways by a remote client
```

pgrx-examples/decoder/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+    #[must_use]
+    pub fn postgresql_conf_options() -> Vec<&'static str> {
+        // return any postgresql.conf settings that are required for your tests
+        vec![]
+    }
+}
+
+// <!-- EOL -->
+
+
```
workingjubilee:
```suggestion
```

pgrx-examples/decoder/README.md
```diff
@@ -0,0 +1,119 @@
...
+     |     "change_count": 1            +
+     | }
+```
+
+## Limitations
+
+This decoder is designed as a basic example and it has the following limitations:
+
+* Only the REPLICA IDENTITY FULL mode is fully supported. Supporting REPLICA IDENTITY DEFAULT
+  would require additionnal work.
```
workingjubilee:
```suggestion
  would require additional work.
```

pgrx-examples/decoder/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+    committed: Option<i64>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    rel: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    old: Option<Tuple>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    new: Option<Tuple>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    change_count: Option<i64>,
+}
```
workingjubilee: Can you add a unit test to this crate that attempts to serialize this struct, with all the Options as None, and then deserialize this struct from the resulting string?

pgrx-examples/decoder/src/lib.rs
workingjubilee, line 1: Could you run `cargo fmt` on this? Thank you.

workingjubilee (changes requested):
Thanks! Looks good to me. I indeed have some mostly-stylistic nits in mind here.

pgrx-examples/decoder/README.md
daamien: I think I meant "cargo-pgrx managed instance"

workingjubilee:
I'm perplexed why we don't implement both deserialization and serialization.

daamien:
> I'm perplexed why we don't implement both deserialization and serialization.

The purpose of a WAL decoding extension is to receive a WAL through the logical decoding mechanism and transform it into text representations of the operations performed. In this example, the extension will produce a JSON output but it will never have to injest any JSON input. It's really a one way road.

The question of who will consume the JSON output and how it will be consumed is out of the scope.

Imagine that you are legally required to archive any change made to a given table in your database. You might want to use a Kafka storage for that.  First the WAL decoder will generate the JSON output in the replication slot and then you'd write a client that consumes events in the replication slot and push them to kafka.

workingjubilee:
Hmm. I guess it just feels weird that we don't check we can reconstruct our type from its own output. If we could, that capability... the interface type... could live in its own library, you see, that both producer and consumer could reference.

pgrx-examples/decoder/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+mod tests {
+    use pgrx::prelude::*;
+
+    #[pg_test]
+    fn test_action_begin() {
+        use crate::Action;
+
+        let input = Action::begin();
+        let json = serde_json::to_string(&input).expect("Serde Error");
+        assert_eq!(json, "{\"typ\":\"BEGIN\"}");
```
workingjubilee: We don't necessarily need vigorous escaping if we use a raw string:
```suggestion
        assert_eq!(json, r#"{"typ":"BEGIN"}"#);
```

workingjubilee (approved):
I think this looks fine.

workingjubilee:
Does it work with `./run-examples.sh`?

pgrx-examples/decoder/Cargo.toml
```diff
@@ -0,0 +1,38 @@
+[package]
+name = "decoder"
```
workingjubilee: oh, only last thought is that it should probably be named `wal_decoder` instead? likelihood of another `decoder` example appearing over sufficient time is approximately 1

daamien: agreed.

Maybe wal2json_decoder ? There may be others types of WAL decoder in the future ?

daamien:
> Does it work with `./run-examples.sh`?

It does :)

daamien:
Renamed , Squashed and Rebased :)

workingjubilee:
Thanks!

Oh, re: soundness, nothing seems obviously wrong? But I haven't mind-numbingly-closely audited it for soundness as it's just an example, and thus is user code, basically: any brain cells I spend on doing that are better off spent on making it easier to write code soundly than torturing you with revisions of some pretty gnarly unsafe code.

PR #1849, opened 2024-09-10
Fixes wrong value for microsecs in a day

### LINKED ISSUE #1850: Conversion from `TimeTzADT` to `TimeWithTimeZone` is broken
It is due to wrong value at https://github.com/pgcentralfoundation/pgrx/blob/b4c444111814a51abb15a14669b95a080c51aed5/pgrx/src/datum/time_with_timezone.rs#L27.

It misses 3 additional zeros.

aykut-bozkurt:
`MICROSECONDS_PER_DAY` constant at `time_with_timezone.rs` misses 3 additional zeros for microseconds part.

That breaks conversion from `(pg_sys::TimeADT, i32)` to `TimeWithTimeZone`. 

Fixes #1850

### DIFF
--- a/pgrx/src/datum/datetime_support/mod.rs
+++ b/pgrx/src/datum/datetime_support/mod.rs
@@ -24,6 +24,9 @@ mod ops;
 
 pub use ctor::*;
 
+pub const USECS_PER_SEC: i64 = 1_000_000;
+pub const USECS_PER_DAY: i64 = pg_sys::SECS_PER_DAY as i64 * USECS_PER_SEC;
+
 /// Tags to identify which "part" of a date or time-type value to extract or truncate to
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
 pub enum DateTimeParts {
--- a/pgrx/src/datum/time_with_timezone.rs
+++ b/pgrx/src/datum/time_with_timezone.rs
@@ -11,7 +11,7 @@ use super::{
     DateTimeConversionError, FromDatum, Interval, IntoDatum, Time, TimestampWithTimeZone,
     ToIsoString,
 };
-use crate::datum::datetime_support::{DateTimeParts, HasExtractableParts};
+use crate::datum::datetime_support::{DateTimeParts, HasExtractableParts, USECS_PER_DAY};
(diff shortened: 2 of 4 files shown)


pgrx/src/datum/time_with_timezone.rs
```diff
@@ -24,7 +24,10 @@ use std::panic::{RefUnwindSafe, UnwindSafe};
 #[derive(Debug, Copy, Clone)]
 #[repr(transparent)]
 pub struct TimeWithTimeZone(pg_sys::TimeTzADT);
-const MICROSECONDS_PER_DAY: pg_sys::TimeADT = 86_400_000;
+
+// 86_400_000_000
```
YohDeadfall: Thanks for the fix! Could you mention that it comes from the `USECS_PER_DAY` constant?

eeeebbbbrrrr:
Thanks @aykut-bozkurt. Could you add a test for this in the pgrx-tests crate?  I believe there’s a module for date/time tests already.

pgrx/src/datum/time_with_timezone.rs
aykut-bozkurt: Makes sense, let's reuse that const across datetime module.

pgrx/src/datum/datetime_support/mod.rs
```diff
@@ -24,6 +24,9 @@ mod ops;
 
 pub use ctor::*;
 
+pub const USECS_PER_SEC: i64 = 1_000_000;
+pub const USECS_PER_DAY: i64 = pg_sys::SECS_PER_DAY as i64 * USECS_PER_SEC;
```
YohDeadfall, line 28: @eeeebbbbrrrr, why does `pg_sys` has `SECS_PER_DAY` generated but not `USECS_PER_SEC`?

eeeebbbbrrrr: hmm.  Probably b/c of how it's defined in `timestamp.h`:

```c
#define USECS_PER_SEC	INT64CONST(1000000)
```

Kinda doubt bindgen can see through even that simple of a #define.

workingjubilee:
I was very tired that week.

PR #1851, opened 2024-09-10
Added into_iter for anyarray

### LINKED ISSUE #1832: AnyArray should be iterable
Promised I'd file an issue about this in Discord.

Alternatively, my bug would be fixed if `Array<AnyElement>` were usable in arguments.

YohDeadfall:
Fixes #1832.

The implementation isn't what I want, but it adds `into_iter` for `&AnyArray` and I hope unblocking @thomcc. Later it would be better to get rid of `AnyArray` in its current state and use `Array<AnyElement>` instead, I will investigate that possibility. But for now that is how it is and without any breaking change.

### DIFF
--- a/pgrx/src/datum/anyarray.rs
+++ b/pgrx/src/datum/anyarray.rs
@@ -92,11 +92,11 @@ where
     Self: 'a,
 {
     type Item = Option<AnyElement>;
-    type IntoIter = AnyArrayIntoIterator<'a>;
+    type IntoIter = AnyArrayIterator<'a>;
 
     fn into_iter(self) -> Self::IntoIter {
         unsafe {
-            AnyArrayIntoIterator {
+            AnyArrayIterator {
                 inner: Array::<pg_sys::Datum>::from_polymorphic_datum(
                     self.datum(),
                     false,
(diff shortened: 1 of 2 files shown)


pgrx/src/datum/anyarray.rs
```diff
@@ -84,3 +86,53 @@ unsafe impl SqlTranslatable for AnyArray {
...
+                    self.oid(),
+                )
+                .map(|a| a.into_iter()),
+                typelem: pg_sys::get_element_type(self.oid()),
+            }
+        }
+    }
+}
+
+pub struct AnyArrayIntoIterator<'a> {
```
workingjubilee:
```suggestion
pub struct AnyArrayIterator<'a> {
```

YohDeadfall:
Should there be a consuming iterator or this one is enough?

workingjubilee:
honestly I expected IntoIterator would have been for AnyArray and be consuming, yeah.

workingjubilee:
Also, to be clear, `Array<'a, T>` has some complicated reasons why it's also Kinda Bad that are outside the scope of this PR, so I'm not too fussed right now.

workingjubilee:
We discussed this and this is probably fine as-is.

PR #1853, opened 2024-09-11
fix segfault converting a null `pg_sys::Datum` to a `String`

### LINKED ISSUE #1852: `String::from_datum()` of a null value led to a segmentation fault
The `FromDatum` implementation for `String` fails when a null value is passed.

Example: 

```
let val: pg_sys::Datum = pg_sys::Datum::null();
let is_null: bool = true;

unsafe {
    match String::from_datum(val, is_null) {   // <- this seg fault
        Some(_) => println!("Returned a value"),
        None => println!("Returned none"),
    };
}
```

In this example, the function should return `None` when a null value is passed. However, the current implementation does not properly handle the null case for String and crash.

Unlike other `from_polymorphic_datum()` implementations, the `String` implementation does not verify if the value is null before attempting to process it. You can see this in the code here:

https://github.com/pgcentralfoundation/pgrx/blob/5eadffb64d37ba08a5cc8060f1ed59153c6f8edd/pgrx/src/datum/from.rs#L400-L422

It used to work on v0.11.4. 
Fails now on v0.12.3. 
pg16

I think the issue was introduced on https://github.com/pgcentralfoundation/pgrx/commit/beb79011a377bb96c34fcc50ae2d652aa1419eb5

eeeebbbbrrrr:
This closes #1852, which reports a segfault when converting an honestly null `Datum` into a `String`.

Looks like this was introduced in https://github.com/pgcentralfoundation/pgrx/commit/beb79011a377bb96c34fcc50ae2d652aa1419eb5 and we didn't have test coverage around this specific case.

I quickly audited `src/datum/from.rs` and everything else there looks fine.

### DIFF
--- a/pgrx-tests/src/tests/from_into_datum_tests.rs
+++ b/pgrx-tests/src/tests/from_into_datum_tests.rs
@@ -43,4 +43,17 @@ mod tests {
         assert_eq!(cstr, expected);
         Ok(())
     }
+
+    #[pg_test]
+    fn null_string_is_none() {
+        let val: pg_sys::Datum = pg_sys::Datum::null();
+        let is_null: bool = true;
+
+        unsafe {
+            if let Some(_) = String::from_datum(val, is_null) {
+                // <- this seg fault
+                panic!("converted a null Datum into a valid string")
+            }
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


workingjubilee (changes requested):
"Fix issue #12090934509" is not an appropriate title.

eeeebbbbrrrr:
> "Fix issue #12090934509" is not an appropriate title.

that's why I used the actual issue number!

I'll fix it and the description once I see that CI passes.  don't worry.  ;)

PR #1854, opened 2024-09-11
use cc to generate cshim

usamoi:
It should enable cross-compiling if `cshim` feature is enabled. It's tested on my machine and it works correctly.


workingjubilee:
I am so glad I added that packaging check.

workingjubilee:
It catches only the weirdest, silliest issues, but man, pgrx sure does have plenty of those.

pgrx-bindgen/Cargo.toml
```diff
@@ -20,3 +20,4 @@ bindgen = { version = "0.70.1", default-features = false, features = ["experimen
 clang-sys = { version = "1", features = ["clang_6_0", "runtime"] }
 quote = "1.0.33"
 shlex = "1.3" # shell lexing, also used by many of our deps
+cc = "1.0"
```
workingjubilee: I know it doesn't matter.

compels me, tho'.
```suggestion
cc = "1.0"

```

workingjubilee (commented):
Thoughts, @eeeebbbbrrrr?

usamoi:
👀, @eeeebbbbrrrr ?

eeeebbbbrrrr (approved):

eeeebbbbrrrr:
I’m sorry. I looked at this when it came in and then… who knows!

I like it. I’ll let @workingjubilee press the merge button, but this LGTM.

eeeebbbbrrrr:
Oh. Does this add any new system dependencies,

I don’t guess it would but wanted to ask.

workingjubilee:
It should effectively reduce them since we no longer need to know about pgxs.mk to build the cshim.

eeeebbbbrrrr:
Hmm. Neat.

would be nice if we could drop the cshim entirely.

PR #1844, opened 2024-09-09
use bindgen to generate wrappers of static inline functions

usamoi:
It generates wrappers of static inline functions if `cshim` is enabled, so that we do not have to write wrappers by translating C to Rust and check if it's correct in all PG versions.

PR #1857, opened 2024-09-11
Update version to 0.12.4

eeeebbbbrrrr:
Welcome to pgrx v0.12.4.

First, and most importantly, it fixes a segfault when converting a NULL `pg_sys::Datum` into a Rust `String`.  It seems you'd need to go out of your way to cause the segfault, but nonetheless, it shouldn't happen.  A NULL Datum should convert to `Option::None` in all situations.

Secondly, @aykut-bozkurt has properly defined microseconds, which is critical for proper Time conversions.

From there, @usamoi enhanced our bindings generation such that we now auto-generate wrappers for any of Postgres' `static inline` functions in the included headers.  This is great as it eliminates our need to manually write these wrappers and it automatically exposes a lot more.

And thanks to @YohDeadfall `AnyArray` is now iterable!

As always, please upgrade using `cargo install cargo-pgrx --version 0.12.4 --locked`.  Then you can use `cargo pgrx upgrade` to update the dependencies in your extension crates.

## What's Changed

* use bindgen to generate wrappers of static inline functions by @usamoi in https://github.com/pgcentralfoundation/pgrx/pull/1844
* Unified argument names of Query methods by @YohDeadfall in https://github.com/pgcentralfoundation/pgrx/pull/1848
* Fixes wrong value for microsecs in a day by @aykut-bozkurt in https://github.com/pgcentralfoundation/pgrx/pull/1849
* fix segfault converting a null `pg_sys::Datum` to a `String` by @eeeebbbbrrrr in https://github.com/pgcentralfoundation/pgrx/pull/1853
* Added into_iter for anyarray by @YohDeadfall in https://github.com/pgcentralfoundation/pgrx/pull/1851

**Full Changelog**: https://github.com/pgcentralfoundation/pgrx/compare/v0.12.3...v0.12.4

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -317,7 +317,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-pgrx"
-version = "0.12.3"
+version = "0.12.4"
 dependencies = [
  "bzip2",
  "cargo-edit",
@@ -1487,7 +1487,7 @@ dependencies = [
 
 [[package]]
 name = "pgrx"
-version = "0.12.3"
+version = "0.12.4"
 dependencies = [
  "atomic-traits",
  "bitflags 2.4.2",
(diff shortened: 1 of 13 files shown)


YohDeadfall:
Forgot to mention that #1849 is critical too.

PR #1858, opened 2024-09-12
Made SPI query arguments type safe

YohDeadfall:


eeeebbbbrrrr:
`&[][..]` is not something you see every day.  Why is it that `&[]` doesn't work?

YohDeadfall:
I guess it's leftover from the time when I tried to play with generics and avoid a breaking change. That was required by the compiler for the type deduction, but can be reduced now.

eeeebbbbrrrr:
I appreciate your efforts on this area of pgrx, btw. 

Spi is incredibly hard to get right in C, and it’s not any easier in Rust.

eeeebbbbrrrr (approved):
This is **really** nice, @YohDeadfall.  Really nice.

Very clean.

Like @workingjubilee said somewhere, I think we'd want to hold this for v0.13, but I think this PR is a motivator for getting there quickly.

workingjubilee:
@YohDeadfall Can you resolve the conflicts?

YohDeadfall:
Sure! Actually, I just wanted to write you about this pull request too (:

workingjubilee (approved):
Thanks!

PR #1864, opened 2024-09-17
Allowed CStr to be used as SPI commands

YohDeadfall:
Made with love to those who doesn't like allocations ❤️

Since the Rust version 1.77 there's a possibility to use `c` and `cr` prefixes to make C string literals. That requires a little bit more characters to type, but reduces the number of allocations and therefore boosts `pgrx` extensions opted in. The feature isn't breaking due to being generic based.


pgrx/src/spi/query.rs
```diff
@@ -41,32 +41,84 @@ pub trait Query<'conn>: Sized {
     ) -> SpiResult<SpiCursor<'conn>>;
 }
 
-impl<'conn> Query<'conn> for &String {
-    type Arguments = Option<Vec<(PgOid, Option<pg_sys::Datum>)>>;
-
-    fn execute(
+/// A trait representing a query which can be prepared.
+pub trait PreparableQuery<'conn>: Query<'conn> {
```
workingjubilee, line 45: This probably seems obvious but can you explain how this distinguishes from Query?

YohDeadfall: To distinguish from the prepared statements which should not be used by the prepare methods of the SPI client.

workingjubilee: Hm, not entirely sure what the benefit is?

workingjubilee: This is probably obvious to you, it's just been ages since I've looked at the Query API and my main dread is it being cemented further, because I would rather it... not exist, tbh. Not in its current form anyway. It's... overly cutesy.

YohDeadfall: The problem that trait solves is prohibiting `PreparedStatement<'conn>` from being used with `prepare` and `prepare_mut` methods of the `SqlClient<'conn>`, so the following code will result in a compiler error:

```rust
let conn = SpiConnection::connect()?;
let client = conn.client();

let once = client.prepare(c"SELECT 1", None)?;
let twice = client.prepare(once, None);
//                 ------- ^^^^ the trait `PreparableQuery<'_>` is not implemented for `PreparedStatement<'_>`
//                              |
//                              required by a bound introduced by this call
```

workingjubilee: Ah.

Can you add a UI test to verify it has a compiler error then?

YohDeadfall: I played with the CI and observed that my UI test fails only for PostgreSQL versions 12, 13 and 14, but not for others. That's totally weird. See #1870.

workingjubilee: ...huh.

workingjubilee: ...oh, I remember. They introduced a type named String in PostgreSQL. Can you defer the test to opening an issue instead?

YohDeadfall: This one, right? https://doxygen.postgresql.org/structString.html

workingjubilee: Yes.

workingjubilee:
Thank you~!

PR #1867, opened 2024-09-18
Use `petgraph::algo::tarjan_scc` to order the schema graph

### LINKED ISSUE #1866: `cargo pgrx schema` output isn't repeatable
On a large extension (ParadeDB's `pg_search` extension), I've found that consecutive runs of `cargo pgrx schema`, with no code changes, produce differently-ordered schemas.  The graph relationship is still correct, but the edges don't seem to emit in the same order every time.  

(I've made sure the few `extension_sql!()` blocks `pg_search` has all have proper `requires = [...]` arrays, just to make sure the graph gods aren't making different decisions for those, which might then upset the whole graph.)

I swear at one point in history the schema generator *did* produce consistent results.

I've been looking through the code, along with diffs in the commit history, and I don't see anything that would introduce randomness.  We make sure to sort the list of sql entities and process them in order when building the graph, and we do a `toposort` to walk the graph to generate the sql.  It's at that point the nodes come out in different orders each time.

@workingjubilee you got any ideas?  I'm happy to fix this (assuming we can?), but I'm kinda stumped on where else to look.

Ultimately, I'd like to be able to do:

```shell
$ diff old-schema.sql new-schema.sql > diff.sql
```

And have a reasonable diff generated, or no diff at all if in fact there were no changes to the schema structure between "old" and "new".

eeeebbbbrrrr:
This seems to give a consistent ordering of the schema through consecutive runs of `cargo pgrx schema`.

It also allows `diff` to produce nice diffs when the schema changes compared to a previous version.

Tossing this up here to see if CI blows up on any of the tests or examples due to the schema being out of order.

I've tested this ordering with the `pg_search` extension (which is quite large) and it still generates a perfectly usable schema.  The same one from run to run, even!

It also adds comment markers around each connected block of sql objects, just to help better visualize the connectedness of the schema.

Closes #1866

### DIFF
--- a/pgrx-sql-entity-graph/src/pgrx_sql.rs
+++ b/pgrx-sql-entity-graph/src/pgrx_sql.rs
@@ -404,6 +404,22 @@ impl PgrxSql {
     pub fn to_sql(&self) -> eyre::Result<String> {
         let mut full_sql = String::new();
 
+        // NB:  A properly we'd *like* to maintain is that the schema generator outputs
+        // consistent results from run-to-run when there are no changes to the schema.
+        // This is to improve change detecting using simple tools like `diff`.
+        //
+        // Historically, we used [`petgraph::algo:toposort`] but its ordering is not at all
+        // consistent.
+        //
+        // [`petgraph::algo::tarjan_scc`] appears to be consistent, although it's not exactly
+        // clear if this is due to an implementation detail or specifics of the algorithm itself.
+        // (I, eeeebbbbrrrr, am not a graph theory expert)
+        //
+        // In any event, if in the future schema generation stops being consistent, this is the
+        // place to look.
+        //
+        // We have no tests around this as it's really just a property we'd like to have, and
+        // it does seem ensuring it is a bit of black magic.
         for nodes in petgraph::algo::tarjan_scc(&self.graph).iter().rev() {
             let mut inner_sql = Vec::with_capacity(nodes.len());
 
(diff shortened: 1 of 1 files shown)


workingjubilee (approved):

pgrx-sql-entity-graph/src/pgrx_sql.rs
```diff
@@ -403,19 +403,27 @@ impl PgrxSql {
...
         let mut full_sql = String::new();
-        for step_id in petgraph::algo::toposort(&self.graph, None).map_err(|e| {
-            eyre!("Failed to toposort SQL entities, node with cycle: {:?}", self.graph[e.node_id()])
-        })? {
-            let step = &self.graph[step_id];
 
-            let sql = step.to_sql(self)?;
-            full_sql.push_str(sql.trim_matches('\n'));
+        for nodes in petgraph::algo::tarjan_scc(&self.graph).iter().rev() {
+            let mut inner_sql = Vec::with_capacity(nodes.len());
```
workingjubilee, line 424: We should probably comment that using this algo is load-bearing for run-to-run ordering.

workingjubilee: ( Especially because as far as we can tell this _isn't_ necessarily the case from executing `tarjan_scc`, it just is almost always so in the current implementation, so the comment sets the reader up for a justifiable ["Are you sure about that?"](https://www.youtube.com/watch?v=oHC1230OpOg) )

eeeebbbbrrrr: done

eeeebbbbrrrr:
Thanks for the merge and your time!

PR #1876, opened 2024-09-24
Added non-atomic SPI connection

YohDeadfall:
Here's my take on making non-atomic SPI connections, but it lacks tests as I don't understand how to properly make them without dealing with portals directly. If I'm not wrong there were tests using a client to issue requests to the server with the required extensions, so it should be like that then.

### DIFF
--- a/pgrx/src/spi.rs
+++ b/pgrx/src/spi.rs
@@ -21,8 +21,8 @@ mod client;
 mod cursor;
 mod query;
 mod tuple;
-pub use client::SpiClient;
 use client::SpiConnection;
+pub use client::{SpiClient, SpiTransaction};
 pub use cursor::SpiCursor;
 pub use query::{OwnedPreparedStatement, PreparedStatement, Query};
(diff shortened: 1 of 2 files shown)


eeeebbbbrrrr:
What's the use case for non-atomic SPI?

YohDeadfall:
It's used primarily if not only by language handlers which take that information from the current call. The problem I see is that the native API allows opting-in into non-atomic connections using `SPI_connect_ext`, but the provided implementation is less flexible, but more sound because with atomic connections no transaction control functions are allowed.

dpxcc:
> What's the use case for non-atomic SPI?

Non-atomic SPI can be quite useful for stored procedures and background workers

By the way, I'm curious why pgrx doesn't offer more native support for stored procedures

YohDeadfall:
> By the way, I'm curious why pgrx doesn't offer more native support for stored procedures

It's purely contributor based now, no company owns it or sponsors it.

eeeebbbbrrrr:
This has been a Draft for quite awhile now.  I'm closing this simply to keep our PR list tidy.  If you decide to come back to this against the current `develop` branch, please open a new PR.

PR #1878, opened 2024-09-25
Trait TypmodInOutFuncs for input(input, oid, typmod) function

KenjiBrown:
https://www.postgresql.org/docs/current/sql-createtype.html
> The input function can be declared as taking one argument of type cstring, or as taking three arguments of types cstring, oid, integer. The first argument is the input text as a C string, the second argument is the type's own OID (except for array types, which instead receive their element type's OID), and the third is the typmod of the destination column, if known (-1 will be passed if not). 

This trait allows for the input function to take three arguments...

### DIFF
--- a/pgrx-macros/src/lib.rs
+++ b/pgrx-macros/src/lib.rs
@@ -769,6 +769,7 @@ enum Animal {
 Optionally accepts the following attributes:
 
 * `inoutfuncs(some_in_fn, some_out_fn)`: Define custom in/out functions for the type.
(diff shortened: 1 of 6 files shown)


KenjiBrown:
Still work in progress.

eeeebbbbrrrr:
Hi @KenjiBrown.  I'm going to close this PR.  Progress here seems to have stalled.  If you decide to come back to working on this, feel free to reopen.

PR #1879, opened 2024-09-25
Added `oids_of!` macro

YohDeadfall:
That's an improvement for SPI making it less cumbersome. Just to feel the difference look at the tests. Who would really want to write something like `PgBuiltInOids::INT4OID.oid()` instead of just telling a Rust type?

I will change the behavior of this macro a bit when you will be ready for shipping `0.13`, so it won't produce any vector but a slice.

### DIFF
--- a/pgrx/src/datum/mod.rs
+++ b/pgrx/src/datum/mod.rs
@@ -217,7 +217,10 @@ pub trait PostgresType {}
 /// ```
 #[macro_export]
 macro_rules! oids_of {
-    ($($t:path),* $(,)?) => ({
-        [$(::pgrx::pg_sys::PgOid::from(<$t>::type_oid())),*]
-    });
+    () =>(
+        [$crate::pg_sys::PgOid::Invalid; 0]
+    );
+    ($($t:path),+ $(,)?) => (
+        [$($crate::pg_sys::PgOid::from(<$t>::type_oid())),*]
+    );
 }
--- a/pgrx/src/datum/mod.rs
+++ b/pgrx/src/datum/mod.rs
@@ -204,3 +204,20 @@ impl<'src, T: IntoDatum> From<T> for DatumWithOid<'src> {
 /// A tagging trait to indicate a user type is also meant to be used by Postgres
 /// Implemented automatically by `#[derive(PostgresType)]`
 pub trait PostgresType {}
+
+/// Creates a [`Vec<pg_sys::Oid>`] containing identifiers of the provded types.
+///
+/// # Examples
+///
+/// ```
+/// use pgrx::{oids_of, datum::IntoDatum};
+/// let v = oids_of![i32, f64];
+/// assert_eq!(v[0], i32::type_oid().into());
+/// assert_eq!(v[1], f64::type_oid().into());
+/// ```
+#[macro_export]
+macro_rules! oids_of {
+    ($($t:path),* $(,)?) => ({
+        [$(::pgrx::pg_sys::PgOid::from(<$t>::type_oid())),*]
+    });
+}
--- a/pgrx/src/datum/mod.rs
+++ b/pgrx/src/datum/mod.rs
@@ -205,12 +205,13 @@ impl<'src, T: IntoDatum> From<T> for DatumWithOid<'src> {
 /// Implemented automatically by `#[derive(PostgresType)]`
 pub trait PostgresType {}
 
-/// Creates a [`Vec<pg_sys::Oid>`] containing identifiers of the provded types.
+/// Creates an array of [`pg_sys::Oid`] with the OID of each provided type
 ///
 /// # Examples
 ///
 /// ```
 /// use pgrx::{oids_of, datum::IntoDatum};
+///
 /// let oids = oids_of![i32, f64];
 /// assert_eq!(oids[0], i32::type_oid().into());
 /// assert_eq!(oids[1], f64::type_oid().into());
@@ -223,6 +224,7 @@ pub trait PostgresType {}
 #[macro_export]
 macro_rules! oids_of {
     () =>(
+        // avoid coercions to an ambiguously-typed array or slice
         [$crate::pg_sys::PgOid::Invalid; 0]
     );
     ($($t:path),+ $(,)?) => (
--- a/pgrx/src/datum/mod.rs
+++ b/pgrx/src/datum/mod.rs
@@ -211,9 +211,14 @@ pub trait PostgresType {}
 ///
 /// ```
 /// use pgrx::{oids_of, datum::IntoDatum};
-/// let v = oids_of![i32, f64];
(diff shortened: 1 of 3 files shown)


pgrx/src/datum/mod.rs
```diff
@@ -149,3 +149,20 @@ impl<'src> Datum<'src> {
...
+/// ```
+/// use pgrx::{oids_of, datum::IntoDatum};
+/// let v = oids_of![i32, f64];
+/// assert_eq!(v[0], i32::type_oid().into());
+/// assert_eq!(v[1], f64::type_oid().into());
+/// ```
+#[macro_export]
+macro_rules! oids_of {
+    ($($t:path),* $(,)?) => ({
+        vec![$(::pgrx::pg_sys::PgOid::from(<$t>::type_oid())),*]
```
workingjubilee: I don't think this should include a `vec![]` when it could instead be invoked as `vec![oids_of!(...)]`

this would allow constructing arrays also, or other structures.

YohDeadfall: Yes, that's the plan when I get rid of `Option<Vec<PgOid>>` because now it's too wordy.

workingjubilee: ...if that's the plan, then?

YohDeadfall: Yeah, I will remove `vec!` from here and will change the methods of `SpiClient` to use slices as soon as you that the breaking change time has come (:

workingjubilee: I'd rather not add this then just break it again.

YohDeadfall: Let it be non-breaking then.

workingjubilee (changes requested):
I don't think this has the right semantics right now. It's not clear to me that `oids_of!()` should be an `InvalidOid`, instead of, say, a compile-time error. Or an empty array.

workingjubilee:
I think we should ship a non-plural `oid_of!` (I'd prefer an fn if we can swing it, tbh, but I'll settle) first and tinker with it a bit before jumping to generating arrays.

YohDeadfall:
If something isn't an `IntoDatum` then a compiler error will appear, so the value of each element purely depends on the implementation of each type used there. `InvalidOid` is just to make an empty version of `oids_of!` to be strongly typed and not to depend on the context.

Let's say, I don't want `oids_of!` to give you an empty array of integers like in the following case:

```rust
let _: &[i32] = oids_of![];
```

That's why there's `[$crate::pg_sys::PgOid::Invalid; 0]`.

YohDeadfall:
> I think we should ship a non-plural `oid_of!`

You have it already, `IntoDatum::type_oid()`, and it's a function. I thought about making a non-plural version of the macro, but I see no real use case for it.

workingjubilee:
tbh, that's kind of the problem.

A macro creates a false impression this will resolve statically.

That fn produces runtime lookups.

YohDeadfall:
It depends, actually. For built-in types, it's static, but for custom types it isn't.

workingjubilee:
I realize that for the existing inputs, it's somewhat incoherent to use. But the reason to make it a zero-sized array, effectively, is to make a macro-call to it... possibly itself from a macro-expansion... valid anywhere we ask for `&[pg_sys::Oid]`, without having to reason about the exact input. For a given input, sure, they might need to do something like
```rust
let oids = oids_of![$($types,)];
oids.is_empty().then_some(&oids)
```
but this enables that.

Basically, we shouldn't overfit it to the existing use-cases.

YohDeadfall:
Sorry, not following. Could you reword it a bit so I can get what I should do?

workingjubilee:
I would still prefer zero arguments to evaluate to a zero-sized slice, or to [`compile_error`](https://doc.rust-lang.org/std/macro.compile_error.html)

YohDeadfall:
Isn't `[$crate::pg_sys::PgOid::Invalid; 0]` a zero sized array?

workingjubilee:
...hm.

I somehow managed to misread this code repeatedly, I suppose.

pgrx/src/datum/mod.rs
```diff
@@ -149,3 +149,23 @@ impl<'src> Datum<'src> {
 /// A tagging trait to indicate a user type is also meant to be used by Postgres
 /// Implemented automatically by `#[derive(PostgresType)]`
 pub trait PostgresType {}
+
+/// Creates a [`Vec<pg_sys::Oid>`] containing identifiers of the provded types.
```
workingjubilee:
```suggestion
/// Creates an array of [`pg_sys::Oid`] with the OID of each provided type
```

pgrx/src/datum/mod.rs
```diff
@@ -149,3 +149,23 @@ impl<'src> Datum<'src> {
...
+///
+/// # Examples
+///
+/// ```
+/// use pgrx::{oids_of, datum::IntoDatum};
```
workingjubilee, line 213:
```suggestion
/// use pgrx::{oids_of, datum::IntoDatum};
///
```

pgrx/src/datum/mod.rs
```diff
@@ -149,3 +149,23 @@ impl<'src> Datum<'src> {
...
+/// let v = oids_of![i32, f64];
+/// assert_eq!(v[0], i32::type_oid().into());
+/// assert_eq!(v[1], f64::type_oid().into());
+/// ```
```
workingjubilee, line 223: mostly for the last line, because I'm silly
```suggestion
///
/// // the usual conversions or coercions are available
/// let oid_vec: Vec<pg_sys::Oid> = oids_of![i8, i16].to_vec();
/// let no_oid: &[pg_sys::Oid] = &oids_of![];
/// assert_eq!(no_oid.len(), 0);
/// ```
```

pgrx/src/datum/mod.rs
```diff
@@ -149,3 +149,23 @@ impl<'src> Datum<'src> {
...
+#[macro_export]
+macro_rules! oids_of {
+    () =>(
+        [$crate::pg_sys::PgOid::Invalid; 0]
```
workingjubilee, line 228:
```suggestion
        // avoid coercions to an ambiguously-typed array or slice
        [$crate::pg_sys::PgOid::Invalid; 0]
```

workingjubilee (approved):
Sorry, I managed to not give this proper attention the past times I glanced at it. Some nits and a rebase should make this good.

pgrx/src/datum/mod.rs
YohDeadfall: Not sure about it, so applied it partly. I made it mostly to demonstrate that the resulting type of `oids_of!` is known even for an empty set of types.

workingjubilee (approved):

PR #1880, opened 2024-09-26
`#include "commands/extension.h"` and `#include "parser/parse_collate.h"`

eeeebbbbrrrr:
A few more `#include`s that I've noticed folks in the wild have declared their own `extern "C" {}` blocks for.


eeeebbbbrrrr:
I approve this PR.

PR #1881, opened 2024-09-26
Move from pg17rc1 to pg17ga

eeeebbbbrrrr:
Not sure if v17 is actually out on the distro repos yet, so lets see what CI does

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -164,7 +164,7 @@ jobs:
           echo ""
 
       - name: Setup release Postgres apt repo
-        if: matrix.postgres != '17'
+        if: matrix.postgres != '18'
         run: |
           sudo apt-get install -y wget gnupg
           sudo sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list'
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -164,15 +164,15 @@ jobs:
           echo ""
 
       - name: Setup release Postgres apt repo
-        if: matrix.postgres != '17'
+        if: matrix.postgres != '18'
```
eeeebbbbrrrr, line 167: queue this up for the next version

.github/workflows/tests.yml
```diff
@@ -164,15 +164,15 @@ jobs:
...
         run: |
           sudo apt-get install -y wget gnupg
           sudo sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list'
           wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | sudo apt-key add -
           sudo apt-get update -y -qq --fix-missing
 
       - name: Setup beta Postgres apt repo
-        if: matrix.postgres == '17'
+        if: matrix.postgres == '18'
```
eeeebbbbrrrr, line 175: queue this up for the next version.  The postgres folks seem to put their beta packages in the same place, so we can just carry this along with us so we don't have to re-learn it again next year.

eeeebbbbrrrr:
Of course.  I'll try again later today

eeeebbbbrrrr:
It's still not out on the repo at http://apt.postgresql.org/pub/repos/apt/dists/jammy-pgdg/17/binary-amd64/.

Which, I don't fully understand.  The directories are there but there's no usable files.  Guess whoever does the needful hasn't yet.

~~I'll come back to this on Friday Sep 27.~~

PR #1884, opened 2024-09-27
Nullable is Copy when T is Copy

workingjubilee:
Discovered we missed this while working on FlatArray.


workingjubilee:
cc @NotGyro There wasn't a reason we withheld this impl, right?

PR #1886, opened 2024-09-27
Macro based IntoDatum for string types

YohDeadfall:
The pull request follows two goals:
* don't repeat yourself,
* implement `IntoDatum` for `&CString` to mirror implementation of the trait for `&String`.

### DIFF
--- a/pgrx/src/datum/into.rs
+++ b/pgrx/src/datum/into.rs
@@ -15,7 +15,12 @@
 use crate::{pg_sys, rust_regtypein, set_varsize_4b, PgBox, PgOid, WhoAllocated};
 use core::fmt::Display;
 use pgrx_pg_sys::panic::ErrorReportable;
-use std::{any::Any, ptr::addr_of_mut, str};
+use std::{
+    any::Any,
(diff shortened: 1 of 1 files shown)


YohDeadfall:
@workingjubilee, I need your opinion on that since I would like to have that in my pull request switch to `Datum<'mcx>` and adjust less code during rebase.

workingjubilee:
on which?

YohDeadfall:
I meant this whole pull request. The thing is that my other unpublished change touches all `IntoDatum` implementations, so I would like to clean up the current code first, have less work later, and avoid potential rebase too.

workingjubilee:
@YohDeadfall Would it work to implement the CStr variations on `Deref<Target = CStr>`?

YohDeadfall:
Yes, that was my first thought, but at the same time a custom type might use that `T` which is `Deref<CStr>` too. That can end up in conflicts, and that's my concern.

workingjubilee:
Hmm.

workingjubilee:
Eh, okay. This is as we currently have and we can make the deref-based change later if we want.

PR #1887, opened 2024-09-27
Removed extra allocation from `IntoDatum for char`

YohDeadfall:


workingjubilee:
It still has to allocate because it needs to palloc the text datum.

tbh I am dubious of this impl but we have it, so.

YohDeadfall:
Yeah, `palloc` will be there forever and comes from the `&str` handel, but `malloc` isn't there anymore.

YohDeadfall:
Thanks for changing the title, by the way, and the raced concern. I saw that thing in the standard library, but I cannot find it once again, and the code is wrong, indeed. It should also use `len_utf8` to get the right number of bytes.

YohDeadfall:
I guess I thought about C strings at that moment, so I completely forgot about the length. But if it would be a `char` to a C string handling then it still had a bug due to the buffer size not being designed for a terminal byte. Anyway, there's a fix provided in #1889 with tests.

PR #1890, opened 2024-09-28
Support refs to unsized SqlTranslatable types

workingjubilee:
This properly separates the question of "does this have an SQL representation?" from "is this a viable function argument?" While it has no immediate benefit aside from as a minor cleanup, it will prove useful in supporting types and functions where an unsized type is used as a generic. Such code will return slice references and the like.


eeeebbbbrrrr (approved):

workingjubilee:
Thanks!

PR #1891, opened 2024-09-29
Add `BorrowDatum` for unsizing borrows of datums

workingjubilee:
Add a trait that allows borrowing Datums from various places, like arguments and arrays, instead of always taking them by-value. This is most important for unsizing borrows, as the sized kind are a simple `.cast()` away.

The motivating rationale is to enable new APIs in the near future like FlatArray and Text. It also brings in some more uniformity because of the blanket-impl of ArgAbi for `&T`. This means that various types that _are_ pass-by-reference may now actually be taken by reference from their Datum, instead of copying them into the Rust function! This is more important for unsized types, but for various reasons, right now it only takes over the ArgAbi of one unsized type: CStr.

### DIFF
--- a/pgrx/src/callconv.rs
+++ b/pgrx/src/callconv.rs
@@ -266,11 +266,11 @@ argue_from_datum! { 'fcx; Date, Interval, Time, TimeWithTimeZone, Timestamp, Tim
 argue_from_datum! { 'fcx; AnyArray, AnyElement, AnyNumeric }
 argue_from_datum! { 'fcx; Inet, Internal, Json, JsonB, Uuid, PgRelation }
 argue_from_datum! { 'fcx; pg_sys::BOX, pg_sys::ItemPointerData, pg_sys::Oid, pg_sys::Point }
-argue_from_datum! { 'fcx; &'fcx str, &'fcx CStr, &'fcx [u8] }
+argue_from_datum! { 'fcx; &'fcx str, &'fcx [u8] }
 
 unsafe impl<'fcx, T> ArgAbi<'fcx> for &'fcx T
 where
-    T: BorrowDatum,
+    T: ?Sized + BorrowDatum,
 {
     unsafe fn unbox_arg_unchecked(arg: Arg<'_, 'fcx>) -> &'fcx T {
         let ptr: *mut u8 = match T::PASS {
--- a/pgrx/src/callconv.rs
+++ b/pgrx/src/callconv.rs
@@ -274,9 +274,8 @@ where
 {
     unsafe fn unbox_arg_unchecked(arg: Arg<'_, 'fcx>) -> &'fcx T {
         let ptr: *mut u8 = match T::PASS {
-            Some(PassBy::Ref) => arg.2.value.cast_mut_ptr(),
-            Some(PassBy::Value) => ptr::addr_of!(arg.0.raw_args()[arg.1].value).cast_mut().cast(),
-            _ => todo!(),
+            PassBy::Ref => arg.2.value.cast_mut_ptr(),
+            PassBy::Value => ptr::addr_of!(arg.0.raw_args()[arg.1].value).cast_mut().cast(),
         };
         unsafe { &*T::point_from(ptr) }
     }
--- a/pgrx/src/callconv.rs
+++ b/pgrx/src/callconv.rs
@@ -277,7 +277,7 @@ where
             PassBy::Ref => arg.2.value.cast_mut_ptr(),
             PassBy::Value => ptr::addr_of!(arg.0.raw_args()[arg.1].value).cast_mut().cast(),
         };
-        unsafe { &*T::point_from(ptr) }
+        unsafe { T::borrow_unchecked(ptr) }
     }
 
     unsafe fn unbox_nullable_arg(arg: Arg<'_, 'fcx>) -> Nullable<Self> {
@@ -288,7 +288,7 @@ where
         if arg.is_null() || ptr.is_null() {
             Nullable::Null
         } else {
-            unsafe { Nullable::Valid(&*T::point_from(ptr)) }
+            unsafe { Nullable::Valid(T::borrow_unchecked(ptr)) }
         }
     }
 }
--- a/pgrx/src/callconv.rs
+++ b/pgrx/src/callconv.rs
@@ -277,18 +277,20 @@ where
             PassBy::Ref => arg.2.value.cast_mut_ptr(),
             PassBy::Value => ptr::addr_of!(arg.0.raw_args()[arg.1].value).cast_mut().cast(),
         };
-        unsafe { T::borrow_unchecked(ptr) }
+        unsafe {
+            let ptr = ptr::NonNull::new_unchecked(ptr);
+            T::borrow_unchecked(ptr)
+        }
     }
 
     unsafe fn unbox_nullable_arg(arg: Arg<'_, 'fcx>) -> Nullable<Self> {
-        let ptr: *mut u8 = match T::PASS {
+        let ptr: Option<ptr::NonNull<u8>> = NonNull::new(match T::PASS {
             PassBy::Ref => arg.2.value.cast_mut_ptr(),
             PassBy::Value => ptr::addr_of!(arg.0.raw_args()[arg.1].value).cast_mut().cast(),
-        };
-        if arg.is_null() || ptr.is_null() {
-            Nullable::Null
-        } else {
-            unsafe { Nullable::Valid(T::borrow_unchecked(ptr)) }
+        });
+        match (arg.is_null(), ptr) {
+            (true, _) | (false, None) => Nullable::Null,
+            (false, Some(ptr)) => unsafe { Nullable::Valid(T::borrow_unchecked(ptr)) },
         }
     }
 }
--- a/pgrx/src/callconv.rs
+++ b/pgrx/src/callconv.rs
@@ -10,14 +10,15 @@
 #![deny(unsafe_op_in_unsafe_fn)]
 //! Helper implementations for returning sets and tables from `#[pg_extern]`-style functions
 
-use crate::datum::Datum;
 use crate::datum::{
     AnyArray, AnyElement, AnyNumeric, Date, FromDatum, Inet, Internal, Interval, IntoDatum, Json,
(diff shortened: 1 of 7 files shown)


pgrx/src/datum/borrow.rs
```diff
@@ -0,0 +1,104 @@
...
+        unsafe { <Self as BorrowDatum>::point_from(ptr.cast()) }
+    }
+
+    /// Optimization for borrowing the referent
+    unsafe fn borrow_unchecked<'dat>(ptr: ptr::NonNull<u8>) -> &'dat Self {
+        unsafe { Self::point_from(ptr).as_ref() }
+    }
+}
+
+macro_rules! borrow_by_value {
```
workingjubilee: `borrow_by_value` is a bit of a misnomer, come to think of it.

pgrx/src/datum/borrow.rs
```diff
@@ -0,0 +1,104 @@
...
+                    ptr.cast()
+                }
+            }
+        )*
+    }
+}
+
+borrow_by_value! {
+    i8, i16, i32, i64, bool, f32, f64, pg_sys::Oid, Date, Time, Timestamp, TimestampWithTimeZone
+}
```
workingjubilee: This needs more tests that specifically exercise BorrowDatum on these "by-value" cases, and not just its interaction with FlatArray or other varlenas. Varlenae?

pgrx/src/callconv.rs
```diff
@@ -265,7 +266,34 @@ argue_from_datum! { 'fcx; Date, Interval, Time, TimeWithTimeZone, Timestamp, Tim
 argue_from_datum! { 'fcx; AnyArray, AnyElement, AnyNumeric }
 argue_from_datum! { 'fcx; Inet, Internal, Json, JsonB, Uuid, PgRelation }
 argue_from_datum! { 'fcx; pg_sys::BOX, pg_sys::ItemPointerData, pg_sys::Oid, pg_sys::Point }
-argue_from_datum! { 'fcx; &'fcx str, &'fcx CStr, &'fcx [u8] }
+argue_from_datum! { 'fcx; &'fcx str, &'fcx [u8] }
```
workingjubilee, line 273:
```suggestion
// We could use the upcoming impl of ArgAbi for `&'fcx T where T: ?Sized + BorrowDatum`
// to support these types by implementing BorrowDatum for them also, but we reject this.
// It would greatly complicate other users of BorrowDatum like FlatArray, which want all impls
// of BorrowDatum to return a borrow of the entire pointee's len.
argue_from_datum! { 'fcx; &'fcx str, &'fcx [u8] }
```

pgrx/src/datum/borrow.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// # Safety
+    /// - This must be correctly invoked for the pointee type, as it may deref and read one or more
+    ///   bytes in its implementation in order to read the inline metadata and unsize the type.
+    ///
+    /// ## For Implementers
+    /// Reading the **first** byte pointed to is permitted if `T::PASS = PassBy::Ref`, assuming you
+    /// are implementing a varlena type. As the other dynamic length type, CStr also does this.
+    /// This function
+    /// - must NOT mutate the pointee
+    /// - must return a pointer with provenance to the entire allocation
```
workingjubilee:
```suggestion
    /// - must point to the entire datum's length (`size_of_val` must not lose bytes)
```

pgrx/src/datum/borrow.rs
workingjubilee:
```suggestion
macro_rules! impl_borrow_fixed_len {
```

pgrx/src/datum/borrow.rs
```diff
@@ -0,0 +1,104 @@
...
+
+                unsafe fn point_from(ptr: ptr::NonNull<u8>) -> ptr::NonNull<Self> {
+                    ptr.cast()
+                }
+            }
+        )*
+    }
+}
+
+borrow_by_value! {
```
workingjubilee:
```suggestion
impl_borrow_fixed_len! {
```

pgrx/src/callconv.rs
YohDeadfall: Do you plan to change `ArgAbi`?

pgrx/src/datum/borrow.rs
```diff
@@ -0,0 +1,104 @@
+#![deny(unsafe_op_in_unsafe_fn)]
+use super::*;
+use crate::layout::PassBy;
+use core::{ffi, mem, ptr};
```
YohDeadfall, line 4: Would it be better to use `std` everywhere since the whole crate isn't `no_std`?

pgrx/src/callconv.rs
```diff
@@ -265,7 +266,38 @@ argue_from_datum! { 'fcx; Date, Interval, Time, TimeWithTimeZone, Timestamp, Tim
...
+            PassBy::Ref => arg.2.value.cast_mut_ptr(),
+            PassBy::Value => ptr::addr_of!(arg.0.raw_args()[arg.1].value).cast_mut().cast(),
+        };
+        unsafe {
+            let ptr = ptr::NonNull::new_unchecked(ptr);
+            T::borrow_unchecked(ptr)
+        }
+    }
+
+    unsafe fn unbox_nullable_arg(arg: Arg<'_, 'fcx>) -> Nullable<Self> {
```
YohDeadfall, line 290: `Nullable` is a kind of `Option`, and methods returning an option and their unchecked variants have the same names not counting the suffix. I would move in that direction.

pgrx/src/callconv.rs
```diff
@@ -265,7 +266,38 @@ argue_from_datum! { 'fcx; Date, Interval, Time, TimeWithTimeZone, Timestamp, Tim
 argue_from_datum! { 'fcx; AnyArray, AnyElement, AnyNumeric }
 argue_from_datum! { 'fcx; Inet, Internal, Json, JsonB, Uuid, PgRelation }
 argue_from_datum! { 'fcx; pg_sys::BOX, pg_sys::ItemPointerData, pg_sys::Oid, pg_sys::Point }
-argue_from_datum! { 'fcx; &'fcx str, &'fcx CStr, &'fcx [u8] }
```
YohDeadfall, line 268: Why `CStr` cannot be borrowed?

YohDeadfall (commented):
Cool, cool, I like it even when I'm not familiar with some parts of `pgrx`. I will take a closer look tomorrow when I come to the keyboard.

pgrx/src/callconv.rs
workingjubilee: It is removed from this set because it is now covered by the blanket impl of `ArgAbi<'fcx> for &'fcx T where T: ?Sized + BorrowDatum`, because we implement `BorrowDatum` for it here:

https://github.com/pgcentralfoundation/pgrx/blob/7c19edc15802cb462e87362d1cba7625313c454b/pgrx/src/datum/borrow.rs#L88-L104

pgrx/src/datum/borrow.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// This version allows you to assume the pointer is aligned to, and readable for, 4 bytes.
+    /// This optimization is not required. When in doubt, avoid implementing it, and rely on your
+    /// `point_from` implementation alone.
+    ///
+    /// # Safety
+    /// - This must be correctly invoked for the pointee type, as it may deref.
+    /// - This must be 4-byte aligned!
+    unsafe fn point_from_align4(ptr: ptr::NonNull<u32>) -> ptr::NonNull<Self> {
+        debug_assert!(ptr.is_aligned());
+        unsafe { <Self as BorrowDatum>::point_from(ptr.cast()) }
```
YohDeadfall: Why does that line use a qualified method call and a few lines below it isn't the case? And which one is actually better in general/for `std` contributions?

workingjubilee: quantity of paranoia I am subjectively experiencing at the time, mostly

pgrx/src/callconv.rs
workingjubilee: I haven't determined yet if I want the base to be a Result type.

pgrx/src/datum/borrow.rs
workingjubilee: @YohDeadfall The main reason I use the fully-qualified form a lot is that this crate has a lot of macros, and in macros, you have to be wary of this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7c81475b351b1ee39cc8cc6be3e330fa

There's some clarity benefit sometimes too, but

pgrx/src/datum/borrow.rs
workingjubilee: Ideally the whole crate would be `no_std`, probably.

workingjubilee:
Current main unknown: does this trait need a lifetime in order to provide a correct API...? Or should I just enforce that in interfaces like `Datum::borrow_as`?

YohDeadfall (approved):

pgrx/src/datum/borrow.rs
workingjubilee: I couldn't find a good way to test this because it requires either FnCall support or SPI support, and I don't want to extend FromDatum + IntoDatum to cover these cases yet.

workingjubilee: nevermind

pgrx/src/datum/borrow.rs
```diff
@@ -0,0 +1,128 @@
...
+    /// - This must be correctly invoked for the pointee type, as it may deref and read one or more
+    ///   bytes in its implementation in order to read the inline metadata and unsize the type.
+    /// - This must be invoked with a pointee initialized for the dynamically specified length.
+    ///
+    /// ## For Implementers
+    /// Reading the **first** byte pointed to is permitted if `T::PASS = PassBy::Ref`, assuming you
+    /// are implementing a varlena type. As the other dynamic length type, CStr also does this.
+    /// This function
+    /// - must NOT mutate the pointee
+    /// - must point to the entire datum's length (`size_of_val` must not lose bytes)
```
workingjubilee, line 40: It may be fine if we relax this requirement eventually.

PR #1892, opened 2024-09-30
Removed an extra alloc in jsonb's IntoDatum

YohDeadfall:


YohDeadfall:
It's not a case since `CString::new` uses specializations under the hood which turns an incoming `String` into a `Vec<u8>` and if it's valid just appends a null byte at the end.

PR #1893, opened 2024-09-30
Improved messages for json errors

YohDeadfall:
Use error messages that better reflect
- https://doc.rust-lang.org/std/option/enum.Option.html#recommended-message-style
- https://rustc-dev-guide.rust-lang.org/diagnostics.html

### DIFF
--- a/pgrx/src/datum/json.rs
+++ b/pgrx/src/datum/json.rs
@@ -123,8 +123,7 @@ impl FromDatum for JsonString {
 /// for json
 impl IntoDatum for Json {
     fn into_datum(self) -> Option<pg_sys::Datum> {
-        let string =
-            serde_json::to_string(&self.0).expect("a json value must be serializable to json");
+        let string = serde_json::to_string(&self.0).unwrap();
         string.into_datum()
     }
 
@@ -136,8 +135,7 @@ impl IntoDatum for Json {
 /// for jsonb
 impl IntoDatum for JsonB {
     fn into_datum(self) -> Option<pg_sys::Datum> {
-        let string =
-            serde_json::to_string(&self.0).expect("a jsonb value must be serializable to json");
+        let string = serde_json::to_string(&self.0).unwrap();
         let cstring = alloc::ffi::CString::new(string)
             .expect("a text version of jsonb must contain no null terminator");
 
--- a/pgrx/src/datum/json.rs
+++ b/pgrx/src/datum/json.rs
@@ -68,7 +68,7 @@ impl FromDatum for JsonB {
                 pg_sys::jsonb_out,
                 &[Some(detoasted.into())],
             )
-            .expect("Ddtum must refer to a valid JSONB varlena");
+            .expect("datum must refer to a valid JSONB varlena");
 
             let value = serde_json::from_str(
                 cstr.to_str().expect("a text version of the JSONB must be a valid UTF8"),
--- a/pgrx/src/datum/json.rs
+++ b/pgrx/src/datum/json.rs
@@ -45,7 +45,7 @@ impl FromDatum for Json {
             let data = vardata_any(varlena);
             let slice = std::slice::from_raw_parts(data as *const u8, len);
             let value =
-                serde_json::from_slice(slice).expect("datum must refer to a valid JSON varlena");
+                serde_json::from_slice(slice).expect("datum must refer to a valid json varlena");
             Some(Json(value))
         }
     }
@@ -68,12 +68,12 @@ impl FromDatum for JsonB {
                 pg_sys::jsonb_out,
                 &[Some(detoasted.into())],
             )
-            .expect("datum must refer to a valid JSONB varlena");
+            .expect("datum must refer to a valid jsonb varlena");
 
             let value = serde_json::from_str(
-                cstr.to_str().expect("a text version of the JSONB must be a valid UTF8"),
+                cstr.to_str().expect("a text version of the jsonb must be a valid utf-8"),
             )
-            .expect("a text version of JSONB must be a vaild JSON");
+            .expect("a text version of jsonb must be a vaild json");
 
             // free the cstring returned from direct_function_call -- we don't need it anymore
             pg_sys::pfree(cstr.as_ptr() as void_mut_ptr);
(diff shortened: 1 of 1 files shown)


pgrx/src/datum/json.rs
```diff
@@ -44,7 +44,7 @@ impl FromDatum for Json {
             let len = varsize_any_exhdr(varlena);
             let data = vardata_any(varlena);
             let slice = std::slice::from_raw_parts(data as *const u8, len);
-            let value = serde_json::from_slice(slice).expect("failed to parse Json value");
+            let value = serde_json::from_slice(slice).expect("failed to parse a json value");
```
workingjubilee: The reason it's called `expect` is that it should encode the _expectation_ instead of _what went wrong_ (because what went wrong is imperfect knowledge, because we often aren't checking and doesn't help the programmer... we want to give a _why_). We've been casual about this, but if we're going to spend time revising these:

Let's rephrase this as "Datum must refer to a JSON varlena" or something similar (if we've failed at parsing JSON, then it's not JSON, according to us).

pgrx/src/datum/json.rs
```diff
@@ -67,12 +67,12 @@ impl FromDatum for JsonB {
                 pg_sys::jsonb_out,
                 &[Some(detoasted.into())],
             )
-            .expect("failed to convert jsonb to a cstring");
+            .expect("failed to convert a jsonb to a cstring");
```
workingjubilee: Likewise, "Datum must refer to JSONB for Postgres to serialize it to a cstring" or something similar.

pgrx/src/datum/json.rs
```diff
@@ -67,12 +67,12 @@ impl FromDatum for JsonB {
...
 
             let value = serde_json::from_str(
-                cstr.to_str().expect("text version of jsonb is not valid UTF8"),
+                cstr.to_str().expect("a text version of the jsonb is not valid UTF8"),
             )
-            .expect("failed to parse JsonB value");
+            .expect("failed to parse a jsonb value");
```
workingjubilee: reframe in terms of requirements, etc.

pgrx/src/datum/json.rs
```diff
@@ -122,7 +122,7 @@ impl FromDatum for JsonString {
 /// for json
 impl IntoDatum for Json {
     fn into_datum(self) -> Option<pg_sys::Datum> {
-        let string = serde_json::to_string(&self.0).expect("failed to serialize Json value");
+        let string = serde_json::to_string(&self.0).expect("failed to serialize a json value");
```
workingjubilee: reframe in terms of requirements, etc.

pgrx/src/datum/json.rs
```diff
@@ -134,9 +134,9 @@ impl IntoDatum for Json {
 /// for jsonb
 impl IntoDatum for JsonB {
     fn into_datum(self) -> Option<pg_sys::Datum> {
-        let string = serde_json::to_string(&self.0).expect("failed to serialize JsonB value");
-        let cstring =
-            alloc::ffi::CString::new(string).expect("string version of jsonb is not valid UTF8");
+        let string = serde_json::to_string(&self.0).expect("failed to serialize a jsonb value");
+        let cstring = alloc::ffi::CString::new(string)
+            .expect("string version of jsonb contains a non-terminating nul");
```
workingjubilee: reframe in terms of requirements, etc.

workingjubilee (changes requested):

pgrx/src/datum/json.rs
YohDeadfall: A follow up question then. Are the other guidelines then how to format messages there? In the standard library I saw that everything is lowercase and in some other crates too.

workingjubilee: That's because of https://rustc-dev-guide.rust-lang.org/diagnostics.html?highlight=style#diagnostic-output-style-guide

YohDeadfall: Should `pgrx` follow it?

workingjubilee: It's usually good style to follow the stdlib, but I'm not that pressed over doing so.

pgrx/src/datum/json.rs
YohDeadfall: An attempt to find a correct wording has been failed completely. The thing here is that `serde_json` should be able to serialize it's own types without any issue. Moreover, the `Value` type says that it's a valid JSON representation.

pgrx/src/datum/json.rs
```diff
@@ -67,12 +68,12 @@ impl FromDatum for JsonB {
                 pg_sys::jsonb_out,
                 &[Some(detoasted.into())],
             )
-            .expect("failed to convert jsonb to a cstring");
+            .expect("Ddtum must refer to a valid JSONB varlena");
```
SteveLauC:
```suggestion
            .expect("Datum must refer to a valid JSONB varlena");
```

A tiny typo:>

YohDeadfall: Thanks! Fixed and lower-cased.

pgrx/src/datum/json.rs
workingjubilee: Ah. Then if we don't have anything useful to say, simply `.unwrap()`.

pgrx/src/datum/json.rs
```diff
@@ -67,12 +68,12 @@ impl FromDatum for JsonB {
                 pg_sys::jsonb_out,
                 &[Some(detoasted.into())],
             )
-            .expect("failed to convert jsonb to a cstring");
+            .expect("datum must refer to a valid jsonb varlena");
 
             let value = serde_json::from_str(
-                cstr.to_str().expect("text version of jsonb is not valid UTF8"),
+                cstr.to_str().expect("a text version of the jsonb must be a valid utf-8"),
```
workingjubilee:
```suggestion
                cstr.to_str().expect("a text version of the jsonb must be valid utf-8"),
```

pgrx/src/datum/json.rs
```diff
@@ -67,12 +68,12 @@ impl FromDatum for JsonB {
...
             )
-            .expect("failed to parse JsonB value");
+            .expect("a text version of jsonb must be a vaild json");
```
workingjubilee:
```suggestion
            .expect("a text version of jsonb must be a valid json");
```

workingjubilee (approved):
thank you for trying. this looks good with the nits addressed.

PR #1894, opened 2024-09-30
Always panic if len of varlena exceeds the maximum

YohDeadfall:
The documentation is fixed due to being incorrect, `4294967295 >> 2` is division by `4`. And the reason of panicking is that it's promised too.

### DIFF
--- a/pgrx/src/datum/into.rs
+++ b/pgrx/src/datum/into.rs
@@ -337,10 +337,12 @@ impl_into_datum_c_str!(&CStr);
 impl<'a> IntoDatum for &'a [u8] {
     /// # Panics
     ///
-    /// This function will panic if the string being converted to a datum is longer than a half of [`i32::MAX`].
+    /// This function will panic if the string being converted to a datum
+    //  is longer than 1 GiB including 4 bytes used for a header.
     #[inline]
     fn into_datum(self) -> Option<pg_sys::Datum> {
-        let len = pg_sys::VARHDRSZ + self.len();
+        let len = self.len().saturating_add(pg_sys::VARHDRSZ);
+        assert!(len < (u32::MAX as usize >> 2));
         unsafe {
             // SAFETY:  palloc gives us a valid pointer and if there's not enough memory it'll raise an error
             let varlena = pg_sys::palloc(len) as *mut pg_sys::varlena;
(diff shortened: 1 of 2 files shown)


workingjubilee:
@YohDeadfall There is a more memorable number that is the limit.

1 gigabyte (specifically, 1024 * 1024 * 1024).

pgrx/src/datum/into.rs
```diff
@@ -352,9 +352,9 @@ impl<'a> IntoDatum for &'a [u8] {
 
             // This is the same as Postgres' `#define SET_VARSIZE_4B` (which have over in
             // `pgrx/src/varlena.rs`), however we're asserting that the input string isn't too big
-            // for a Postgres varlena, since it's limited to 32bits -- in reality it's about half
+            // for a Postgres varlena, since it's limited to 32bits -- in reality it's a quarter
             // that length, but this is good enough
-            debug_assert!(len < (u32::MAX as usize >> 2));
+            assert!(len < (u32::MAX as usize >> 2));
```
workingjubilee: if we are going to always assert then we should raise it up to before `palloc`, so we don't waste time trying to allocate something we're going to reject anyways.

pgrx/src/datum/into.rs
```diff
@@ -337,7 +337,7 @@ impl_into_datum_c_str!(&CStr);
 impl<'a> IntoDatum for &'a [u8] {
     /// # Panics
     ///
-    /// This function will panic if the string being converted to a datum is longer than a `u32`.
+    /// This function will panic if the string being converted to a datum is longer than a half of [`i32::MAX`].
     #[inline]
     fn into_datum(self) -> Option<pg_sys::Datum> {
         let len = pg_sys::VARHDRSZ + self.len();
```
workingjubilee: And we will want to make this a `.checked_add` or `.overflowing_add`

pgrx/src/datum/into.rs
```diff
@@ -337,7 +337,7 @@ impl_into_datum_c_str!(&CStr);
 impl<'a> IntoDatum for &'a [u8] {
     /// # Panics
     ///
-    /// This function will panic if the string being converted to a datum is longer than a `u32`.
+    /// This function will panic if the string being converted to a datum is longer than a half of [`i32::MAX`].
```
workingjubilee: Let's be precise: greater than or equal to 1GB *after* adding the varlena header (which is currently 4 bytes).

workingjubilee (changes requested):

pgrx/src/datum/into.rs
YohDeadfall: Had that thought, but I assumed that the current place for assertion was intentional. Will fix.

workingjubilee: nah, I think it just kinda was organic and not thought about heavily because it was a `debug_assert`.

pgrx/src/datum/into.rs
YohDeadfall: Used `GiB` to be more precise since.

workingjubilee (approved):
Thanks!

PR #1896, opened 2024-10-01
`impl FromDatum for CString`

workingjubilee:
As much as I would like FromDatum to unexist, it should probably be usable for all std types we even half-support, especially since it's less dubious, in several ways, than the `&'a CStr` impl.

### DIFF
--- a/pgrx/src/datum/from.rs
+++ b/pgrx/src/datum/from.rs
@@ -449,7 +449,7 @@ impl FromDatum for char {
 }
 
 /// for cstring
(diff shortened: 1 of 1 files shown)


YohDeadfall:
What do you have in your mind about `FromDatum`? How should it e done otherwise? It might be a dumb question, but I really would like to get the direction you're moving on to avoid conflicts.

workingjubilee:
@YohDeadfall It can't enforce lifetimes, that's all.

PR #1900, opened 2024-10-03
Implement UnboxDatum for ItemPointerData

syvb:
`ItemPointerData` can be directly converted from a Datum, so `ItemPointerData` should have it implemented.


workingjubilee:
Thanks!

Issue #1897, opened 2024-10-01
Build failed on ppc64el


tucnak:
I know ppc64el is probably not natively supported by this library, however I'd tried building it regardless as both Rust as well as Postgres support it pretty well, however I'd ran into the following issue when building one of the extensions I'm interested in. Even though admittedly it seems `cee-scape` is to blame, I wonder if it's possible to build without it.

Is there any reason in particular why inline assembly is required here?

Related #1144

```
error: cannot find macro `maybesig_setjmp_asm` in this scope
   --> /home/badt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cee-scape-0.2.0/src/asm_based.rs:164:9
    |
164 |         maybesig_setjmp_asm!(setjmp, jbuf_ptr, unused_savemask, closure_env_ptr, c2r, ret);
    |         ^^^^^^^^^^^^^^^^^^^

error: cannot find macro `maybesig_setjmp_asm` in this scope
   --> /home/badt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cee-scape-0.2.0/src/asm_based.rs:215:9
    |
215 |         maybesig_setjmp_asm!(sigsetjmp, jbuf_ptr, savemask, closure_env_ptr, c2r, ret);
    |         ^^^^^^^^^^^^^^^^^^^

error[E0425]: cannot find value `JMP_BUF_SIZE` in this scope
 --> /home/badt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cee-scape-0.2.0/src/macos_compat.rs:8:33
  |
8 | const SIG_JMP_BUF_SIZE: usize = JMP_BUF_SIZE + 1;
  |                                 ^^^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `JMP_BUF_SIZE` in this scope
  --> /home/badt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cee-scape-0.2.0/src/macos_compat.rs:14:17
   |
8  | const SIG_JMP_BUF_SIZE: usize = JMP_BUF_SIZE + 1;
   | ------------------------------------------------- similarly named constant `SIG_JMP_BUF_SIZE` defined here
...
14 |     _buf: [u32; JMP_BUF_SIZE],
   |                 ^^^^^^^^^^^^ help: a constant with a similar name exists: `SIG_JMP_BUF_SIZE`

For more information about this error, try `rustc --explain E0425`.
[...]

PR #1901, opened 2024-10-04
fix compilation on some CPUs and Windows

usamoi:
This patch contains two part:

* use cshim for sigsetjmp on CPUs/OSes that are not supported by `cee-scape`'s asm (needed by #1897 and Windows)
* fix linking error on Windows in #1133


pgrx-macros/src/rewriter.rs
```diff
@@ -135,19 +136,52 @@ fn foreign_item_fn(func: &ForeignItemFn, abi: &syn::Abi) -> syn::Result<proc_mac
...
+                    } else {
+                        false
+                    }
+                } else {
+                    false
+                }
+            } else {
+                false
+            }
+        }
```
workingjubilee: HADOUKEN!

pgrx-macros/src/rewriter.rs
```diff
@@ -135,19 +136,52 @@ fn foreign_item_fn(func: &ForeignItemFn, abi: &syn::Abi) -> syn::Result<proc_mac
     let arg_list = rename_arg_list(&func.sig)?;
     let arg_list_with_types = rename_arg_list_with_types(&func.sig)?;
     let return_type = func.sig.output.clone();
+    let link_with_cshim = func.attrs.iter().any(|attr| match &attr.meta {
+        syn::Meta::NameValue(kv) => {
+            if kv.path.get_ident().map(|x| x.to_string()) == Some("link_name".to_string()) {
```
workingjubilee: Ident implements PartialEq for str, you don't need to `.to_string()` this

pgrx-macros/src/rewriter.rs
```diff
@@ -135,19 +136,52 @@ fn foreign_item_fn(func: &ForeignItemFn, abi: &syn::Abi) -> syn::Result<proc_mac
...
+                if let syn::Expr::Lit(lit) = &kv.value {
+                    if let syn::Lit::Str(value) = &lit.lit {
```
workingjubilee: Just do the `syn` imports here so we can make this
```suggestion
                if let Expr::Lit(ExprLit { lit: Lit::Str(value), .. }) = &kv.value {
```

pgrx-macros/src/rewriter.rs
```diff
@@ -135,19 +136,52 @@ fn foreign_item_fn(func: &ForeignItemFn, abi: &syn::Abi) -> syn::Result<proc_mac
     let arg_list = rename_arg_list(&func.sig)?;
     let arg_list_with_types = rename_arg_list_with_types(&func.sig)?;
     let return_type = func.sig.output.clone();
+    let link_with_cshim = func.attrs.iter().any(|attr| match &attr.meta {
+        syn::Meta::NameValue(kv) => {
```
workingjubilee: the `if` should be able to be combined into the match arm, something like:
```suggestion
        syn::Meta::NameValue(kv) if kv.path.get_ident().filter(|x| x == "link_name").is_some() => {
```

workingjubilee (changes requested):

usamoi:
HADOUKEN has gone.

pgrx-bindgen/src/build.rs
```diff
@@ -1116,18 +1125,11 @@ fn apply_pg_guard(items: &Vec<syn::Item>) -> eyre::Result<proc_macro2::TokenStre
...
-                    |item| match item {
-                        ForeignItem::Fn(func) => extern_funcs.push(func),
-                        item => others.push(item),
-                    },
-                );
+                let items = block.items.iter().filter(|&item| !is_blocklisted_item(item));
                 out.extend(quote! {
                     #[pgrx_macros::pg_guard]
-                    #abi { #(#extern_funcs)* }
+                    #abi { #(#items)* }
```
workingjubilee, line 1131: Ah, so we're now handling `extern static`s under this. I see.

pgrx-macros/src/rewriter.rs
```diff
@@ -135,19 +136,44 @@ fn foreign_item_fn(func: &ForeignItemFn, abi: &syn::Abi) -> syn::Result<proc_mac
...
+                false
+            }
+        }
+        _ => false,
+    });
+    let link = if link_with_cshim {
+        quote! {}
+    } else {
+        quote! { #[cfg_attr(target_os = "windows", link(name = "postgres"))] }
+    };
```
workingjubilee, line 153: Hrm. What should I be seeing and when? I am trying to force this to trigger while building the bindings and it doesn't seem to be happening?

workingjubilee: Ah, wait, this will only happen during macro expansion, sec...

workingjubilee: ...which happens after `cfg`, hmmm...

workingjubilee: Okay, forced this through `cargo expand` after some fiddling and it works as expected.

pgrx-pg-sys/src/submodules/ffi.rs
```diff
@@ -9,7 +9,64 @@
...
 #![deny(unsafe_op_in_unsafe_fn)]
 
-use cee_scape::SigJmpBufFields;
+#[cfg(not(all(
+    any(target_os = "linux", target_os = "macos"),
+    any(target_arch = "x86_64", target_arch = "aarch64")
+)))]
+mod cee_scape {
+    #[cfg(not(feature = "cshim"))]
+    compile_error!("target platform cannot work without feature cshim");
```
workingjubilee, line 18: Hm. Is there a reason we're not just using the `use_c_to_interface_with_setjmp` feature of `cee_scape`?

usamoi: 1. To support more CPUs and Windows, we need a patch of `cee_scape`.
2. In `postgres/src/include/c.h`, sigsetjmp is defined as:
```c
/*
 * When there is no sigsetjmp, its functionality is provided by plain
 * setjmp.  We now support the case only on Windows.  However, it seems
 * that MinGW-64 has some longstanding issues in its setjmp support,
 * so on that toolchain we cheat and use gcc's builtins.
 */
#ifdef WIN32
#ifdef __MINGW64__
typedef intptr_t sigjmp_buf[5];
#define sigsetjmp(x,y) __builtin_setjmp(x)
#define siglongjmp __builtin_longjmp
#else							/* !__MINGW64__ */
#define sigjmp_buf jmp_buf
#define sigsetjmp(x,y) setjmp(x)
#define siglongjmp longjmp
#endif							/* __MINGW64__ */
#endif							/* WIN32 */
```
I feel it's better to include the header and use `sigsetjmp` in `pgrx-cshim.c` than adding a `__builtin_setjmp` version of `call_closure_with_setjmp` in `cee-scape` and gating the variants by cfg...

workingjubilee: Ah, interesting.

workingjubilee (approved):
Thank you!

PR #1905, opened 2024-10-09
docs: update generic type name of TableIterator

SteveLauC:
### What does this PR do

Updates the doc of `TableIterator`, the generic type has been renamed in #1701 but the doc was not updated


workingjubilee (approved):
Thanks!

PR #1910, opened 2024-10-09
Adds more headers

aykut-bozkurt:
`access/detoast.h`: useful to get detoasted size of varlena datums
`utils/rls.h`: useful for check if row level security is enabled for a relation


workingjubilee:
Thank you!

PR #1913, opened 2024-10-11
Added connect_mut for data changing SPI operations

YohDeadfall:
The idea is to make the `connect_mut` function which then can be used for mutating operations with SPI. That should be used instead of the current `SqlClient` passed by value and `mut` modifier in closure arguments.

That's a breaking change, so waits till `0.13`.

### DIFF
--- a/pgrx-examples/schemas/src/lib.rs
+++ b/pgrx-examples/schemas/src/lib.rs
@@ -101,7 +101,7 @@ mod tests {
 
     #[pg_test]
     fn test_my_some_schema_type() -> Result<(), spi::Error> {
-        Spi::connect(|mut c| {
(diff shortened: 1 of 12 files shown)


eeeebbbbrrrr:
This one appears to have bitrotted a little bit.  Yoh, if you wanna bring it back to life we can probably get it merged.

YohDeadfall:
@eeeebbbbrrrr, it's mergeable now.

YohDeadfall:
@eeeebbbbrrrr, hope you are doing well and have a bit of time to review that tiny pull requests. If you cannot, may @workingjubilee check it?

workingjubilee (approved):

PR #1917, opened 2024-10-12
fix compilation on emscripten

usamoi:


usamoi:
Notes for someone who is interested in this.

Examples in `pgrx-examples` work with [pglite](https://github.com/electric-sql/pglite/tree/main) if the following is set correctly:

1. `rustc` flag `-Clink-arg=-sSIDE_MODULE=2`
2. environment variable `BINDGEN_EXTRA_CLANG_ARGS_wasm32_unknown_emscripten="-isystem /home/usamoi/.emscripten_cache/sysroot/include/fakesdl -isystem /home/usamoi/.emscripten_cache/sysroot/include/compat -isystem /opt/emscripten-llvm/lib/clang/20/include -isystem /home/usamoi/.emscripten_cache/sysroot/include -fvisibility=default"` set (you can find these paths by `emcc test.c --verbose`)
3. environment variable `PGRX_BINDGEN_NO_DETECT_INCLUDES=1`
4. add needed emscripten/postgres symbols in pglite (see https://github.com/usamoi/pglite/pull/2).
5. `cargo` flag `--features pgrx/cshim`

Known issues:

1. `_emscripten_memcpy_js` is needed by the environment (but it does not emit errors util this symbol is used at run-time, and I'm not sure if it could be solved by adding `rustc` flags `-Ctarget-feature=+bulk-memory` and `-Clink-arg=-mbulk-memory`)

PGRX does not support running on emscripten and pglite does not support loading pgrx extensions officially. So there could be more potential bugs and you can **not** expect it works without your extra efforts.

eeeebbbbrrrr (approved):

PR #1921, opened 2024-10-16
upgrade all dependencies

eeeebbbbrrrr:
A run of our `./upgrade-deps.sh` script.


eeeebbbbrrrr:
well, this didn't work.  #ugh

eeeebbbbrrrr:
I did this because i noticed there's two different `toml_edit` versions in the dependency tree, but of course that's the one that can't be upgraded for some reason b/c of `cargo-toml`.  Very annoying, but it is what it is for today, I suppose.

PR #1924, opened 2024-10-21
add `commands/progress.h`

usamoi:
https://github.com/postgres/postgres/blob/master/src/include/commands/progress.h
It contains constants used with the progress reporting facilities.


workingjubilee (approved):
Thanks!

PR #1925, opened 2024-10-23
Imply `NOT_IN_SAMPLE` for `GucFlags::NO_SHOW_ALL`

Ngalstyan4:
Hi all, 

firstly, thanks for giving pgrx to the world!

We use pgrx at Lantern[^0] and when configuring GUCs noticed a small issue with GUC flags. Postgres requires `GUC_NOT_IN_SAMPLE` GUC flag to be set whenever `NO_SHOW_ALL` is set[^1]. Make applying `GucFlags::NO_SHOW_ALL` also apply `NOT_IN_SAMPLE`, as `NOT_IN_SAMPLE` cannot be set.

After applying this diff, we no longer have a direct name matching between pgrx GUC names and postgres ones. An alternative might be to deprecate `NO_SHOW_ALL` and introduce something like `NO_SHOW_ALL_AND_NO_SAMPLE`, but not sure if it is worth it.

[^0]: https://github.com/lanterndata/lantern
[^1]: https://github.com/postgres/postgres/blob/dbedc461b4e7a9cb4d6f5777174bdf180edb95fd/src/backend/utils/misc/guc.c#L1506-L1519

### DIFF
--- a/pgrx/src/guc.rs
+++ b/pgrx/src/guc.rs
@@ -60,7 +60,7 @@ bitflags! {
     /// descriptions below for their behaviour.
     pub struct GucFlags: i32 {
         /// Exclude from SHOW ALL
-        const NO_SHOW_ALL = pg_sys::GUC_NO_SHOW_ALL as i32;
+        const NO_SHOW_ALL = pg_sys::GUC_NO_SHOW_ALL as i32 | pg_sys::GUC_NOT_IN_SAMPLE as i32;
         /// Exclude from RESET ALL
         const NO_RESET_ALL = pg_sys::GUC_NO_RESET_ALL as i32;
         /// Auto-report changes to client
--- a/pgrx/src/guc.rs
+++ b/pgrx/src/guc.rs
@@ -55,7 +55,7 @@ pub enum GucContext {
 }
 
 bitflags! {
-    #[derive(Default)]
+    #[derive(Default, Copy, Clone)]
     /// Flags to control special behaviour for the GUC that these are set on. See their
     /// descriptions below for their behaviour.
     pub struct GucFlags: i32 {
--- a/pgrx-tests/src/tests/guc_tests.rs
+++ b/pgrx-tests/src/tests/guc_tests.rs
@@ -174,6 +174,9 @@ mod tests {
 
     #[pg_test]
     fn test_guc_flags() {
+        // variable ensures that GucFlags is Copy, so single name can be used when defining
+        // multiple gucs
+        let no_show_flag = GucFlags::NO_SHOW_ALL;
         static GUC_NO_SHOW: GucSetting<bool> = GucSetting::<bool>::new(true);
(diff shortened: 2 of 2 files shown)


workingjubilee:
@Ngalstyan4 your "source" link doesn't work at all, it seems.

Ngalstyan4:
sorry, this is what I was referring to: https://github.com/postgres/postgres/blob/dbedc461b4e7a9cb4d6f5777174bdf180edb95fd/src/backend/utils/misc/guc.c#L1506-L1519

Fixed the link in the description above as well.

workingjubilee:
Thank you.

...I wonder why they don't just make one flag imply... well, anyway.

pgrx/src/guc.rs
```diff
@@ -55,12 +55,12 @@ pub enum GucContext {
...
 
 bitflags! {
-    #[derive(Default)]
+    #[derive(Default, Copy, Clone)]
     /// Flags to control special behaviour for the GUC that these are set on. See their
     /// descriptions below for their behaviour.
     pub struct GucFlags: i32 {
         /// Exclude from SHOW ALL
-        const NO_SHOW_ALL = pg_sys::GUC_NO_SHOW_ALL as i32;
+        const NO_SHOW_ALL = pg_sys::GUC_NO_SHOW_ALL as i32 | pg_sys::GUC_NOT_IN_SAMPLE as i32;
```
workingjubilee, line 63: This technically prevents independently setting these. Or at least not in an easy and error-free way.

Not sure I or anyone else cares about that.

Ngalstyan4: Just to clarify, you mean it prevents independently setting `pg_sys::NO_SHOW_ALL`, right?

`pg_sys::GUC_NOT_IN_SAMPLE` is not exposed via a safe wrapper as `GucFlags` member, but my understanding is that if/when that is exposed, then that and all the other flags can still be set independently. 

`pg_sys::NO_SHOW_ALL` cannot be set independently but that seems to be a postgres invariant (even though the invariant is not enforced in C)

workingjubilee: by set independently, I mean if one did this:
```rust
let guc_flags = GucFlags::NO_SHOW_ALL;
let guc_flags = guc_flags & !GucFlags::NOT_IN_SAMPLE;
```
that would be obviously incorrect.
more subtle:
```rust
let guc_flags = GucFlags::NOT_IN_SAMPLE;
let guc_flags = guc_flags & !GucFlags::NO_SHOW_ALL;
```

...I think I'd have to use some function calls on the bitflags API surface instead, since I think it doesn't auto-implement operators, but you get the idea. It does implement a trait that has the equivalent functions as those operators.

workingjubilee: Anyways, we will worry this slightly unintuitive behavior if we ever have anyone ask for `NOT_IN_SAMPLE`.

workingjubilee (approved):

PR #1929, opened 2024-10-27
fix: crash from drop impls that panic via `leak_and_drop_on_delete()`

eeeebbbbrrrr:
If `PgMemoryContexts::leak_and_drop_on_delete()` leaks a type that is `impl Drop`, and that implementation happens to panic, then Postgres will segfault because the callback function was not properly guarded.

As a drive-by, slightly cleanup error reporting in the test framework.

### DIFF
--- a/pgrx/src/memcxt.rs
+++ b/pgrx/src/memcxt.rs
@@ -555,6 +555,8 @@ impl PgMemoryContexts {
     /// this MemoryContext, the original instance of `T` will be resurrected and its `impl Drop`
     /// will be called.
     pub fn leak_and_drop_on_delete<T>(&mut self, v: T) -> *mut T {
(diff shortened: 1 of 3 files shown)


pgrx/src/memcxt.rs
```diff
@@ -555,6 +555,8 @@ impl PgMemoryContexts {
     /// this MemoryContext, the original instance of `T` will be resurrected and its `impl Drop`
     /// will be called.
     pub fn leak_and_drop_on_delete<T>(&mut self, v: T) -> *mut T {
+        use crate as pgrx;
+        #[pgrx::pg_guard]
```
eeeebbbbrrrr, line 559: the actual fix!

PR #1930, opened 2024-10-27
Teach `#[pg_cast]` how to pass down arguments useful to `#[pg_extern]`

eeeebbbbrrrr:
The `#[pg_cast]` macro does two things.  It creates a regular UDF and it then generates the appropriate `CREATE CAST` sql.

Ultimately, the cast is just a function so we should support all the attibutes that `#[pg_extern]` supports.

### DIFF
--- a/pgrx-tests/tests/compile-fail/invalid_pgcast_options.stderr
+++ b/pgrx-tests/tests/compile-fail/invalid_pgcast_options.stderr
@@ -1,7 +1,11 @@
-error: custom attribute panicked
- --> tests/compile-fail/invalid_pgcast_options.rs:3:1
+error: Invalid option `invalid_opt` inside `invalid_opt `
+ --> tests/compile-fail/invalid_pgcast_options.rs:3:11
   |
 3 | #[pg_cast(invalid_opt)]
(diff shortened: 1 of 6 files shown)


pgrx-tests/tests/compile-fail/invalid_pgcast_options.stderr
```diff
@@ -1,7 +1,11 @@
...
 3 | #[pg_cast(invalid_opt)]
-  | ^^^^^^^^^^^^^^^^^^^^^^^
+  |           ^^^^^^^^^^^
+
+error: failed parsing pg_extern arguments
+ --> tests/compile-fail/invalid_pgcast_options.rs:3:11
   |
-  = help: message: Unrecognized pg_cast option: invalid_opt.
+3 | #[pg_cast(invalid_opt)]
+  |           ^^^^^^^^^^^
```
eeeebbbbrrrr, line 11: This changed because it's now the `#[pg_extern]` macro that's deciding what is valid.  I don't know that this new error message is objectively better from a UX perspective, but it's still an error, and that's good enough for this PR.

PR #1933, opened 2024-10-29
Changed args of prepare methods to be slices

YohDeadfall:
Since there's an opportunity to break more things for `0.13` I decided to improve the `prepare` and `prepare_mut` methods as well. The motivation for that is the same as for #1858, `&[]` requires no heap allocations on the heap if there's a fixed set of arguments.


workingjubilee:
sure.

eeeebbbbrrrr (approved):

PR #1942, opened 2024-11-12
feat: auto-generate type alignment (#[pgrx(alignment = "on")])

JamesGuthrie:
Postgres allows for types to specify their alignment in the `CREATE TYPE` statement. This change adds the ability to derive Postgres' alignment configuration from the type's Rust alignment (`std::mem::align_of::<T>()`). This functionality is opt-in through the `#[pgrx(alignment = "on")]` attribute:

```
#[derive(PostgresType)]
#[pgrx(alignment = "on")]
struct AlignedTo8Bytes {
  v1: u64,
  v2: [u64; 3]
}
```

### DIFF
--- a/pgrx-macros/src/lib.rs
+++ b/pgrx-macros/src/lib.rs
@@ -781,6 +781,7 @@ Optionally accepts the following attributes:
 
 * `inoutfuncs(some_in_fn, some_out_fn)`: Define custom in/out functions for the type.
 * `pgvarlena_inoutfuncs(some_in_fn, some_out_fn)`: Define custom in/out functions for the `PgVarlena` of this type.
+* `pgrx(alignment = "<align>")`: Derive Postgres alignment from Rust type. One of `"on"`, or `"off"`.
 * `sql`: Same arguments as [`#[pgrx(sql = ..)]`](macro@pgrx).
 */
 #[proc_macro_derive(
--- a/pgrx-sql-entity-graph/src/lib.rs
+++ b/pgrx-sql-entity-graph/src/lib.rs
@@ -94,6 +94,8 @@ pub trait SqlGraphIdentifier {
     fn line(&self) -> Option<u32>;
 }
 
+pub use postgres_type::Alignment;
+
 /// An entity corresponding to some SQL required by the extension.
 #[derive(Debug, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
 pub enum SqlGraphEntity {
--- a/pgrx-sql-entity-graph/src/postgres_type/entity.rs
+++ b/pgrx-sql-entity-graph/src/postgres_type/entity.rs
@@ -16,13 +16,82 @@
 
 */
 use crate::mapping::RustSqlMapping;
+use crate::pgrx_attribute::{ArgValue, PgrxArg, PgrxAttribute};
 use crate::pgrx_sql::PgrxSql;
 use crate::to_sql::entity::ToSqlConfigEntity;
 use crate::to_sql::ToSql;
 use crate::{SqlGraphEntity, SqlGraphIdentifier, TypeMatch};
-use std::collections::BTreeSet;
-
 use eyre::eyre;
+use proc_macro2::TokenStream;
+use quote::{format_ident, quote, ToTokens, TokenStreamExt};
+use std::collections::BTreeSet;
+use syn::spanned::Spanned;
+use syn::{AttrStyle, Attribute, Lit};
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub enum Alignment {
+    On,
+    Off,
+}
+
+const INVALID_ATTR_CONTENT: &str =
+    r#"expected `#[pgrx(alignment = align)]`, where `align` is "on", or "off""#;
+
+impl ToTokens for Alignment {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        let value = match self {
+            Alignment::On => format_ident!("On"),
+            Alignment::Off => format_ident!("Off"),
+        };
+        let quoted = quote! {
+            ::pgrx::pgrx_sql_entity_graph::Alignment::#value
+        };
+        tokens.append_all(quoted);
+    }
+}
+
+impl Alignment {
+    pub fn from_attribute(attr: &Attribute) -> Result<Option<Self>, syn::Error> {
+        if attr.style != AttrStyle::Outer {
+            return Err(syn::Error::new(
+                attr.span(),
+                "#[pgrx(alignment = ..)] is only valid in an outer context",
+            ));
+        }
+
+        let attr = attr.parse_args::<PgrxAttribute>()?;
+        for arg in attr.args.iter() {
+            let PgrxArg::NameValue(ref nv) = arg;
+            if !nv.path.is_ident("alignment") {
+                continue;
+            }
+
+            return match nv.value {
+                ArgValue::Lit(Lit::Str(ref s)) => match s.value().as_ref() {
+                    "on" => Ok(Some(Self::On)),
+                    "off" => Ok(Some(Self::Off)),
+                    _ => Err(syn::Error::new(s.span(), INVALID_ATTR_CONTENT)),
+                },
+                ArgValue::Path(ref p) => Err(syn::Error::new(p.span(), INVALID_ATTR_CONTENT)),
+                ArgValue::Lit(ref l) => Err(syn::Error::new(l.span(), INVALID_ATTR_CONTENT)),
+            };
+        }
+
+        Ok(None)
+    }
+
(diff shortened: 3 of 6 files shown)


pgrx-sql-entity-graph/src/lib.rs
```diff
@@ -94,6 +94,8 @@ pub trait SqlGraphIdentifier {
     fn line(&self) -> Option<u32>;
 }
 
+pub use postgres_type::Alignment;
```
JamesGuthrie, line 97: I'm not sure that this is the right way to expose the `Alignment` enum (no other modules do it this way). I didn't want to change the visibility of the `postgres_type` module to global, for fear of leaking other types which we don't want to.

pgrx-sql-entity-graph/src/postgres_type/entity.rs
```diff
@@ -21,8 +21,88 @@ use crate::to_sql::entity::ToSqlConfigEntity;
...
+    /// Used to parse a generator config from a set of item attributes
+    pub fn from_attributes(attrs: &[Attribute]) -> Result<Option<Self>, syn::Error> {
+        if let Some(attr) = attrs.iter().find(|attr| attr.path().is_ident("pgrx")) {
+            Self::from_attribute(attr)
+        } else {
+            Ok(None)
+        }
+    }
+}
+
```
JamesGuthrie: Not sure if this should live here, or elsewhere?

pgrx-sql-entity-graph/src/postgres_type/entity.rs
```diff
@@ -163,7 +256,7 @@ impl ToSql for PostgresTypeEntity {
                     \tINTERNALLENGTH = variable,\n\
                     \tINPUT = {schema_prefix_in_fn}{in_fn}, /* {in_fn_path} */\n\
                     \tOUTPUT = {schema_prefix_out_fn}{out_fn}, /* {out_fn_path} */\n\
-                    \tSTORAGE = extended\n\
+                    \tSTORAGE = extended{alignment}\n\
```
JamesGuthrie, line 255: This is kinda icky, but it works.

pgrx-macros/src/lib.rs
```diff
@@ -781,6 +781,7 @@ Optionally accepts the following attributes:
 
 * `inoutfuncs(some_in_fn, some_out_fn)`: Define custom in/out functions for the type.
 * `pgvarlena_inoutfuncs(some_in_fn, some_out_fn)`: Define custom in/out functions for the `PgVarlena` of this type.
+* `pgrx(alignment = "<align>")`: Define alignment for this type. One of `char`, `int2`, `int4`, or `double` (see https://www.postgresql.org/docs/current/sql-createtype.html).
```
JamesGuthrie: I'm not sure if this should go here. Seems like the best place.

workingjubilee:
could you just make this a property of `repr(align)`?

JamesGuthrie:
> could you just make this a property of `repr(align)`?

Sorry, I'm not sure I understand the suggestion. Are you saying that instead of adding the (custom) attribute `#[pgrx(alignment = "...")]`, we should piggy-back on the presence of a `#[repr(align(x))]`, and use that to generate the `ALIGNMENT` field of the generated SQL?

workingjubilee:
exactly.

workingjubilee:
if the Rust align and the Postgres align disagree, UB is probably about to occur anyways.

JamesGuthrie:
On second thought: what I would really want is for pgrx to auto-generate the correct alignment for my type `T` based off the the value of `std::mem::align_of::<T>()`.

What do you think about that idea?

I'm not sure if that is something that we'd want to "turn on" by default. If not, then I'll strawman this:

```
#[derive(PostgresType)]
#[pgrx(alignment = "on")]
struct AlignedTo8Bytes {
  v1: u64,
  v2: [u64; 3]
}
```

workingjubilee:
> On second thought: what I would really want is for pgrx to auto-generate the correct alignment for my type T based off the the value of std::mem::align_of::<T>().
> 
> What do you think about that idea?

Yes, that sounds fine if you can make it happen.

JamesGuthrie:
> > On second thought: what I would really want is for pgrx to auto-generate the correct alignment for my type T based off the the value of std::mem::align_of::().
> > What do you think about that idea?
> 
> Yes, that sounds fine if you can make it happen.

I can! I've pushed my changes.

workingjubilee:
Musing on design thoughts:
- Perhaps we should do this fully-implicitly and bail silently for types that request an invalid alignment, but then use the attribute as a sort of "I **require** this alignment, fail if I can't get it"?
- Alternatively, we can view this as an early opt-in.

JamesGuthrie:
I've realized that this alignment functionality is only really relevant to people who are implementing `IntoDatum` and `FromDatum` manually (activated through the `#[bikeshed_postgres_type_manually_impl_from_into_datum]` attribute). (Note: I believe that the default CBOR-based `FromDatum`/`IntoDatum` implementations are not alignment-sensitive).

> Musing on design thoughts:
> 
> * Perhaps we should do this fully-implicitly and bail silently for types that request an invalid alignment, but then use the attribute as a sort of "I **require** this alignment, fail if I can't get it"?
> * Alternatively, we can view this as an early opt-in.

My initial motivation for making this opt-in is backwards-compatibility. Extensions that have a custom type's data stored are in a difficult situation, because the generated sql is not backwards-compatible with the previously-generated sql, and there's no real path out of this situation.

To clarify: if you have a type which is 8-byte aligned in Rust, but 4-byte aligned in postgres (the default for varlena), then (going forward) the SQL we would generate for the type would have the 8-byte alignment, but all instances of custom types which are stored on disk are already "locked-in" at 4 byte alignment. It's not possible to change alignment via `ALTER TYPE`; the only way to "upgrade" a type to the correct alignment is to drop and re-create it.

In our extension we intend to support the incorrectly-aligned types that already exist in the wild, and generate the 8-byte alignment for all new installations.

On the other hand one could argue that by always emitting the alignment, the extension developer is in no worse a situation than today. Developers of existing extensions may already unwittingly face the "alignment problem". By emitting the correct alignment for their types, we fix all new installations. The developer then still needs to support incorrectly-aligned types for extensions which have already been released.

Specifically on this:

> use the attribute as a sort of "I **require** this alignment, fail if I can't get it"?

I could imagine that there is somebody who still wants 4-byte alignment for a type which Rust would 8-byte align, mostly for ultimate compactness (maybe the type is mostly 12 bytes, and the additional 4 bytes of overhead is too much to stomach). Obviously in this case the onus is on the developer to ensure that they implement to/from datum correctly, to account for potentially un-aligned data (do-able with [`read_unaligned`](https://doc.rust-lang.org/std/ptr/fn.read_unaligned.html)). Note: there may be more gotchas here that I haven't considered.

In summary:

- I think it would make sense to emit the alignment by default, but only when `#[bikeshed_postgres_type_manually_impl_from_into_datum]` is used.
- I think that there is a use-case around being able to specify 4-byte alignment in Postgres, even when the type is 8-byte aligned. I would propose that we kick this can down the road though.

eeeebbbbrrrr:
Could you please add a test in `pgrx-tests` to assert that the desired alignment value gets set on the CREATE TYPE statement?

eeeebbbbrrrr (changes requested):
Could you please add a test in `pgrx-tests` to assert that the desired alignment value gets set on the CREATE TYPE statement?

eeeebbbbrrrr:
Hi @JamesGuthrie!  Would you mind adding a test in `pgrx-tests` that asserts the desired alignment value is set on the CREATE TYPE statement?

A new pgrx release is incoming, and it'd be good to have this included.

JamesGuthrie:
> Hi @JamesGuthrie! Would you mind adding a test in `pgrx-tests` that asserts the desired alignment value is set on the CREATE TYPE statement?
> 
> A new pgrx release is incoming, and it'd be good to have this included.

Hey @eeeebbbbrrrr, I'm not sure that I understand what you would like me to do. There are already tests in the `pgrx-examples` code which (indirectly) assert that the alignment value is set (by defining types with and without the alignment set, and then asking the database what the type's alignment is).

Would you like me to move/duplicate those tests to `pgrx-tests`? Or did you want a test which actually inspects the generated `CREATE TYPE` statement and validates that we're setting the alignment correctly? If so, I'm not sure how to do that in `pgrx-tests`, I don't see how to access the generated SQL. Otherwise that could be tested with a unit test of the `PostgresTypeEntity`'s `ToSql` impl?

eeeebbbbrrrr:
> There are already tests in the pgrx-examples code which (indirectly) assert that the alignment value is set 

I have looked at this PR about 4 times now, and missed that every time.

Okay, those are just fine and essentially what I imagined they'd look like.

Gonna merge now, and this will be included in the next release -- v0.13.1

eeeebbbbrrrr:
Thank you for the contribution, btw.  And for your patience!

JamesGuthrie:
It's a pleasure!

PR #1956, opened 2024-12-03
Implemented DerefMut for PgRelation

### LINKED ISSUE #1955: Expose `pgstat_count_index_scan` and friends, or make `PgRelation` implement `DerefMut`
Access method implementations are supposed to call `pgstat_count_index_scan` to keep `pg_stat_user_indexes` up to date, but since this is a `#define` macro rather than a function, there is no wrapper function in `pgrx`.  Meanwhile updating `pgstat_info` directly given a `PgRelation` is not possible because `PgRelation` is not `DerefMut`.

YohDeadfall:
Closes #1955, but it still makes sense to implement `pgstat` macros as functions.

### DIFF
--- a/pgrx/src/rel.rs
+++ b/pgrx/src/rel.rs
@@ -14,7 +14,7 @@ use crate::{
 use pgrx_sql_entity_graph::metadata::{
     ArgumentError, Returns, ReturnsError, SqlMapping, SqlTranslatable,
 };
-use std::ops::Deref;
+use std::ops::{Deref, DerefMut};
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr:
I appreciate the effort, but I'm not sure making `PgRelation` mutable is really a great idea.  Doing so implies that those mutations actually do things, and in most cases they don't.

If someone wants to port those `pg_stat_count_*` macros, I'd happily merge that.  Seems I could use them in my day job, so it might be me that ends up doing it eventually.

Issue #681, opened 2022-09-10
Supporting `cargo binstall`?


workingjubilee:
At least one user mentioned they like to use `cargo binstall`, and it seems sensible enough, but that user also encountered mysterious problems when updating a project which leads me to suspect we may need to [tweak some details in our Cargo.toml metadata in order to fully support binstalls](https://github.com/cargo-bins/cargo-binstall/blob/main/SUPPORT.md). 

PGX and thus `cargo pgx` has some security concerns given the nature of the projects it builds, but our CI is secure enough that `cargo binstall` piggybacking off of it seems like a fine way of delivering a binary for common devtool usage. High-security operations which trust other networks less will want to build from our Rust source after obtaining a trusted Rust compiler, naturally, but our README and the like is effectively talking to the _average_ user who hasn't brought in any other considerations yet, which includes a lot of Jackie Devvers.

But, honestly, given that our next action is to recommend the user **fetch and build from source** all versions of Postgres that we support, uh... well, the "resource constrained devices" reasoning for supporting binstalls doesn't apply as much.

Important detail: We would also like our users to [use the lockfile we actually checked in](https://github.com/tcdi/pgx/pull/680), so we would want that to also be done by the binstall build. If that can be arranged, that itself is actually a reason to support them.

workingjubilee:
Okay, I don't think this is actually viable given issues in cargo-pgx mentioned in #687. We might be able to fix that so cargo-pgx is independent eventually, but not today.

usamoi:
Will we reopen this issue since we have removed `dlopen`?

PR #1958, opened 2024-12-09
add support for cargo-binstall

usamoi:
#681

The support could be added to all `0.12.*` versions.

See https://github.com/usamoi/pgrx/releases/tag/v0.12.7 for downloading binaries generated by this pull request.

### DIFF
--- a/.github/workflows/upload-release.yml
+++ b/.github/workflows/upload-release.yml
@@ -1,0 +1,64 @@
+name: upload release
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        type: string
+        description: Tag
+        required: true
+
+permissions:
+  contents: write
+
+jobs:
+  upload-release:
+    name: upload-release
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-22.04
+            target: x86_64-unknown-linux-musl
+            pkg_fmt: .tar.gz
+            binary_ext: ""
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr:
I'm very confused about what this PR is meant to do.  It adds a new GH workflow file, modifies some tests and (seemingly) reformats `unbox.rs`?

usamoi:
> I'm very confused about what this PR is meant to do. It adds a new GH workflow file, modifies some tests and (seemingly) reformats `unbox.rs`?

When this PR was submitted, without modifying tests and formatting `unbox.rs`, CI cannot pass.

It adds a new GH workflow file to build binaries for cargo-pgrx so that cargo-binstall could discover them.

Edit: removed now

eeeebbbbrrrr:
The new workflow attaches binaries to our tagged releases, right?

#681 was kinda left in a bit of a nebulous state.  @workingjubilee do we want to do this?  I can only see this being useful if the consumer doesn't intend to upgrade cargo-pgrx very often.

The "resource constrained machines" bit over in #681 is compelling, I suppose, but it also takes some serious resources to use cargo-pgrx in any meaningful way, so... still not sure of the value.

There's little cost on our end, so I won't block this if you think it's a good idea.

usamoi:
I want to install cargo-pgrx without building it in CI, and it's why I prefer `cargo binstall`...

Issue #1694, opened 2024-05-06
`prgx new` should generate `.github/workflow/ci.yaml` file


nyurik:
I think there should be a simple CI generated as part of the `prgx new`. Since most people tend to use github anyway, might as well give a github CI example - and it can be easily adjusted for other CI platforms if needed.

P.S. I think an example of packaging (on the release page) would also be very helpful, as it would save a lot of time for developers on figuring out how to package for different platforms.

```yaml
name: CI

on:
  push:
    branches: [ main ]
  pull_request:
    branches: [ main ]
  release:
    types: [ published ]
  workflow_dispatch:

defaults:
  run:
    shell: bash

jobs:
  test:
    name: Test
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: Swatinem/rust-cache@v2
        if: github.event_name != 'release' && github.event_name != 'workflow_dispatch'
      # fmt test can be done before installing pgrx
      - run: cargo fmt --all -- --check
      # PRGX must be installed in order to compile pgrx-pg-sys
      # PRGX version must match the version in Cargo.toml
      - run: cargo install cargo-pgrx --version 0.11.4 --locked --force
      - run: cargo pgrx init
      - run: RUSTFLAGS='-D warnings' cargo build --workspace --all-targets --bins --tests --lib --benches --examples
      - run: RUSTDOCFLAGS="-D warnings" cargo doc --no-deps
      - run: cargo clippy -- -D warnings
      - run: cargo clippy --workspace --all-targets --bins --tests --lib --benches --examples -- -D warnings
      - run: cargo pgrx test
```

eeeebbbbrrrr:
This idea has come up before.  Your example .yml file isn't quite enough.

It really needs to figure out the pgrx version used in `Cargo.toml` and `cargo install cargo-pgrx --version $THAT_VERSION --locked`.  Ideally it would also configure a matrix for all the supported postgres versions, use `apt` to install that version of postgres and `cargo pgrx init --pgXX=/path/to/$PGVER/bin/pg_config`.

It might be useful to have macos included in the build matrix too.  macos is a bit more difficult to configure, but it's doable.

nyurik:
[...]

PR #1959, opened 2024-12-09
Issue #1694/Github Actions ci.yaml file create

TimMensch:
Hi! This is a PR that addresses Issue #1694. I performed this work as part of an interview exercise, and I believe it is correct, but that's the end of my interest in the feature.

If you'd like any modifications, I happily pass the baton on to anyone else who wants to do them. If you don't like the PR as-is and don't want to change it, feel free to delete it. I did the work and would be happy for your project to benefit from it, but I'm not invested in the project at all; I don't even use Rust at the moment. 

So enjoy, or not, as you prefer. 😄

### DIFF
--- a/cargo-pgrx/src/command/new.rs
+++ b/cargo-pgrx/src/command/new.rs
@@ -54,6 +54,7 @@ pub(crate) fn create_crate_template(
     create_directory_structure(path.clone())?;
     create_control_file(path.clone(), name)?;
     create_cargo_toml(path.clone(), name)?;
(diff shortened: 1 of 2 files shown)


eeeebbbbrrrr:
I hope your interview went well.

I don't believe this is something we want to accept.  I don't think it's up to us, as maintainers of a framework, to dictate (or even suggest) how a user's CI should operate.  This will lead us down a path of "what about this?  what about that?" that isn't productive.

So, with my thanks, I'm going to close this.

PR #1934, opened 2024-11-01
fix cargo-pgrx and pgrx-tests on Windows

usamoi:
Notes:
1. postgres must run without administrator privileges, so switch calling `postgres` directly to starting postgres by `pg_ctl` in testing
2. Windows CI is added
3. known issue: `pgrx-cshim.c` is compiled with `-flto`, because, without this flag, `pgrx_embed.exe` is linked to `postgres.exe`
4. known issue: `cargo-pgrx` downloads, instead of building, postgres binaries from EDB website
5. known issue: `cargo pgrx run` and `cargo pgrx test` always print logs for `pg_ctl start` on Windows (pipes generated by `std::process::Command` are leaked to postgres, a command with `Stdio::piped()` hangs, so we use `Stdio::inherit()` on Windows)
6. breaking change: this pull request fixes `PgLwLock` and `PgAtomic` but introduces a breaking change about `PgLwLock` and `PgAtomic`: name must be provided at `PgLwLock/PgAtomic::new`, `PgLwLock::from_named` is removed and a parameter of `PgLwLock::attach` changes
7. breaking change: this pull request makes everything `C-unwind` and it's necessary for downstream to switch from `C` to `C-unwind`
8. regression: `cargo-pgrx` users need to execute `sudo sysctl fs.protected_fifos=0` before using `--runas` on Linux

PR #1981, opened 2025-02-24
Updated expected UI test outputs

YohDeadfall:


YohDeadfall:
It's broken for the same reason as #1934. That's a compiler regression.

eeeebbbbrrrr (approved):
Thank you for this, @YohDeadfall.

PR #1982, opened 2025-02-24
Port `bufpage.h` functions

rebasedming:
Ports functions from `bufpage.h` that are used to interact with block storage

### DIFF
--- a/pgrx-pg-sys/src/port.rs
+++ b/pgrx-pg-sys/src/port.rs
@@ -389,14 +389,34 @@ pub unsafe fn MemoryContextSwitchTo(context: crate::MemoryContext) -> crate::Mem
     old
 }
 
-pub use crate::PGERROR as ERROR;
+#[allow(non_snake_case)]
+#[inline(always)]
+#[cfg(any(feature = "pg12", feature = "pg13", feature = "pg14", feature = "pg15"))]
+pub unsafe fn BufferGetPageSize(_buffer: pg_sys::Buffer) -> pg_sys::Size {
+    // #define BufferGetPageSize(buffer) \
+    // ( \
+    //     AssertMacro(BufferIsValid(buffer)), \
+    //     (Size)BLCKSZ \
+    // )
+    pg_sys::BLCKSZ as pg_sys::Size
+}
+
+
+#[allow(non_snake_case)]
+#[inline(always)]
+#[cfg(any(feature = "pg12", feature = "pg13", feature = "pg14", feature = "pg15"))]
+pub unsafe fn ItemIdGetOffset(item_id: pg_sys::ItemId) -> u32 {
+    // #define ItemIdGetOffset(itemId) \
+    // ((itemId)->lp_off)
+    (*item_id).lp_off()
+}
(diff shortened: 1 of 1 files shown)


pgrx-pg-sys/src/port.rs
```diff
@@ -389,4 +389,158 @@ pub unsafe fn MemoryContextSwitchTo(context: crate::MemoryContext) -> crate::Mem
...
+#[allow(non_snake_case)]
+#[inline(always)]
+#[cfg(any(feature = "pg12", feature = "pg13", feature = "pg14", feature = "pg15"))]
+pub unsafe fn BufferGetPageSize(_buffer: pg_sys::Buffer) -> pg_sys::Size {
+    // #define BufferGetPageSize(buffer) \
+    // ( \
+    //     AssertMacro(BufferIsValid(buffer)), \
+    //     (Size)BLCKSZ \
+    // )
+    pg_sys::BLCKSZ as pg_sys::Size
```
eeeebbbbrrrr: For completeness, this should do what the `BufferIsValid()` macro does too.  Can you add another top-level function for that and then `assert!(BufferIsValid(buffer))` here, too?

```C
/*
 * BufferIsValid
 *		True iff the given buffer number is valid (either as a shared
 *		or local buffer).
 *
 * Note: For a long time this was defined the same as BufferIsPinned,
 * that is it would say False if you didn't hold a pin on the buffer.
 * I believe this was bogus and served only to mask logic errors.
 * Code should always know whether it has a buffer reference,
 * independently of the pin state.
 *
 * Note: For a further long time this was not quite the inverse of the
 * BufferIsInvalid() macro, in that it also did sanity checks to verify
 * that the buffer number was in range.  Most likely, this macro was
 * originally intended only to be used in assertions, but its use has
 * since expanded quite a bit, and the overhead of making those checks
 * even in non-assert-enabled builds can be significant.  Thus, we've
 * now demoted the range checks to assertions within the macro itself.
 */
static inline bool
BufferIsValid(Buffer bufnum)
{
	Assert(bufnum <= NBuffers);
	Assert(bufnum >= -NLocBuffer);

	return bufnum != InvalidBuffer;
}
```

I think the rust port would basically be:

```rust
/// <comments from above, please>
pub unsafe fn BufferIsValid(bufnum: pg_sys::Buffer) -> bool {
   assert!(bufnum <= pg_sys::NBuffers);
   assert!(bufnum >= -pg_sys::NLocBuffer);
   bufnum != pg_sys::InvalidBuffer
}

eeeebbbbrrrr (approved):
Just the one change and I think this will be good to merge.

Thanks, @rebasedming

eeeebbbbrrrr (changes requested):
oops, requesting changes

eeeebbbbrrrr:
I'll figure out what's wrong with CI -- appears to be unrelated to this PR.

eeeebbbbrrrr (approved):

eeeebbbbrrrr:
Thank you @rebasedming.  Merging!

Issue #1977, opened 2025-02-07
Install extension to a pgrx-managed Postgres instance


SergeiPatiakin:
It would be nice to have a pgrx command to install an extension into pgrx-managed instance $PG_MAJOR, without starting that instance or starting psql.

- `cargo pgrx install` installs to wherever pg_config points, not necessarily a pgrx-managed instance. A shell command like `cargo pgrx install "$PGRX_HOME"/$PG_MAJOR*/pgrx-install/bin/pg_config` does what I want, but it would be nice if it was a built-in command.
- `cargo run $PG_MAJOR` does what I want, but it also starts the instance and psql. 

The syntax for this improvement could be:
- a new optional PG_VERSION positional argument in `cargo pgrx install` (i.e. it would become `cargo pgrx install [OPTIONS] [PG_VERSION]`)
- a new option in `cargo pgrx run`, but I can't think of an appropriate name. `cargo pgrx run $PG_MAJOR --install-only`?

Does this makes sense? Maybe it's already built in but I couldn't find it?

eeeebbbbrrrr:
> a new option in cargo pgrx run, but I can't think of an appropriate name.` cargo pgrx run $PG_MAJOR --install-only`

This one seems reasonable.  Would you like to PR it?

SergeiPatiakin:
Yes, I can give it a go

YohDeadfall:
@eeeebbbbrrrr, it's completed by #1988.

PR #1988, opened 2025-02-27
--install-only option for cargo pgrx run command

SergeiPatiakin:
This PR closes issue #1977 

We allow the user to install an extension into a pgrx-managed Postgres instance without starting the instance or spawning a PSQL shell.

### DIFF
--- a/cargo-pgrx/src/command/run.rs
+++ b/cargo-pgrx/src/command/run.rs
@@ -48,6 +48,9 @@ pub(crate) struct Run {
     /// Use an existing `pgcli` on the $PATH.
     #[clap(env = "PGRX_PGCLI", long)]
     pgcli: bool,
+    /// Install without running
+    #[clap(long)]
+    install_only: bool,
 }
 
 impl CommandExecute for Run {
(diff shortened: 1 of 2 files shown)


cargo-pgrx/src/command/run.rs
```diff
@@ -120,6 +125,10 @@ pub(crate) fn run(
         features,
     )?;
 
+    if install_only {
+        return Ok(());
```
SergeiPatiakin, line 129: Alternatively this early return could be after `start_postgres(pg_config)?;`, I don't have a strong opinion

eeeebbbbrrrr:
Super simple! 

Would you mind updating https://github.com/pgcentralfoundation/pgrx/blob/develop/cargo-pgrx/README.md too?  It's good to have all the cli flags documented with a sentence or two about when/why they're useful.

SergeiPatiakin:
@eeeebbbbrrrr good idea, done!

eeeebbbbrrrr (approved):
This looks fine and good.  Thanks @SergeiPatiakin.  Once CI passes (one job might fail, CI has been acting up recently) I'll merge and you'll see this in the next release!

PR #1991, opened 2025-03-01
`pg_sys::Oid::from_datum()` should cast to `u32`

eeeebbbbrrrr:
As implemented in Postgres, the OID version of a `Datum` is a cast of the Datum's `uintptr_t` value to a `u32`, not a bounds checked version of that pointer value.

Adds a test for this.  It would fail because `pg_sys::Oid::try_from()` would return an Err on sufficiently large OID values, which we'd coerece into a None, causing our (at least) function argument handling to panic, indicating that it received a NULL post-FromDatum when it didn't.

Related, the `PgRelation::from_datum()` implementation is actually a REGCLASS behind the scenes, which is an OID, and it was similarly bugged.  This is kinda hard to write a test for as you'd need a relation of some kind in the schema with an OID over i32::MAX, and that's really tough to materialize without mucking around with the `pg_resetwal` tool, which is way outside the scope of what our test suite can handle.

### DIFF
--- a/pgrx-tests/src/tests/oid_tests.rs
+++ b/pgrx-tests/src/tests/oid_tests.rs
@@ -1,0 +1,39 @@
+//LICENSE Portions Copyright 2019-2021 ZomboDB, LLC.
+//LICENSE
+//LICENSE Portions Copyright 2021-2023 Technology Concepts & Design, Inc.
+//LICENSE
+//LICENSE Portions Copyright 2023-2023 PgCentral Foundation, Inc. <contact@pgcentral.org>
+//LICENSE
+//LICENSE All rights reserved.
+//LICENSE
+//LICENSE Use of this source code is governed by the MIT license that can be found in the LICENSE file.
+#[cfg(any(test, feature = "pg_test"))]
+#[pgrx::pg_schema]
+mod tests {
+    #[allow(unused_imports)]
+    use crate as pgrx_tests;
+
+    use pgrx::prelude::*;
+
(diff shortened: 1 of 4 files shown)


eeeebbbbrrrr:
@workingjubilee I'm gonna merge this once CI passes and cut a new release.  I'm confident in the changes here, but I wanted to make sure you see this.  It's kinda... funny.  :(

workingjubilee:
"fun"

pgrx-tests/src/tests/oid_tests.rs
```diff
@@ -0,0 +1,39 @@
...
+        assert_eq!(oid.as_u32(), 42);
+        Ok(())
+    }
+
+    #[pg_test]
+    fn test_completely_unreasonable_but_still_valid_oid() -> spi::Result<()> {
+        // nb:  this stupid value is greater than i32::MAX
+        let oid = Spi::get_one::<pg_sys::Oid>("SELECT tests.oid_roundtrip(2147483648)")?
+            .expect("SPI result was null");
+        assert_eq!(oid.as_u32(), 2_147_483_648);
```
workingjubilee, line 36: I think you might have mixed up a value here. `2_147_483_648` is above `i32::MAX` but the `.value()` is a usize, and then we converted via `TryInto` into `u32`. That should handle all values up to **`u32::MAX`**, which is much higher.

what has to happen instead is a garbage Datum is passed: the **higher** bits are set.

PR #1994, opened 2025-03-06
Enable cross compiling extensions

rhelmot:
Two pieces here in order to enable my successful build of some pgrx extensions from x86_64-linux to x86_64-freebsd:

- `cargo pgrx init --no-run` does not attempt to run the discovered postgreses, since they could be for the target architecture or OS.
- `--target` flag to some other commands is passed along to cargo build invocations and injected into the calculation of the build directory path. The builds done in order to generate bindings (?) are still done native, since the result has to run at build time.

Note that I was doing this within nix and nixpkgs, but I think it should generalize pretty well to the normal rust cross toolchains.

Note also that I haven't actually tested this diff precisely, since I was working with something that pinned, of all things, 0.12.0-alpha1. The rebase was pretty straightforward but if you want to see the diff that I know for sure works, check [here](https://github.com/rhelmot/pgrx/tree/alpha-patched).

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,26 @@ my_extension=# SELECT hello_my_extension();
 
 For more details on how to manage pgrx extensions see [Managing pgrx extensions](cargo-pgrx/README.md).
 
+## Cross-compiling
+
+So far, cross-compiling extensions with pgrx has only been demonstrated under [nix](https://nixos.org/).
+Proper support in nixpkgs is still in flux, so watch this space.
+
+In order to cross-compile outside of nix, you will need two ingredients:
+
+1) A sysroot for the cross architecture, with rust configured to invoke the linker with flags to link against this sysroot.
+2) A `pg_config` program that provides values associated with postgres compiled for the desired cross architecture.
+   This can be achieved by either emulating pg_config with qemu-user or similar, or replicating pg_config with a shell script.
+
+Then, modify the standard build process in the following ways:
+
+1) Pass `--no-run` to `cargo pgrx init`, since we cannot run postgres for the cross machine.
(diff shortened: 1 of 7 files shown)


wolfgangwalther:
@rhelmot did you look at #981, too? Maybe that could be revived.

rhelmot:
I had not seen that. Based on the diff, it seems pretty similar to this, with three additional knobs, two for sysroots and one for an alternative pg_config. All of those are pretty unnecessary if you're compiling though nixpkgs, and I would have thought that the sysroots wouldn't be necessary because of the comprehensive nature of the rust cross toolchain, though maybe the C pieces in the dependency tree care about this?

pg_config is concerning though. I am not actually familiar enough with pgrx to understand why we need pg_config for the build platform in addition to the host.

wolfgangwalther:
> pg_config is concerning though. I am not actually familiar enough with pgrx to understand why we need pg_config for the build platform in addition to the host.

Yes, that confuses me as well, and I am not familiar enough with it either.

eeeebbbbrrrr (changes requested):
This looks okay.  Before merging, however, I'd like to see the `cargo-pgrx/README.md` updated with these new command line options.

Ideally, adding a short section there to discuss cross-compilation would be nice too.

eeeebbbbrrrr:
Hi @rhelmot, I'd like to merge this if you can address my comment here:  https://github.com/pgcentralfoundation/pgrx/pull/1994#pullrequestreview-2698133000

rhelmot:
Rebased and added a README section. I'll come back and update the nix stuff when the dust has settled over at nixpkgs.

eeeebbbbrrrr (approved):
Approved.  I'll merge once CI passes

PR #1999, opened 2025-03-13
`DatumWithOid`: Add `new_from_datum(Option<Datum>, Oid)` and `null_oid(Oid)`

### LINKED ISSUE #1993: DatumWithOid doesn't allow dynamic creation of NULLs anymore
I have some code which does the following (inelegant, I know): 

```rust
enum SqlValue {
  Int4(i32),
  Text(String),
  ...
  Null(Oid)
}
```

That is, for non-null values, the OID is inferred from the enum variant. For null, it's encoded in `.0` of the `Null` variant.

The old API allowed me to dynamically construct `NULL`s of any type by doing `(PgBuiltinOids::TEXTOID.oid(), None)`. The new API has no way to do that, as it's only possible to create `DatumWithOid` from a `T: IntoDatum`. 

Can we get a new (unsafe) function `DatumWithOid` which takes an `Oid` and an `Option<Datum>` directly instead of going through `IntoDatum`? That would support this use case. 

I know this API is suboptimal and I could refactor it to represent `NULL` more nicely, but that abstraction spans multiple projects and it's a large amount of work.

the-kenny:
This allows the creation of dynamically typed `DatumWithOid` values where the type isn't known at compile time.

Fixes #1993

### DIFF
--- a/pgrx/src/datum/mod.rs
+++ b/pgrx/src/datum/mod.rs
@@ -171,7 +171,22 @@ impl<'src> DatumWithOid<'src> {
     ///
     /// [`Oid`]: pg_sys::Oid
     pub unsafe fn new<T: IntoDatum>(value: T, oid: pg_sys::Oid) -> Self {
-        Self { datum: value.into_datum().map(|d| Datum(d, PhantomData::default())), oid }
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr (approved):

eeeebbbbrrrr:
Thanks for the contribution!

PR #2002, opened 2025-03-17
Instruct `rustfmt` to use `--edition 2021`

eeeebbbbrrrr:
The problem here is that `pgrx-bindgen`'s `rust_fmt()` step was (silently) failing because `rustfmt` without `--edition 2021` doesn't understand rust `c"C Cstrings"`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -214,14 +214,15 @@ dependencies = [
 [[package]]
 name = "bindgen"
 version = "0.71.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5f58bf3d7db68cfbac37cfc485a8d711e87e064c3d0fe0435b92f7a407f9d6b3"
+source = "git+https://github.com/rust-lang/rust-bindgen.git?rev=20aa65a#20aa65a0b9edfd5f8ab3e038197da5cb2c52ff18"
 dependencies = [
  "annotate-snippets",
  "bitflags 2.8.0",
  "cexpr",
  "clang-sys",
  "itertools",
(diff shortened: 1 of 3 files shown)


YohDeadfall:
> It's incredibly annoying to not have our pgXX.rs file be formatted.

That explains the issue I faced today where one of these files contained a single line of code and therefore was useless.

eeeebbbbrrrr:
I've updated this PR to set

```toml
bindgen = { git = "https://github.com/rust-lang/rust-bindgen.git", rev = "20aa65a", features = ["experimental", "runtime"] }
```

In `pgrx-bindgen/Cargo.toml`.  We gotta do it.  It's surprising to me that the bindgen folks haven't released something that's compatible with Edition 2024 yet, but here we are.  

Assuming CI passes, I'm going to merge this.

eeeebbbbrrrr:
```
error: all dependencies must have a version specified when packaging.
dependency `bindgen` does not specify a version
Note: The packaged dependency will use the version from crates.io,
the `git` specification will be removed from the dependency declaration.
```

Dangit.
giphy

YohDeadfall:
Sign together with Niko about your dependencies: https://youtu.be/thdpaw_3VTw

usamoi:
`pgrx-pg-sys` still uses `edition = "2021"`. What about just using `--edition 2021`? A few months later, when we try to update pgrx to `edition = "2024"`, bindgen might have released a new version by then.

eeeebbbbrrrr:
Hmm. I didn’t try that. I’ll give it a shot and see what happens.

PR #2003, opened 2025-03-17
Added mapping for `xid`

YohDeadfall:


pgrx/src/xid.rs
```diff
@@ -36,3 +43,84 @@ fn convert_xid_common(xid: pg_sys::TransactionId, last_xid: u32, epoch: u32) ->
 
     (epoch << 32) | xid as u64
 }
+
+/// An `inet` type from PostgreSQL
+#[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq)]
+pub struct Xid(pg_sys::TransactionId);
```
eeeebbbbrrrr: Is this what we want?  

I'm not _super_ keen on wrapping such a fundamental internal Postgres type.  I feel like doing so implies there's some difference between this new `Xid` type and the Postgres internal `pg_sys::TransactionId` -- and there's not.

I'm assuming you did this mostly for the serde support?  `pg_sys::TransactionId` will serialize just fine as it's a plain `u32`.

eeeebbbbrrrr: Additionally, this isn't "An `inet` type from PostgreSQL"

YohDeadfall: Not because of `serde`, but for proper scheme generation, to have `xid` type in function signatures and results.

Will fix the comment, thanks!

eeeebbbbrrrr: > Not because of serde, but for proper scheme generation

The Rust type and SQL type don't need to be spelled the same.

You can `impl SqlTranslatable for pg_sys::TransactionId`, which is what I'd prefer.  I really don't like the idea of us having a wrapper type around `pg_sys::TransactionId`.

YohDeadfall: I would do that, but then any `u32` will be `xid`. Do you think that `CommandId` or any other alias of `u32` is `xid`? I don't.

https://docs.rs/pgrx-pg-sys/0.13.1/src/pgrx_pg_sys/include/pg14.rs.html#5762

eeeebbbbrrrr: You are correct.  I had forgotten that type aliases are presented to our schema generator as the type they alias.

Then what we need is some handing around `pg_sys::TransactionId` in a manner similar to how we have our own `pg_sys::Datum` impl.  This way, the type is still called `pg_sys::TransactionId` internally, which is my primary requirement here, and we can add whatever From traits or otherwise we think we want directly in the `pgrx-pg-sys` crate.

pgrx-pg-sys/src/submodules/transaction_id.rs
```diff
@@ -0,0 +1,74 @@
...
+//LICENSE
+//LICENSE Use of this source code is governed by the MIT license that can be found in the LICENSE file.
+use pgrx_sql_entity_graph::metadata::{
+    ArgumentError, Returns, ReturnsError, SqlMapping, SqlTranslatable,
+};
+
+pub type MultiXactId = TransactionId;
+
+/// An `xid` type from PostgreSQL
+#[repr(C)]
```
eeeebbbbrrrr: I think we want this to be `#[repr(transparent)]`.  

The reasons being that it communicates to the user that our `struct TransactionId` is wrapping some other type and it will also cause some friction for anyone that tries to PR changing this struct by adding another field.

eeeebbbbrrrr (changes requested):

eeeebbbbrrrr:
@YohDeadfall I think this looks good, outside of the one minor change.  I think #2014 is going to force a pgrx v0.14.0, so now is a great time to go ahead and get this finished and committed.

pgrx-pg-sys/src/submodules/transaction_id.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub const FIRST_NORMAL: Self = Self::from_u32(3);
+    pub const MAX: Self = Self::from_u32(u32::MAX);
+
+    #[inline]
+    pub const fn from_u32(xid: u32) -> Self {
+        Self(xid)
+    }
+
+    #[inline]
+    pub const fn to_u32(self) -> u32 {
```
YohDeadfall: Is it really fine? I would like to avoid any further name bikeshading as it's a little disaster even if another method added and this marked as deprecated. Shouldn't it be `into_u32`?

eeeebbbbrrrr: honestly, we should call this `into_inner(self)` and then do the same over on `pg_sys::Oid` and close #2011.

And there's also no need for this `from_u32(xid: u32)` function either, because of `impl From<u32> for TransactionId` down below.

pgrx-pg-sys/src/submodules/transaction_id.rs
```diff
@@ -0,0 +1,74 @@
...
+//LICENSE Use of this source code is governed by the MIT license that can be found in the LICENSE file.
+use pgrx_sql_entity_graph::metadata::{
+    ArgumentError, Returns, ReturnsError, SqlMapping, SqlTranslatable,
+};
+
+pub type MultiXactId = TransactionId;
+
+/// An `xid` type from PostgreSQL
+#[repr(transparent)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
eeeebbbbrrrr: I think we also want to `impl Display` too.  Have it just be the u32 value.  It's not uncommon to want to convert a transaction id into a string for logging purposes

eeeebbbbrrrr (changes requested):

pgrx-pg-sys/src/submodules/transaction_id.rs
YohDeadfall: `from_u32` is for `const` only since traits cannot have that modifier on the functions.  It can be removed, but then it won't be possible to use during compiler time.

If there's no need in `const fn`s then I see no reason in `from_smthg` and `into_smthg` at all as there are `from` and `into`. Makes sense?

eeeebbbbrrrr (approved):
Thanks for your diligence on this one, @YohDeadfall !

PR #2004, opened 2025-03-17
Pinned cargo-edit version 

YohDeadfall:
Since `cargo install` doesn't use the lockfile by default, the CI can fail if any dependency is updated like here: https://github.com/pgcentralfoundation/pgrx/actions/runs/13908033860/job/38915524055?pr=2003.

This pull request fixes the issue without bumping the dependency.


workingjubilee:
I think this is the wrong fix. `cargo_edit`'s Cargo.toml doesn't track versioning for the library: it tracks versioning for the binary. They recommend pinning `cargo_edit`, so let's do that.

YohDeadfall:
There's an issue with the CI, but it's unrelated to my changes at all. It's something with Microsoft/GitHub itself.

workingjubilee:
- [ ]

PR #2006, opened 2025-03-18
tell C compiler to put code and data in separate sections

usamoi:
Fixes a known issue (cshim requires LTO) in https://github.com/pgcentralfoundation/pgrx/pull/1934.


workingjubilee:
...huh.

PR #2011, opened 2025-03-21
Renamed Oid::as_u32 to to_u32

YohDeadfall:
The reason behind that is that `as_*` methods doesn't consume `self` but a reference, and `to_*` methods do conversion by consuming `self`.

### DIFF
--- a/pgrx-pg-sys/src/submodules/oids.rs
+++ b/pgrx-pg-sys/src/submodules/oids.rs
@@ -82,7 +82,7 @@ impl Oid {
         }
     }
 
-    pub const fn as_u32(self) -> u32 {
+    pub const fn to_u32(self) -> u32 {
         self.0
     }
 }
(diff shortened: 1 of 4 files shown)


vrmiguel:
This doesn't conform to the [Rust API guidelines](http://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv).

"Conversions prefixed as_ and into_ typically decrease abstraction, either **exposing a view into the underlying representation (as)** or deconstructing data into its underlying representation (into).
Conversions prefixed to_, on the other hand, typically stay at the same level of abstraction but do some work to change from one representation to another."

YohDeadfall:
Hard to disagree, but [there are many uses of `to_` in `std` itself](https://doc.rust-lang.org/std/convert/trait.TryFrom.html?search=to_) even for cheap things like `to_bits`. Saw the same thing in `bevy` too. Anyway, as there's a guideline, it's better to follow it. Thanks!

YohDeadfall:
Though maybe it's better to ask @workingjubilee first to avoid unnecessary changes.

eeeebbbbrrrr:
I appreciate the effort but I don't think this is a change we should make now.  Perhaps when we get ready for a larger v0.14.0 with some breaking changes this could make sense, but in our current v0.13.x release series it seems like it'll be more detrimental to users than helpful.

YohDeadfall:
The thing here is that we need to decide how the method should be named as #2003 also does that too. As long as you are here, decide what's the correct option.

eeeebbbbrrrr (approved):
I'm good with this.  Thanks

PR #2012, opened 2025-03-21
feat: add `PgHeapTuple::set_by_index_unchecked`

silver-ymz:
This PR adds an unsafe version `PgHeapTuple::set_by_index` which doesn't check if the datum is compatible with the attribute's type.

I need to set a general datum in the heap tuple. And the datum may not implement `FromDatum` and `IntoDatum`. So original `PgHeapTuple::set_by_index` seems to be unusable.


eeeebbbbrrrr (approved):
Thanks, @silver-ymz!

PR #2014, opened 2025-03-22
force using extern "C-unwind" for functions marked with #[pg_guard]

usamoi:
`pgrx` is switched from "C" to "C-unwind" in #1934 so it's fine to emit an error here

Without this error, someone who updates dependencies without reading changelog, would break extensions silently, if their only usage of `#[pg_guard]` is `_PG_init`. It's bad.

### DIFF
--- a/pgrx-macros/src/rewriter.rs
+++ b/pgrx-macros/src/rewriter.rs
@@ -47,6 +47,10 @@ pub fn item_fn_without_rewrite(mut func: ItemFn) -> syn::Result<proc_macro2::Tok
     let attrs = mem::take(&mut func.attrs);
     let generics = func.sig.generics.clone();
 
+    if !sig.abi.clone().and_then(|abi| abi.name).is_some_and(|name| name.value() == "C-unwind") {
+        panic!("#[pg_guard] must be combined with extern \"C-unwind\"");
+    }
+
     if attrs.iter().any(|attr| attr.path().is_ident("no_mangle"))
         && generics.params.iter().any(|p| match p {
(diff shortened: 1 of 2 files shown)


eeeebbbbrrrr:
What's the down-stream impact to users with this change?  I'm hesitant to merge because it's not clear to me if extensions out in the wild will need to be updated because of this.

usamoi:
> What's the down-stream impact to users with this change?

The downstream must manually switch the ABI of functions marked with `#[pg_guard]` from `C` to `C-unwind`. Otherwise there are errors.

If we don't emit an error and a downstream function marked with `#[pg_guard]` uses `C` ABI, then when the extension is compiled for Windows, all errors will be converted into an abort.

Even if we don't emit an error, downstream cannot simply do nothing, because the ABI of the callback function types generated by bindgen has changed from `C` to `C-unwind`.

eeeebbbbrrrr:
> Even if we don't emit an error, downstream cannot simply do nothing, because the ABI of the callback function types generated by bindgen has changed from C to C-unwind.

I think this forces us to move to pgrx v0.14.0.  Not that that's a problem, but it seems like a big enough change to warrant that.

What do you think, @usamoi?

usamoi:
I also think the next version must be `0.14.0`. #1934 did another breaking change about shmem API. This is also a reason to bump the minor version.

eeeebbbbrrrr:
Okay.  I'll do that this weekend.  I appreciate your time and input!

And I'll just go ahead and merge this now...

PR #2017, opened 2025-03-26
Dedup array `IntoDatum`, drop intermediate Vecs for `tuples`

vrmiguel:
Moves duplicated `IntoDatum` code from  for `&[T]` and `Vec<T>` into a single function, and makes tuple's `IntoDatum` not allocate intermediate Vecs

### DIFF
--- a/pgrx/src/datum/array.rs
+++ b/pgrx/src/datum/array.rs
@@ -1019,6 +1019,7 @@ fn array_datum_from_iter<T: IntoDatum>(elements: impl Iterator<Item = T>) -> Opt
         pg_sys::initArrayResult(
             T::type_oid(),
             PgMemoryContexts::CurrentMemoryContext.value(),
+            // All elements use the same memory context
             false,
         )
     };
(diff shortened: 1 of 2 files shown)


pgrx/src/datum/array.rs
```diff
@@ -1012,41 +1012,47 @@ where
...
+                isnull,
+                T::type_oid(),
+                PgMemoryContexts::CurrentMemoryContext.value(),
+            );
+        }
+    }
+
+    if state.is_null() {
+        // shouldn't happen
+        None
```
eeeebbbbrrrr: I'm probably being overly pedantic because yes, this shouldn't happen.  

So if it does, lets `panic!()` here instead.  Silently returning `None` would be very confusing.

eeeebbbbrrrr (changes requested):
Thanks @vrmiguel!  Just one little nit and I think this will be ready to merge.

pgrx/src/datum/array.rs
vrmiguel: Done!

We could also switch to `initArrayResultWithSize` eventually

eeeebbbbrrrr (approved):

PR #2020, opened 2025-03-28
Fix formatting of bindgenned files by piping through `prettyplease`.

thomcc:
Currently, some bindgenned files (the pg17.rs in my work codebase) end up too big for rustfmt to handle, so it just gives up and spits out the input directly. The tool designed for this purpose is the [`prettyplease`](https://docs.rs/prettyplease/latest/prettyplease/) crate, which produces semi-formatted output (not identical to rustfmt), that rustfmt can take as input.


eeeebbbbrrrr:
@thomcc I believe that https://github.com/pgcentralfoundation/pgrx/pull/2002 fixes this properly.

thomcc:
Ah, alright. I'll assume it does.

PR #2022, opened 2025-03-29
Disabled Will It Bend on forks

YohDeadfall:
As Will It Bend is fixed now that change isn't important as should be no GitHub notifications about CI failures every week. Still. it's better to disable the workflow for forks as it's designed for maintainers to keep the project healthy.

### DIFF
--- a/.github/workflows/will-it-blend-develop.yml
+++ b/.github/workflows/will-it-blend-develop.yml
@@ -14,6 +14,7 @@ env:
 
 jobs:
   distro_tests:
+    if: github.repository_owner == 'pgcentralfoundation'
     name: Distro tests
     runs-on: ubuntu-22.04
     strategy:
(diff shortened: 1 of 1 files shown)


usamoi:
Screenshot_20250330-071800_Trebuchet

It can be disabled manually on forks.

YohDeadfall:
Yes, but why keep it for forks when it's intended for the owners? It can be disabled manually and even GitHub will do that automatically if the fork isn't active for a while, so I'm unsure about the change. And yet, it's not a story for forks as I said. That's why I'm not sure that it's a worth change. Though that's up for @eeeebbbbrrrr to decide what is right here.

YohDeadfall:
On the other hand without spamming the fork owners about failures the chances of it being fixed for this repo goes down to zero. So, let's keep it, there's a good reason for that.

PR #2023, opened 2025-03-31
include `utils/spccache.h`

usamoi:
https://github.com/postgres/postgres/blob/master/src/include/utils/spccache.h

It's used to get `spc_random_page_cost`, `spc_seq_page_cost`, `io_concurrency` and `maintenance_io_concurrency` for a table.

### DIFF
--- a/pgrx-pg-sys/include/pg13.h
+++ b/pgrx-pg-sys/include/pg13.h
@@ -191,6 +191,7 @@
 #include "utils/selfuncs.h"
 #include "utils/snapmgr.h"
 #include "utils/sortsupport.h"
+#include "utils/spccache.h"
 #include "utils/catcache.h"
 #include "utils/syscache.h"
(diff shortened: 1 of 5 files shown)


vrmiguel:
How does the bindgen process work for pgrx? (cc  @eeeebbbbrrrr)

Would `spc_random_page_cost` only appear in `pgrx-pg-sys/src/include/pg{version}.rs` whenever pgrx-bindgen gets a bump?

YohDeadfall:
There is documentation about that:
 https://github.com/pgcentralfoundation/pgrx/blob/develop/docs%2Fsrc%2Fdesign-decisions.md#L71-L75

So, updating the headers list should be enough to get new units included into `pg_sys`, but for consumers there should be a new version of `pgrx` to include that change or a repo reference instead as a dependency.

PR #2025, opened 2025-04-04
Elide needless lifetime

nyurik:
Used clippy to find `clippy::needless_lifetimes` and fix them


nyurik:
CI fails due to a breaking change in zip2 - see https://github.com/zip-rs/zip2/issues/333

eeeebbbbrrrr:
Looks like this PR doesn't want to pass CI.

eeeebbbbrrrr:
It's been awhile since there was an update here.  Closing.

PR #2029, opened 2025-04-06
Implemented `Display` for `Oid`

YohDeadfall:
The purpose of this change is to forward `Debug` and `Display` implementations of the `Oid` type to its underlying field, so it's possible to use it for logging.


eeeebbbbrrrr (approved):
Thank you.  I appreciate and welcome this change.

PR #2033, opened 2025-04-11
Fixed Discord invitation link

YohDeadfall:
It fixes the link, but we also have a badge that looks broken as it's disabled by the server administrator. Should it then removed at all or enabled in the server settings?

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
 
 ## Want to chat with us or get a question answered?
 
-   + **Please join our [Discord Server](https://discord.gg/PMrpdJsqcJ).**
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr:
I dunno.  Getting the invite link correct seems good enough for now.  Thanks.  Can't say I've ever clicked on it!

YohDeadfall:
That's because there are two links. The one I fixed led to `ZomboDB` resting in peace, but the second is correct though the badge is deactivated:

image

Again, it would be cool to fix it or remove.

PR #2034, opened 2025-04-11
Fixed error message if no artifact found

YohDeadfall:


cargo-pgrx/src/command/install.rs
```diff
@@ -418,12 +418,12 @@ pub(crate) fn find_library_file(
     manifest_path: &Path,
     build_command_messages: &Vec<CargoMessage>,
 ) -> eyre::Result<PathBuf> {
-    use std::env::consts::{DLL_EXTENSION, DLL_SUFFIX};
+    use std::env::consts::DLL_EXTENSION;
 
     // cargo sometimes decides to change whether targets are kebab-case or snake_case in metadata,
     // so normalize away the difference
-    let target_name = manifest.target_name()?.replace('-', "_");
```
eeeebbbbrrrr: What's the reason for removing line 425 but not also updating the comment?  Now I don't know what the intent is supposed to be.

eeeebbbbrrrr (changes requested):

eeeebbbbrrrr:
Going to hold this until after v0.14.0 is released.

cargo-pgrx/src/command/install.rs
YohDeadfall: Removed it as it's useless now according to the selected logic.

eeeebbbbrrrr (approved):

PR #2039, opened 2025-04-12
update dependencies and fix compilation problem on pg13

eeeebbbbrrrr:
upgrades all our dependencies that needed it and also fixes a compilation problem with pg13 related to date/time part extraction.  

I don't know why that didn't show up in CI, but it did during the "cargo check" our `upgrade-deps.sh` script does.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -44,6 +44,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "allocator-api2"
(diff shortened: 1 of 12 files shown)


eeeebbbbrrrr:
I approve my changes and will merge once CI agrees with me.

Then it'll be v0.14.0 time!!!!!

PR #2041, opened 2025-04-12
fix `cargo pgrx install` error if there is a build script

usamoi:
If there is a build script for extension, `cargo pgrx install` would error because `cargo-pgrx` thinks the build script artifact is expected dynamic library artifact.

### DIFF
--- a/cargo-pgrx/src/command/install.rs
+++ b/cargo-pgrx/src/command/install.rs
@@ -436,7 +436,10 @@ pub(crate) fn find_library_file(
             _ => None,
         })
         // normalize being flattened and low to the ground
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr:
Is this something that always happens with v0.14.0?  Do we need to put out v0.14.1?  What change caused this to start happening?

usamoi:
> Is this something that always happens with v0.14.0?

It happens with v0.14.0, if there is a build script for the extension.

> Do we need to put out v0.14.1?

I feel yes, because it breaks my code😆

> What change caused this to start happening?

https://github.com/pgcentralfoundation/pgrx/pull/2032

eeeebbbbrrrr (approved):

PR #2051, opened 2025-04-28
Remove the $libdir/ prefix from module_pathname

theory:
It limits the search to PKGLIBDIR; if anyone installs it into another directory in `dynamic_library_path`, Postgres won't be able to find it.

[Details](https://justatheory.com/2025/04/update-control/).

I think pgrx should also consider adding something equivalent to the PGRX `prefix` param to allow the sharedir and pkglibdir files to be installed somewhere other than where `pg_config` suggests to properly support `dynamic_library_path` and the new `extension_control_path` configs.

### DIFF
--- a/cargo-pgrx/README.md
+++ b/cargo-pgrx/README.md
@@ -699,7 +699,7 @@ AS 'MODULE_PATHNAME', 'hello_extension_wrapper';
 ```
 
 `MODULE_PATHNAME` is replaced by Postgres with the configured value in the `.control` file. For pgrx-based extensions,
-this is  usually set to `$libdir/<extension-name>`.
+this is  usually set to `<extension-name>`.
 
 When using versioned shared-object support, the same SQL would look as follows:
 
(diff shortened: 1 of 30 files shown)


eeeebbbbrrrr:
This is probably just fine to merge, so I'll go ahead and do that.  

I'm not up-to-date on `dynamic_library_path` and `extension_control_path`, so I'll have to do some reading.  What did you have in mind, @theory?

theory:
Right now pgrx exclusively installs shared libraries in PKGLIBDIR and control files and SQL file sin SHAREDIR. What it needs, I think, is a `--prefix` option that installs them under a different prefix. The analog is the PGXS `prefix=` option, documented [here](https://www.postgresql.org/docs/devel/extend-pgxs.html) (search for "You can select a separate directory prefix"). For better or worse, it should probably have the same behavior, adding `postgresql` to the end if neither `postgres` nor `pgsql` appear in the value. Then pgrx should append `share` for the sharedir, `share/extensions` for the control and SQL files, `share/$dir` if the Control file `directory` option is specified. It should append `lib` for pkglibdir. Presumably doc files would go under `$prefix/docs`.

PR #2054, opened 2025-05-04
pass bundle and bundle_loader to linker on macos

### LINKED ISSUE #731: Avoid needing extensions to pass `-Wl,-undefined,dynamic_lookup` to the linker
`pgx`-using extensions currently need to [pass `-Wl,-undefined,dynamic_lookup` to the linker](https://github.com/tcdi/pgx/blob/master/.cargo/config). This is a flag that fixes an error on targets that use Mach-O as their binary format (e.g. Apple targets, of which we only care about macOS), but it looks like we pass it always, and the other linkers seem to ignore it.

This flag works around a issue caused by symbol resolution on Mach-O. Specifically: due to the two-level namespace the linker/loader can't fully delay symbol resolution until runtime efficiently unless it knows the dylib it will appear in, so the linker errors if it hits undefined symbol by default (rather than defaulting to the slow path). We pass a flag that changes the handling of undefined symbols to `dynamic_lookup`, forcing the loader to resolve each undefined symbol by searching every dylib in the program (including ones referenced transitively) for a symbol with a matching name (erroring iff it fails to locate the symbol in any of them).

This is both easy and completely avoids the link-time unknown symbol errors that we'd be otherwise guaranteed to get for our use case. But makes loading the extension slower and is quite discouraged[^2] by Apple (although it's very unlikely they'll change anything on macOS here).

[^2]: It's deprecated on all of Apples targets except macOS, but it's probably not possible to run postgres on iOS, tvOS, or watchOS, and on macOS there are some tools that will warn about it due to the performance cost.

The real impact is more along the lines of the following:

1. It's a point of friction for users of `pgx` on these platforms, as they need configure that flag. Failing to do it results in an inscrutable error. Thankfully, `cargo pgx new` does a lot to help solve this. Still, it would be nice to remove such a mysterious (for most) piece of the generated template.
2. It makes it more difficult for users to detect issues elsewhere in their code, as it turns most linker errors (missing library, typo in `extern "C" { ... }`, ...) aren't caught until load time, when normally they'd be caught at link time. More generally, it is likely to make other linker/loader issues harder to track down
3. Some perf, as mentioned. Because this cost is `O(undef_symbols * dylibs_in_program)` it may be quite expensive, as there seem to be a lot of symbols in Postgres, and `dylibs_in_program` can be surprisingly large. But also it might not be actually noticable in practice and is a one-time cost inside the loader (e.g. dlopen) anyway.

I think the best solution is to build a `.tbd` file[^3] describing the symbols provided by the postgres dylibs, and link against that. I am pretty sure the CommandLineTools has something for creating tbd files from dylibs, but we may need to tweak it for this purpose, or craft the file ourselves (it is not especially complex YAML) after getting the symbol info for each dylib ourselves. Linking with such a TBD in the build.rs would remove the need for dynamic lookup (with its associated downsides). It's probably overkill to do this at the moment, but this flag causes weird issues and is discouraged, so I figured I'd file a bug about it while it was on my mind.

[^3]: TBD is a YAML-based file format that can be passed to the linker with as if it were a dylib, and allows resolving the symbols to which dylib they come from (it is basically the equivalent to import libraries on windows, except with 100% more yaml). It allows linking aginst dylibs that do not actually exist at link time, which is the case for almost all system libraries on macOS these days (dyld shared cache and all).

usamoi:
fix #731

Since this pull request requires all extensions to include a build script, it introduces a breaking change to the API.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -3,7 +3,3 @@ b = "build --features"
 c = "check --features"
 t = "test --features"
 r = "run --features"
-
-[target.'cfg(target_os="macos")']
-# Postgres symbols won't be available until runtime
-rustflags = ["-Clink-arg=-Wl,-undefined,dynamic_lookup"]
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -466,10 +466,15 @@ jobs:
       # hack Cargo.toml to use this version of pgrx from github
       - name: hack Cargo.toml
         run: |
-          echo "[patch.crates-io]" >> /tmp/sample/Cargo.toml
-          echo "pgrx        = { path = \"${GITHUB_WORKSPACE}/pgrx\"        }" >> /tmp/sample/Cargo.toml
-          echo "pgrx-macros = { path = \"${GITHUB_WORKSPACE}/pgrx-macros\" }" >> /tmp/sample/Cargo.toml
-          echo "pgrx-tests  = { path = \"${GITHUB_WORKSPACE}/pgrx-tests\"  }" >> /tmp/sample/Cargo.toml
+          cat <<EOF >> /tmp/sample/Cargo.toml
+          [patch.crates-io]
+          pgrx = { path = "${GITHUB_WORKSPACE}/pgrx" }
+          pgrx-macros = { path = "${GITHUB_WORKSPACE}/pgrx-macros" }
+          pgrx-pg-config = { path = "${GITHUB_WORKSPACE}/pgrx-pg-config" }
+          pgrx-pg-sys = { path = "${GITHUB_WORKSPACE}/pgrx-pg-sys" }
+          pgrx-sql-entity-graph = { path = "${GITHUB_WORKSPACE}/pgrx-sql-entity-graph" }
+          pgrx-tests = { path = "${GITHUB_WORKSPACE}/pgrx-tests" }
+          EOF
 
       - name: show Cargo.toml
         run: cat /tmp/sample/Cargo.toml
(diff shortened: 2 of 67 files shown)


eeeebbbbrrrr:
@usamoi I'm not sure this is a good idea.  I'm not keen on forcing every downstream user to have to add a `build.rs` that's essentially only for MacOS.  I feel like this will cause a ton of friction for pgrx.

Are there any other options here?

usamoi:
> Are there any other options here?

* `rustc-cdylib-link-arg` works for dependents, but emits an warning (https://github.com/rust-lang/cargo/issues/9562)
* https://github.com/rust-lang/cargo/issues/14903 allows extensions to run pgrx's build script by editing `Cargo.toml` only, but we don't know when it's available
* adding support for importing symbols from executable files by `raw-dylib` for Mach-O targets; I can't see anyone is going to implement this feature for Rust

eeeebbbbrrrr:
@usamoi I kinda think that continuing to require a `.cargo/config.toml` file that passes the necessary compiler flags is superior to needing a `build.rs` and a `[build-dependencies]` section for every pgrx-based extension.

If you've got a better idea, I'm all ears, but `cargo pgrx new` sets up the necessary `.cargo/config.toml` (and has for a few years now), so I just don't see a lot of value in such a breaking change.

usamoi:
> I just don't see a lot of value in such a breaking change.

Maybe. After all, there haven't been any bug reports in pgrx indicating that this actually causes issues for the extension. Let's leave this PR here and see if anyone needs it in the future.

eeeebbbbrrrr:
hey @usamoi, I guess we basically decided not to do anything on this issue so I'm going to close it.  I need to protect myself from accidentally merging it.  ;)

Issue #2052, opened 2025-04-30
PG18 Support


anth0nyleung:
Hello. The first pg18 beta version is going to be release in about 3 weeks. I could not find any issues tracking the support for Pg18, hence creating a new issue to track this. 

I am curious about whether there's any timeline on supporting that or would it be possible to start on that with the current 18_devel branch.

Thank you.

YohDeadfall:
I plan to work on that, but there's a rule about supporting only final versions. Therefore, it won't be merged till the last moment.

anth0nyleung:
Thanks @YohDeadfall ! Im curious where this rule is defined and the reason behind it. Could you also please clarify on what you mean by final version? do you mean the first beta version / rc / .0 version?

YohDeadfall:
The 18.0 version. See https://github.com/paradedb/paradedb/issues/1019#issuecomment-2836716401.

VoVAllen:
Hi, we have a branch to support PG18 and already tested it with our pgrx extension. Check it at https://github.com/silver-ymz/pgrx/tree/support-18-devel cc @silver-ymz

YohDeadfall:
Oh, that's cool! Could you please open a pull request then?

silver-ymz:
> Could you please open a pull request then?

Yes, I'd like to open a PR then.

jakob:
Sorry for asking stupid questions, but does that mean that we have to wait until PostgreSQL 18 is released in autumn before we can build extensions with pgrx for it?

I'm unfamiliar with Rust / crago / pgrx, I just want to build extensions for PostgreSQL 18beta1 for testing, so that we can release it simultaneously with PostgreSQL 18.

The following commands fail (I assume that is because PostgreSQL 18 is not supported yet)

```sh
cargo install cargo-pgrx # as far as I can tell this installs pgrx 14.3
cargo pgrx init --pg18 # suggests using the argument --pg17 instead :)
```

As I said before, I'm unfamiliar how rust, cargo and pgrx work together, I'm just trying to understand if I can test building extensions for PostgreSQL 18 ahead of time, or if I need to wait until after it was released.

usamoi:
@jakob It's supported in `develop` branch. Use `cargo-pgrx` and `pgrx` in this repo.

jakob:
> Use cargo-pgrx and pgrx in this repo

[...]

PR #2056, opened 2025-05-06
Support Postgres 18beta1

silver-ymz:
related #2052

## Notable Change

- Postgres changes the structure of `TupleDescData`. It stores `CompactAttribute` rather than `FormData_pg_attribute`. All related code have been updated.
- Postgres changes `executor_start` and `executor_run` hook API, so `PgHook` trait will have 2 versions, `pg13-17` and `pg18+`.
- Postgres introduces a new C macro `PG_MODULE_MAGIC_EXT`, so we add a rust macro `pg_module_magic_ext!($name, $version)` to keep the same API with it.


pgrx-pg-config/src/lib.rs
```diff
@@ -67,6 +67,7 @@ pub enum PgMinorVersion {
     Release(u16),
     Beta(u16),
     Rc(u16),
+    Devel,
```
eeeebbbbrrrr: We don't need this and I don't want it.  :)

When pg18beta1 is released we can use the `::Beta(1)` variant, which is how we've handled this in the past.

eeeebbbbrrrr (changes requested):
I suspect this is generally fine except for the new `::Devel` version variant.  This isn't how we've handled things in the past and it's not how I want to start doing it.

When the Postgres folks actually release beta1, which allows us to use it in CI and such, then we can merge support for it.

Which reminds me... this PR will need to update our github actions to support 18 too.

eeeebbbbrrrr:
I sincerely appreciate your work here, @silver-ymz.  It's quite a daunting task to add support for a new Postgres version!

silver-ymz:
> When the Postgres folks actually release beta1, which allows us to use it in CI and such, then we can merge support for it.

Ok, when Postgres folks release beta1, I'll continue to work on this PR, update CI and update with new break changes.

eeeebbbbrrrr:
https://www.postgresql.org/about/news/postgresql-18-beta-1-released-3070/

Look what just happened!

pgrx-examples/aggregate/src/lib.rs
```diff
@@ -14,7 +14,16 @@ use pgrx::StringInfo;
...
+#[cfg(any(
+    feature = "pg13",
+    feature = "pg14",
+    feature = "pg15",
+    feature = "pg16",
+    feature = "pg17"
+))]
+pgrx::pg_module_magic!();
+#[cfg(any(feature = "pg18"))]
+pgrx::pg_module_magic_ext!(c"aggregate", pgrx::pg_sys::PG_VERSION);
```
eeeebbbbrrrr: I'm just picking one of these at random, but this isn't a desirable user-facing change.

Can we instead overload the `pgrx::pg_module_magic!()` macro to have a second form that that takes the extra arguments?

Forcing all downstream users to make this change, simply because of pg18 support, is not ideal.

eeeebbbbrrrr (changes requested):

pgrx-examples/aggregate/src/lib.rs
silver-ymz: `pgrx::pg_module_magic_ext!()` have been removed. Users can use `pgrx::pg_module_magic!()` or `pgrx::pg_module_magic!($name, $version)`.

eeeebbbbrrrr (approved):
This looks great, @silver-ymz!  Sincerely, thank you for your work and please accept my apologizes for letting this sit for so long -- RealLife tends to get in the way sometimes.

PR #2058, opened 2025-05-07
initial support for `pg_regress`-based tests

eeeebbbbrrrr:
This adds a new `cargo pgrx regress` subcommand that knows how to manage/run sql-based tests compatible with Postgres' `pg_regress` tool (which this uses).

I'd expect there to be some UX issues here that will need to be addressed in future commits/releases, but this feature needs to start somewhere.


eeeebbbbrrrr:
Worth noting that there's 3 new entries existing extensions will want to add to their `.gitignore` file if they adopt regression tests.
Screenshot 2025-05-07 at 2 23 25 PM

Our `cargo pgrx new ...` command will do this for new extensions.  I didn't want to teach `cargo pgrx upgrade` how to do this because many folks are real fussy about how their `.gitignore` is organized.

eeeebbbbrrrr:
CI is red because I cancelled the jobs.  

I'm going to merge this through and then start a new release.

PR #2075, opened 2025-05-19
Expose guc hooks

thesuhas:
Exposing the GUC hooks to insert extern C code that will execute before assigning a GUC/while assigning a GUC/`show` of GUC.


thesuhas:
@charmitro can I get a review on this whenever you get a chance?

pgrx/src/guc.rs
```diff
@@ -244,6 +365,9 @@ impl GucRegistry {
         max_value: i32,
         context: GucContext,
         flags: GucFlags,
+        check_hook: Option<unsafe extern "C-unwind" fn(*mut i32, *mut *mut c_void, u32) -> bool>,
```
tristan957: u32 is a Rust type. You should use `libc::c_int`.

tristan957: You can probably use: `pg_sys::GucIntCheckHook` instead of defining it yourself though.

eeeebbbbrrrr (approved):
I'll merge soon as CI passes (which I'm approving a run for now).

usamoi:
```rust
unsafe extern "C-unwind" fn f(_: *mut bool, _: *mut *mut c_void, _: u32) -> bool {
    panic!()
}
GucRegistry::define_bool_guc_with_hooks(..., f)
```

should cause undefined behavior, since Rust panic unwinds without being caught?

eeeebbbbrrrr:
Looks like CI is unhappy, there's a conflict, and there's some review suggestions here that ought to be considered.

I'm not sure of the best way to handle @usamoi's point here:  https://github.com/pgcentralfoundation/pgrx/pull/2075#issuecomment-2925422388

It would be the programmer's responsibility to add a `#[pg_guard]` on that function -- that doesn't seem unreasonable.  Maybe these new `xxx_with_hooks()` functions should all be declared `unsafe` with a "# Safety" note on each about this.

pgrx/src/guc.rs
charmitro: IMO `pg_sys::GucSource::Type` should be the best fit here. Match the pg_sys bindings and ofc the type safety we need.

pgrx/src/guc.rs
```diff
@@ -211,13 +212,133 @@ where
...
+            context,
+            flags,
+            None,
+            None,
+            None,
+        );
+    }
+
+    // GUC Registration functions that expose hooks
+    pub fn define_bool_guc_with_hooks(
```
charmitro:
```suggestion
    /// Define a boolean GUC with custom hooks.
    /// 
    /// # Hooks
    /// 
    /// * `check_hook` - Validates new values. Return false to reject.
    /// * `assign_hook` - Called after value is set. Use for side effects.
    /// * `show_hook` - Returns custom display string for SHOW commands.
    pub fn define_bool_guc_with_hooks(
```

I would love to have something like this here, if you agree.

charmitro (commented):
What do you think about a test case? As simple as it can be.

  1. Check hooks being called
  2. Assign hooks performing side effects
  3. Show hooks providing custom display output

thesuhas:
> Looks like CI is unhappy, there's a conflict, and there's some review suggestions here that ought to be considered.
> 
> I'm not sure of the best way to handle @usamoi's point here: [#2075 (comment)](https://github.com/pgcentralfoundation/pgrx/pull/2075#issuecomment-2925422388)
> 
> It would be the programmer's responsibility to add a `#[pg_guard]` on that function -- that doesn't seem unreasonable. Maybe these new `xxx_with_hooks()` functions should all be declared `unsafe` with a "# Safety" note on each about this.

Done! But won't all GUC registration calls with hooks have to be in unsafe blocks/functions?

thesuhas:
> I'll merge soon as CI passes (which I'm approving a run for now).

Can you approve another run?

eeeebbbbrrrr:
> But won't all GUC registration calls with hooks have to be in unsafe blocks/functions?

yes,  because they fundamentally are unsafe. 

We can’t make them safe because the functions can’t guarantee the user declares their hook function correctly.

There’s nothing wrong with unsafe, but there is something wrong with not declaring something unsafe that actually is.

pgrx-tests/src/tests/guc_tests.rs
```diff
@@ -231,4 +232,114 @@ mod tests {
             assert_eq!(GUC_NO_SHOW.get(), true, "'no_show' should reset after 'RESET ALL'");
         });
     }
+
+    #[pg_test]
+    fn test_guc_check_hook() {
+        static SIDE_EFFECT: std::sync::RwLock<i32> = std::sync::RwLock::new(0);
+
+        // Define hooks
+        unsafe extern "C-unwind" fn check_hook(
```
eeeebbbbrrrr, line 242: All the user-defined hook functions, like this one, need to be decorated with `#[pg_guard]`.

That's the reason why the `define_xxx_with_hooks()` functions need to be `unsafe` -- because it can't confirm the programmer did the right thing.  Which didn't happen here and in the other tests.

eeeebbbbrrrr: In fact, as a followup, please define a a check_hook function that simply does `panic!("oh no!");` and write a `#[pg_test]` for it that expects the test to panic.

You'll see that without a `#[pg_guard]` on that hook function Postgres will instead crash.

eeeebbbbrrrr (changes requested):

eeeebbbbrrrr (approved):
Thank you for your work!  Soon as CI passes I'll merge this.

eeeebbbbrrrr:
Looks like this latest round of changes to the tests caused Postgres to crash.  :(

pgrx-tests/src/tests/guc_tests.rs
```diff
@@ -231,4 +232,178 @@ mod tests {
...
+            newval: *mut bool,
+            _extra: *mut *mut std::ffi::c_void,
+            _source: pg_sys::GucSource::Type,
+        ) -> bool {
+            if *newval {
+                panic!("should panic!");
+            }
+            *newval
+        }
```
eeeebbbbrrrr: yeah, this test is probably the one that's causing Postgres to crash.  This is not a test we want!  haha.  :)

Just remove this entire `test_check_hook_fail_unguarded()`.  We know doing the unsafe thing will cause a crash -- no need to prove that.

thesuhas: Ah I see, apologies for that. Interestingly it passed locally when I added the `#[should_panic(expected = "connection closed")]` macro. Removing it now!

eeeebbbbrrrr: The tests run in parallel across concurrent backend connections and you probably got "lucky" whereas CI didn't.

If this last CI run passes then I'll merge!

eeeebbbbrrrr:
Thanks for your work and tenacity, @thesuhas.  This stuff ain’t easy!

PR #2078, opened 2025-05-22
Allows to create multiple aggregates for the same Rust type

### LINKED ISSUE #2077: Multiple aggregate on same type
Hello, I'm I would like to define multiple aggregate functions on one of my types: https://github.com/JeanChristopheMorinPerso/data_experiments/blob/c47bca89aab29d96a25a96a475c63a438707a9cd/conda_pgsql_rust_ext/src/lib.rs#L7-L12

I can add at least one aggregate function using the `pg_aggregate` macro like this:
```rust
#[pg_aggregate]
impl Aggregate for CondaVersion {
    type State = CondaVersion;
    type Args = CondaVersion;
    const NAME: &'static str = "min";
    const SORT_OPERATOR: Option<&'static str> = Some("condaversion_lt");

    fn state(
        current: Self::State,
        arg: Self::Args,
        _fcinfo: pg_sys::FunctionCallInfo
    ) -> Self::State {
        // NOTE THAT I HAVE NOT YET TESTED THIS IMPLEMENTATION
        if current <= arg {
            return current;
        }
        return arg;
    }
}
```

This results in
```sql
CREATE  FUNCTION "conda_version_state"(
        "this" CondaVersion, /* conda_pgsql_rust_ext::CondaVersion */
        "arg_one" CondaVersion /* conda_pgsql_rust_ext::CondaVersion */
) RETURNS CondaVersion /* conda_pgsql_rust_ext::CondaVersion */
STRICT
LANGUAGE c /* Rust */
AS 'MODULE_PATHNAME', 'conda_version_state_wrapper';


CREATE AGGREGATE min (
	CondaVersion /* conda_pgsql_rust_ext::CondaVersion */
)
(
	SFUNC = "conda_version_state", /* conda_pgsql_rust_ext::CondaVersion::state */
	STYPE = CondaVersion, /* conda_pgsql_rust_ext::CondaVersion */
	SORTOP = "condaversion_lt" /* conda_pgsql_rust_ext::CondaVersion::SORT_OPERATOR */
);
```
But I'm kind of failing to see how I could define a second one... Say, a `max` aggregate.

Is this possible with pgrx? I'm not very good at rust, so forgive me if this is more a rust question than a pgrx question.

Thanks!

if0ne:
Closes #2077 

Unfortunately, the rust doesn't allow using `&'static str` in const generic yet. This would allow simply moving an associative constant to a generic trait. Therefore an auxiliary trait `ToAggregateName` with associative constant `NAME` was added. This allows `Aggregate` to be implemented multiple times for the same type with different names.

This change introduces a bit of boilerplate. In order to simplify the use of this trait, a derive macro `AggregateName` has been developed, which automatically implements this trait. In this case, `NAME` becomes equal to the name of the structure for which this macro is used. To change the name of the name you should use attribute `#[aggregate_name = “custom_name”]`.

Example:
```rust
#[derive(Copy, Clone, Default, Debug, PostgresType, Serialize, Deserialize)]
pub struct DemoOps {
    count: i32,
}

struct DemoSumName;

// Using derive macro
#[derive(AggregateName)]
#[aggregate_name = "demo_sub"]
struct DemoSubName;

// Explicit trait implementation
impl ToAggregateName for DemoSumName {
    const NAME: &'static str = "demo_sum";
}

// The first aggregate:
#[pg_aggregate]
impl Aggregate<DemoSumName> for DemoOps {
    /* implementation */
}

// The second aggregate:
#[pg_aggregate]
impl Aggregate<DemoSubName> for DemoOps {
    /* implementation */
}
```

**BREAKING CHANGES**: Now for the `Aggregate` trait you need to specify a generic parameter that implements the `ToAggregateName` trait. I tried to leave backwards compatibility, but decided to break it after all so that developers using `pgrx` would see that the semantics of the `Aggregate` trait had changed.


eeeebbbbrrrr (approved):
Thanks.  

As soon as you have a chance to resolve the conflicts (likely caused by recent PRs) I'll be happy to merge.

if0ne:
@eeeebbbbrrrr Do I need to update postgresql-aggregates-with-rust.md?

charmitro:
@eeeebbbbrrrr let's see if CI passes now? Seems @if0ne did some work for it.

PR #2081, opened 2025-05-25
Added quotes around name of types and schema in CREATE TYPE

### LINKED ISSUE #2080: Add quotes around the names of the generated types and enums
At this time, enums and types are generated without quotes around the names.

The following will be inserted in `pg_type` as `my_type`.

```sql
CREATE TYPE My_Type AS ENUM ('foo', 'bar');
```

While the following will be inserted in `pg_type` as `My_Type`.

```sql
CREATE TYPE "My_Type" AS ENUM ('foo', 'bar');
```

I believe it is desirable to conserve the casing for the types, as the type `My_Type` is considered a distinct type from the type `my_type`, so either the names are to be lowercased, or quotes should be added.

I believe the better and easier solution would be for the second option, i.e. to add the quotes so to ensure that the Postgres name matches the rust name. **I expect this change to potentially be API breaking.**

Do let me know your opinion on the matter.

Best,
Luca

LucaCappelletti94:
Closes #2080

### DIFF
--- a/articles/forging-sql-from-rust.md
+++ b/articles/forging-sql-from-rust.md
@@ -861,7 +861,7 @@ impl ToSql for InventoryPostgresEnum {
             "\n\
                     -- {file}:{line}\n\
                     -- {full_path}\n\
-                    CREATE TYPE {schema}{name} AS ENUM (\n\
+                    CREATE TYPE {schema}\"{name}\" AS ENUM (\n\
                         {variants}\
                     );\
                 ",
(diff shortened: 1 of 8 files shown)


LucaCappelletti94:
I am afraid the approach needs to be much more precise and replace in the functions solely the custom types, as by adding quotes around all types as I am currently doing I end up adding quotes to `"bigint[]"`, which should not be quoted.

LucaCappelletti94:
Most likely I will refactor this to just update the `SqlTranslatable`, so to make it return a string for custom types.

eeeebbbbrrrr:
~~I've marked this as a Draft.  When you've got what you want to accomplish dialed in, just mark it as ready for review.~~

See below.

eeeebbbbrrrr:
I added some short commentary on #2080, but quoting type names is **not** something I want in pgrx as it will make types much more difficult to work with from SQL and also cause significant downstream backwards incompatibility impacts.

PR #2083, opened 2025-06-02
Pass `LLVM_*` variables to `--runas` command

theory:
To allow variables like `LLVM_PROFILE_FILE`, which is used for code coverage profiling, to be passed through.


eeeebbbbrrrr (approved):

theory:
Depressingly, with this fix, the `--runas` process no longer writes profile files at all! :-( I'm updating my test coverage script to omit `LLVM_PROFILE_FILE` now.

PR #2084, opened 2025-06-11
Decoding command output in Windows

### LINKED ISSUE #2069: Panic in non-utf-8 code pages in Windows
All lines of code `String::from_utf8(output.stdout).unwrap()` in pgrx-pg-config cause a panic if Windows uses a non utf-8 code page, e.g. Cyrillic Windows uses 866. 

Possible solutions:

1. Use `from_utf8_lossy`
2. Add additional step in Getting Started for Windows: [Using UTF-8 in the Windows Terminal](https://akr.am/blog/posts/using-utf-8-in-the-windows-terminal)

if0ne:
Closes #2069

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -597,6 +597,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f46ad14479a25103f283c0f10005961cf086d8dc42205bb44c46ac563475dca6"
 
 [[package]]
+name = "codepage"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "48f68d061bc2828ae826206326e61251aca94c1e4a5305cf52d9138639c918b4"
+dependencies = [
+ "encoding_rs",
+]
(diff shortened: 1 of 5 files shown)


usamoi:
As I understand it, using `from_utf8_lossy` is not suitable for double-byte encodings. A better approach would be to use `GetACP` to retrieve the system code page and then use the corresponding encoding from `encoding_rs` to convert it to UTF-8.

if0ne:
`from_utf8_lossy` will help avoid crashes from panic. The fact that `REPLACEMENT_CHARACTER`s will be displayed is a different problem imo.

I could solve the encoding issue with a third-party crate, but I have little experience with open-source and don't know if it's ok to add an additional dependency without any discussion from the major contributors.

YohDeadfall:
It can be a Windows only dependency then as it requires special handling anyway, and `pgrx` already has some Windows specific fixes. So, why not?

if0ne:
Ok, I'll fix the Windows encoding problem then.

PR #2086, opened 2025-06-24
WIP: Adding support for contrib/intarray style fast-allocated arrays

mhov:
## Faster arrays!
Now that `Array<T>` is stable, this PR ports in performance enhancements for creating PostgreSQL `int[]` arrays—based on the optimizations found in PostgreSQL’s `contrib/intarray` extension (specifically [`new_intArrayType(int num)`](https://github.com/postgres/postgres/blob/master/contrib/intarray/_int_tool.c#L224)).

The goal is to significantly speed up the conversion from `&[i32]` to a PostgreSQL `ArrayType` Datum.

## Background

Currently, to return a PostgreSQL `int[]` from a Rust function, we typically return a `Vec<i32>`, which then gets converted to a Datum via `array_datum_from_iter(..)`  
([implementation here](https://github.com/pgcentralfoundation/pgrx/blob/c0de15d29d2ef175f3b56b630f85af09448094e5/pgrx/src/datum/array.rs#L1017)).

This approach uses:
- `pg_sys::initArrayResult`
- `pg_sys::accumArrayResult`
- `pg_sys::makeArrayResult`

These APIs build a varlena array by accumulating values and repeatedly reallocating memory as the capacity grows. For larger arrays, this results in substantial performance overhead.

## Optimization

PostgreSQL's `contrib/intarray` takes a different approach for fixed-size datums: it pre-allocates the entire array up front and directly `memcpy`s the data into `ARR_DATA_PTR`. This avoids the need for reallocations and is **much faster**.

## What I've added

- Implemented `BoxRet` for `Array<'a, T>` when `T` is a fixed-size numeric type (`i8`, `i16`, `i32`, `f32`, `f64`)
  - Enables using `Array<T>` directly as a return type from `#[pg_extern]` functions.
- Adds optimized allocation logic for these numeric types
  - Pre-allocates the full `ArrayType`
  - Uses `memcpy`-like behavior for fast construction from slices
- Adds helper functions for:
  - Creating empty arrays
  - Creating `Array<T>` from `&[T]`

## Benchmarks

I've benchmarked two approaches for turning `Vec<i32> -> Array<i32>`:
1. The current `array_datum_from_iter(..)`
2. The new fast pre-allocated method

Each test used 10,000 rows of randomly generated `int[]` in a temp table, work_mem = '1GB'.  I used Instant::now() to time only the microseconds needed to convert and accumulated all those timings

```
    #[pg_extern]
    fn accum_alloc(input: Vec<i32>) -> i32 {
        let start = std::time::Instant::now();
        let array = input.into_datum().unwrap();  // this uses array_datum_from_iter(..)
        let duration = start.elapsed();
        duration.subsec_micros() as i32
    }
    
    #[pg_extern]
    fn fast_alloc(input: Vec<i32>) -> i32 {
        let start = std::time::Instant::now();
        let array = Array::<i32>::new_from_slice(input.as_slice()).expect("couldn't allocate");
        let duration = start.elapsed();
        duration.subsec_micros() as i32
    }
```

| Row Count | int[] Length | accum_alloc (total μs) | fast_alloc (total μs) | Improvement |
|-----------|--------------|----------------------------------|------------------------|-------------|
| 10,000    | 10           | 16,933.00                        | 2,120.00               | 8.0×        |
| 10,000    | 100          | 155,352.00                       | 10,228.00              | 15.2×       |
| 10,000    | 1,000        | 1,375,613.00                     | 30,443.00              | 45.2×       |
| 10,000    | 10,000       | 13,666,835.00                    | 251,825.00             | 54.3×       |

It's been a while since I contributed, so I'm a little rusty on matching the teams style/conventions, and as usual lifetime differences between PG allocated and rust allocated still escape me.  I'm sure the ergonomics of the new Array<T> functions could be improved.  Am I doing anything the wrong way here?

### DIFF
--- a/pgrx/src/array.rs
+++ b/pgrx/src/array.rs
@@ -8,7 +8,7 @@
 //LICENSE
 //LICENSE Use of this source code is governed by the MIT license that can be found in the LICENSE file.
 #![allow(clippy::precedence)]
-use crate::datum::{Array, IntoDatum, UnboxDatum};
+use crate::datum::{Array, ArrayFastAllocSubType};
 use crate::toast::{Toast, Toasty};
 use crate::{layout, pg_sys, set_varsize_4b, varlena, PgMemoryContexts};
 use bitvec::ptr::{self as bitptr, BitPtr, BitPtrError, Mut};
@@ -378,9 +378,7 @@ impl RawArray {
     /// Slightly faster than new_array_type_with_len(0)
     pub fn new_empty_array_type<T>() -> Result<RawArray, ArrayAllocError>
     where
-        T: IntoDatum,
-        T: UnboxDatum,
-        T: Sized,
+        T: ArrayFastAllocSubType,
     {
         unsafe {
             let array_type = pg_sys::construct_empty_array(T::type_oid());
@@ -393,9 +391,7 @@ impl RawArray {
     /// Rustified version of new_intArrayType(int num) from https://github.com/postgres/postgres/blob/master/contrib/intarray/_int_tool.c#L219
     pub fn new_array_type_with_len<T>(len: usize) -> Result<RawArray, ArrayAllocError>
     where
-        T: IntoDatum,
-        T: UnboxDatum,
-        T: Sized,
+        T: ArrayFastAllocSubType,
     {
         if len == 0 {
             return Self::new_empty_array_type::<T>();
--- a/pgrx/src/array.rs
+++ b/pgrx/src/array.rs
@@ -431,4 +431,6 @@ impl Toasty for RawArray {
 pub enum ArrayAllocError {
     #[error("Failed to allocate memory for Array")]
     MemoryAllocationFailed,
+    #[error("Expected len {0}, found {1}")]
+    IterLenMismatch(usize, usize),
 }
--- a/pgrx/src/array.rs
+++ b/pgrx/src/array.rs
@@ -8,13 +8,14 @@
 //LICENSE
 //LICENSE Use of this source code is governed by the MIT license that can be found in the LICENSE file.
 #![allow(clippy::precedence)]
-use crate::datum::Array;
+use crate::datum::{Array, IntoDatum, UnboxDatum};
 use crate::toast::{Toast, Toasty};
-use crate::{layout, pg_sys, varlena};
+use crate::{layout, pg_sys, set_varsize_4b, varlena, PgMemoryContexts};
 use bitvec::ptr::{self as bitptr, BitPtr, BitPtrError, Mut};
 use bitvec::slice::BitSlice;
 use core::ptr::{self, NonNull};
 use core::slice;
+use pgrx_pg_sys::ArrayType;
 
 mod port;
 
@@ -373,10 +374,65 @@ impl RawArray {
         let ptr = self.ptr.as_ptr().cast::<u8>();
         ptr.wrapping_add(unsafe { varlena::varsize_any(ptr.cast()) })
     }
+
+    /// Slightly faster than new_array_type_with_len(0)
+    pub fn new_empty_array_type<T>() -> Result<RawArray, ArrayAllocError>
+    where
+        T: IntoDatum,
+        T: UnboxDatum,
+        T: Sized,
+    {
+        unsafe {
+            let array_type = pg_sys::construct_empty_array(T::type_oid());
+            let array_type =
+                NonNull::new(array_type).ok_or(ArrayAllocError::MemoryAllocationFailed)?;
+            Ok(RawArray::from_ptr(array_type))
+        }
+    }
+
+    /// Rustified version of new_intArrayType(int num) from https://github.com/postgres/postgres/blob/master/contrib/intarray/_int_tool.c#L219
+    pub fn new_array_type_with_len<T>(len: usize) -> Result<RawArray, ArrayAllocError>
+    where
+        T: IntoDatum,
+        T: UnboxDatum,
+        T: Sized,
+    {
+        if len == 0 {
+            return Self::new_empty_array_type::<T>();
+        }
+        let elem_size = std::mem::size_of::<T>();
+        let nbytes: usize = port::ARR_OVERHEAD_NONULLS(1) + elem_size * len;
+
+        unsafe {
+            let array_type = PgMemoryContexts::For(pg_sys::CurrentMemoryContext).palloc0(nbytes)
+                as *mut ArrayType;
+            if array_type.is_null() {
+                return Err(ArrayAllocError::MemoryAllocationFailed);
+            }
+            set_varsize_4b(array_type as *mut pg_sys::varlena, nbytes as i32);
+            (*array_type).ndim = 1;
+            (*array_type).dataoffset = 0; /* marker for no null bitmap */
+            (*array_type).elemtype = T::type_oid();
+
+            let ndims = port::ARR_DIMS(array_type);
+            *ndims = len as i32; // equivalent of ARR_DIMS(r)[0] = num;
+            let arr_lbound = port::ARR_LBOUND(array_type);
+            *arr_lbound = 1;
+
+            let array_type = NonNull::new_unchecked(array_type);
+            Ok(RawArray::from_ptr(array_type))
+        }
+    }
 }
 
 impl Toasty for RawArray {
     unsafe fn drop_toast(&mut self) {
         unsafe { pg_sys::pfree(self.ptr.as_ptr().cast()) }
     }
+}
+
+#[derive(thiserror::Error, Debug, Copy, Clone, Eq, PartialEq)]
+pub enum ArrayAllocError {
+    #[error("Failed to allocate memory for Array")]
+    MemoryAllocationFailed,
 }
--- a/pgrx/src/array.rs
+++ b/pgrx/src/array.rs
@@ -388,7 +388,7 @@ impl RawArray {
         }
     }
 
-    /// Rustified version of new_intArrayType(int num) from https://github.com/postgres/postgres/blob/master/contrib/intarray/_int_tool.c#L219
+    /// Rustified version of new_intArrayType(int num) from [https://github.com/postgres/postgres/blob/master/contrib/intarray/_int_tool.c#L219]
     pub fn new_array_type_with_len<T>(len: usize) -> Result<RawArray, ArrayAllocError>
     where
         T: ArrayFastAllocSubType,
--- a/pgrx/src/array/port.rs
+++ b/pgrx/src/array/port.rs
@@ -126,3 +126,22 @@ pub(super) unsafe fn ARR_DATA_PTR(a: *mut pg_sys::ArrayType) -> *mut u8 {
 
     unsafe { a.cast::<u8>().add(ARR_DATA_OFFSET(a)) }
 }
+
+/// Returns a pointer to the lower bounds of the array.
+/// # Safety
+/// Does a field access, but doesn't deref out of bounds of ArrayType.  The caller asserts that
+/// `a` is a properly allocated [`pg_sys::ArrayType`]
+///
+/// [`pg_sys::ArrayType`] is typically allocated past its size, and its somewhere in that region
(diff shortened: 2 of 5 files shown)


pgrx/src/array.rs
```diff
@@ -373,10 +374,65 @@ impl RawArray {
...
+        }
+    }
+
+    /// Rustified version of new_intArrayType(int num) from https://github.com/postgres/postgres/blob/master/contrib/intarray/_int_tool.c#L219
+    pub fn new_array_type_with_len<T>(len: usize) -> Result<RawArray, ArrayAllocError>
+    where
+        T: IntoDatum,
+        T: UnboxDatum,
+        T: Sized,
+    {
```
usamoi, line 395: It's not okay to block something like:

```rust
let a = pgrx::datum::new_array_with_len::<&[u8]>(5).expect("failed to create array");
```

It's better to mark this function as unsafe, and avoid exposing anything other than those specific `Array::new_with_len` methods.

pgrx/src/array.rs
```diff
@@ -373,10 +374,65 @@ impl RawArray {
...
+        if len == 0 {
+            return Self::new_empty_array_type::<T>();
+        }
+        let elem_size = std::mem::size_of::<T>();
+        let nbytes: usize = port::ARR_OVERHEAD_NONULLS(1) + elem_size * len;
+
+        unsafe {
```
workingjubilee, line 402: In new code, a safe function with internal `unsafe` should explain why the code is sound.

pgrx/src/datum/array.rs
```diff
@@ -503,6 +653,52 @@ fn as_slice<'a, T: Sized>(array: &'a Array<'_, T>) -> Result<&'a [T], ArraySlice
...
+    unsafe {
+        let raw_array = RawArray::new_empty_array_type::<T>()?;
+        let datum: pgrx_pg_sys::Datum = pg_sys::Datum::from(raw_array.into_ptr().as_ptr());
+        Array::<'a, T>::from_polymorphic_datum(datum, false, pg_sys::get_array_type(T::type_oid()))
+            .ok_or(ArrayAllocError::MemoryAllocationFailed)
+    }
+}
+
+/// Creates an `Array<T>` of a fixed len, with 0 for all elements
+/// Uses a single PG allocation rather than
```
workingjubilee: rather than?

pgrx/src/datum/array.rs
```diff
@@ -503,6 +653,52 @@ fn as_slice<'a, T: Sized>(array: &'a Array<'_, T>) -> Result<&'a [T], ArraySlice
...
+#[inline(always)]
+pub fn new_array_with_len<'a, T: Sized>(len: usize) -> Result<Array<'a, T>, ArrayAllocError>
+where
+    T: IntoDatum,
+    T: UnboxDatum,
```
workingjubilee: I do not believe `T: IntoDatum + UnboxDatum + Sized` is a sufficient bound to establish that it is sound to create that type with all-zero initialization. This function is probably `unsafe` in actuality, albeit with a maybe-trivial precondition.

workingjubilee: I prefer this sort of bound to be written like this
```suggestion
pub fn new_array_with_len<'a, T>(len: usize) -> Result<Array<'a, T>, ArrayAllocError>
where
    T: Sized,
    T: IntoDatum,
    T: UnboxDatum,
```

or this if it works:

```suggestion
pub fn new_array_with_len<'a, T>(len: usize) -> Result<Array<'a, T>, ArrayAllocError>
where
    T: IntoDatum + UnboxDatum + Sized,
```

pgrx/src/array.rs
workingjubilee: Hum. Could you elaborate?

pgrx/src/array/port.rs
```diff
@@ -126,3 +126,22 @@ pub(super) unsafe fn ARR_DATA_PTR(a: *mut pg_sys::ArrayType) -> *mut u8 {
...
+/// that the returned pointer points, so don't attempt to `pfree` it.
+#[inline(always)]
+pub(super) unsafe fn ARR_LBOUND(a: *mut pg_sys::ArrayType) -> *mut i32 {
+    //  #define ARR_LBOUND(a) \
+    //        ((int *) (((char *) (a)) + sizeof(ArrayType) + \
+    //                  sizeof(int) * ARR_NDIM(a)))
+
+    a.cast::<u8>()
+        .add(std::mem::size_of::<pg_sys::ArrayType>())
+        .add(std::mem::size_of::<i32>() * ((*a).ndim as usize))
```
workingjubilee: style nit
```suggestion
        .add(mem::size_of::<pg_sys::ArrayType>())
        .add(mem::size_of::<i32>() * ((*a).ndim as usize))
```

pgrx/src/array/port.rs
```diff
@@ -126,3 +126,22 @@ pub(super) unsafe fn ARR_DATA_PTR(a: *mut pg_sys::ArrayType) -> *mut u8 {
 
     unsafe { a.cast::<u8>().add(ARR_DATA_OFFSET(a)) }
 }
+
+/// Returns a pointer to the lower bounds of the array.
```
workingjubilee: ...Maybe this should specify whether it actually points to the _list_ of lower bounds given by Postgres as a series of integers, or the actual lower bound position in the actual data.

( The answer is the former. )

workingjubilee (changes requested):
Some initial comments, mostly style nits. Let's discuss the motivation more.

If memory serves, I believe an array is a varlena allocation that does not have to match the precise size of the array's length, yes? That is, the varlena can be allocated bigger than the array needs. Why not something more like the [`spare_capacity_mut`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.spare_capacity_mut) API that Vec has, so we can construct things without first zeroing them? I believe the zeroing overhead is in many cases inconsequential and often optimized-out anyways, but we can always make our code easier to optimize.

pgrx/src/array/port.rs
workingjubilee: For `ARR_DIMS` I just left off a comment on that entirely on the tacit assumption that people shouldn't use it or they will read the original source enough to understand how to use it if they touch it. I find that also acceptable here. Basically, the text should be correct or simply absent, not misleading.

pgrx/src/datum/array.rs
mhov: oops, rather than `pg_sys::accumArrayResult(..)`

pgrx/src/array.rs
usamoi: There is

```rust
let elem_size = std::mem::size_of::<T>(); // T = &[u8]
```

in the function body. This looks unreasonable. `IntoDatum + UnboxDatum + Sized` doesn't block types that is passed by value, and not all types of zero-initialization are valid.

mhov:
@usamoi @workingjubilee  instead of all the `T: Sized + IntoDatum + UnboxDatum` stuff should we just make a opt-in trait like we do for [RangeSubType](https://github.com/pgcentralfoundation/pgrx/blob/0b04f1ad5488210377f5925e3d80e1c12438ee1f/pgrx/src/datum/range.rs#L485) that we specifically implement for `i8,i16,i32,i64,f32,f64` ?

workingjubilee:
@mhov Probably so, I think that would be a better starting point anyways as then it would be clearer what we're asking for.

pgrx/src/datum/array.rs
mhov: replaced with an opt in ArrayFastAllocSubType trait

mhov: replaced with an opt in ArrayFastAllocSubType trait, types are specifically opted in

pgrx/src/datum/array.rs
mhov: typo fixed

pgrx/src/array.rs
mhov: replaced with an opt in ArrayFastAllocSubType trait

pgrx/src/datum/array.rs
```diff
@@ -1349,7 +1349,7 @@ where
     }
 }
 
-/// This trait allows for arrays of certain numeric types to use Array<T>'s single allocation strategy
+/// This trait allows for arrays of certain numeric types to use Array&lt;T&gt;'s single allocation strategy
```
workingjubilee: erm, this should be
```suggestion
/// This trait allows for arrays of certain numeric types to use `Array<T>`'s single allocation strategy
```

pgrx/src/datum/array.rs
```diff
@@ -424,6 +424,31 @@ impl Array<'_, f64> {
...
+    /// Returns a mutable slice of `f64`s which comprise this [`Array`].
+    ///
+    /// # Errors
+    ///
+    /// Returns a [`ArraySliceError::ContainsNulls`] error if this [`Array`] contains one or more
+    /// SQL "NULL" values.  In this case, you'd likely want to fallback to using [`Array::iter()`].
+    #[inline]
+    pub fn as_mut_slice(&mut self) -> Result<&mut [f64], ArraySliceError> {
+        as_mut_slice(self)
+    }
```
workingjubilee: as discussed elsewhere, this is currently unsound because we do not make any effort to distinguish a borrowed-from-Postgres Array from a newly-allocated one, so this allows mutating data that Postgres may assume was unmutated.

mhov: If this fast allocation technique is using postgres's allocation anyways, aren't we still borrowing it from Postgres in both cases?

workingjubilee: If we create the array we notionally have `&mut` to it.

mhov: Ok I finally see the risk, I encountered this situation in the C version of my extension where I mutated an array from a result set, instead of taking a copy of it first.  The mutated version ended up being read into other subsequent queries because I was modifying the cache layer's data.  

As we discussed, I could make this private (+ unsafe?) and only to be used by the `pub fn new_from_slice(slice: &'_ [T]) -> Result<Array<'mcx, T>, ArrayAllocError>`  constructor.  Sound good?

mhov:
i made as_mut_slice internal, and reworked the Array::<T:ArrayFastAllocSubType> constructors under a single impl<'mcx, T> Array<'mcx, T> where T: ArrayFastAllocSubType { .. } including one that can construct from any Iterator<Item = T:ArrayFastAllocSubType> and one with a FnOnce to allow a limited shot at the allocated Array's mut slice.

Let me know what you think about the ergonomics and soundness of `Array::<T>::new_from_iter(..)` and `Array::<T>::new_with_init_fn(..)`

pgrx/src/array.rs
```diff
@@ -373,10 +374,63 @@ impl RawArray {
         let ptr = self.ptr.as_ptr().cast::<u8>();
         ptr.wrapping_add(unsafe { varlena::varsize_any(ptr.cast()) })
     }
+
+    /// Slightly faster than new_array_type_with_len(0)
+    pub fn new_empty_array_type<T>() -> Result<RawArray, ArrayAllocError>
+    where
+        T: ArrayFastAllocSubType,
```
workingjubilee, line 381: This can just use the `Scalar` trait bound. Hmm. I should probably move that into `pgrx::array` I guess, since it will outlast `datum::Array`.

workingjubilee (commented):
With this new API, is it possible to create an `Array<'static, T>`?

workingjubilee (changes requested):
Unfortunately, I believe this PR isn't suitable, because while talking about it with @mhov we discovered it's possible to pick the `'static` lifetime for a new `Array<'static, T>` made this way. That shouldn't work, basically, which means we can't accept this PR even in its current form, sadly. I'm working on fixing up an older branch I had made to be its replacement.

I'm leaving this open because I'm treating it as an issue/feature request, essentially.

mhov:
All good, thanks @workingjubilee

workingjubilee:
I have landed #2207, thus addressing this.

PR #2087, opened 2025-06-25
fix ci on beta rust (1.89)

usamoi:


reneleonhardt:
It's always good to run CI on beta to catch upcoming problems early on 👍

PR #2089, opened 2025-06-27
Use if-let to unpack Options

stuhood:
Use if-let to unpack Options, to avoid tripping the `for_loops_over_fallibles` warning in Rust `1.88.0`.


eeeebbbbrrrr:
pgrx has no MSRV policy so I think that this will be fine to merge assuming CI is smart enough to pick up the new rust version.

we'll find out here in an hour...

stuhood:
> pgrx has no MSRV policy so I think that this will be fine to merge assuming CI is smart enough to pick up the new rust version.
> 
> we'll find out here in an hour...

Looks like CI will want us to do the Rust upgrade over here first. Should that be a separate PR, or should I just convert this into a "get building under `1.88.0`" PR?

eeeebbbbrrrr:
> Looks like CI will want us to do the Rust upgrade over here first. Should that be a separate PR, or should I just convert this into a "get building under `1.88.0`" PR?

man, don't worry about it.  I'll get into the weeds this weekend and whip it into shape.  Thanks tho!

YohDeadfall:
That's done by #2087.

eeeebbbbrrrr (approved):

PR #2100, opened 2025-06-28
chore: consolidate package settings in workspace

nyurik:
allow the same version, edition, and other settings to be centrally managed. Also bump to the newer workspace resolver to simplify future 2024 migration

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@
 #LICENSE Use of this source code is governed by the MIT license that can be found in the LICENSE file.
 
 [workspace]
-resolver = "2"
+resolver = "3"
 members = [
     "cargo-pgrx",
     "pgrx",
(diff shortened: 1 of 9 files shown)


nyurik:
This PR fails on the `pgrx-version-updater` - which tries to parse Cargo.toml and obviously fails - as it tries to mimic what `cargo` itself does with all of its logic.  Should the updater be deprecated?  Should it call `cargo metadata` using the https://crates.io/crates/cargo_metadata crate?

eeeebbbbrrrr:
I just merged #2101 -- will that take care of the failure here?

nyurik:
@eeeebbbbrrrr i don't think it will - the issue here is that the version updater is a bit of a hacky code - which tried to do what is now possible with the workspaces - i.e. keep `version` field the same across all crates.  I can modify this PR to not use the workspace version in crates, but going forward, I think the updater needs to be either reworked or deleted alltogether - I don't have enough context to know how much you needed it or not

eeeebbbbrrrr:
hmm.

If we can define everything in the workspace Cargo.toml such that the individual crates don't need to be updated whenever we run the `./prepare-release.sh` script, and that we get the exact same result, then sure, I'm fine with getting rid of the version updater.  The best code is the code that we don't have to maintain.

If that's the direction you're headed in, please, continue!

nyurik:
@eeeebbbbrrrr exactly.  Please merge this PR as it passes and moves in that direction.  I will work on another PR to get it further cleaned up

nyurik:
P.S. My biggest concern is that i do not have a full grasp on what changes `./prepare-release.sh` and update-version do, and I am not certain how to test it locally. Otherwise I would have deleted the update-version right away :)

eeeebbbbrrrr (approved):

PR #2107, opened 2025-07-04
improve shmem api

### LINKED ISSUE #2106: shmem api is unsound
1. `PGRXSharedMemory` implementation of `heapless` is unsound.

```rust
unsafe impl<T, const N: usize> PGRXSharedMemory for heapless::Vec<T, N> {}
unsafe impl<T, const N: usize> PGRXSharedMemory for heapless::Deque<T, N> {}
unsafe impl<K: Eq + Hash, V: Default, S, const N: usize> PGRXSharedMemory
    for heapless::IndexMap<K, V, S, N>
{
}
```

Here `T: PGRXSharedMemory, K: PGRXSharedMemory, V: PGRXSharedMemory` constraint is not enforced.

`heapless` itself does not guarantee that these data structures are suitable for use with mmap. It might be a better approach to remove these implementations and require users to manually review `heapless` and lock version of `heapless` themselves.

2. `PgAtomic` requires `atomic_traits::Atomic` without requiring `PGRXSharedMemory`, which is unsound.

`atomic_traits::Atomic` is a safe trait and any type could implement this. Even if it's an sealed or unsafe trait, it's unsound, since it supports loom atomics, which are not real machine atomics.

3. PostgreSQL does not check for duplicate shared memory or lock names.

https://github.com/postgres/postgres/blob/REL_17_5/src/backend/storage/ipc/shmem.c#L370-L486

https://github.com/postgres/postgres/blob/REL_17_5/src/backend/storage/lmgr/lwlock.c#L659-L703

https://github.com/postgres/postgres/blob/REL_17_5/src/backend/storage/lmgr/lwlock.c#L567-L600

`pg_shmem_init` should be unsafe, since users must ensure that there are not name conflicts. But it's safe to call it.

4. `PgSpinLock` is not constrained by `T: PGRXSharedMemory` too.

usamoi:
closes #2106

1. `heapless` is removed due to unsoundness
2. `atomic_traits` is removed due to unsoundness
3. `PgAtomic::new` and `PgLwLock::new` is marked unsafe; caller must be confident that there are no name conflicts (including `pg_shmem_init` is called for a variable twice)
4. `unsafe impl<T: PGRXSharedMemory> PGRXSharedMemory for PgSpinLock<T> {}`, in order to allow it work on shared memory
5. `pg_shmem_init!(ATOMIC = AtomicUsize::new(0));` is allowed now
6. checks if the pointer to the allocated shared memory is properly aligned

It breaks api.

### DIFF
--- a/pgrx-examples/bgworker/src/lib.rs
+++ b/pgrx-examples/bgworker/src/lib.rs
@@ -30,7 +30,7 @@ pgrx::pg_module_magic!(name, version);
 
 #[pg_guard]
 pub extern "C-unwind" fn _PG_init() {
-    if unsafe { pgrx::pg_sys::IsUnderPostmaster } {
+    if unsafe { !pgrx::pg_sys::process_shared_preload_libraries_in_progress } {
         pgrx::error!("this extension must be loaded via shared_preload_libraries.");
     }
     BackgroundWorkerBuilder::new("Background Worker Example")
--- a/pgrx-examples/bgworker/src/lib.rs
+++ b/pgrx-examples/bgworker/src/lib.rs
@@ -30,6 +30,9 @@ pgrx::pg_module_magic!(name, version);
 
 #[pg_guard]
 pub extern "C-unwind" fn _PG_init() {
+    if unsafe { pgrx::pg_sys::IsUnderPostmaster } {
+        pgrx::error!("this extension must be loaded via shared_preload_libraries.");
+    }
     BackgroundWorkerBuilder::new("Background Worker Example")
         .set_function("background_worker_main")
         .set_library("bgworker")
@@ -84,4 +87,13 @@ pub extern "C-unwind" fn background_worker_main(arg: pg_sys::Datum) {
     }
 
     log!("Goodbye from inside the {} BGWorker! ", BackgroundWorker::get_name());
+}
+
+#[cfg(test)]
+pub mod pg_test {
+    pub fn setup(_options: Vec<&str>) {}
+
+    pub fn postgresql_conf_options() -> Vec<&'static str> {
+        vec!["shared_preload_libraries='bgworker'"]
+    }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -149,16 +149,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atomic-traits"
-version = "0.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "707f750b93bd1b739cf9ddf85f8fe7c97a4a62c60ccf8b6f232514bd9103bedc"
-dependencies = [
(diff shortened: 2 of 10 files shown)


pgrx-examples/bgworker/src/lib.rs
```diff
@@ -30,6 +30,9 @@ pgrx::pg_module_magic!(name, version);
 
 #[pg_guard]
 pub extern "C-unwind" fn _PG_init() {
+    if unsafe { pgrx::pg_sys::IsUnderPostmaster } {
+        pgrx::error!("this extension must be loaded via shared_preload_libraries.");
+    }
```
cbandy: I've also [seen](https://github.com/pgaudit/pgaudit/blob/6d61cb7df9a9f2cfcf80ea5bbdf01427be0c85c0/pgaudit.c#L2122-L2125) `!process_shared_preload_libraries_in_progress` used for this.

```c
    /* Must be loaded with shared_preload_libraries */
    if (!process_shared_preload_libraries_in_progress)
        ereport(ERROR, (errcode(ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE),
                errmsg("pgaudit must be loaded via shared_preload_libraries")));
```

eeeebbbbrrrr:
Hey @usamoi!  Can you elaborate on heapless being unsound?

usamoi:
> Can you elaborate on heapless being unsound?

```
static S: PgLwLock<Vec<&'static str>>;

// process 1
S.write().push(Box::leak(String::from("aaaaaa")));

// process 2
dbg!(S.read().pop().unwrap()); // boom!
```

eeeebbbbrrrr:
Doesn't removing heapless drastically limit the usefulness of our shared memory support?

On the other hand, does anyone even use it, I wonder?

I'm going to cut a new release today and I'm on the fence about if we should merge this PR now or later.

usamoi:
> Doesn't removing heapless drastically limit the usefulness of our shared memory support?

Now `AssertPGRXSharedMemory` is added, so they could still use it by:

```rust
static DEQUE: PgLwLock<AssertPGRXSharedMemory<heapless::Deque<Pgtest, 400>>> =
    unsafe { PgLwLock::new(c"shmem_deque") };

pg_shmem_init!(DEQUE = unsafe { AssertPGRXSharedMemory::new(Default::default()) });
```

compared to before:

```rust
static DEQUE: PgLwLock<heapless::Deque<Pgtest, 400>> =
    unsafe { PgLwLock::new(c"shmem_deque") };

pg_shmem_init!(DEQUE);
```

eeeebbbbrrrr:
> Now AssertPGRXSharedMemory is added, so they could still use it by:

Okay, I see.  Then yeah, I suppose this is fine.  I'll merge it now and then start on a v0.16.0

PR #2113, opened 2025-07-22
`cargo pgrx regress --resetdb` will run `setup.sql`

ccleve:
cargo pgrx regress --resetdb doesn't re-run setup.sql if there is an up-to-date /expected/setup.out. We definitely want to run it if the existing testing database gets blown away.


eeeebbbbrrrr (approved):
Thanks @ccleve.  I've noticed something like this myself but haven't had the opportunity to track it down.

PR #2120, opened 2025-08-01
delete `pgrx::hooks`

usamoi:
followup of #2118

`pgrx::hooks` is deprecated since 0.12.1 (Aug 22, 2024). We have now discovered a new unsoundness in it recently. It's time to delete `pgrx::hooks` now.


eeeebbbbrrrr (approved):
I concur!

Thanks for doing this.

It'll cause the next release to be 0.16.0, but that's okay.

ccleve:
I'm curious -- what was the unsoundness?

usamoi:
> what was the unsoundness?

`pgrx::hooks` lack proper `pg_guard_ffi_boundary`. If there is a PostgreSQL error from hooks, `sigsetjmp` would cross Rust frames. If any one of them is not [plain old frame](https://rust-lang.github.io/rfcs/2945-c-unwind-abi.html#plain-old-frames), it's UB.

https://github.com/pgcentralfoundation/pgrx/pull/1752 presents another reason why it is unsound.

PR #2125, opened 2025-08-12
Example : register PostgreSQL hooks

daamien:
Introducing a basic safety catch example to demonstrate how to implement and register PostgreSQL hooks.

* A post_parse_analyze hook will prevent users from using the DELETE command without a WHERE clause.
* A process_utility hook will prevent non-superuser from using TRUNCATE.

Following a discussion on the discord channel, I wrote this code to understand how to migrate from PGRX 0.14 to PGRX 0.16 and replace the former PgHooks trait that was removed recently. 

I thought this could help other extension developers in the same situation...

Let me know if that's the proprer way to do this....

### DIFF
--- a/pgrx-examples/hooks/src/lib.rs
+++ b/pgrx-examples/hooks/src/lib.rs
@@ -1,0 +1,236 @@
+//
+// pg_guard_ffi_boundary is necessary when calling any PostgreSQL-exported
+// functions, especially the hooks functions pointers coming from PostgreSQL
+// cf. https://github.com/pgcentralfoundation/pgrx/blob/develop/pgrx-pg-sys/src/submodules/ffi.rs.
+//
+
+use pgrx::pg_sys::ffi::pg_guard_ffi_boundary;
+use pgrx::prelude::*;
+
+::pgrx::pg_module_magic!(name, version);
+
+fn delete_must_have_a_where(query: PgBox<pg_sys::Query>) {
+    if query.commandType != pg_sys::CmdType::CMD_DELETE {
+        return ();
+    }
+
+    // SAFETY: the jointree is always defined in a DELETE Query
+    let jointree = unsafe { PgBox::from_pg(query.jointree) };
+
+    if jointree.quals.is_null() {
+        panic!("DELETE queries must have a WHERE clause");
+    }
+}
+
+fn only_superusers_can_truncate(pstmt: PgBox<pg_sys::PlannedStmt>) {
+    if !unsafe { pgrx::is_a(pstmt.utilityStmt, pg_sys::NodeTag::T_TruncateStmt) } {
+        return ();
+    }
+
+    if !unsafe { pg_sys::superuser() } {
+        panic!("Only superusers can truncate")
+    }
+}
+
+unsafe fn register_hooks() {
+    //
+    // Post Parse Analyze hook
+    //
+    static mut PREV_POST_PARSE_ANALYZE_HOOK: pg_sys::post_parse_analyze_hook_type = None;
+    PREV_POST_PARSE_ANALYZE_HOOK = pg_sys::post_parse_analyze_hook;
+    pg_sys::post_parse_analyze_hook = Some(post_parse_analyze_hook);
+
+    // The hook functions signatures may change between major version
+    // For instance: in the post_parse_analyze hook, the JumbleState struct
+    // appeared in Postgres 14
+    // In that case, we need some conditional compilation to declare the
+    // proper signature for each version
+    #[cfg(feature = "pg13")]
+    #[pg_guard]
+    unsafe extern "C-unwind" fn post_parse_analyze_hook(
+        parse_state: *mut pg_sys::ParseState,
+        query: *mut pg_sys::Query,
+    ) {
+        delete_must_have_a_where(PgBox::from_pg(query));
+        if let Some(prev_hook) = PREV_POST_PARSE_ANALYZE_HOOK {
+            pg_guard_ffi_boundary(|| prev_hook(parse_state, query));
+        }
+    }
+    #[cfg(any(
+        feature = "pg14",
+        feature = "pg15",
+        feature = "pg16",
+        feature = "pg17",
+        feature = "pg18"
+    ))]
+    #[pg_guard]
+    unsafe extern "C-unwind" fn post_parse_analyze_hook(
+        parse_state: *mut pg_sys::ParseState,
+        query: *mut pg_sys::Query,
+        jumble_state: *mut pg_sys::JumbleState,
+    ) {
+        delete_must_have_a_where(PgBox::from_pg(query));
+        if let Some(prev_hook) = PREV_POST_PARSE_ANALYZE_HOOK {
+            pg_guard_ffi_boundary(|| prev_hook(parse_state, query, jumble_state));
+        }
+    }
+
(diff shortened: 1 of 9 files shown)


pgrx-examples/hooks/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+    // proper signature for each version
+    #[cfg(feature = "pg13")]
+    #[pg_guard]
+    unsafe extern "C-unwind" fn post_parse_analyze_hook(
+        parse_state: *mut pg_sys::ParseState,
+        query: *mut pg_sys::Query,
+    ) {
+        delete_must_have_a_where(PgBox::from_pg(query));
+        if let Some(prev_hook) = PREV_POST_PARSE_ANALYZE_HOOK {
+            prev_hook(parse_state, query);
```
usamoi:
```suggestion
            pgrx::pg_sys::ffi::pg_guard_ffi_boundary(|| prev_hook(parse_state, query));
```

The other three calls to `prev_hooks` also need to be modified in the same way.

pgrx-examples/hooks/src/lib.rs
```diff
@@ -0,0 +1,225 @@
+use pgrx::prelude::*;
+
+::pgrx::pg_module_magic!(name, version);
+
+fn delete_must_have_a_where(query: PgBox<pg_sys::Query>) {
+    if query.commandType != pg_sys::CmdType::CMD_DELETE {
+        return ();
+    }
+
+    // SAFETY: the jointree is always defined in a DELETE Query
```
charmitro, line 17: Do you think a defensive check here is sane? Like,
```rs
      if query.jointree.is_null() {
          return...;
      }
```

pgrx-examples/hooks/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+    PREV_PROCESS_UTILITY_HOOK = pg_sys::ProcessUtility_hook;
+    pg_sys::ProcessUtility_hook = Some(process_utility_hook);
+
+    // Until Postgres 13, the process utility hook didn't have a read_only_tree param
+    #[cfg(feature = "pg13")]
+    #[pg_guard]
+    unsafe extern "C-unwind" fn process_utility_hook(
+        pstmt: *mut pg_sys::PlannedStmt,
+        query_string: *const i8,
+        context: u32,
```
charmitro, line 91: Could be `pg_sys::ProcessUtilityContext::Type`?

pgrx-examples/hooks/src/lib.rs
cbandy: 👋🏻 (I'm new here)

The documentation on `pg_guard_ffi_boundary` is all internal. When/Where is it appropriate to use? When/Where is it _necessary_ to use?

usamoi: > When/Where is it _necessary_ to use?

It's **necessary** to use if you're calling any PostgreSQL-exported functions or extension-exported functions.

In most cases, you do not need to explicitly use this function because the functions in `pgrx::pg_sys` are rewritten through macros as follows:

```rust
#[pgrx::pg_guard]
extern "C" {
    fn psql_fn();
}
// The code expands to:
/************************
fn psql_fn() {
    extern "C" {
        fn psql_fn();
    }
    pgrx::pg_sys::ffi::pg_guard_ffi_boundary(|| psql_fn())
}
************************/
```

However, for any function pointers coming from PostgreSQL, we cannot rewrite them in this way. Therefore, you need to manually call `pg_guard_ffi_boundary`.

```rust
#[pgrx::pg_guard]
#[unsafe(no_mangle)]
extern "C-unwind" fn this_fn_can_be_called_by_postgresql(p: extern "C-unwind" fn()) {
    // `p` comes from PostgreSQL, so calling `p()` is dangerous!
    // let's use `pg_guard_ffi_boundary`
    pgrx::pg_sys::ffi::pg_guard_ffi_boundary(|| p())
}
```

As for why we need to wrap all PostgreSQL-exported function-call with `pg_guard_ffi_boundary`, please refer to https://github.com/pgcentralfoundation/pgrx/blob/develop/pgrx-pg-sys/src/submodules/ffi.rs.

pgrx-examples/hooks/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+}
+
+unsafe fn register_hooks() {
+    //
+    // Post Parse Analyze hook
+    //
+    static mut PREV_POST_PARSE_ANALYZE_HOOK: pg_sys::post_parse_analyze_hook_type = None;
+    PREV_POST_PARSE_ANALYZE_HOOK = pg_sys::post_parse_analyze_hook;
+    pg_sys::post_parse_analyze_hook = Some(post_parse_analyze_hook);
```
cbandy, line 41: ❓ Is it reasonable to use `replace` here?


```suggestion
    PREV_POST_PARSE_ANALYZE_HOOK =
        pg_sys::post_parse_analyze_hook.replace(post_parse_analyze_hook);
```

pgrx-examples/hooks/src/lib.rs
daamien: The context param is defined as u32 here: 

https://docs.rs/pgrx/latest/pgrx/pg_sys/static.ProcessUtility_hook.html

I'm not sure why but I do as I'm told.... ;-)

pgrx-examples/hooks/src/lib.rs
daamien: defensive against what ?  :-)

At this point in the function, we know the query is a `DELETE`, so there's no way the jointree pointer would be NULL

pgrx-examples/hooks/src/lib.rs
daamien: Imho this is the most impressive feature of PRGX.

pgrx-examples/hooks/src/lib.rs
daamien: @cbandy : it produces the following warning

```rust

warning: creating a mutable reference to mutable static
  --> src/lib.rs:77:5
   |
77 |     pg_sys::ProcessUtility_hook.replace(process_utility_hook);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ mutable reference to mutable static
   |

Comment (cbandy) [reply pgrx-examples/hooks/src/lib.rs:41]: Thanks for trying. I see the same here.

I _don't_ get that in my extension using `emit_log_hook` with `#![deny(static_mut_refs)]` and `#![warn(unsafe_op_in_unsafe_fn)]`. 😩

```rust
    static mut HOOK_EMIT_LOG: pg_sys::emit_log_hook_type = None;
    HOOK_EMIT_LOG = pg_sys::emit_log_hook.replace(my_emit_log_hook);
```

eeeebbbbrrrr:
@daamien thanks for taking this upon yourself.  I apologize for not doing this myself, as I said I would on Discord.  My excuse is that I'm a busy boy!

daamien:
> @daamien thanks for taking this upon yourself. I apologize for not doing this myself, as I said I would on Discord. My excuse is that I'm a busy boy!

No apologies needed :) 

Thanks for your initial example, I just built a basic extension around it :)

PR #2131, opened 2025-08-21
give exported `__pgrx_internals_fn_*` functions unique names

eeeebbbbrrrr:
It's entirely possible for a pgrx extension to have functions like this:

```rust
mod a {
   #[pg_extern]
   fn foo() { }
}

mod b {
   #[pg_extern]
   fn foo() { }
}
```

These are two distinct functions in Rust, and should be in the SQL schema, but the naming of the sql entity graph `__pgrx_internals_fn_*` function isn't unique enough and causes compilation problems.

We could maybe include the schema name here, but I don't think that's quite enough, so instead we just append a UUIDv4 value.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -74,6 +74,9 @@ jobs:
       - name: Print sccache stats (before)
         run: sccache --show-stats
 
+      - name: Install rustfmt
+        run: rustup component add rustfmt
+
       - name: Run rustfmt
         run: cargo fmt --all -- --check
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -380,6 +380,9 @@ jobs:
       - name: Print sccache stats (before)
         run: sccache --show-stats
 
+      - name: Install clippy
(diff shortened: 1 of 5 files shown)


usamoi:
> These are two distinct functions in Rust, and should be in the SQL schema, but the naming of the sql entity graph `__pgrx_internals_fn_*` function isn't unique enough and causes compilation problems.

Why don't two `_wrapper` functions and two `pg_finfo_` functions conflict too?

eeeebbbbrrrr:
> Why don't two `_wrapper` functions and two `pg_finfo_` functions conflict too?

They do!  But that's not what I'm trying to fix here.  There'd be no way to "fix" what you're describing  without changing how we name *every* exported `#[pg_extern]` symbol, and that seems like a bad idea.

eeeebbbbrrrr:
I'll leave this PR for now.  

I'm trying to get f-ing CI to work so I can release support for pg18.

eeeebbbbrrrr:
ehh.  this was a bad idea.

PR #2132, opened 2025-08-29
fix big endian

usamoi:
Tests passed on an S390X machine.


eeeebbbbrrrr (approved):
I'll merge but this is going to be very difficult for us to ensure correctness going forward due to lack of CI resources for big endian.

PR #2133, opened 2025-08-31
remove home, supports-color 0.2 and update toml, color-eyre

usamoi:
1. Remove dependency on `home`, since `std::env::home_dir` is fixed and undeprecated since Rust 1.85.
2. Remove dependency on `supports-color 0.2`, since it's outdated and not used.
3. `toml 0.8` is oudated. Updated to `0.9`.
4. `color-eyre 0.6.3` is outdated. Updated to `0.6.5`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,12 @@ version = 4
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.24.2"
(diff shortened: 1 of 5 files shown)


usamoi:
CI fails due to https://github.com/rust-lang/rust/issues/146045 and it would be fixed in tomorrow's nightly rustc.

workingjubilee:
Thanks!

PR #2143, opened 2025-09-25
Fixing CI

eeeebbbbrrrr:
CI has bitrotted and it looks like the new rust 1.90.0 lights up some of our "ui" tests, and I'd prefer to not deal with that, so this also adds a rust-toolchain.toml to set us to 1.89.0

### DIFF
--- a/.github/workflows/runas.yml
+++ b/.github/workflows/runas.yml
@@ -86,6 +86,11 @@ jobs:
           cargo --version
           pg_config --version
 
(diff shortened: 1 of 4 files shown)


usamoi:
~This is actually caused by https://github.com/actions/runner-images/pull/13044. Since GitHub downgraded the Rust version from 1.90 to 1.89, UI tests are broken.~

I made a mistake here.

PR #2153, opened 2025-10-16
Add Scalar trait for generic handling of primitive arrays

workingjubilee:
This deletes a bunch of repeated code in favor of an `unsafe trait`, which will become important if we want to add more things that essentially share this bound.


workingjubilee:
macOS failures are just brew failures, it seems.

eeeebbbbrrrr (approved):

PR #2155, opened 2025-10-16
include `access/tsmapi.h`

usamoi:
`access/tsmapi.h` is the header for [Writing a Table Sampling Method](https://www.postgresql.org/docs/current/tablesample-method.html).


workingjubilee:
Seems fine!

PR #2156, opened 2025-10-16
Remove testing for deprecated Intel Macs

workingjubilee:
Includes clarifying the list of targets down to a "simple" enumeration.


workingjubilee:
If anyone wants to fix this another way, they are welcome to post a revert of my PR combined with a proper fix. Otherwise, this seems best: we can't easily obtain the resources to maintain this testing, so don't.

PR #2161, opened 2025-10-22
Set BGWORKER_SHMEM_ACCESS when building background workers

### LINKED ISSUE #2160: BackgroundWorkerBuilder quietly registers broken workers
## What happened

In the early stages of developing a background worker on Postgres 17.6, I built a worker that logged a few messages:

```rust
…
    BackgroundWorkerBuilder::new("worker")
        .set_start_time(BgWorkerStartTime::PostmasterStart)
        .set_function("worker_main")
        .set_library("worker")
        .load();
…

#[pg_guard]
#[unsafe(no_mangle)]
pub extern "C-unwind" fn worker_main(_arg: pg_sys::Datum) {
    log!("{} is starting", BackgroundWorker::get_name());
    log!("{} stopped", BackgroundWorker::get_name());
}
```

Postgres started successfully, but instead of my messages, Postgres logged:

```
[2025-10-22 00:21:52.679 CDT] [25769] [68f869f0.64a9]: LOG:  background worker "worker": background workers without shared memory access are not supported
```

I had to call `enable_shmem_access(None)` for Postgres to start the worker.

```rust
    BackgroundWorkerBuilder::new("worker")
        .set_start_time(BgWorkerStartTime::PostmasterStart)
        .set_function("worker_main")
        .set_library("worker")
        .enable_shmem_access(None)
        .load();
```

```                                                                  
[2025-10-22 00:23:00.205 CDT] [26443] [68f86a34.674b]: LOG:  worker is starting
```

## What I expected

1. I expected my initial set of builder calls to produce a valid background worker.
2. I expected an error when I produced a background worker that Postgres didn't like. The `LOG` level from Postgres honestly surprised me.

## Other information

That message led straight to this discussion about why shmem access is necessary now: https://postgr.es/m/20210802065116.j763tz3vz4egqy3w@alap3.anarazel.de

cbandy:
Background workers must attach to shared memory since Postgres 15.

Fixes: pgcentralfoundation/pgrx#2160
Discussion:  https://postgr.es/m/20210802065116.j763tz3vz4egqy3w@alap3.anarazel.de
See: https://github.com/postgres/postgres/commit/80a8f95b3bca6a80672d1766c928cda34e979112

### DIFF
--- a/pgrx/src/bgworkers.rs
+++ b/pgrx/src/bgworkers.rs
@@ -510,7 +510,7 @@ impl BackgroundWorkerBuilder {
         BackgroundWorkerBuilder {
             bgw_name: name.to_string(),
             bgw_type: name.to_string(),
-            bgw_flags: BGWflags::empty(),
+            bgw_flags: BGWflags::BGWORKER_SHMEM_ACCESS, // required since Postgres 15
             bgw_start_time: BgWorkerStartTime::PostmasterStart,
             bgw_restart_time: None,
             bgw_library_name: name.to_string(),


pgrx/src/bgworkers.rs
```diff
@@ -510,7 +510,7 @@ impl BackgroundWorkerBuilder {
         BackgroundWorkerBuilder {
             bgw_name: name.to_string(),
             bgw_type: name.to_string(),
-            bgw_flags: BGWflags::empty(),
+            bgw_flags: BGWflags::BGWORKER_SHMEM_ACCESS, // required since Postgres 15
```
cbandy, line 513: I considered leaving this `empty()` for pg13 and pg14, but the thread on the mailing list suggested that omitting this flag is bad for some platforms.

> If you're saying that this code has been 100% broken for 7 years and
> nobody's noticed until now, then that suggests that nobody actually
> uses non-shmem-connected bgworkers.

> Well, the problem only exists on Windows so maybe this indeed
> escaped notice.

workingjubilee:
Thanks for investigating that and the PR! We'll try to get to this soon.

workingjubilee (approved):
weird!

workingjubilee:
@cbandy I can accept this as-is, but what would be nice is if you could add some text describing this requirement to https://github.com/pgcentralfoundation/pgrx/blob/develop/pgrx-examples/bgworker/README.md which is effectively where our bgworker documentation de facto lives.

I believe it's implied by the requirement to use `shared_preload_libraries` but making it explicit would be great.

This can be addressed in a followup as I will likely hit "merge" on this anyways unless you happen to see this response in the moments before I do so.

PR #2162, opened 2025-10-23
Manually bind `static MyProcPort` and `struct Port`

### LINKED ISSUE #2157: Bindings for client information in Port
Postgres logs include information about the connected client from the global [`MyProcPort` variable](https://github.com/postgres/postgres/blob/REL_18_0/src/backend/utils/init/globals.c#L51), which is allocated in `TopMemoryContext`. I want to read some of those fields from [an `emit_log_hook`](https://github.com/postgres/postgres/blob/REL_18_0/src/include/utils/elog.h#L476-L478) to be consistent with built-in log messages, but bindings for the [`Port` type](https://github.com/postgres/postgres/blob/REL_18_0/src/include/libpq/libpq-be.h#L108-L128) are not currently generated.

Because the type describes the connected client, it utilizes some GSS and SSL types, which cascade into [several other dependencies](https://github.com/pgcentralfoundation/pgrx/pull/2117). I'm not interested in those values, so I'm hopeful there's a way to access these fields without unnecessary includes:

- `user_name`, [log_line_prefix `%u`](https://github.com/postgres/postgres/blob/REL_18_0/src/backend/utils/error/elog.c#L2922-L2937)
- `database_name`, [log_line_prefix `%d`](https://github.com/postgres/postgres/blob/REL_18_0/src/backend/utils/error/elog.c#L2938-L2953)
- `remote_host` and `remote_port`, [log_line_prefix `%r`](https://github.com/postgres/postgres/blob/REL_18_0/src/backend/utils/error/elog.c#L3103-L3152)

workingjubilee:
This is my proposed alternative to pgcentralfoundation/pgrx#2117 building on @usamoi's suggestion. In order to enable interacting with `static MyProcPort`, we introduce a definition of the `struct Port` type in `mod libpq` that is available on all versions of Postgres. Several fields are left private to discourage access to them, as in general enabling access to Postgres's cryptography APIs for authentication and authorization is outside the desired project scope of PGRX.

In addition, in one (but only one) version, Postgres 17, the result of `size_of::<Port>()` is not even guaranteed to be correct. This is because we don't immediately know, without implementing `cfg(accessible(..))`, whether or not these fields should exist or not. Because of these difficulties, we leave those dubious fields deprecated on Postgres 17 so that no one interacts with them until such time as we gain more confidence in that.

Fixes pgcentralfoundation/pgrx#2157

cc @cbandy

### DIFF
--- a/pgrx-pg-sys/src/libpq.rs
+++ b/pgrx-pg-sys/src/libpq.rs
@@ -71,10 +71,11 @@ pub mod be {
 
         ssl_in_use: bool,
         peer_cn: *mut core::ffi::c_char,
+        #[cfg(any(feature = "pg14", feature = "pg15", feature = "pg16", feature = "pg17", feature = "pg18"))]
         peer_dn: *mut core::ffi::c_char,
         peer_cert_valid: bool,
 
-        #[cfg(any(feature = "pg18", feature = "pg17"))]
+        #[cfg(any(feature = "pg17", feature = "pg18"))]
         alpn_used: bool,
         #[cfg(feature = "pg18")]
         last_read_was_eof: bool,
--- a/pgrx-pg-sys/src/libpq.rs
+++ b/pgrx-pg-sys/src/libpq.rs
@@ -71,7 +71,13 @@ pub mod be {
 
         ssl_in_use: bool,
         peer_cn: *mut core::ffi::c_char,
-        #[cfg(any(feature = "pg14", feature = "pg15", feature = "pg16", feature = "pg17", feature = "pg18"))]
+        #[cfg(any(
+            feature = "pg14",
+            feature = "pg15",
+            feature = "pg16",
+            feature = "pg17",
+            feature = "pg18"
+        ))]
         peer_dn: *mut core::ffi::c_char,
         peer_cert_valid: bool,
 
--- a/pgrx-pg-sys/src/libpq.rs
+++ b/pgrx-pg-sys/src/libpq.rs
@@ -1,0 +1,98 @@
+/** Manually-constructed bindings to libpq
+
+Because pgrx is for extensions which run in the Postgres server, it rarely needs access to libpq.
+However, some server-side extensions need to interact with the reality that clients exist.
+Unfortunately, doing that means acknowledging that clients need authentication and authorization,
+areas of concern that are far beyond what pgrx wants to involve itself with or be responsible for.
+
+We define some types and signatures here to allow a minimal amount of usage of items from libpq,
+while largely rejecting the notion that we should involve ourselves in security-laden concerns.
+
+*/
+
+pub mod be {
+
+    unsafe extern "C" {
+        pub static mut MyProcPort: *mut Port;
+    }
+
+    /// #define SCRAM_MAX_KEY_LEN          PG_SHA256_DIGEST_LENGTH
+    /// #define PG_SHA256_DIGEST_LENGTH    32
+    const SCRAM_MAX_KEY_LEN: usize = 32;
+
+    #[repr(C)]
+    pub struct Port {
+        pub sock: crate::pgsocket,
+        pub noblock: bool,
+        pub proto: crate::ProtocolVersion,
+        pub laddr: crate::SockAddr,
+        pub raddr: crate::SockAddr,
+        pub remote_host: *mut core::ffi::c_char,
+        pub remote_hostname: *mut core::ffi::c_char,
+        pub remote_hostname_resolv: core::ffi::c_int,
+        pub remote_hostname_errcode: core::ffi::c_int,
+        pub remote_port: *mut core::ffi::c_char,
+        #[cfg(any(feature = "pg13", feature = "pg14", feature = "pg15", feature = "pg16"))]
+        pub canAcceptConnections: core::ffi::c_uint,
+        #[cfg(feature = "pg18")]
+        pub local_host: [core::ffi::c_char; 64],
+        pub database_name: *mut core::ffi::c_char,
+        pub user_name: *mut core::ffi::c_char,
+        pub cmdline_options: *mut core::ffi::c_char,
+        pub guc_options: *mut core::ffi::c_char,
+        pub application_name: *mut core::ffi::c_char,
+
+        // The remainder is for completeness, so Rust sees Port's layout as correctly as possible.
+        // Ideally we would use `extern type` so the remainder of this was seen as of unknown size.
+        // An alternative is to simply treat them as private fields, so we do.
+        HbaLine: *mut core::ffi::c_void,
+
+        #[cfg(any(feature = "pg14", feature = "pg15"))]
+        authn_id: *const core::ffi::c_char,
+
+        default_keepalives_idle: core::ffi::c_int,
+        default_keepalives_interval: core::ffi::c_int,
+        default_keepalives_count: core::ffi::c_int,
+        default_tcp_user_timeout: core::ffi::c_int,
+        keepalives_idle: core::ffi::c_int,
+        keepalives_interval: core::ffi::c_int,
+        keepalives_count: core::ffi::c_int,
+        tcp_user_timeout: core::ffi::c_int,
+
+        #[cfg(feature = "pg18")]
+        scram_ClientKey: [u8; SCRAM_MAX_KEY_LEN],
+        #[cfg(feature = "pg18")]
+        scram_ServerKey: [u8; SCRAM_MAX_KEY_LEN],
+        #[cfg(feature = "pg18")]
+        has_scram_keys: bool,
+
+        // as if ENABLE_GSS == false && ENABLE_SSPI == false
+        gss: *mut core::ffi::c_void,
+
+        ssl_in_use: bool,
+        peer_cn: *mut core::ffi::c_char,
+        peer_dn: *mut core::ffi::c_char,
+        peer_cert_valid: bool,
+
+        #[cfg(any(feature = "pg18", feature = "pg17"))]
+        alpn_used: bool,
+        #[cfg(feature = "pg18")]
+        last_read_was_eof: bool,
+
+        // NOTE: 5 fields remain on PG17, but two are `#ifdef USE_OPENSSL` in PG17, so treat all
+        // as conditioned on PG18, even if that is not strictly accurate for PG17
+
+        // as if USE_OPENSSL == false && ENABLE_SSPI == false
+        #[cfg(feature = "pg18")]
+        ssl: *mut core::ffi::c_void,
+        #[cfg(feature = "pg18")]
+        peer: *mut core::ffi::c_void,
+
+        #[cfg(feature = "pg18")]
+        raw_buf: *mut core::ffi::c_char,
+        #[cfg(feature = "pg18")]
+        raw_buf_consumed: isize,
+        #[cfg(feature = "pg18")]
+        raw_buf_remaining: isize,
+    }
+}
--- a/pgrx-pg-sys/src/libpq.rs
+++ b/pgrx-pg-sys/src/libpq.rs
@@ -97,11 +97,14 @@ pub mod be {
         #[cfg(feature = "pg18")]
         peer: *mut core::ffi::c_void,
 
-        #[cfg(feature = "pg18")]
+        #[cfg(any(feature = "pg17", feature = "pg18"))]
+        #[cfg_attr(feature = "pg17", deprecated(since = "0.17.0", note = "may be unsound to access on Postgres 17 depending on build `#define`s")]
         raw_buf: *mut core::ffi::c_char,
-        #[cfg(feature = "pg18")]
+        #[cfg(any(feature = "pg17", feature = "pg18"))]
+        #[cfg_attr(feature = "pg17", deprecated(since = "0.17.0", note = "may be unsound to access on Postgres 17 depending on build `#define`s")]
         raw_buf_consumed: isize,
-        #[cfg(feature = "pg18")]
+        #[cfg(any(feature = "pg17", feature = "pg18"))]
+        #[cfg_attr(feature = "pg17", deprecated(since = "0.17.0", note = "may be unsound to access on Postgres 17 depending on build `#define`s")]
         raw_buf_remaining: isize,
     }
 }
--- a/pgrx-pg-sys/src/libpq.rs
+++ b/pgrx-pg-sys/src/libpq.rs
@@ -39,13 +39,15 @@ pub mod be {
         pub database_name: *mut core::ffi::c_char,
         pub user_name: *mut core::ffi::c_char,
         pub cmdline_options: *mut core::ffi::c_char,
-        pub guc_options: *mut core::ffi::c_char,
+        pub guc_options: *mut crate::List,
         pub application_name: *mut core::ffi::c_char,
 
         // The remainder is for completeness, so Rust sees Port's layout as correctly as possible.
         // Ideally we would use `extern type` so the remainder of this was seen as of unknown size.
         // An alternative is to simply treat them as private fields, so we do.
-        HbaLine: *mut core::ffi::c_void,
+
+        // This should be `*mut crate::HbaLine` if we ever bind that
+        hba: *mut core::ffi::c_void,
 
         #[cfg(any(feature = "pg14", feature = "pg15"))]
         authn_id: *const core::ffi::c_char,
--- a/pgrx-pg-sys/src/libpq.rs
+++ b/pgrx-pg-sys/src/libpq.rs
@@ -83,7 +83,7 @@ pub mod be {
         // NOTE: 5 fields remain on PG17, but two are `#ifdef USE_OPENSSL` in PG17, so treat all
         // as conditioned on PG18, even if that is not strictly accurate for PG17
 
-        // as if USE_OPENSSL == false && ENABLE_SSPI == false
+        // as if USE_OPENSSL == false
         #[cfg(feature = "pg18")]
         ssl: *mut core::ffi::c_void,
         #[cfg(feature = "pg18")]
--- a/pgrx-pg-sys/src/libpq.rs
+++ b/pgrx-pg-sys/src/libpq.rs
@@ -98,13 +98,13 @@ pub mod be {
         peer: *mut core::ffi::c_void,
 
         #[cfg(any(feature = "pg17", feature = "pg18"))]
-        #[cfg_attr(feature = "pg17", deprecated(since = "0.17.0", note = "may be unsound to access on Postgres 17 depending on build `#define`s")]
+        #[cfg_attr(feature = "pg17", deprecated(since = "0.17.0", note = "may be unsound to access on Postgres 17 depending on build `#define`s"))]
         raw_buf: *mut core::ffi::c_char,
         #[cfg(any(feature = "pg17", feature = "pg18"))]
-        #[cfg_attr(feature = "pg17", deprecated(since = "0.17.0", note = "may be unsound to access on Postgres 17 depending on build `#define`s")]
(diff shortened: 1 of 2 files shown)


pgrx-pg-sys/src/libpq.rs
```diff
@@ -0,0 +1,98 @@
+/** Manually-constructed bindings to libpq
+
+Because pgrx is for extensions which run in the Postgres server, it rarely needs access to libpq.
```
cbandy, line 3: 👍🏻

pgrx-pg-sys/src/libpq.rs
```diff
@@ -0,0 +1,98 @@
...
+
+pub mod be {
+
+    unsafe extern "C" {
+        pub static mut MyProcPort: *mut Port;
+    }
+
+    /// #define SCRAM_MAX_KEY_LEN          PG_SHA256_DIGEST_LENGTH
+    /// #define PG_SHA256_DIGEST_LENGTH    32
+    const SCRAM_MAX_KEY_LEN: usize = 32;
```
cbandy, line 22: 👍🏻

pgrx-pg-sys/src/libpq.rs
```diff
@@ -0,0 +1,98 @@
...
+        pub raddr: crate::SockAddr,
+        pub remote_host: *mut core::ffi::c_char,
+        pub remote_hostname: *mut core::ffi::c_char,
+        pub remote_hostname_resolv: core::ffi::c_int,
+        pub remote_hostname_errcode: core::ffi::c_int,
+        pub remote_port: *mut core::ffi::c_char,
+        #[cfg(any(feature = "pg13", feature = "pg14", feature = "pg15", feature = "pg16"))]
+        pub canAcceptConnections: core::ffi::c_uint,
+        #[cfg(feature = "pg18")]
+        pub local_host: [core::ffi::c_char; 64],
```
cbandy: 👍🏻

pgrx-pg-sys/src/libpq.rs
```diff
@@ -0,0 +1,98 @@
...
+        pub guc_options: *mut core::ffi::c_char,
+        pub application_name: *mut core::ffi::c_char,
+
+        // The remainder is for completeness, so Rust sees Port's layout as correctly as possible.
+        // Ideally we would use `extern type` so the remainder of this was seen as of unknown size.
+        // An alternative is to simply treat them as private fields, so we do.
+        HbaLine: *mut core::ffi::c_void,
+
+        #[cfg(any(feature = "pg14", feature = "pg15"))]
+        authn_id: *const core::ffi::c_char,
```
cbandy, line 52: 👍🏻

pgrx-pg-sys/src/libpq.rs
```diff
@@ -0,0 +1,98 @@
...
+        #[cfg(feature = "pg18")]
+        has_scram_keys: bool,
+
+        // as if ENABLE_GSS == false && ENABLE_SSPI == false
+        gss: *mut core::ffi::c_void,
+
+        ssl_in_use: bool,
+        peer_cn: *mut core::ffi::c_char,
+        peer_dn: *mut core::ffi::c_char,
+        peer_cert_valid: bool,
```
cbandy, line 189: 🔧 pg13 lacks `peer_dn` here. https://github.com/postgres/postgres/blob/REL_13_0/src/include/libpq/libpq-be.h#L187


```suggestion
        ssl_in_use: bool,
        peer_cn: *mut core::ffi::c_char,
        #[cfg(any(feature = "pg14", feature = "pg15", feature = "pg16", feature = "pg17", feature = "pg18"))]
        peer_dn: *mut core::ffi::c_char,
        peer_cert_valid: bool,
```

pgrx-pg-sys/src/libpq.rs
```diff
@@ -0,0 +1,98 @@
...
+
+        #[cfg(any(feature = "pg18", feature = "pg17"))]
+        alpn_used: bool,
```
cbandy: ⛏️ nit


```suggestion
        #[cfg(any(feature = "pg17", feature = "pg18"))]
        alpn_used: bool,
```

pgrx-pg-sys/src/libpq.rs
```diff
@@ -0,0 +1,98 @@
...
+        #[cfg(feature = "pg18")]
+        last_read_was_eof: bool,
```
cbandy, line 192: 👍🏻

pgrx-pg-sys/src/libpq.rs
```diff
@@ -0,0 +1,98 @@
...
+
+        // NOTE: 5 fields remain on PG17, but two are `#ifdef USE_OPENSSL` in PG17, so treat all
+        // as conditioned on PG18, even if that is not strictly accurate for PG17
+
+        // as if USE_OPENSSL == false && ENABLE_SSPI == false
+        #[cfg(feature = "pg18")]
+        ssl: *mut core::ffi::c_void,
+        #[cfg(feature = "pg18")]
+        peer: *mut core::ffi::c_void,
```
cbandy: 🤔 I see only `USE_OPENSSL` affecting these. https://github.com/postgres/postgres/blob/REL_18_0/src/include/libpq/libpq-be.h#L215


```suggestion
        // as if USE_OPENSSL == false
        #[cfg(feature = "pg18")]
        ssl: *mut core::ffi::c_void,
        #[cfg(feature = "pg18")]
        peer: *mut core::ffi::c_void,
```

pgrx-pg-sys/src/libpq.rs
```diff
@@ -0,0 +1,98 @@
...
+
+        #[cfg(feature = "pg18")]
+        raw_buf: *mut core::ffi::c_char,
+        #[cfg(feature = "pg18")]
+        raw_buf_consumed: isize,
+        #[cfg(feature = "pg18")]
+        raw_buf_remaining: isize,
```
cbandy: 👍🏻

pgrx-pg-sys/src/libpq.rs
```diff
@@ -0,0 +1,98 @@
...
+        peer_dn: *mut core::ffi::c_char,
+        peer_cert_valid: bool,
+
+        #[cfg(any(feature = "pg18", feature = "pg17"))]
+        alpn_used: bool,
+        #[cfg(feature = "pg18")]
+        last_read_was_eof: bool,
+
+        // NOTE: 5 fields remain on PG17, but two are `#ifdef USE_OPENSSL` in PG17, so treat all
+        // as conditioned on PG18, even if that is not strictly accurate for PG17
```
cbandy: 🤔 So we're too small in pg17. I'm not sure what to do about that or how to better explain it to the reader. 👍🏻

cbandy (commented):
There's one more field that needs an annotation for pg13, but everything else LGTM.

It was tedious to check, but I suppose we'll only need to compare when adding a major version.

Thank you for working on this!

pgrx-pg-sys/src/libpq.rs
workingjubilee: Ah, thank you. Missed that one.

pgrx-pg-sys/src/libpq.rs
workingjubilee: think that was a copypaste error, yeah.

pgrx-pg-sys/src/libpq.rs
workingjubilee: I'm currently inclined to write it off as "if someone used `size_of` on a random Postgres struct it was probably wrong anyways" because of Postgres gleefully (ab)using the "pointer-to-header" pattern.

workingjubilee:
@eeeebbbbrrrr I don't intend to merge this without your review because we ought to be aligned on this: if this feels like it's flying too close to the sun, then let's not. But I figure that this works out well-enough for our purposes.

pgrx-pg-sys/src/libpq.rs
```diff
@@ -0,0 +1,105 @@
...
+        pub remote_hostname_errcode: core::ffi::c_int,
+        pub remote_port: *mut core::ffi::c_char,
+        #[cfg(any(feature = "pg13", feature = "pg14", feature = "pg15", feature = "pg16"))]
+        pub canAcceptConnections: core::ffi::c_uint,
+        #[cfg(feature = "pg18")]
+        pub local_host: [core::ffi::c_char; 64],
+        pub database_name: *mut core::ffi::c_char,
+        pub user_name: *mut core::ffi::c_char,
+        pub cmdline_options: *mut core::ffi::c_char,
+        pub guc_options: *mut core::ffi::c_char,
```
usamoi: The type of the field should be `*mut crate::List`.

https://github.com/postgres/postgres/blob/REL_18_0/src/include/libpq/libpq-be.h#L153

pgrx-pg-sys/src/libpq.rs
```diff
@@ -0,0 +1,105 @@
...
+        ssl: *mut core::ffi::c_void,
+        #[cfg(feature = "pg18")]
+        peer: *mut core::ffi::c_void,
+
+        #[cfg(feature = "pg18")]
+        raw_buf: *mut core::ffi::c_char,
+        #[cfg(feature = "pg18")]
+        raw_buf_consumed: isize,
+        #[cfg(feature = "pg18")]
+        raw_buf_remaining: isize,
```
usamoi: It's also included in PostgreSQL 17.

https://github.com/postgres/postgres/blob/REL_17_0/src/include/libpq/libpq-be.h#L226-L228

pgrx-pg-sys/src/libpq.rs
```diff
@@ -0,0 +1,105 @@
...
+        pub database_name: *mut core::ffi::c_char,
+        pub user_name: *mut core::ffi::c_char,
+        pub cmdline_options: *mut core::ffi::c_char,
+        pub guc_options: *mut core::ffi::c_char,
+        pub application_name: *mut core::ffi::c_char,
+
+        // The remainder is for completeness, so Rust sees Port's layout as correctly as possible.
+        // Ideally we would use `extern type` so the remainder of this was seen as of unknown size.
+        // An alternative is to simply treat them as private fields, so we do.
+        HbaLine: *mut core::ffi::c_void,
```
usamoi: `HbaLine` is the name of the type, not the name of the field.

https://github.com/postgres/postgres/blob/REL_18_0/src/include/libpq/libpq-be.h#L165

pgrx-pg-sys/src/libpq.rs
```diff
@@ -0,0 +1,105 @@
...
+
+        #[cfg(any(feature = "pg17", feature = "pg18"))]
+        alpn_used: bool,
+        #[cfg(feature = "pg18")]
+        last_read_was_eof: bool,
+
+        // NOTE: 5 fields remain on PG17, but two are `#ifdef USE_OPENSSL` in PG17, so treat all
+        // as conditioned on PG18, even if that is not strictly accurate for PG17
+
+        // as if USE_OPENSSL == false
```
usamoi, line 194: We can implement `#[cfg(accessible(crate::USE_OPENSSL))]` here through `bindgen::callbacks::ParseCallbacks`.

workingjubilee: hm, that can fix the size on pg17, I guess

pgrx-pg-sys/src/libpq.rs
workingjubilee: I can't believe my eyes crossed at that. Thanks!

workingjubilee:
```suggestion

        // This should be `*mut crate::HbaLine` if we ever bind that
        hba: *mut core::ffi::c_void,
```

pgrx-pg-sys/src/libpq.rs
workingjubilee: Thanks!
```suggestion
        pub guc_options: *mut crate::List,
```

pgrx-pg-sys/src/libpq.rs
workingjubilee: I'm not sure how, at least not right away, @usamoi. I'm not seeing an obvious header to include that doesn't also try to pull in OpenSSL.

usamoi: > I'm not sure how, at least not right away, @usamoi. I'm not seeing an obvious header to include that doesn't also try to pull in OpenSSL.

It's `pg_config.h`, which is already included.

Screenshot_20251028_140609

workingjubilee: huh. then why doesn't our code define USE_OPENSSL as a constant?

usamoi: > huh. then why doesn't our code define USE_OPENSSL as a constant?

It's undefined if PostgreSQL is not compiled with OpenSSL. If we set the environment variable `PGRX_PG_CONFIG_PATH` to the `pg_config` of a PostgreSQL build compiled with OpenSSL, then `pgrx_pg_sys::USE_OPENSSL` will be present.

workingjubilee: oh right, duh.

pgrx-pg-sys/src/libpq.rs
eeeebbbbrrrr: If I'm tracking correctly... can we add some `_padding` bytes in the correct spots?

Or would it be better to just define this struct once for every Postgres version where it's different?  It is a bit difficult to reason about as currently written whereas defining it N times would be easier to see what's going on at the expense of duplicating portions of the struct.  I think that's probably a fair compromise, yeah?

eeeebbbbrrrr:
> @eeeebbbbrrrr I don't intend to merge this without your review because we ought to be aligned on this: if this feels like it's flying too close to the sun, then let's not. But I figure that this works out well-enough for our purposes.

I'm aligned in spirit.  I don't really have any care about how this is done, other than _maybe_ [this comment](https://github.com/pgcentralfoundation/pgrx/pull/2162#discussion_r2487367113).  It seems to me that `size_of::<T>()` ought to be correct for this struct across the various Postgres versions.

pgrx-pg-sys/src/libpq.rs
workingjubilee: I think for Postgres 17 we can get as close as making sure there's some padding to at least match the minimum size, but it will be inexact. I'll add some with a comment.

pgrx-pg-sys/src/libpq.rs
workingjubilee: in `ParseCallbacks::int_macro`, right? I see, I see.

pgrx-pg-sys/src/libpq.rs
```diff
@@ -0,0 +1,107 @@
...
+        ssl: *mut core::ffi::c_void,
+        #[cfg(feature = "pg18")]
+        peer: *mut core::ffi::c_void,
+
+        #[cfg(feature = "pg18")]
+        raw_buf: *mut core::ffi::c_char,
+        #[cfg(feature = "pg18")]
+        raw_buf_consumed: isize,
+        #[cfg(feature = "pg18")]
+        raw_buf_remaining: isize,
```
workingjubilee: So, currently I am going to do this, so that we can land this with documentation that there's a problem that people will run into. Even if they access these fields somehow, this way I've absolutely forced them to look directly at the warning. I will look into the `int_macro` step as an improvement.
```suggestion
        #[cfg(any(feature = "pg17", feature = "pg18"))]
        #[cfg_attr(feature = "pg17", deprecated(since = "0.17.0", note = "may be unsound to access on Postgres 17 depending on build `#define`s")]
        raw_buf: *mut core::ffi::c_char,
        #[cfg(any(feature = "pg17", feature = "pg18"))]
        #[cfg_attr(feature = "pg17", deprecated(since = "0.17.0", note = "may be unsound to access on Postgres 17 depending on build `#define`s")]
        raw_buf_consumed: isize,
        #[cfg(any(feature = "pg17", feature = "pg18"))]
        #[cfg_attr(feature = "pg17", deprecated(since = "0.17.0", note = "may be unsound to access on Postgres 17 depending on build `#define`s")]
        raw_buf_remaining: isize,
```

pgrx-pg-sys/src/libpq.rs
```diff
@@ -0,0 +1,110 @@
...
+
+        #[cfg(any(feature = "pg17", feature = "pg18"))]
+        #[cfg_attr(feature = "pg17", deprecated(since = "0.17.0", note = "may be unsound to access on Postgres 17 depending on build `#define`s")]
+        raw_buf: *mut core::ffi::c_char,
+        #[cfg(any(feature = "pg17", feature = "pg18"))]
+        #[cfg_attr(feature = "pg17", deprecated(since = "0.17.0", note = "may be unsound to access on Postgres 17 depending on build `#define`s")]
+        raw_buf_consumed: isize,
+        #[cfg(any(feature = "pg17", feature = "pg18"))]
+        #[cfg_attr(feature = "pg17", deprecated(since = "0.17.0", note = "may be unsound to access on Postgres 17 depending on build `#define`s")]
+        raw_buf_remaining: isize,
```
workingjubilee: oops
```suggestion
        #[cfg(any(feature = "pg17", feature = "pg18"))]
        #[cfg_attr(feature = "pg17", deprecated(since = "0.17.0", note = "may be unsound to access on Postgres 17 depending on build `#define`s"))]
        raw_buf: *mut core::ffi::c_char,
        #[cfg(any(feature = "pg17", feature = "pg18"))]
        #[cfg_attr(feature = "pg17", deprecated(since = "0.17.0", note = "may be unsound to access on Postgres 17 depending on build `#define`s"))]
        raw_buf_consumed: isize,
        #[cfg(any(feature = "pg17", feature = "pg18"))]
        #[cfg_attr(feature = "pg17", deprecated(since = "0.17.0", note = "may be unsound to access on Postgres 17 depending on build `#define`s"))]
        raw_buf_remaining: isize,
```

pgrx-pg-sys/src/libpq.rs
workingjubilee: Hm. Yeah, I'm going to split this along the Postgres 17 "divider".

pgrx-pg-sys/src/libpq.rs
```diff
@@ -0,0 +1,202 @@
...
+        tcp_user_timeout: core::ffi::c_int,
+
+        // as if ENABLE_GSS == false && ENABLE_SSPI == false
+        gss: *mut core::ffi::c_void,
+
+        ssl_in_use: bool,
+        peer_cn: *mut core::ffi::c_char,
+        peer_cert_valid: bool,
+
+        alpn_used: bool,
```
cbandy, line 117: Should this have a `peer_dn` field here?

https://github.com/postgres/postgres/blob/REL_17_0/src/include/libpq/libpq-be.h#L199-L206

```c
	/*
	 * SSL structures.
	 */
	bool		ssl_in_use;
	char	   *peer_cn;
	char	   *peer_dn;
	bool		peer_cert_valid;
	bool		alpn_used;
```

pgrx-pg-sys/src/libpq.rs
```diff
@@ -0,0 +1,202 @@
...
+        last_read_was_eof: bool,
```
cbandy, line 192: Same here; `peer_dn` field?

https://github.com/postgres/postgres/blob/REL_18_0/src/include/libpq/libpq-be.h#L205-L213

```c
	/*
	 * SSL structures.
	 */
	bool		ssl_in_use;
	char	   *peer_cn;
	char	   *peer_dn;
	bool		peer_cert_valid;
	bool		alpn_used;
	bool		last_read_was_eof;
```

cbandy (commented):
👍🏻 I like carving pg17 out as the... bad one from upstream.

🔧 I think I see two missing fields, but I may be mistaken.

pgrx-pg-sys/src/libpq.rs
workingjubilee: Thanks for the catch.

PR #2167, opened 2025-10-28
Restore so-called needless lifetimes

workingjubilee:
These were removed in pgcentralfoundation/pgrx#2097 but some instances wound up erasing deliberate, meaningful names for lifetimes. Eliding them as technically-meaningless at some use-sites is unhelpful.


workingjubilee:
Hm. I should probably go over this to restore the ones that actually contained information, and not the ones that are just `'a`

PR #2165, opened 2025-10-28
Use Rust Edition 2024

workingjubilee:
There are many cases pgrx can benefit from let-chains. Format the world to allow using Edition 2024.

NOTE: This is, strictly speaking, a breaking change because bindgen will alter the name for the `gen` field on certain structs to `gen_`.

PR #2169, opened 2025-10-28
Replace with let-chains in pgrx-sql-entity-graph

workingjubilee:
The real reason for pgcentralfoundation/pgrx#2165

My vengeance has come at last.

### DIFF
--- a/pgrx-sql-entity-graph/src/used_type.rs
+++ b/pgrx-sql-entity-graph/src/used_type.rs
@@ -274,16 +274,13 @@ impl UsedType {
 
         // if the Type is like `Result<T, E>`, this finds the `T`
         let mut resolved_ty_inner: Option<syn::Type> = None;
-        if result {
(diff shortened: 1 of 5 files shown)


pgrx-sql-entity-graph/src/used_type.rs
```diff
@@ -274,16 +274,13 @@ impl UsedType {
...
-                }
-            }
+        if result
+            && let syn::Type::Path(tp) = &resolved_ty
+            && let Some(first_segment) = tp.path.segments.first()
+            && let syn::PathArguments::AngleBracketed(ab) = &first_segment.arguments
+            && let Some(syn::GenericArgument::Type(ty)) = ab.args.first()
+        {
+            resolved_ty_inner = Some(ty.clone());
         }
```
workingjubilee, line 284: I WIN. YOU LOSE!

"uh jubilee who are you arguing with"

the rightwards drift, probably

PR #2175, opened 2025-10-28
Show line number of failed asserts

piki:
A little quality-of-life improvement: when a unit test fails, the test runner now shows the line number.  For tests with several assertions, this is quite handy.

Before:
Screenshot 2025-10-28 at 3 59 27 PM

After:
Screenshot 2025-10-28 at 3 57 15 PM

### DIFF
--- a/pgrx-tests/src/framework.rs
+++ b/pgrx-tests/src/framework.rs
@@ -150,8 +150,11 @@ pub fn run_test(
                     return Ok(());
                 }
 
-                let pg_location = dberror.file().unwrap_or("<unknown>").to_string();
+                let mut pg_location = dberror.file().unwrap_or("<unknown>").to_string();
                 let rust_location = dberror.where_().unwrap_or("<unknown>").to_string();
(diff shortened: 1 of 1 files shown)


workingjubilee:
and `column()` would not do what we expect, it seems

workingjubilee (approved):
Thanks!

piki:
> and `column()` would not do what we expect, it seems

`dberror.column()` is the name (a `&str`) of the column in the relation, not a column number where Rust panicked.  `panic.rs` does retain the panic's column number for a couple of display formatters, but there isn't really a place for it to go in the `dberror`, so it's not available in `framework.rs`.

If you want `framework.rs` to display the column number, we could append it to the message field, like `assertion 'left == right' failed at {file}:{line} column {col}`, but IMO the line number is almost always enough to debug a broken test.

workingjubilee:
If you wanted I could review that PR but this seems good enough to add on its own!

PR #2179, opened 2025-10-29
Add rust-analyzer.toml

workingjubilee:
This allows the repo to work for anyone using an r-a-based editor, regardless of whether it's VS Code or Helix or whatever.


workingjubilee:
Note that Helix still winds up trying build commands that don't work but that's fine, rust-analyzer still gets the information it wants from Flycheck

PR #2182, opened 2025-10-29
Replace `impl AsRef<T>` with `&T`

workingjubilee:
A pattern that was popular some years ago, when `impl Trait` arguments were first introduced, was to make everything an `impl Trait` arg. This can be undesirable because it weakens type inference and spreads whatever indirection is used throughout the codebase, demanding repeat invocations of `.as_ref()` or whatever. Unfortunately, pgrx was first made in the period of that being en vogue.

The alternative, of requiring things like `Option<&T>`, does demand using `Option::as_deref` a bit, but fortunately that convenience exists nowadays. It also means that the indirection is handled once at the top level and not addressed after that, which is more appropriate for internals. ~~None~~ **Almost** none of the relevant functions in pgrx are considered public API, after all, as they are found throughout cargo-pgrx and pgrx-sql-entity-graph.

Remove almost all instances of `impl AsRef`, usually with `Path` or `str`, and replace them with `&Path` or `&str` as appropriate. This deletes a few lines from the codebase outright, and in those that remain, it usually simplifies the existing code. In a couple cases, the benefit involves actually removing `clone`s of data!

The remaining instances linger with a note about their shame.

### DIFF
--- a/cargo-pgrx/src/command/install.rs
+++ b/cargo-pgrx/src/command/install.rs
@@ -71,9 +71,9 @@ impl CommandExecute for Install {
             return sudo_install.execute();
         }
 
-        let metadata = crate::metadata::metadata(&self.features, self.manifest_path.as_ref())
+        let metadata = crate::metadata::metadata(&self.features, self.manifest_path.as_deref())
             .wrap_err("couldn't get cargo metadata")?;
-        crate::metadata::validate(self.manifest_path.as_ref(), &metadata)?;
+        crate::metadata::validate(self.manifest_path.as_deref(), &metadata)?;
         let package_manifest_path =
             crate::manifest::manifest_path(&metadata, self.package.as_ref())
                 .wrap_err("Couldn't get manifest path")?;
--- a/cargo-pgrx/src/command/install.rs
+++ b/cargo-pgrx/src/command/install.rs
@@ -100,7 +100,7 @@ impl CommandExecute for Install {
 
         display_version_info(&pg_config, &PgVersionSource::PgConfig(pg_config.label()?));
         install_extension(
-            self.manifest_path.as_ref(),
+            self.manifest_path.as_deref(),
             self.package.as_ref(),
             &package_manifest_path,
             &pg_config,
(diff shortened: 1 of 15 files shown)


workingjubilee:
My mistake: it is semi-publicly exposed in a way that I do not care to fix in _one_ place.

cargo-pgrx/src/command/install.rs
```diff
@@ -364,21 +364,21 @@ fn copy_sql_files(
...
             user_package,
-            &package_manifest_path,
+            package_manifest_path.as_ref(),
             profile,
             is_test,
             features,
             target,
             Some(&dest),
-            Option::<String>::None,
+            None,
```
workingjubilee, line 373: This is a perfect example of what I mean by "using `impl AsRef` everywhere weakens type inference". It can't even infer the type of an Option, here!

cargo-pgrx/src/command/install.rs
```diff
@@ -364,21 +364,21 @@ fn copy_sql_files(
...
             None,
             skip_build,
             output_tracking,
         )?;
     }
 
     // now copy all the version upgrade files too
-    if let Ok(dir) = fs::read_dir(package_manifest_path.as_ref().parent().unwrap().join("sql/")) {
+    if let Ok(dir) = fs::read_dir(package_manifest_path.parent().unwrap().join("sql/")) {
```
workingjubilee, line 381: A lot of the churn is in cases like this, where we just are straight-up deleting intermediate function calls.

cargo-pgrx/src/command/schema.rs
```diff
@@ -97,15 +97,15 @@ impl CommandExecute for Schema {
...
-            package_manifest_path,
+            &package_manifest_path,
             &profile,
             self.test,
             &self.features,
             self.target.as_deref(),
-            self.out.as_ref(),
-            self.dot,
+            self.out.as_deref(),
+            self.dot.as_deref(),
```
workingjubilee, line 108: there's the occasional straight loss, but they don't justify the rest.

workingjubilee:
All this was done because it made it excessively painful to split up functions, which is probably why this codebase has so many huge ones.

PR #2200, opened 2025-11-06
Add back `peer_dn` field to `Port` on `pg17..`

workingjubilee:
Catch the field I accidentally dropped while splitting `Port` into different written-out versions, thanks to @cbandy!


cbandy:
I'm excited to try this soon. 🚀

Issue #1765, opened 2024-07-08
Enforcing array type structure


workingjubilee:
Some datatypes in Postgres are inductive types. Notably, rows/composite types are: a composite types is made up of a selection of fields of other types, including, potentially, a composite type! However, *arrays* are not an inductive datatype in Postgres. There is no representation for "array of array of T" that is sensical: there are only arrays of higher dimensionality. And an array must form a matrix, see #1762 for more on that.

...however, you can make an array that is made of composite types, and you can make a composite type that has an array as a field, making the fact it isn't implemented as a potentially-inductive datatype... rather pointless?

Whatever, I don't really care.

...Except I do care that I get reports of people trying to return `Vec<Vec<T>>`, but that doesn't have a coherent repr in Postgres, so they get something they didn't expect. If we made it so this didn't work unless it was something an array considered a valid type, this would fail.

workingjubilee:
well, it turns out bounding this is in fact is necessary to make it possible to use `FlatArray` correctly, as otherwise we trust the wrong thing. lol.

PR #2207, opened 2025-11-17
Add `FlatArray<'_, T>`

workingjubilee:
Introduce a new Rust type for what ArrayTypes "really are": FlatArray. As an unsized type, it will primarily be used as `&FlatArray<'_, T>` or similar in actual extension code. The long-term plan is for FlatArray to phase out `pgrx::datum::Array<'_, T>` entirely, as it has several advantages over its predecessor:

- It uses `Element`, a subset of `BorrowDatum`, to power its simplicity in iteration, making it more easily verifiable and maintainable.
- It is the unsized type itself, allowing composing it with various Rusty pointer types like `&`, `&mut`, or `PBox`.
- It is no longer endangered by performance cliffs or hidden allocations depending on the type of `T`.
- Because of these previous traits, it is possible to soundly and safely allocate a fresh `FlatArray`, mutate it, and return that one allocation, instead of allocating an intermediate Vec in Rust memory to mutate!

It also has an associated cost in that it no longer supports any `T` that requires complex unboxing of items like `datum::Array<'a, String>`. You must instead use `array.iter().map(closure)` to accomplish that. This itself summarizes the difference: `datum::Array` is effectively an iteration protocol, but `array::FlatArray` is an actual data structure.

This addresses the implied feature request in, and thus supercedes, pgcentralfoundation/pgrx#2086.


workingjubilee:
There's one mystery of why the lifetime isn't something I can use, here, but I'll leave that one for tomorrow. It will require digging into our macro expansion code, which is always "fun".

workingjubilee:
Dismantling this incrementally into other PRs so that it becomes more obvious what is actually a key part of this PR overall, and also to make it easier to pinpoint the macro/lifetime issue.

workingjubilee:
Biggest remaining question for this PR that I've been thinking about:
- Is `BorrowDatum` the right constraint, or should it be `ArrayElement`?

workingjubilee:
The answer is yes, partly because `FlatArray` itself is `BorrowDatum`, which means it runs afoul of https://github.com/pgcentralfoundation/pgrx/issues/1765

workingjubilee:
I think the other Q is "can we get rid of the inner lifetime?"

workingjubilee:
In order to make this work, we need to break `BorrowDatum` apart and introduce `Element`, which provides most of its functionality.

`Element` then receives a blanket implementation of `BorrowDatum` for it.

workingjubilee:
## TODO

- [x] Make less tests in `array_borrowed` commented-out
- [ ] ~~Write some more functional tests~~ walled on SPI
- [x] Think about if there's any more compile-fails that should be added
- [x] Rebase on https://github.com/pgcentralfoundation/pgrx/pull/2218 when that is landed
- [x] Rewrite PR description
- [x] Investigate multidimensional array support

workingjubilee:
Current status re: multidimensional arrays:

Postgres currently still claims that MDAs are not different from other arrays and that it doesn't enforce dimensionality to arguments, so I have to have an array type that handles MDAs anyways. This remains funny to me because Postgres has to several times, while creating arrays, enforce a sort of boundary on dimensionality (because Postgres arrays are rigid in length-per-dimension).

workingjubilee:
( The reason this is so is because dimensionality is not encoded on the type axis, which I have... Opinions about. )

workingjubilee:
Hm. So, actually writing more useful tests is at a bit of a wall due to SPI. And others depend on Text, instead. So I think this is where this all comes to a head, now.

pgrx/src/array/flat_array.rs
```diff
@@ -0,0 +1,340 @@
...
+    pub fn contains_nulls(&self) -> bool {
+        // SAFETY: Constructive validity from ref and function is non-mutating
+        unsafe { pg_sys::array_contains_nulls((&raw const self.head).cast_mut()) }
+    }
+}
+
+#[derive(Debug)]
+pub struct ArrayAllocError {
+    _stuff: (),
+}
```
workingjubilee: This allows the `match err.kind()` pattern without committing us to a specific internal representation of the error type here, which may be useful if Postgres creates new restrictions on array parameters (e.g. individual dimension length).

pgrx/src/array/flat_array.rs
```diff
@@ -0,0 +1,340 @@
...
+    Some problems with the design of an iter_mut for FlatArray:
+    In order to traverse the array, we need to assume well-formedness of e.g. cstring/varlena elements,
+    but &mut would allow safely updating varlenas within their span, e.g. injecting \0 into cstrings.
+    making it so that nothing allows making an ill-formed varlena via &mut seems untenable, also?
+    probably only viable to expose &mut for fixed-size types, then
+    */
+    pub fn iter_mut(&mut self) -> ArrayIterMut<'mcx, T> {
+        ???
+    }
+    */
```
workingjubilee: I should remove this.
```suggestion
```

pgrx/src/array/flat_array.rs
```diff
@@ -0,0 +1,340 @@
...
+    /// Borrow the nth element (0-indexed)
+    ///
+    /// `FlatArray::get` may have to iterate elements, so this is `O(n)` in the general case
+    pub fn get(&self, index: usize) -> Option<Nullable<&T>> {
+        self.iter().nth(index)
+    }
```
workingjubilee, line 228: note that because we require `T: Element` for FlatArray, we will be able to add functions to `Element` that enable faster indexing based on knowledge of `T`'s layout

pgrx/src/array/flat_array.rs
```diff
@@ -0,0 +1,340 @@
...
+        let index = 0;
+        let offset = 0;
+        let align = Layout::lookup_oid(self.head.elemtype).align;
+
+        ArrayIter { data, nulls, nelems, arr, index, offset, align }
+    }
+
+    pub fn iter_non_null(&self) -> impl Iterator<Item = &T> {
+        self.iter().filter_map(|elem| elem.into_option())
+    }
```
workingjubilee, line 220: This can be significantly optimized later.

pgrx/src/array/flat_array.rs
```diff
@@ -0,0 +1,340 @@
...
+            product = product.checked_mul(dim).unwrap();
+            product + 1
+        });
+        let nelems = product as usize;
+
+        let null_size = if has_nulls { nelems.div_ceil(8) } else { 0 };
+
+        let prefix_size = base_size + dims_size * 2 + null_size;
+        const MAX_ELEM_ALIGN: usize = pg_sys::MAXIMUM_ALIGNOF as _;
+        const { assert!(align_of::<T>() <= MAX_ELEM_ALIGN) };
```
workingjubilee, line 143: Hm, this assert is starting to appear in a bunch of places...

pgrx/src/array/flat_array.rs
```diff
@@ -0,0 +1,340 @@
...
+{
+    pub fn new_zeroed_in<'cx, const N: usize>(
+        dims: [usize; N],
+        has_nulls: bool,
+        memcx: &MemCx<'cx>,
+    ) -> Result<PBox<'cx, FlatArray<'cx, T>>, ArrayAllocError> {
+        let base_size = size_of::<pg_sys::ArrayType>();
+        let ndims = N;
+        let dims_size = size_of::<ffi::c_int>() * ndims;
+        const { assert!(N != 0) };
```
workingjubilee: This could also return an unconditional ArrayAllocError but that seems somewhat pointless as it just makes the error dynamic.

pgrx/src/array/flat_array.rs
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    pub fn contains_nulls(&self) -> bool {
+        // SAFETY: Constructive validity from ref and function is non-mutating
+        unsafe { pg_sys::array_contains_nulls((&raw const self.head).cast_mut()) }
+    }
+}
+
+#[derive(Debug)]
+pub struct ArrayAllocError {
```
workingjubilee: Maybe this should be `ArrayBuildError` so that it can be used by `ArrayBuilder`?

`NewArrayError`?

pgrx/src/array/flat_array.rs
workingjubilee: I don't know if this is actually useful though, since it would effectively be a breaking change anyways, so people would want to update their extensions for it anyways?

pgrx/src/array/flat_array.rs
```diff
@@ -103,16 +112,16 @@ where
...
-            return Err(ArrayAllocError { _stuff: () });
-        };
+
+        if let Some((zlen_dim, _)) = dims.into_iter().enumerate().find(|(i, len)| *len == 0) {
+            return Err(ArrayAllocError::ZeroLenDim(zlen_dim));
+        }
+        if let Some(over) = nelems.checked_sub(MAX_ARRAY_SIZE) {
+            return Err(ArrayAllocError::TooManyElems { over });
+        }
+        if let Some(over) = size.checked_sub(MAX_ALLOC_SIZE) {
```
workingjubilee: this is probably off by one

workingjubilee:
At first I couldn't think of any more tests to write, but I can at least write down the compile-time error cases for this API.

workingjubilee:
TODO:
- [x] `from_slice` constructor
- [x] `new_empty` constructor

workingjubilee:
Adding the `from_slice` constructor in particular turned out to be a particularly healthy exercise: it caught `as_non_null_slice_mut` having `&[T]` as its inner return type, heh. Oops.

workingjubilee:
`trybuild` does not support build-fail tests, despite the name, so mono-time checks aren't viable to test for, annoyingly.

pgrx/src/array/flat_array.rs
```diff
@@ -64,14 +64,10 @@ where
...
-        over: usize,
-    },
-    TooManyElems {
-        over: usize,
-    },
+    TooManyBytes,
+    TooManyElems,
     /// One or more dimensions are zero
-    ZeroLenDim(usize),
+    ZeroLenDim,
```
workingjubilee, line 79: While returning the information we had was cute, it wasn't viable in all cases without compromising the code significantly, and it was always at risk of being incomplete, so I stopped doing that.

pgrx/src/array/flat_array.rs
```diff
@@ -0,0 +1,377 @@
...
+            // COMPAT: assign so fields must be initialized even if ArrayType changes
+            // SAFETY: _ARRAY_TYPE_IS_PADDING_FREE means we will not deinitialize any bytes
+            (*head_ptr) = pg_sys::ArrayType {
+                vl_len_: varlena::encode_vlen_4b(nbytes) as i32,
+                ndim: ndims as ffi::c_int,
+                dataoffset,
+                elemtype,
+            };
+            *(head_ptr.add(base_size).cast()) = dim_ints;
+            *(head_ptr.add(base_size + dims_size).cast()) = lbounds;
```
workingjubilee: The previous `cast` makes these `add`s incorrect. Should use `byte_add` here.

workingjubilee:
Hm. I think I should/could probably add testing that only relies on direct calls instead of SPI, and that would be enough for some basic functionality, even though it doesn't test the "full cycle".

PR #2208, opened 2025-11-17
Hoist varlena encoding fn

workingjubilee:
They may see more use in this factored-out form.


workingjubilee:
Extracted from https://github.com/pgcentralfoundation/pgrx/pull/2207

PR #2210, opened 2025-11-17
Add `MemCx<'current, T>` arguments and `PBox<'current, T>` returns

workingjubilee:
tl;dr: `caller_picks_lifetime + macro_expansion == unsoundness`,
so prevent that by adding API for new, lifetime-bound allocations.

For functions exposed to PostgreSQL via `pg_extern`, signatures often resemble these:

```rust
#[pg_extern]
fn do_something(
    floats: Array<'a, f32>,
    hms: Time,
    day: TimestampTz
) -> Vec<TimestampTz> {
```

```sql
-- pgrx will generate something like this
CREATE FUNCTION do_something(
    floats real[],
    hms time,
    day timestamptz
) RETURNS timestamptz[]
```

In SQL, we accept an array and return an array of a different type, but in Rust, we return a `Vec`. This is because we lack API to allocate and return a fresh `Array<'a, T>`s. Behind the scenes, we *do* allocate an `ArrayType` at the translation boundary, unseen to a Rust programmer. That, plus the Vec, means two allocations just to mutate an array, even for a fixed-size type!

Unfortunately, pgrx-offered type translations like `Array<'a, T>`, cannot have a lifetime *and* expose a new *safe* function to make it from lifetime-free values! Such API must have the `unsafe` requirement of access-exclusivity for its lifetime, described in functions like `slice::from_raw_parts_mut`[^1], because the lifetime becomes caller-selected. This makes it easy to use a `'static` lifetime with `pg_extern` and expand `unsafe` code that trusts that lifetime.

A `'static` lifetime is always wrong for `palloc`ated types, even with `TopMemoryContext`, since it dies whenever Postgres resets the context and Postgres does have conditions under which it performs such a global reset. The result is various extremely subtle soundness problems are possible, yet the root cause lies in code that most never read. We can solve this by denying the possibility of a caller-selected lifetime. This will require removing some types, ultimately, but first we need an API to migrate to.

This new API is `MemCx<'mcx, T>` and `PBox<'mcx, T>`. Importantly, there is no *safe* route to creating `MemCx<'mcx, T>` with a caller-selected lifetime. One can only reach `MemCx<'mcx, T>` safely as an argument: either pass a closure to `memcx::current_context` or call a new `pg_extern` function via SQL. Either way gives you `&MemCx<'current>`. These let you create a lifetime-bound allocation that can live long enough to return it.

We can already return fixed-size pass-by-reference types, so right now `PBox` is only an optimization. It allows not putting a type on the stack and then immediately moving it directly into a heap allocation. More importantly, it enables writing tests, so it is important ground work for returning de facto unsized values, as they can hide behind `PBox` which is always `Sized`.

[^1]: https://doc.rust-lang.org/std/slice/fn.from_raw_parts_mut.html

### DIFF
--- a/pgrx-tests/src/tests/memcxt_tests.rs
+++ b/pgrx-tests/src/tests/memcxt_tests.rs
@@ -37,10 +37,7 @@ mod tests {
     ) -> PBox<'mcx, TimeWithTimeZone> {
         let palloc = memcx.alloc_bytes(size_of::<TimeWithTimeZone>());
         let timetz = TimeWithTimeZone::new(4, 20, 0.0).unwrap();
-        unsafe {
-            *(palloc as *mut _) = timetz;
-            PBox::from_raw_in(NonNull::new(palloc.cast()).unwrap(), memcx)
-        }
+        PBox::new_in(timetz, memcx)
     }
 
     #[pg_test]
--- a/pgrx-tests/src/tests/memcxt_tests.rs
+++ b/pgrx-tests/src/tests/memcxt_tests.rs
@@ -18,7 +18,6 @@ mod tests {
     use pgrx::palloc::PBox;
     use pgrx::pg_test;
     use pgrx::prelude::*;
-    use std::ptr::NonNull;
     use std::sync::Arc;
     use std::sync::atomic::{AtomicBool, Ordering};
 
--- a/pgrx-tests/src/tests/memcxt_tests.rs
+++ b/pgrx-tests/src/tests/memcxt_tests.rs
@@ -14,9 +14,10 @@ mod tests {
     use crate as pgrx_tests;
 
     use pgrx::PgMemoryContexts;
-    use pgrx::memcx::MemCx;
+    use pgrx::memcx::{MemCx, PBox};
     use pgrx::pg_test;
     use pgrx::prelude::*;
+    use std::ptr::NonNull;
     use std::sync::Arc;
     use std::sync::atomic::{AtomicBool, Ordering};
 
@@ -31,8 +32,15 @@ mod tests {
     }
 
     #[pg_extern]
-    pub fn accept_return_memcx<'mcx>(memcx: &'mcx MemCx<'mcx>) -> &'mcx MemCx<'mcx> {
-        memcx
+    pub fn accept_mcx_return_timetz<'mcx>(
+        memcx: &'mcx MemCx<'mcx>,
+    ) -> PBox<'mcx, TimeWithTimeZone> {
+        let palloc = memcx.alloc_bytes(size_of::<TimeWithTimeZone>());
+        let timetz = TimeWithTimeZone::new(4, 20, 0.0).unwrap();
+        unsafe {
+            *(palloc as *mut _) = timetz;
+            PBox::from_raw_in(NonNull::new(palloc.cast()).unwrap(), memcx)
+        }
     }
 
     #[pg_test]
--- a/pgrx-tests/src/tests/memcxt_tests.rs
+++ b/pgrx-tests/src/tests/memcxt_tests.rs
@@ -14,6 +14,7 @@ mod tests {
     use crate as pgrx_tests;
 
     use pgrx::PgMemoryContexts;
+    use pgrx::memcx::MemCx;
     use pgrx::pg_test;
     use pgrx::prelude::*;
     use std::sync::Arc;
@@ -27,6 +28,11 @@ mod tests {
         fn drop(&mut self) {
             self.did_drop.store(true, Ordering::SeqCst);
         }
+    }
+
+    #[pg_extern]
+    pub fn accept_return_memcx<'mcx>(memcx: &'mcx MemCx<'mcx>) -> &'mcx MemCx<'mcx> {
+        memcx
     }
 
     #[pg_test]
--- a/pgrx-tests/src/tests/memcxt_tests.rs
+++ b/pgrx-tests/src/tests/memcxt_tests.rs
@@ -14,7 +14,8 @@ mod tests {
     use crate as pgrx_tests;
 
     use pgrx::PgMemoryContexts;
-    use pgrx::memcx::{MemCx, PBox};
+    use pgrx::memcx::MemCx;
+    use pgrx::palloc::PBox;
     use pgrx::pg_test;
     use pgrx::prelude::*;
     use std::ptr::NonNull;
--- a/pgrx/src/memcx.rs
+++ b/pgrx/src/memcx.rs
@@ -10,8 +10,7 @@ use pgrx_sql_entity_graph::metadata::{
 // - mctx
 // Search engines will see "memc[tx]{2}" and assume you mean memcpy!
 // And it's nice-ish to have shorter lifetime names and have 'mcx consistently mean the lifetime.
-use crate::callconv::{Arg, ArgAbi, BoxRet, FcInfo};
-use crate::datum::{BorrowDatum, Datum};
+use crate::callconv::{Arg, ArgAbi};
 use crate::pg_sys;
 use core::{marker::PhantomData, ptr::NonNull};
 
@@ -150,40 +149,5 @@ unsafe impl<'mcx> SqlTranslatable for &MemCx<'mcx> {
 
     fn return_sql() -> Result<Returns, ReturnsError> {
         Ok(Returns::One(SqlMapping::Skip))
-    }
-}
-
-/// A type pallocated in a context
-pub struct PBox<'mcx, T: ?Sized> {
-    ptr: NonNull<T>,
-    _cx: PhantomData<MemCx<'mcx>>,
-}
-
-impl<'mcx, T: ?Sized> PBox<'mcx, T> {
-    pub unsafe fn from_raw_in(ptr: NonNull<T>, _cx: &MemCx<'mcx>) -> PBox<'mcx, T> {
-        PBox { ptr, _cx: PhantomData }
-    }
-}
-
-unsafe impl<'mcx, T> BoxRet for PBox<'mcx, T>
-where
-    T: ?Sized + BorrowDatum,
-{
-    unsafe fn box_into<'fcx>(self, fcinfo: &mut FcInfo<'fcx>) -> Datum<'fcx> {
-        // SAFETY: by proxy
-        unsafe { fcinfo.return_raw_datum(pg_sys::Datum::from(self.ptr.cast::<u8>().as_ptr())) }
-    }
-}
-
-unsafe impl<'mcx, T> SqlTranslatable for PBox<'mcx, T>
-where
-    T: SqlTranslatable + ?Sized,
-{
-    fn argument_sql() -> Result<SqlMapping, ArgumentError> {
-        T::argument_sql()
-    }
-
-    fn return_sql() -> Result<Returns, ReturnsError> {
-        T::return_sql()
     }
 }
--- a/pgrx/src/memcx.rs
+++ b/pgrx/src/memcx.rs
@@ -175,5 +175,18 @@ where
     }
 }
 
+unsafe impl<'mcx, T> SqlTranslatable for PBox<'mcx, T>
+where
+    T: SqlTranslatable + ?Sized,
+{
+    fn argument_sql() -> Result<SqlMapping, ArgumentError> {
+        T::argument_sql()
+    }
+
+    fn return_sql() -> Result<Returns, ReturnsError> {
+        T::return_sql()
+    }
+}
+
 /// An "owning" palloc.
 pub struct Palloc<'mcx, T>(T, &'mcx MemCx<'mcx>);
--- a/pgrx/src/memcx.rs
+++ b/pgrx/src/memcx.rs
@@ -171,7 +171,7 @@ where
 {
     unsafe fn box_into<'fcx>(self, fcinfo: &mut FcInfo<'fcx>) -> Datum<'fcx> {
         // SAFETY: by proxy
-        unsafe { fcinfo.return_raw_datum(mem::transmute(self.ptr)) }
+        unsafe { fcinfo.return_raw_datum(pg_sys::Datum::from(self.ptr.cast::<u8>().as_ptr())) }
     }
 }
 
--- a/pgrx/src/memcx.rs
+++ b/pgrx/src/memcx.rs
@@ -152,3 +152,28 @@ unsafe impl<'mcx> SqlTranslatable for &MemCx<'mcx> {
         Ok(Returns::One(SqlMapping::Skip))
     }
 }
+
+/// A type pallocated in a context
+pub struct PBox<'mcx, T: ?Sized> {
+    ptr: NonNull<T>,
+    _cx: PhantomData<MemCx<'mcx>>,
+}
+
+impl<'mcx, T: ?Sized> PBox<'mcx, T> {
+    pub unsafe fn from_raw_in(ptr: NonNull<T>, _cx: &MemCx<'mcx>) -> PBox<'mcx, T> {
+        PBox { ptr, _cx: PhantomData }
+    }
+}
+
+unsafe impl<'mcx, T> BoxRet for PBox<'mcx, T>
+where
+    T: ?Sized + BorrowDatum,
+{
+    unsafe fn box_into<'fcx>(self, fcinfo: &mut FcInfo<'fcx>) -> Datum<'fcx> {
+        // SAFETY: by proxy
+        unsafe { fcinfo.return_raw_datum(mem::transmute(self.ptr)) }
+    }
+}
+
+/// An "owning" palloc.
+pub struct Palloc<'mcx, T>(T, &'mcx MemCx<'mcx>);
--- a/pgrx/src/memcx.rs
+++ b/pgrx/src/memcx.rs
@@ -187,6 +187,3 @@ where
         T::return_sql()
     }
 }
-
-/// An "owning" palloc.
-pub struct Palloc<'mcx, T>(T, &'mcx MemCx<'mcx>);
--- a/pgrx/src/memcx.rs
+++ b/pgrx/src/memcx.rs
@@ -1,4 +1,8 @@
 //! Memory Contexts in PostgreSQL, now with lifetimes.
+use pgrx_sql_entity_graph::metadata::{
+    ArgumentError, Returns, ReturnsError, SqlMapping, SqlTranslatable,
+};
+
 // "Why isn't this pgrx::mem or pgrx::memcxt?"
 // Postgres actually uses all of:
 // - mcxt
@@ -136,5 +140,15 @@ unsafe impl<'fcx> ArgAbi<'fcx> for &MemCx<'fcx> {
 
     fn is_virtual_arg() -> bool {
         true
+    }
+}
+
+unsafe impl<'mcx> SqlTranslatable for &MemCx<'mcx> {
+    fn argument_sql() -> Result<SqlMapping, ArgumentError> {
+        Ok(SqlMapping::Skip)
+    }
+
+    fn return_sql() -> Result<Returns, ReturnsError> {
+        Ok(Returns::One(SqlMapping::Skip))
     }
 }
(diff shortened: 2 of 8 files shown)


workingjubilee:
moving PBox work into this PR since it should probably work for things that aren't just FlatArray.

workingjubilee:
```rust
        fn_args: <[_]>::into_vec(::alloc::boxed::box_new([
            ::pgrx::pgrx_sql_entity_graph::PgExternArgumentEntity {
                pattern: "memcx",
                used_ty: ::pgrx::pgrx_sql_entity_graph::UsedTypeEntity {
                    ty_source: "& '_ MemCx < 'mcx >",
                    ty_id: core::any::TypeId::of::<&'_ MemCx<'mcx>>(),
                    full_path: core::any::type_name::<&'_ MemCx<'mcx>>(),
                    module_path: {
                        let ty_name = core::any::type_name::<&'_ MemCx<'mcx>>();
                        let mut path_items: Vec<_> = ty_name.split("::").collect();
                        let _ = path_items.pop();
                        path_items.join("::")
                    },
                    composite_type: None,
                    variadic: false,
                    default: None,

                    #[doc = r" Set via the type being an `Option`."]
                    optional: false,
                    metadata: {
                        use ::pgrx::pgrx_sql_entity_graph::metadata::SqlTranslatable;
                        <&'_ MemCx<'mcx>>::entity()
                    },
                },
            },
        ])),
```
I believe it is this part of the expansion that is "buggy" for our purposes: this should be an anonymized or higher-rank lifetime in these cases.

workingjubilee:
Ah, just a missing recursion for TypeReference in `anonymize_lifetimes`!

pgrx-tests/src/tests/memcxt_tests.rs
```diff
@@ -29,6 +31,18 @@ mod tests {
...
+    #[pg_extern]
+    pub fn accept_mcx_return_timetz<'mcx>(
+        memcx: &'mcx MemCx<'mcx>,
+    ) -> PBox<'mcx, TimeWithTimeZone> {
+        let palloc = memcx.alloc_bytes(size_of::<TimeWithTimeZone>());
+        let timetz = TimeWithTimeZone::new(4, 20, 0.0).unwrap();
+        unsafe {
+            *(palloc as *mut _) = timetz;
+            PBox::from_raw_in(NonNull::new(palloc.cast()).unwrap(), memcx)
+        }
```
workingjubilee: This... is an overly finicky pattern. I should probably provide a safe convenience.

pgrx-sql-entity-graph/src/lifetimes.rs
```diff
@@ -51,6 +51,8 @@ pub fn anonymize_lifetimes(value: &mut syn::Type) {
             if let Some(lifetime) = type_ref.lifetime.as_mut() {
                 lifetime.ident = syn::Ident::new("_", lifetime.ident.span());
             }
+            // recurse
+            anonymize_lifetimes(&mut type_ref.elem);
```
workingjubilee: the offender

pgrx/src/memcx.rs
```diff
@@ -116,3 +123,65 @@ mod nightly {
         }
     }
 }
+
+unsafe impl<'fcx> ArgAbi<'fcx> for &MemCx<'fcx> {
+    unsafe fn unbox_arg_unchecked(_arg: Arg<'_, 'fcx>) -> Self {
+        // SAFETY: We are called to unbox an argument, which means the backend was initialized.
+        // We use this horrific expression to allow the lifetime to be extended arbitrarily
+        // and achieve an "in-place" transformation of CurrentMemoryContext's pointer.
+        unsafe { &*((&raw mut pg_sys::CurrentMemoryContext).cast()) }
```
workingjubilee, line 136: I was sort of disturbed when I figured this out

workingjubilee:
TODO:
- [x] provide more allocation functions for `MemCx` and `PBox` to cover some common cases
- [x] provide a few random things from `Box`'s API that seem useful
- [x] move `PBox` to `palloc` module (it will eventually be joined by a `PStr` and `PVec`, in all likelihood)

workingjubilee:
I initially thought I could add more conveniences, more easily, but it actually has a bit of API design required to do well at it. My biggest add without much design effort is to offer a `PBox::new_in` equivalent.

pgrx/src/palloc/pbox.rs
```diff
@@ -0,0 +1,60 @@
...
+
+unsafe impl<'mcx, T> BoxRet for PBox<'mcx, T>
+where
+    T: ?Sized + BorrowDatum,
+{
+    unsafe fn box_into<'fcx>(self, fcinfo: &mut FcInfo<'fcx>) -> Datum<'fcx> {
+        // SAFETY: by proxy
+        unsafe { fcinfo.return_raw_datum(pg_sys::Datum::from(self.ptr.cast::<u8>().as_ptr())) }
+    }
+}
```
workingjubilee, line 71: This is subtly incorrect: it doesn't account for whether the return has a by-ref or by-value convention!

workingjubilee: Fixing this is subtly tricky, because the answer "should" be to use `BoxRet::box_into(self.ptr.read(), fcinfo)`, except we haven't bounded `T: BoxRet`, somewhat intentionally.

workingjubilee: Oh, I can copy values into a raw Datum and return that via the same approach.

pgrx/src/memcx.rs
```diff
@@ -116,3 +124,30 @@ mod nightly {
         }
     }
 }
+
+unsafe impl<'fcx> ArgAbi<'fcx> for &MemCx<'fcx> {
+    unsafe fn unbox_arg_unchecked(_arg: Arg<'_, 'fcx>) -> Self {
+        // SAFETY: We are called to unbox an argument, which means the backend was initialized.
+        // We use this horrific expression to allow the lifetime to be extended arbitrarily
+        // and achieve an "in-place" transformation of CurrentMemoryContext's pointer.
```
workingjubilee, line 133:
```suggestion
        // and achieve an "in-place" transformation of CurrentMemoryContext's pointer.
        // The soundness of this is riding on the lifetimes used for `unbox_arg_unchecked` in our macros,
        // as the expanded code is designed so that `fcinfo` and each `arg` are truly "borrowed" in rustc's eyes.
```

pgrx/src/palloc/pbox.rs
```diff
@@ -0,0 +1,78 @@
...
+                    debug_assert!(size <= size_of::<pg_sys::Datum>());
+                    // using `BorrowDatum::point_from` handles endianness
+                    let datum_ptr = T::point_from(NonNull::from_mut(&mut datum).cast::<u8>());
+                    datum_ptr.cast::<u8>().copy_from_nonoverlapping(self.ptr.cast(), size);
+                }
+                datum
+            }
+            PassBy::Ref => pg_sys::Datum::from(self.ptr.cast::<u8>().as_ptr()),
+        };
+        // SAFETY: by proxy, BorrowDatum is an `unsafe traitw so the above impl must be correct
```
workingjubilee:
```suggestion
        // SAFETY: by proxy, BorrowDatum is an `unsafe trait` so the above impl must be correct
```

pgrx/src/palloc/pbox.rs
```diff
@@ -0,0 +1,78 @@
...
+
+[stdbox]: alloc::boxed::Box
+*/
+pub struct PBox<'mcx, T: ?Sized> {
+    ptr: NonNull<T>,
+    _cx: PhantomData<MemCx<'mcx>>,
+}
+
+impl<'mcx, T: ?Sized> PBox<'mcx, T> {
+    pub unsafe fn from_raw_in(ptr: NonNull<T>, _cx: &MemCx<'mcx>) -> PBox<'mcx, T> {
```
workingjubilee, line 30:
```suggestion
    // # Safety
    // The same constraints as [`Box::from_raw`], AND 
    // - you assert the pointer was allocated in the `MemCx`
    // - you assert the pointer may be freed by `pfree`
    pub unsafe fn from_raw_in(ptr: NonNull<T>, _cx: &MemCx<'mcx>) -> PBox<'mcx, T> {
```

pgrx/src/palloc/pbox.rs
```diff
@@ -0,0 +1,78 @@
...
+        PBox { ptr, _cx: PhantomData }
+    }
+}
+
+impl<'mcx, T: Sized> PBox<'mcx, T> {
+    pub fn new_in(val: T, memcx: &MemCx<'mcx>) -> Self {
+        const { assert!(align_of::<T>() <= 8) };
```
workingjubilee: This can be relaxed on higher Postgres versions where we can use aligned allocations.

pgrx/src/memcx.rs
```diff
@@ -116,3 +124,30 @@ mod nightly {
...
+        unsafe { &*((&raw mut pg_sys::CurrentMemoryContext).cast()) }
+    }
+
+    unsafe fn unbox_nullable_arg(_arg: Arg<'_, 'fcx>) -> crate::nullable::Nullable<Self> {
+        crate::nullable::Nullable::Null
+    }
```
workingjubilee:
```suggestion
    unsafe fn unbox_nullable_arg(_arg: Arg<'_, 'fcx>) -> Nullable<Self> {
        // SAFETY: Should never happen in actuality, but as long as we're here...
        if unsafe { pg_sys::CurrentMemoryContext.is_null() } {
            Nullable::Null
        } else {
            Nullable::Valid(Self::unbox_arg_unchecked(_arg))
        }
```

pgrx/src/memcx.rs
```diff
@@ -1,15 +1,21 @@
...
+};
+
 // "Why isn't this pgrx::mem or pgrx::memcxt?"
 // Postgres actually uses all of:
 // - mcxt
 // - memcxt
 // - mctx
 // Search engines will see "memc[tx]{2}" and assume you mean memcpy!
 // And it's nice-ish to have shorter lifetime names and have 'mcx consistently mean the lifetime.
+use crate::callconv::{Arg, ArgAbi};
```
workingjubilee, line 13:
```suggestion
use crate::callconv::{Arg, ArgAbi};
use crate::nullable::Nullable;
```

pgrx/src/palloc/pbox.rs
```diff
@@ -0,0 +1,78 @@
...
+where
+    T: ?Sized + BorrowDatum,
+{
+    unsafe fn box_into<'fcx>(self, fcinfo: &mut FcInfo<'fcx>) -> Datum<'fcx> {
+        let datum = match T::PASS {
+            PassBy::Value => {
+                // start with a zeroed Datum, just to minimize funny business
+                let mut datum = pg_sys::Datum::null();
+                // SAFETY: we know this type is Sized and has a size less than the Datum, and
+                // a PBox must have an initialized pointee, so a copy is sound-by-construction
```
workingjubilee:
```suggestion
                // SAFETY: Due to BorrowDatum, this type has a definite size less than a Datum,
                // and PBox must have an initialized pointee, so a copy is sound-by-construction.
```

pgrx/src/memcx.rs
```diff
@@ -116,3 +124,30 @@ mod nightly {
...
+    unsafe fn unbox_nullable_arg(_arg: Arg<'_, 'fcx>) -> crate::nullable::Nullable<Self> {
+        crate::nullable::Nullable::Null
+    }
+
+    fn is_virtual_arg() -> bool {
+        true
+    }
+}
+
+unsafe impl<'mcx> SqlTranslatable for &MemCx<'mcx> {
```
workingjubilee, line 154:
```suggestion
/// SAFETY: virtual argument
unsafe impl<'mcx> SqlTranslatable for &MemCx<'mcx> {
```

pgrx/src/palloc/pbox.rs
```diff
@@ -0,0 +1,78 @@
...
+                datum
+            }
+            PassBy::Ref => pg_sys::Datum::from(self.ptr.cast::<u8>().as_ptr()),
+        };
+        // SAFETY: by proxy, BorrowDatum is an `unsafe traitw so the above impl must be correct
+        unsafe { fcinfo.return_raw_datum(datum) }
+    }
+}
+
+unsafe impl<'mcx, T> SqlTranslatable for PBox<'mcx, T>
```
workingjubilee, line 75:
```suggestion
/// SAFETY: SQL has no "pointers" so by-val and by-ref calling conventions are identical,
/// and all `PBox` truly does is enable pass-by-ref returns of unsized values.
unsafe impl<'mcx, T> SqlTranslatable for PBox<'mcx, T>
```

pgrx/src/palloc/pbox.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub unsafe fn from_raw_in(ptr: NonNull<T>, _cx: &MemCx<'mcx>) -> PBox<'mcx, T> {
+        PBox { ptr, _cx: PhantomData }
+    }
+}
+
+impl<'mcx, T: Sized> PBox<'mcx, T> {
+    pub fn new_in(val: T, memcx: &MemCx<'mcx>) -> Self {
+        const { assert!(align_of::<T>() <= 8) };
+        let ptr = memcx.alloc_bytes(size_of::<T>()).cast();
+        // We were guaranteed an appropriately sized allocation to write to,
```
workingjubilee:
```suggestion
        // SAFETY: We were guaranteed an appropriately sized allocation to write to,
```

workingjubilee:
Oh, I should also include `Deref`

pgrx/src/palloc/pbox.rs
```diff
@@ -0,0 +1,107 @@
...
+use pgrx_sql_entity_graph::metadata::{
+    ArgumentError, Returns, ReturnsError, SqlMapping, SqlTranslatable,
+};
+
+/** As [`Box<T, A>`][stdbox] where `A` is a [`MemCx`]
+
+
+[stdbox]: alloc::boxed::Box
+*/
+pub struct PBox<'mcx, T: ?Sized> {
```
eeeebbbbrrrr, line 20: Hi.  I don't like the name `PBox` but I don't have a better suggestion.  I think I don't like it because it's too similar to the existing `PgBox`.

pgrx/src/palloc/pbox.rs
```diff
@@ -0,0 +1,107 @@
...
+    // - you assert the pointer was allocated in the `MemCx`
+    // - you assert the pointer may be freed by `pfree`
+    pub unsafe fn from_raw_in(ptr: NonNull<T>, _cx: &MemCx<'mcx>) -> PBox<'mcx, T> {
+        PBox { ptr, _cx: PhantomData }
+    }
+}
+
+impl<'mcx, T: Sized> PBox<'mcx, T> {
+    pub fn new_in(val: T, memcx: &MemCx<'mcx>) -> Self {
+        const { assert!(align_of::<T>() <= 8) };
```
eeeebbbbrrrr: instead of `<= 8` wouldn't we want `<= size_of::<usize>()`?

pgrx/src/memcx.rs
```diff
@@ -116,3 +125,38 @@ mod nightly {
...
+    unsafe fn unbox_arg_unchecked(_arg: Arg<'_, 'fcx>) -> Self {
+        // SAFETY: We are called to unbox an argument, which means the backend was initialized.
+        // We use this horrific expression to allow the lifetime to be extended arbitrarily
+        // and achieve an "in-place" transformation of CurrentMemoryContext's pointer.
+        // The soundness of this is riding on the lifetimes used for `unbox_arg_unchecked` in our macros,
+        // as the expanded code is designed so that `fcinfo` and each `arg` are truly "borrowed" in rustc's eyes.
+        unsafe { &*((&raw mut pg_sys::CurrentMemoryContext).cast()) }
+    }
+
+    unsafe fn unbox_nullable_arg(_arg: Arg<'_, 'fcx>) -> Nullable<Self> {
```
eeeebbbbrrrr: nit: `_arg` is actually used.  call it `arg`?

pgrx/src/palloc/pbox.rs
eeeebbbbrrrr: instead of `<= 8` wouldn't we want `<= size_of::<usize>()`?

pgrx/src/memcx.rs
```diff
@@ -116,3 +125,38 @@ mod nightly {
...
+        // SAFETY: Should never happen in actuality, but as long as we're here...
+        if unsafe { pg_sys::CurrentMemoryContext.is_null() } {
+            Nullable::Null
+        } else {
+            Nullable::Valid(Self::unbox_arg_unchecked(_arg))
+        }
+    }
+
+    fn is_virtual_arg() -> bool {
```
eeeebbbbrrrr, line 148: What is a "virtual arg"?  I guess it's one we inject and not one the user explicitly listed in the `#[pg_extern]` function?

pgrx/src/datum/borrow.rs
```diff
@@ -116,7 +116,7 @@ macro_rules! impl_borrow_fixed_len {
 impl_borrow_fixed_len! {
     i8, i16, i32, i64, bool, f32, f64,
     pg_sys::Oid, pg_sys::Point,
-    Date, Time, Timestamp, TimestampWithTimeZone
+    Date, Time, Timestamp, TimestampWithTimeZone, TimeWithTimeZone
```
eeeebbbbrrrr: Looks like an accident.

pgrx/src/palloc/pbox.rs
eeeebbbbrrrr, line 20: Any benefit to this being `#[repr(transparent)]`?

eeeebbbbrrrr (approved):
I don't like the name `PBox` but I don't have any better ideas that aren't annoyingly long.  So I'm not going to let that hold up forward progress here.  feel free to merge!  Thanks!

pgrx/src/palloc/pbox.rs
workingjubilee: Oh, true.

pgrx/src/memcx.rs
workingjubilee: Yes, to be clear it's one that doesn't appear in the generated SQL: it's real to Rust, "fake" to SQL, and not a Datum.

pgrx/src/palloc/pbox.rs
workingjubilee: Yes, actually!
```suggestion
#[repr(transparent)]
pub struct PBox<'mcx, T: ?Sized> {
```

pgrx/src/datum/borrow.rs
workingjubilee: Oddly, isn't! Though I may pull it out so it is clearer as a separate change.

pgrx/src/palloc/pbox.rs
workingjubilee: I think this is correctest:
```suggestion
        const { assert!(align_of::<T>() <= size_of::<pg_sys::Datum>()) };
```

pgrx/src/memcx.rs
workingjubilee:
```suggestion
    unsafe fn unbox_nullable_arg(arg: Arg<'_, 'fcx>) -> Nullable<Self> {
```

pgrx/src/memcx.rs
```diff
@@ -116,3 +125,38 @@ mod nightly {
...
+        // as the expanded code is designed so that `fcinfo` and each `arg` are truly "borrowed" in rustc's eyes.
+        unsafe { &*((&raw mut pg_sys::CurrentMemoryContext).cast()) }
+    }
+
+    unsafe fn unbox_nullable_arg(_arg: Arg<'_, 'fcx>) -> Nullable<Self> {
+        // SAFETY: Should never happen in actuality, but as long as we're here...
+        if unsafe { pg_sys::CurrentMemoryContext.is_null() } {
+            Nullable::Null
+        } else {
+            Nullable::Valid(Self::unbox_arg_unchecked(_arg))
```
workingjubilee:
```suggestion
            Nullable::Valid(Self::unbox_arg_unchecked(arg))
```

workingjubilee:
The macOS failures are utterly mysterious.

PR #2212, opened 2025-11-22
Recurse through `T` while anonymizing `&T`

workingjubilee:


pgrx-sql-entity-graph/src/lifetimes.rs
```diff
@@ -51,6 +51,8 @@ pub fn anonymize_lifetimes(value: &mut syn::Type) {
             if let Some(lifetime) = type_ref.lifetime.as_mut() {
                 lifetime.ident = syn::Ident::new("_", lifetime.ident.span());
             }
+            // recurse
```
workingjubilee:
```suggestion
            // &'a T could be &'a &'b U so recurse to get &&U
```

PR #2213, opened 2025-11-22
`impl BorrowDatum for TimeTz`

workingjubilee:
It was overlooked.


pgrx/src/datum/borrow.rs
```diff
@@ -116,7 +116,7 @@ macro_rules! impl_borrow_fixed_len {
 impl_borrow_fixed_len! {
     i8, i16, i32, i64, bool, f32, f64,
     pg_sys::Oid, pg_sys::Point,
-    Date, Time, Timestamp, TimestampWithTimeZone
+    Date, Time, Timestamp, TimestampWithTimeZone, TimeWithTimeZone
```
workingjubilee: Oh, I guess it's the wrong way around. Hm.
```suggestion
    Date, Time, TimeWithTimeZone, Timestamp, TimestampWithTimeZone
```

PR #2214, opened 2025-11-22
Change `MemCx::alloc_bytes` to return `NonNull`

workingjubilee:
Postgres will throw if it gets NULL here, so roll with that.

### DIFF
--- a/pgrx/src/memcx.rs
+++ b/pgrx/src/memcx.rs
@@ -29,8 +29,10 @@ impl<'mcx> MemCx<'mcx> {
 
     /// Allocate a raw byte buffer `size` bytes in length
     /// and returns a pointer to the new allocation.
-    pub fn alloc_bytes(&self, size: usize) -> *mut u8 {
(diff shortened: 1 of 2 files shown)


pgrx/src/memcx.rs
```diff
@@ -29,8 +29,10 @@ impl<'mcx> MemCx<'mcx> {
 
     /// Allocate a raw byte buffer `size` bytes in length
     /// and returns a pointer to the new allocation.
-    pub fn alloc_bytes(&self, size: usize) -> *mut u8 {
-        unsafe { pg_sys::MemoryContextAlloc(self.ptr.as_ptr(), size).cast() }
+    pub fn alloc_bytes(&self, size: usize) -> NonNull<u8> {
+        let ptr = unsafe { pg_sys::MemoryContextAlloc(self.ptr.as_ptr(), size).cast() };
+        // SAFETY: `pg_sys::MemoryContextAlloc` is not allowed to return NULL, it must error.
```
workingjubilee, line 34:
```suggestion
        // SAFETY: `pg_sys::MemoryContextAlloc` is not allowed to return NULL, it errors instead.
```

PR #2221, opened 2025-11-28
Add `commands/publicationcmds.h` to generated `pg_sys` bindings

Sinjo:
I was looking to use `pg_sys::AlterPublicationOwner` in a utility extension at PlanetScale and noticed it wasn't included in the generated bindings.

The functions in `publicationcmds.h` seem similarly useful to other command headers like `event_trigger.h`, so I figure including them won't be too controversial.


workingjubilee:
Sure!

Most bindings aren't controversial, just "does someone actually need this?"

Ideally one day we move to a model of bindings where we don't have them all piled in the same namespace and can select between them, but, well,

PR #2226, opened 2025-12-02
Allow explicit handling of allocation errors

### LINKED ISSUE #2225: Should `ArrayAllocError` include an `Oom` variant?
There's a few reasons that this might make sense to do, including
- People can make arbitrarily limited memory contexts and may want to deliberately use that to constrain a set of array allocations without immediately failing the current transaction if those don't work out.
- It may simply provide perf benefits due to being able to bypass some error-handling overhead.

If the answer is "Yes", then `new_empty` should be changed to return `Result` (because even such a tiny allocation can error) and the allocation error risk should also inform the relevant `MemCx` APIs

workingjubilee:
Right now, we have several fallible constructors for Postgres types. When this is an allocation error, we don't allow handling it. As I expect there to be more such fallible constructors in the future, and transient memory contexts can be arbitrarily limited in size, I think we should expose this problem in otherwise-fallible code. We can retain some infallible constructors elsewhere, though. If nothing else, this can allow better diagnostics via `track_caller`.

Resolves #2225

### DIFF
--- a/pgrx/src/array/flat_array.rs
+++ b/pgrx/src/array/flat_array.rs
@@ -1,7 +1,7 @@
 #![deny(missing_docs)]
 use crate::datum::{Array, BorrowDatum, Datum};
 use crate::layout::{Align, Layout};
-use crate::memcx::MemCx;
+use crate::memcx::{MemCx, OutOfMemory};
 use crate::nullable::Nullable;
 use crate::palloc::PBox;
 use crate::pgrx_sql_entity_graph::metadata::{
@@ -77,6 +77,8 @@ pub enum ArrayAllocError {
     TooManyElems,
     /// One or more dimensions are zero
     ZeroLenDim,
+    /// There is insufficient memory in this context
+    OutOfMemory,
 }
 
 const MAX_ALLOC_SIZE: usize = 0x3fffffff;
@@ -104,7 +106,7 @@ where
 
         let ndims = N;
         if N == 0 {
-            return Ok(FlatArray::new_empty(memcx));
+            return FlatArray::new_empty(memcx);
         }
         const { assert!(N <= MAX_DIMS) };
 
@@ -151,7 +153,7 @@ where
         let elemtype = <T as Scalar>::OID;
         let tail_size = size - base_size;
 
-        let ptr = alloc_zeroed_head(memcx, tail_size, ndims as i32, dataoffset, elemtype);
+        let ptr = alloc_zeroed_head(memcx, tail_size, ndims as i32, dataoffset, elemtype)?;
         // SAFETY: we allocated enough for our dimensions and lbounds
         unsafe {
             ptr.byte_add(base_size).cast().write(dim_ints);
@@ -164,11 +166,13 @@ where
     }
 
     /// Allocate a 0-dimension array
-    pub fn new_empty<'cx>(memcx: &MemCx<'cx>) -> PBox<'cx, FlatArray<'cx, T>> {
+    pub fn new_empty<'cx>(
+        memcx: &MemCx<'cx>,
+    ) -> Result<PBox<'cx, FlatArray<'cx, T>>, ArrayAllocError> {
         let nbytes = mem::size_of::<pg_sys::ArrayType>();
-        let ptr = alloc_zeroed_head(memcx, 0, 0, 0, <T as Scalar>::OID);
+        let ptr = alloc_zeroed_head(memcx, 0, 0, 0, <T as Scalar>::OID)?;
         // SAFETY: it's valid, if 0-dimensional
-        unsafe { PBox::from_raw_in(FlatArray::cast_tailed(ptr), memcx) }
+        Ok(unsafe { PBox::from_raw_in(FlatArray::cast_tailed(ptr), memcx) })
     }
 
     /// Allocate an array sized to fit a slice and copy it
@@ -179,7 +183,7 @@ where
         memcx: &MemCx<'cx>,
     ) -> Result<PBox<'cx, FlatArray<'cx, T>>, ArrayAllocError> {
         match data.len() {
-            0 => Ok(FlatArray::new_empty(memcx)),
+            0 => FlatArray::new_empty(memcx),
             len => {
                 let mut array = FlatArray::new_zeroed_in([len], false, memcx)?;
                 array.as_non_null_slice_mut().unwrap().copy_from_slice(data);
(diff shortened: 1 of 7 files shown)


workingjubilee:
Huh. I expected it to be more intrusive than this, tbh. Feels hard to argue against.

pgrx/src/memcx.rs
```diff
@@ -36,18 +36,18 @@ impl<'mcx> MemCx<'mcx> {
 
     /// Allocate a raw byte buffer `size` bytes in length
     /// and returns a pointer to the new allocation.
-    pub fn alloc_bytes(&self, size: usize) -> NonNull<u8> {
-        let ptr = unsafe { pg_sys::MemoryContextAlloc(self.ptr.as_ptr(), size).cast() };
-        // SAFETY: `pg_sys::MemoryContextAlloc` is not allowed to return NULL, it must error.
-        unsafe { NonNull::new_unchecked(ptr) }
+    pub fn alloc_bytes(&self, size: usize) -> Result<NonNull<u8>, OutOfMemory> {
```
workingjubilee, line 39: I just changed the signature on this one because breaking users of this "direct allocation" API is something I already did once this cycle: it's now `NonNull<u8>`, not `*mut u8`. Also, they probably want to be able to make decisions about error-handling-or-not themselves since they're allocating raw buffers!

pgrx/src/palloc/pbox.rs
```diff
@@ -33,13 +33,18 @@ impl<'mcx, T: ?Sized> PBox<'mcx, T> {
 }
 
 impl<'mcx, T: Sized> PBox<'mcx, T> {
+    #[track_caller]
     pub fn new_in(val: T, memcx: &MemCx<'mcx>) -> Self {
+        PBox::try_new_in(val, memcx).unwrap()
```
workingjubilee, line 38: With `track_caller`, this way if it panics we point to the place someone called `PBox::new_in`.

pgrx/src/array/flat_array.rs
```diff
@@ -104,7 +106,7 @@ where
 
         let ndims = N;
         if N == 0 {
-            return Ok(FlatArray::new_empty(memcx));
+            return FlatArray::new_empty(memcx);
```
workingjubilee, line 109: Making code like this simpler, even if it's subtle right now, seems like it will be an increasingly bigger win down the line.

pgrx/src/array/flat_array.rs
workingjubilee, line 71: I think this actually provides the full list of errors we can expect for this type, even in most future APIs.
```suggestion
/// Errors occurring when constructing fresh arrays
```

workingjubilee: There's one more possible variant, maybe: if we add ways to use a type without a guaranteed-to-be-a-valid-element `pg_sys::Oid`. But we can apply that as a restriction to implementations of `T: Element`, so I don't think so?

pgrx/src/memcx.rs
```diff
@@ -68,6 +68,16 @@ impl<'mcx> MemCx<'mcx> {
...
+#[derive(Debug)]
+pub struct OutOfMemory {
+    _reserve: (),
+}
+impl OutOfMemory {
+    pub fn new() -> OutOfMemory {
+        OutOfMemory { _reserve: () }
+    }
+}
+
```
eeeebbbbrrrr, line 80: Is it worth it to have this be an actual Error and store the size it tried to allocate that it couldn't, and have that reported in its Display impl?

pgrx/src/memcx.rs
```diff
@@ -36,18 +36,18 @@ impl<'mcx> MemCx<'mcx> {
 
     /// Allocate a raw byte buffer `size` bytes in length
     /// and returns a pointer to the new allocation.
-    pub fn alloc_bytes(&self, size: usize) -> NonNull<u8> {
-        let ptr = unsafe { pg_sys::MemoryContextAlloc(self.ptr.as_ptr(), size).cast() };
-        // SAFETY: `pg_sys::MemoryContextAlloc` is not allowed to return NULL, it must error.
-        unsafe { NonNull::new_unchecked(ptr) }
+    pub fn alloc_bytes(&self, size: usize) -> Result<NonNull<u8>, OutOfMemory> {
+        let flags = (pg_sys::MCXT_ALLOC_NO_OOM) as i32;
+        let ptr = unsafe { pg_sys::MemoryContextAllocExtended(self.ptr.as_ptr(), size, flags) };
```
eeeebbbbrrrr, line 41: Should we preemptively check `AllocSizeIsValid` and/or `AllocHugeSizeIsValid`.  If we don't, then `MemoryContextAllocExtended` can raise an ERROR.

And how do we expose MCXT_ALLOC_HUGE?

pgrx/src/memcx.rs
```diff
@@ -36,18 +36,18 @@ impl<'mcx> MemCx<'mcx> {
...
 
     /// Allocate a raw byte buffer `size` bytes in length
     /// and returns a pointer to the new allocation.
-    pub fn alloc_zeroed_bytes(&self, size: usize) -> NonNull<u8> {
-        let ptr = unsafe { pg_sys::MemoryContextAllocZero(self.ptr.as_ptr(), size).cast() };
-        // SAFETY: `pg_sys::MemoryContextAlloc` is not allowed to return NULL, it must error.
-        unsafe { NonNull::new_unchecked(ptr) }
+    pub fn alloc_zeroed_bytes(&self, size: usize) -> Result<NonNull<u8>, OutOfMemory> {
+        let flags = (pg_sys::MCXT_ALLOC_NO_OOM | pg_sys::MCXT_ALLOC_ZERO) as i32;
+        let ptr = unsafe { pg_sys::MemoryContextAllocExtended(self.ptr.as_ptr(), size, flags) };
```
eeeebbbbrrrr, line 49: ditto

pgrx/src/list/flat_list.rs
```diff
@@ -112,7 +112,8 @@ impl<'cx, T: Enlist> List<'cx, T> {
                 // No silly reasoning, simply allocate ~2 cache lines for a list
                 let list_size = 128;
                 unsafe {
-                    let list: *mut pg_sys::List = mcx.alloc_bytes(list_size).cast().as_ptr();
+                    let list: *mut pg_sys::List =
+                        mcx.alloc_bytes(list_size).unwrap().cast().as_ptr();
```
eeeebbbbrrrr, line 116: do we want to blindly `.unwrap()` here?  Having the function return `Result<&mut ListHead<'cx, T>, OutOfMemory>` would be a little awkward but it would also indicate to the caller that it's possible it will allocate.

workingjubilee: I need to rethink how this function works given I explicitly marked it as "unstable" as I didn't have much allocation infrastructure going at the time.

pgrx/src/memcx.rs
workingjubilee: Some types will never want to use `MCXT_ALLOC_HUGE`, so I think we should handle it as its own `alloc_extended` function and have a newtyped input for the flags.

workingjubilee: We could also skip past `Extended` and expose `MemoryContextAllocAligned` as the "most complicated alloc" function, but have alignments over the limit return a (richer) Err type... so on pg16 that lets you go up to the `2.pow(26)` limit, and on pg15 and under that's 8 (well, `align_of::<ffi::c_double>()` technically, I believe).

pgrx/src/memcx.rs
workingjubilee: definitely a possibility. I haven't added it yet because I figure the caller usually has that information, but maybe for bubbling up purposes?

essence-groove:
> Huh. I expected it to be more intrusive than this, tbh. Feels hard to argue against.

Yeah, hardly touching explicit. Not exactly graceful either. 👍

PR #2231, opened 2025-12-15
Add the ClientAuthentication_hook

daamien:


daamien:
I am still unsure that `libpq.rs` is the right place for this, let me know if it should be elsewhere.

pgrx-pg-sys/src/libpq.rs
```diff
@@ -203,3 +203,9 @@ pub mod be {
         raw_buf_remaining: isize,
     }
 }
+
+/// The ClientAuthentication Hook cannot be simply implemented with bindgen
+pub type ClientAuthentication_hook_type =
+    Option<unsafe extern "C-unwind" fn(port: *mut be::Port, status: u32)>;
+
+pub static mut ClientAuthentication_hook: ClientAuthentication_hook_type = None;
```
workingjubilee: This needs to be, itself, wrapped in `extern "C"`.

pgrx-pg-sys/src/libpq.rs
```diff
@@ -203,3 +203,12 @@ pub mod be {
         raw_buf_remaining: isize,
     }
 }
+
+/// The ClientAuthentication Hook cannot be implemented simply with bindgen
```
eeeebbbbrrrr: This is a nit, but this isn’t useful as a doccomment. 

Maybe add a short doccomment about what this type is for, and change this into a regular comment.

eeeebbbbrrrr (changes requested):

pgrx-pg-sys/src/libpq.rs
daamien: done and squashed the commits into 1

eeeebbbbrrrr (approved):

PR #2236, opened 2025-12-30
Place lbound before updating array len product

workingjubilee:
Without this, Postgres decides the array starts at 5 or whatever.


workingjubilee:
Caught thanks to @mhov taking a look!

workingjubilee:
Test failure seems to be purely a timeout issue.

PR #2238, opened 2026-01-05
fix: use same profile for embed binary to avoid double compilation

mildbyte:
`second_build` and `compute_sql` always use the dev/debug profile for the pgrx_embed binary, regardless of the profile used for the main extension build.

This causes cargo to recompile the entire dependency tree in debug mode even when the extension had already been compiled in release mode.

When compiling an extension in release mode, this results in:
- A release compilation
- Another debug compilation
- Debug artifacts often larger than release artifacts
- CI cache bloat (sccache/GHA cache has to store these debug artifacts)

Now the embed binary uses the same profile as the library build, allowing cargo to reuse artifacts from the first build. The second compilation now only links the small pgrx_embed binary itself.

This shouldn't change the dev experience, where both the extension and the embed binary will compile in the debug mode.

### DIFF
--- a/cargo-pgrx/src/command/schema.rs
+++ b/cargo-pgrx/src/command/schema.rs
@@ -207,9 +207,9 @@ pub(crate) fn generate_schema_implicit(
         tracing::info!(dot = %dot_path.display(), "Writing Graphviz DOT");
     }
 
-    second_build(cargo, &features_arg, embed.path(), &manifest)?;
+    second_build(cargo, &features_arg, embed.path(), &manifest, profile)?;
 
-    compute_sql(&manifest)?;
+    compute_sql(&manifest, profile)?;
 
     Ok(())
 }
@@ -452,6 +452,7 @@ fn second_build(
     features_arg: &str,
     embed_path: &Path,
     manifest: &Manifest,
+    profile: &CargoProfile,
 ) -> eyre::Result<()> {
     // We do pass cfg to the binary and do not pass cfg to dependencies to avoid recompilation
     // The only cargo command respecting our need is `cargo rustc`
@@ -459,6 +460,8 @@ fn second_build(
         .subcommand("rustc")
         .flag_args("--bin", vec![pgrx_embed_name(manifest)?])
         .into_command();
+
+    command.args(profile.cargo_args());
 
     command.arg("--");
 
@@ -468,7 +471,7 @@ fn second_build(
 
     let command_str = format!("{command:?}");
     eprintln!(
-        "{} {}, in debug mode, for SQL generation with features {}",
+        "{} {} for SQL generation with features {}",
         "  Rebuilding".bold().green(),
         "pgrx_embed".cyan(),
         features_arg.cyan(),
@@ -487,9 +490,9 @@ fn second_build(
     Ok(())
 }
 
-fn compute_sql(manifest: &Manifest) -> eyre::Result<()> {
+fn compute_sql(manifest: &Manifest, profile: &CargoProfile) -> eyre::Result<()> {
     let mut bin = get_target_dir()?;
-    bin.push("debug"); // pgrx_embed_ is always compiled in debug mode
+    bin.push(profile.target_subdir());
     bin.push(pgrx_embed_name(manifest)?);
 
     let mut command = std::process::Command::new(bin);


eeeebbbbrrrr:
There’s a reason pgrx does what it does, but I cannot remember why. 

Let me dig through git tomorrow and refresh my memory. 

That said, I’m not seeing a reason why we shouldn’t merge this or something similar.

workingjubilee:
cc @usamoi

usamoi:
> There’s a reason pgrx does what it does, but I cannot remember why.

We did this because if fat LTO is enabled for profile `release`, compiling `pgrx_embed` is very slow.

For this issue, I have a solution: passing `-Clto=off` when invoking `cargo rustc`. I have been using this approach consistently in both CI and local development, but I am not confident that it can guarantee reliability in all cases.

eeeebbbbrrrr:
This is a bit of a tangent, but I really want some method to generate the schema that doesn't require building another binary.  I have some vague ideas about how to do that but haven't sat down to explore them.

That said, I think merging this PR is probably okay.  I think folks that build a pgrx extension in release mode already expect a longer compilation time.

eeeebbbbrrrr:
> That said, I think merging this PR is probably okay. I think folks that build a pgrx extension in release mode already expect a longer compilation time.

Then again, this PR essentially reverts #1838, which was done in response to issue #1835.

Now I'm not so sure.

workingjubilee:
> For this issue, I have a solution: passing -Clto=off when invoking cargo rustc. I have been using this approach consistently in both CI and local development, but I am not confident that it can guarantee reliability in all cases.

I think we should try the ol' "throw it over and see if anyone screams" shipping strat, here.

workingjubilee:
That said,

@mildbyte Can you actually like... cite numbers, here? Also, why do you assume people don't locally develop using `--release`, exactly? It is very often that people build and test an optimized version also if they are doing anything performance-sensitive, but may omit the more intense optimizations that their final release artifact is subjected to in the interest of timeliness.

mildbyte:
> @mildbyte Can you actually like... cite numbers, here? Also, why do you assume people don't locally develop using `--release`, exactly? It is very often that people build and test an optimized version also if they are doing anything performance-sensitive, but may omit the more intense optimizations that their final release artifact is subjected to in the interest of timeliness.

Yep, I'll get some clean / incremental build timings when I have the chance - thanks for looking at the PR!

On dev vs release - I was just thinking about the original reasoning in https://github.com/pgcentralfoundation/pgrx/issues/1835 when saying that. So, with this PR someone relying on the current behaviour of using the dev profile for development will still avoid the double compilation, and someone using --release will now also avoid it.

For context, our use case is a bit gnarly: we're working on an extension for querying lakehouse data in object storage from within PG, so we pull in a lot of heavy crates like DataFusion, object store libraries, Iceberg/Delta etc (technically, there's a background worker that we compile as a separate binary that handles the actual queries, but we still rely on those crates in the core extension for loading table metadata inside of PG). And, we have to support it for a large matrix of operating systems / PG versions / our custom PG fork / x86/arm architectures.

Most of our integration tests and even development is in --release for exactly the reason you mentioned, there are some tests that I don't think work at all in --dev mode due to stack optimizations interacting with DataFusion's planner and recursion depth.

So we end up doing a lot of non-incremental builds in CI. We use sccache with the GitHub Actions Cache to alleviate the pain a bit, but sccache effectively ends up caching the compilation artifacts twice - the release-profile version and the dev-profile version due to the pgrx_embed build (which runs as part of the full `cargo pgrx package` invocation). The dev artifacts are bigger by size and the GHA cache is limited to 10GB, so I conjecture there's a lot of cache pressure/evictions from storing those. I guess we could do a cached `cargo build --release`, turn off the cache and do `cargo pgrx package` as the last mile build.

> This is a bit of a tangent, but I really want some method to generate the schema that doesn't require building another binary. I have some vague ideas about how to do that but haven't sat down to explore them.

I was thinking about it too as an alternative for this pgrx change. My only idea so far is using the the pgrx_embed cfg feature flag that the schema build injects:

- we'd have a thin schema-only crate as our entrypoint that just has the PG function definitions
- there's an implementation crate that has all the dependencies
- the schema-only crate doesn't depend on the implementation crate when we're compiling with the embed feature
- the schema-only crate has some cfg! flags to disable all imports and to turn the function bodies into a noop when we're compiling with the embed feature

We're still building a binary but it might be more efficient.

eeeebbbbrrrr:
This is no longer relevant b/c we got rid of the pgrx_embed binary entirely!

Issue #2228, opened 2025-12-08
Allow a way to change the "main thread" or disable the threading errors


JelteF:
Hi again, I'm working on [pg_duckdb](https://github.com/duckdb/pg_duckdb) these days. In that project I'm actually calling into Postgres FFI from different threads. This is safe because I manage a global lock there. I'm running into the problem though, that when calling SPI, rust extensions on the same system will abort the whole Postgres process tree [due to this check](https://github.com/pgcentralfoundation/pgrx/blob/0d8e30223f434f0942cdcd394bba7854fa49d2ff/pgrx-pg-sys/src/submodules/thread_check.rs#L29-L41).

So I'm wondering if there's a way in which pg_duckdb could either tell pgrx extensions what the "current" thread id is. Or a way to temporarily disable that check. Because in this environment the FFI calling behavior is actually safe, but pgrx extensions are a so conservative that it still doesn't work.

eeeebbbbrrrr:
I suppose we could add a feature flag to wholesale turn this check off at compile time.  I'm not sure we can easily provide a way to lie about what the "main" thread is.  

I wonder if you could instead invert your process and use a channel to communicate messages from your thread(s) back to the main thread so it can do the work.  I think that's a pretty common approach to ensure a specific thread is doing certain work.

piki:
Hello from PlanetScale.  I'm the author/maintainer of `pg_insights`, the pgrx-based extension that provides [statement-level visibility](https://planetscale.com/docs/postgres/monitoring/query-insights) for PlanetScale dashboards.  I'm at least part of the reason @JelteF opened this issue, as we were working to get `pg_duckdb` working on PlanetScale.

I don't think I would use a flag to turn off `check_active_thread` entirely.  That assertion protects too much, and without it, every multithreaded behavior in our extension (we use librdkafka, which is multithreaded) or anyone else's becomes UB and risks much worse outcomes than a server crash.  `pg_duckdb` is careful about locking around their FFI calls, but I can't assume every other extension will be.

What I ended up building instead is a `would_fail_thread_check` call, [merged](https://github.com/planetscale/pgrx/pull/1) into our public pgrx fork.  That allows code in hooks like
```rust
fn raw_XXX_hook(...) {
  if !pg_sys::thread_check::would_fail_thread_check() {
    handle_XXX_hook(...);  // our handling with various FFI calls
  }
  prev_hook(...);
}
```

[...]

PR #2242, opened 2026-02-03
introduce `#[pg_guard(unsafe_entry_thread)]`

eeeebbbbrrrr:
initial commit of rejiggering how pgrx tracks the "ACTIVE_THREAD".   This is to address issue #2228.

Functions annotated with `#[pg_guard(unsafe_entry_thread)]` do some trickery to reset pgrx' understanding of the "ACTIVE_THREAD" so that whatever thread happens to be calling that function then becomes our understanding of the "ACTIVE_THREAD", thereby allowing pgrx to do Postgres FFI on a thread that isn't the main thread.

The situation behind #2228 is that pg_duckdb, an extension written in C++, uses threads (presumably "safely") within Postgres and if one of those threads does something like SPI, which then causes planner/executor hooks to be called, and one of those hooks is implemented with pgrx, this new `unsafe_entry_thread` tag will let that callback run.  It's basically making an assumption that the caller is properly managing the thread it's on.

There's very little code changes here -- the bulk of the PR is a new example crate/extension with its own little c-shim so that I could manually test Postgres itself kicking off a thread that does SPI that ends up calling into a planner hook setup by the rust/pgrx extension.

I have a need to push this through for $DayJob, so I'm probably going to do that later this week.

### DIFF
--- a/pgrx-pg-sys/src/submodules/thread_check.rs
+++ b/pgrx-pg-sys/src/submodules/thread_check.rs
@@ -22,7 +22,7 @@
 //! future...
 
 use std::num::NonZeroUsize;
-use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 
 static ACTIVE_THREAD: AtomicUsize = AtomicUsize::new(0);
 #[track_caller]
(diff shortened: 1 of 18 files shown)


pgrx-pg-sys/src/submodules/thread_check.rs
```diff
@@ -40,6 +40,23 @@ pub fn check_active_thread() {
     }
 }
 
+pub mod active_thread {
```
eeeebbbbrrrr, line 44:
```suggestion
#[doc(hidden)]
pub mod active_thread {
```

PR #2247, opened 2026-02-11
chore: minor lints

nyurik:
I ran this just to see if clippy would flag anything:

```
cargo clippy --all-targets --workspace --fix --allow-dirty --features pg18 -- -W clippy::uninlined_format_args
```

I have a suspicion that a few of these are incorrect - i.e. type casting is not always guaranteed - but if so, there should be an `#[allow(...)]` to highlight that.  I'll wait for the CI to pass first.

### DIFF
--- a/cargo-pgrx/src/cargo.rs
+++ b/cargo-pgrx/src/cargo.rs
@@ -83,11 +83,10 @@ impl Cargo {
         let mut cmd = cargo();
 
         // subcommand *must* go first
-        if self.subcmd != "" {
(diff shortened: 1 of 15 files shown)


eeeebbbbrrrr:
It's not that I don't appreciate the effort behind these "lint" PRs, but they cause a lot of churn in the codebase and are difficult for me to review in a timely manner and, as evidenced here, tend to bitrot quickly.

Please accept my sincere thanks for your efforts, but I need to close this without merging.

PR #2252, opened 2026-03-05
Feature Request: PGXN-ready Artifact Generation in cargo pgrx package

Nirajkashyap:
### Overview
This proposal adds support in **pgrx** to generate a PGXN-ready distributable artifact directly from the packaging command.

With a dedicated flag, `cargo pgrx package` will:

- Generate a valid `META.json` (PGXN metadata) from Cargo.toml file **OR** use existing META.json file present in root folder in PGRX extension.
- Bundle the extension into a `.zip` or `.tar.gz` archive
- Produce an artifact ready for upload to PGXN

---

## Motivation

Publishing a pgrx extension to PGXN currently requires several manual steps:

1. Creating a `meta.json`
2. Packaging files into the correct archive layout

Automating this workflow will:

- Reduce friction for extension authors
- Ensure consistent packaging
- Simplify release automation (CI/CD)
- Encourage more PGXN adoption

---

## Proposed Usage

```bash
cargo pgrx package --pgxn --format zip

### CODE REVIEW

**Comment by eeeebbbbrrrr** [comment]
A few notes here.  First, I would not want `cargo pgrx new` to create a license file.  That's for the user to do on their own -- it is not pgrx's place to even suggest a license to the user.

I also generally feel the same way about a readme file, but I'd be willing to let that happen so long as it's as minimal as possible.  As in only lines 1-3 of what's in this PR.

Also, I'd prefer the PGXN-specific code be moved into its own `pgxn.rs` file.

If you're up for doing these things I'll take a more in-depth look once you've updated the PR.  Thanks!

**Comment by Nirajkashyap** [comment]
HI @eeeebbbbrrrr,
updated Pull request as per feedback and happy to update it. please review PR.

- Introduce pgxn module for PGXN package generation
- Remove unused license template and update README template
- Refactor package.rs to utilize new pgxn module for PGXN package creation


happy to contribute in this project  under your guidance.

Please help me to test this PR. please forward any link or doc where i can test this.

**Comment by eeeebbbbrrrr** [comment]
I wonder if @theory has a minute to review this and offer suggestions.  He *is* PGXN.

File: cargo-pgrx/src/command/pgxn.rs [hunk 1]
```diff
@@ -0,0 +1,352 @@
...
+        version: package.version.as_ref().unwrap_or(&"0.0.0".to_string()).clone(),
+        maintainer: package
+            .authors
+            .clone()
+            .unwrap_or_else(|| vec!["Unknown <unknown@example.com>".to_string()]),
+        license: package.license.clone().unwrap_or_else(|| "postgresql".to_string()),
+        provides: None,
+        prereqs: None,
+        resources: None,
+        tags: package.keywords.clone(),
```
theory: Use PGXN tags first, fall back on keywords.

cargo-pgrx/src/command/pgxn.rs
```diff
@@ -0,0 +1,352 @@
...
+        generated_by: Some("cargo-pgrx".to_string()),
```
theory, line 168: Add version.

cargo-pgrx/src/command/pgxn.rs
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    // Add provides section
+    let mut provides = BTreeMap::new();
+    provides.insert(
+        extname.to_string(),
+        ProvideEntry {
+            abstract_: meta.abstract_.clone(),
+            file: Some(format!("share/postgresql/extension/{extname}-{}.sql", meta.version)),
+            docfile: Some("README.md".to_string()),
```
theory: Support other documentation file.

cargo-pgrx/src/command/pgxn.rs
```diff
@@ -0,0 +1,352 @@
...
+                r#type: Some("git".to_string()),
+            }),
+        });
+    }
+
+    // Add provides section
+    let mut provides = BTreeMap::new();
+    provides.insert(
+        extname.to_string(),
+        ProvideEntry {
```
theory, line 249: Does pgrx support multiple extensions? If so, then so should the `META.json`.

cargo-pgrx/src/command/pgxn.rs
```diff
@@ -0,0 +1,352 @@
...
+    meta.provides = Some(provides);
+
+    // Write META.json
+    let meta_json = serde_json::to_string_pretty(&meta)
+        .wrap_err("Failed to serialize META.json")?;
+    let meta_path = out_dir.join("META.json");
+    fs::write(&meta_path, &meta_json)
+        .wrap_err("Failed to write META.json")?;
+
+    // Copy README.md if it exists
```
theory: Also any other doc files.

theory (commented):
Looks like it will be handy for simple configurations. A few notes:

* I suggest adding unit tests to cover each execution branch
* PGXN does not require a README, but personally I welcome any and all encouragement to provide documentation
* Need to support other documentation files than README (which really should not be documentation)
* Can it handle the case where there are multiple extensions?
* Document fallback values (e.g., description if no abstract) in comments
* Maybe support longer description
* Add a comment to always use semantic versions, including for dependencies and like Postgres itself.
* Version may be semver prerelease

Nirajkashyap:
HI @theory ,
I have tried to address the comments. Please have look and provide feedback.

cargo-pgrx/src/command/pgxn.rs
Nirajkashyap: used fallback if pgxn meta tags are not there use package.keywords

cargo-pgrx/src/command/pgxn.rs
Nirajkashyap: version is added from package version

cargo-pgrx/src/command/pgxn.rs
Nirajkashyap: updated code to do fallback for one of files as docfile.

cargo-pgrx/src/command/pgxn.rs
Nirajkashyap: currently only single extension is supported.

cargo-pgrx/src/command/pgxn.rs
Nirajkashyap: updated code to do fallback for one of files as docfile.

eeeebbbbrrrr:
~~I see this teaches `cargo pgrx new` to generate a file named `readme.md`.  This is just my preference but I prefer `README.md`.~~

EDIT:  Sorry, I was mistaken.  This PR does indeed create a file named `README.md`.  A++, that's fine.  Still curious if that's okay for PGXN...

My question is probably for @theory, but does PGXN dictate a spelling for the "readme" file?  I noticed that you wrote `README` in a comment above.

eeeebbbbrrrr:
@Nirajkashyap regarding licenses, I see that you've adjusted the `Cargo.toml` template to use the "postgres" license and I also see that the code will fallback to "postgres".

Both of these are strictly incorrect.  As I said before, it's not pgrx's place to suggest a license.  If we assign a license up-front, the user may not realize that and initially publish their code under a license they did not intend to use.  And that _could_ have bad outcomes for that user. 

So I would prefer that we do not default any license at all, anywhere.  Not in `Cargo.toml` and not in the code.  If PGXN requires a license then we should panic when the user tries to build the PGXN package, informing the user that they need to pick a license.

cargo-pgrx/src/command/pgxn.rs
```diff
@@ -0,0 +1,731 @@
...
+pub struct MetaSpec {
+    pub version: String,
+    pub url: String,
+}
+
+/// Find the best documentation file in the given directory
+/// PGXN does not require a README, but personally I welcome any and all encouragement to provide documentation
+fn find_documentation_file(manifest_dir: &PathBuf) -> Option<String> {
+    // Check for common documentation files in order of preference
+    let doc_files = [
```
eeeebbbbrrrr, line 104: this can be a `const`.

cargo-pgrx/src/command/pgxn.rs
```diff
@@ -0,0 +1,731 @@
...
+            .to_string(),
+        // Maybe support longer description - currently same as abstract, but could be extracted from README
+        description: package.description.clone().unwrap_or_default(),
+        // Always use semantic versions, including for dependencies and Postgres itself.
+        // Version may be semver prerelease (e.g., 1.0.0-alpha.1)
+        version: package.version.as_ref().unwrap_or(&"0.0.0".to_string()).clone(),
+        maintainer: package
+            .authors
+            .clone()
+            .unwrap_or_else(|| vec!["Unknown <unknown@example.com>".to_string()]),
```
eeeebbbbrrrr, line 161: if a maintainer/author is required by PGXN then we should panic if there isn't one, we should not provide a default.

cargo-pgrx/src/command/pgxn.rs
```diff
@@ -0,0 +1,731 @@
...
+        // Fallback to "postgresql" license if none specified
+        license: package.license.clone().unwrap_or_else(|| "postgresql".to_string()),
```
eeeebbbbrrrr, line 163: As I noted in a comment, panic instead of defaulting to a license.

cargo-pgrx/src/command/pgxn.rs
```diff
@@ -0,0 +1,731 @@
...
+            .and_then(|m| m.get("pgxn"))
+            .and_then(|pgxn| pgxn.get("abstract"))
+            .and_then(|v| v.as_str())
+            .unwrap_or(&package.description)
+            .to_string(),
+        // Maybe support longer description - currently same as abstract, but could be extracted from README
+        description: package.description.clone().unwrap_or_default(),
+        // Always use semantic versions, including for dependencies and Postgres itself.
+        // Version may be semver prerelease (e.g., 1.0.0-alpha.1)
+        version: package.version.as_ref().unwrap_or(&"0.0.0".to_string()).clone(),
```
eeeebbbbrrrr, line 157: if a version is required we should panic if we can't figure it out.  defaulting to `0.0.0` isn't ideal.

cargo-pgrx/src/command/pgxn.rs
```diff
@@ -0,0 +1,731 @@
...
+        let doc_src = manifest_dir.join(docfile_name);
+        if doc_src.exists() {
+            let doc_dst = out_dir.join(docfile_name);
+            fs::copy(&doc_src, &doc_dst)
+                .wrap_err(format!("Failed to copy {}", docfile_name))?;
+        }
+    }
+
+    // Copy LICENSE if it exists
+    let license_src = manifest_dir.join("LICENSE");
```
eeeebbbbrrrr, line 276: can we not get the license filename from the `Cargo.toml` manifest itself?

cargo-pgrx/src/command/pgxn.rs
```diff
@@ -0,0 +1,731 @@
...
+    
+    provides.insert(
+        extname.to_string(),
+        ProvideEntry {
+            abstract_: meta.abstract_.clone(),
+            file: Some(format!("share/postgresql/extension/{extname}-{}.sql", meta.version)),
+            docfile,
+            version: meta.version.clone(),
+        },
+    );
```
eeeebbbbrrrr, line 255: pgrx also supports schema version upgrade scripts and this will need to include all of those too.  If you look at the code paths for `cargo pgrx install` there's a place where those are detected and enumerated for copying.  If you need to pull that code out to share it here, that's fine, but they need to be included too.

eeeebbbbrrrr (changes requested):

theory:
> My question is probably for @theory, but does PGXN dictate a spelling for the "readme" file? I noticed that you wrote `README` in a comment above.

Looks like it's [case-insensitive](https://github.com/pgxn/pgxn-api/blob/b6c62076dcae632b34829e1dc2aa9294b8037bf0/lib/PGXN/API/Indexer.pm#L566).

theory:
> So I would prefer that we do not default any license at all, anywhere. Not in `Cargo.toml` and not in the code. If PGXN requires a license then we should panic when the user tries to build the PGXN package, informing the user that they need to pick a license.

[License is indeed required](https://pgxn.org/spec/#license), although `postgres` isn't a valid value; `postgresql` is.

cargo-pgrx/src/command/pgxn.rs
theory: Yes, [required](https://pgxn.org/spec/#maintainer). The [required fields](https://pgxn.org/spec/#Required.Fields) are

* abstract
* maintainer
* license
* provides
* meta-spec
* name
* version

cargo-pgrx/src/command/pgxn.rs
theory: Might need a mapping to the [valid PGXN values](https://pgxn.org/spec/#license) (scroll a bit to the table).

Nirajkashyap:
Hi @eeeebbbbrrrr  
is it good if i discard this PR and raise two separate small Pull Request ?

1. First Pull Request 
 
```cargo pgrx  new```  with flag ```--pgxn``` Should do following 
 a. use sample META.json which is required by pgxn and populate below in cargo.toml file 
 
 ```
 # PGXN Metadata
# See https://pgxn.org/meta/spec.txt for complete specification
[package.metadata.pgxn]
```
OR
b. simply create META.json to keep cargo.toml  file clean 



2. Second Pull Request
```cargo pgrx package```  with flag ```--pgxn``` Should do following 
if Previously A approach is followed package should create META.json in output folder.
if Previously B approach is followed just copy META.json to output folder.

eeeebbbbrrrr:
> Hi @eeeebbbbrrrr is it good if i discard this PR and raise two separate small Pull Request ?
> 
> 1. First Pull Request
> 
> `cargo pgrx new` with flag `--pgxn` Should do following a. use sample META.json which is required by pgxn and populate below in cargo.toml file
> 
> ```
> # PGXN Metadata
> # See https://pgxn.org/meta/spec.txt for complete specification
> [package.metadata.pgxn]
> ```
> 
> OR b. simply create META.json to keep cargo.toml file clean
> 
> 2. Second Pull Request
>    `cargo pgrx package`  with flag `--pgxn` Should do following
>    if Previously A approach is followed package should create META.json in output folder.
>    if Previously B approach is followed just copy META.json to output folder.

Hi @Nirajkashyap!  Please feel free to do what's easiest and correct for you.  this sounds like a fine idea!

PR #2253, opened 2026-03-05
Add mapping from PostgreSQL's CIRCLE type to Rust

blogh:
Hi, 

We are adding point anonymisation functions in [PostgreSQL anonymizer](https://gitlab.com/dalibo/postgresql_anonymizer/-/merge_requests/629) and I noticed that the PostgreSQL's `Circle` type is not mapped to the rust `CIRCLE` type  whereas `BOX` and `Point` are.  This patch attempts to add support for this.


eeeebbbbrrrr (approved):
Hey!  Thanks.  This looks good.  Soon as CI passes I'll merge

Issue #2257, opened 2026-03-14
`cargo pgrx install` discovers SQL entities but does not emit extension SQL file


rioriost:
I'm seeing what looks like a SQL emission / install-time projection problem in `cargo-pgrx`.

## Environment

- macOS Apple Silicon
- PostgreSQL 17
- pgrx tested with:
  - `0.16.1`
  - `0.17.0`
- cargo-pgrx version matched to crate version in each test
- clean local repro on the same machine for both versions

## Summary

`cargo pgrx install` reports:

- `Discovered 4 SQL entities`

It successfully:

- copies the control file
- copies the shared library
- rebuilds `pgrx_embed`

But it **does not generate the extension SQL file** for the extension being installed.

There is also no `Writing ... <extension>--<version>.sql` message in the install log.

## Minimal reproduction

I created a minimal pgrx extension crate with only four `#[pg_extern]` functions.

`src/lib.rs`:

```rust
use pgrx::prelude::*;

::pgrx::pg_module_magic!(name, version);

#[pg_extern]
fn demo_bool_arg(input: i32) -> bool {
    input > 100
}

#[pg_extern]
fn demo_text_noarg() -> &'static str {
    "ok"
}

#[pg_extern]
fn demo_i32_noarg() -> i32 {
    42
}

#[pg_extern]
fn demo_bool_noarg() -> bool {
    true
}

#[pg_guard]
pub extern "C-unwind" fn _PG_init() {}

#[pg_guard]
pub extern "C-unwind" fn _PG_fini() {}
```

Cargo.toml (example for 0.17.0; I reproduced similarly with 0.16.1):

```toml
[package]
name = "pgrx_sql_emit_min"
version = "0.1.0"
edition = "2021"

[lib]
crate-type = ["cdylib", "lib"]

[[bin]]
name = "pgrx_embed_pgrx_sql_emit_min"
path = "./src/bin/pgrx_embed.rs"

[features]
default = ["pg17"]
pg17 = ["pgrx/pg17", "pgrx-tests/pg17"]
pg18 = ["pgrx/pg18", "pgrx-tests/pg18"]
pg_test = []

[dependencies]
pgrx = "=0.17.0"

[dev-dependencies]
pgrx-tests = "=0.17.0"
```

Control file (`pgrx_sql_emit_min.control`):

```
comment = 'minimal pgrx SQL emission reproducer'
default_version = '0.1.0'
[...]

PR #2258, opened 2026-03-14
Fix schema SQL generation manifest resolution and improve output

rioriost:
For issue #2257

## Summary

Fix SQL generation during `cargo pgrx install` / `package` / `schema` by ensuring schema-time cargo invocations use the resolved package manifest instead of the user-supplied manifest-path context.

Also preserve `pgrx_embed` output during SQL generation so emission failures are visible.

## Root cause

`generate_schema_for_cli()` resolves the target crate as `package_manifest_path`, but previously constructed the follow-up cargo command using `user_manifest_path`.

That created a context mismatch:

- SQL entity discovery used the resolved package manifest and built extension library
- but the `pgrx_embed` rebuild/run phase could execute under a different cargo manifest context

This matches the observed behavior in #2257:

- `Discovered N SQL entities`
- `Rebuilding pgrx_embed`
- but no `Writing ... <ext>--<version>.sql`
- and no emitted extension SQL file

## Changes

- use `package_manifest_path` when constructing schema-generation cargo commands
- use `Command::status()` in `compute_sql()` so `pgrx_embed` stdout/stderr are not swallowed
- add a regression test asserting the schema cargo command uses the resolved package manifest path

## Testing

- added a regression test for schema manifest resolution
- checked file diagnostics for `cargo-pgrx/src/command/schema.rs`

### DIFF
--- a/cargo-pgrx/src/command/schema.rs
+++ b/cargo-pgrx/src/command/schema.rs
@@ -122,7 +122,7 @@ impl CommandExecute for Schema {
     features = ?features.features,
 ))]
 pub(crate) fn generate_schema_for_cli(
-    user_manifest_path: Option<&Path>,
+    _user_manifest_path: Option<&Path>,
     user_package: Option<&str>,
     package_manifest_path: &Path,
     profile: &CargoProfile,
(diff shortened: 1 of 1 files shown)


rioriost:
this patch could not solve the problem.

PR #2262, opened 2026-03-20
Capture Rust backtraces for Postgres ERRORs raised through pg_sys FFIcalls

eeeebbbbrrrr:
When Rust code calls a pg_sys function (e.g. pg_sys::relation_open()) and that function internally raises an ERROR via elog/ereport, the longjmp is caught by pg_guard_ffi_boundary and converted to a Rust panic. The panic hook captures a Rust backtrace at that point, but it was never attached to the error — the CaughtError::PostgresError path went through pg_re_throw() which bypassed do_ereport() entirely, so the backtrace was discarded.

Fix this by:

- In `downcast_panic_payload()`, when unwrapping a `CaughtError::PostgresError`, attach the backtrace from PANIC_LOCATION to the error report. This is the common path used by both `PgTryBuilder` and `run_guarded`, so the backtrace is available regardless of how the error is caught.

- In `run_guarded(),` route CaughtError::PostgresError through GuardAction::Report (and thus do_ereport) instead of GuardAction::ReThrow (and pg_re_throw). This ensures the backtrace appears in the ERROR's DETAIL line when the error propagates all the way out.

- In `pg_guard_ffi_boundary`, call `FlushErrorState(`) after` CopyErrorData` to properly clean the original error off Postgres' fixed-size errordata[] stack before do_ereport pushes a new entry.

- Remove the now-unused `GuardAction::ReThrow` variant and `pg_re_throw()` call.

### DIFF
--- a/pgrx-pg-sys/src/submodules/ffi.rs
+++ b/pgrx-pg-sys/src/submodules/ffi.rs
@@ -235,9 +235,12 @@ unsafe fn pg_guard_ffi_boundary_impl<T, F: FnOnce() -> T>(f: F) -> T {
             };
             let line = errdata.lineno as _;
 
-            // clean up after ourselves by freeing the result of [CopyErrorData] and restoring
-            // Postgres' understanding of where its next longjmp should go
+            // clean up after ourselves by freeing the result of [CopyErrorData], flushing the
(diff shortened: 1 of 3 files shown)


eeeebbbbrrrr:
(I'm not ashamed)

PR #2263, opened 2026-03-22
Add in-process pg_bench benchmarking for pgrx extensions

eeeebbbbrrrr:
Introduce a new `#[pg_bench]` benchmarking workflow for pgrx extensions, backed by a new optional pgrx-bench crate and a new cargo pgrx bench subcommand.

This adds Criterion-driven, in-process benchmarks that run inside a Postgres backend, using the same managed cluster lifecycle as cargo pgrx run rather than the ephemeral pg_test harness. Benchmarks are installed into a managed database named $extname_benches, and each cargo pgrx bench invocation is recorded as a persistent benchmark group with environment metadata, git metadata, pg_settings snapshots, and normalized benchmark results.

Highlights

- add #[pg_bench] proc macro in pgrx-macros
- require benchmarks to live in #[cfg(feature = "pg_bench")] #[pg_schema] mod benches
- add per-benchmark setup support via #[pg_bench(setup = some_path)]
- add transaction modes:
  - shared
  - subtransaction_per_batch
  - subtransaction_per_iteration
- add new pgrx-bench runtime crate exposing:
  - Bencher
  - BatchSize
  - black_box
- add cargo pgrx bench command
- build/install bench-enabled extensions through the normal install path, including control files and generated SQL scripts
- refresh the extension in $extname_benches with DROP EXTENSION / CREATE EXTENSION while preserving runner-owned benchmark history
- store benchmark history in persistent pgrx_bench schema tables and views
- stream per-benchmark results as each benchmark finishes
- emit Criterion-like CLI output for time/change/slope/mean/median
- persist raw Criterion artifacts and replay them as baselines for later comparisons
- switch comparison reporting to Criterion-driven change data instead of host-side approximations
- warn, but do not fail, when pg_bench is explicitly enabled for cargo pgrx run/install/package

Result storage

Benchmark history is stored outside extension ownership in the pgrx_bench schema so it survives extension refreshes. Each run group captures:

- benchmark group name and comparison group
- extension/package/profile metadata
- full command line
- host and toolchain metadata
- git metadata
- pg_settings snapshot
- benchmark case/run/estimate/sample/throughput rows
- raw Criterion JSON artifacts for later comparison replay

Comparison behavior

Comparisons now use persisted Criterion artifacts from a prior benchmark group. The backend reconstructs Criterion’s base/ layout, reruns the benchmark, reads Criterion’s change/estimates.json, and mirrors Criterion 0.5.1’s mixed-bootstrap T-test path for p-value reporting. This keeps change intervals and status messages aligned with Criterion’s own model.

Examples, docs, and CI

- add new pgrx-examples/benching example crate
- add example README showing #[pg_bench], setup functions, and transaction handling
- add PG_BENCH-SPEC.md design/spec document
- expand pgrx-bench README with feature overview, workflow guidance, and black_box discussion
- update cargo-pgrx README to document cargo pgrx bench
- add CI smoke coverage for the new benching example
- update publish.sh for the new pgrx-bench crate

Tests

- add unit tests in pgrx-bench for:
  - Criterion new-vs-base directory selection
  - baseline artifact materialization
  - artifact collection
  - Criterion-style summary classification
  - deterministic p-value behavior
  - comparison parsing from persisted baseline artifacts

This establishes pg_bench as the in-process benchmarking counterpart to pg_test for pgrx extensions, with persistent historical storage and Criterion-compatible comparison reporting.

--- 

NOTE:  100% AI, but this is a super cool feature.  Something I've always wanted to do but just never had the time.  Only took a few hours with Codex.


cargo-pgrx/src/command/bench.rs
```diff
@@ -0,0 +1,1523 @@
...
+        .arg("-lc")
+        .arg(command)
+        .output()
+        .wrap_err_with(|| format!("failed to run `{command}`"))?;
+    if output.status.success() {
+        Ok(String::from_utf8_lossy(&output.stdout).trim().to_string())
+    } else {
+        Err(eyre!("command `{command}` failed: {}", String::from_utf8_lossy(&output.stderr).trim()))
+    }
+}
```
eeeebbbbrrrr: What's going on here?  Why would we be doing anything through `zsh`?

A requirement is that this also work on Windows, assuming it has the command/tool being executed.

cargo-pgrx/src/command/bench.rs
```diff
@@ -0,0 +1,1523 @@
...
+        command_output(&format!("git -C {} describe --always --dirty --tags", shell_escape(&root)))
+            .ok();
+    let git_dirty = command_output(&format!(
+        "git -C {} status --porcelain --untracked-files=no",
+        shell_escape(&root)
+    ))
+    .map(|status| !status.is_empty())
+    .unwrap_or(false);
+
+    Ok(GitMetadata { git_commit, git_branch, git_describe, git_dirty })
```
eeeebbbbrrrr, line 1574: this should be using Command's .arg() builder APIs, not formatting strings that then need to be escaped using `shell_escape()` below.

cargo-pgrx/src/command/bench.rs
```diff
@@ -0,0 +1,1523 @@
...
+}
+
+fn shell_escape(value: &str) -> String {
+    let mut escaped = String::from("'");
+    escaped.push_str(&value.replace('\'', "'\\''"));
+    escaped.push('\'');
+    escaped
+}
```
eeeebbbbrrrr: See previous comment.  this should be 100% unnecessary

cargo-pgrx/src/command/bench.rs
```diff
@@ -0,0 +1,1523 @@
...
+    fn as_str(&self) -> &'static str {
+        match self {
+            Self::Shared => "shared",
+            Self::SubtransactionPerBatch => "subtransaction_per_batch",
+            Self::SubtransactionPerIteration => "subtransaction_per_iteration",
+        }
+    }
+}
+
+const PERSISTENT_SCHEMA_SQL: &str = r#"
```
eeeebbbbrrrr: Prefer this to be in a "pgrx-bench.sql" artifact in the source tree and included with `include_bytes!()`

cargo-pgrx/src/command/install.rs
```diff
@@ -114,6 +115,25 @@ impl CommandExecute for Install {
...
+    }
+
+    eprintln!(
+        "{}",
+        format!(
+            "WARNING: building with feature `pg_bench`\nbenchmark functions and helper dependencies will be included in this `cargo pgrx {command}` build\nthis is usually not intended for packaged releases"
+        )
+        .red()
+        .bold()
+    );
```
eeeebbbbrrrr, line 130: just put the `WARNING:` bit in red().bold().

pgrx-bench/src/lib.rs
```diff
@@ -0,0 +1,1305 @@
...
+enum Routine<'a> {
+    Iter(Box<dyn FnMut() + 'a>),
+    IterBatched {
+        setup: Box<dyn FnMut() -> Box<dyn Any> + 'a>,
+        routine: Box<dyn FnMut(Box<dyn Any>) + 'a>,
+        batch_size: BatchSize,
+    },
+}
+
+impl<'a> Bencher<'a> {
```
eeeebbbbrrrr, line 99: detailed usage comments for all the public members is necessary.

pgrx-bench/src/lib.rs
```diff
@@ -0,0 +1,1305 @@
...
+
+    fn into_routine(self) -> Result<Routine<'a>, String> {
+        self.routine.ok_or_else(|| {
+            "benchmark function did not register a timing loop; call b.iter(...) or b.iter_batched(...)"
+                .to_string()
+        })
+    }
+}
+
+pub fn execute_benchmark(
```
eeeebbbbrrrr: This is public for pgrx to use not not intended for users.  `#[doc(hidden)]`

pgrx-bench/src/lib.rs
```diff
@@ -0,0 +1,1305 @@
...
+                throughput: None,
+                comparison: None,
+                artifacts: Vec::new(),
+            })
+            .expect("BenchResult should serialize"),
+        ),
+    }
+}
+
+pub fn describe_benchmark(definition: BenchDefinition) -> JsonB {
```
eeeebbbbrrrr: `#[doc(hidden)]`

pgrx-bench/src/lib.rs
```diff
@@ -0,0 +1,1305 @@
...
+            )),
+        }
+    }
+}
+
+thread_local! {
+    static SEED_RNG: RefCell<Rand64> = RefCell::new(Rand64::new(
+        SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("time went backwards")
```
eeeebbbbrrrr: .expect() messages should be written in affirmative.  and in this case it should be an `.unwrap_or_else(|| panic!("time went backwards"))`

pgrx-macros/src/lib.rs
```diff
@@ -136,13 +141,282 @@ pub fn pg_test(attr: TokenStream, item: TokenStream) -> TokenStream {
     stream.into()
 }
 
+/// `#[pg_bench]` functions are in-process Criterion-driven benchmarks that run inside Postgres
+/// during `cargo pgrx bench`.
+#[proc_macro_attribute]
```
eeeebbbbrrrr, line 142: there's enough new code added to this file to warrant a new `pg_bench.rs` file to contain all these things.

eeeebbbbrrrr (commented):
- any item in the new `pgrx-bench` crate that is meant for a user to use should have doccomments with at least one example of how to use the item.  All others that need to be public for internal use by pgrx should also be concisely documented but behind `#[doc(hidden)]`

- the bench runner in cargo-pgrx needs to be compatible with Windows and should not be formatting full command strings but instead should be using `Command`'s argument apis.

pgrx-bench/src/lib.rs
```diff
@@ -41,13 +70,17 @@ const ARTIFACT_KIND_CHANGE_ESTIMATES_JSON: &str = "criterion_change_estimates_js
 
 #[derive(Debug, Clone, Copy, Serialize, Deserialize, Eq, PartialEq)]
 #[serde(rename_all = "snake_case")]
+#[doc(hidden)]
```
eeeebbbbrrrr: so there's a lot of these in this file.  lets make a new module called `pgrx.rs` here at the root and move all these over there.  document that module as being for pgrx-only and then still keep all these `#[doc(hidden)]` in the new module

cargo-pgrx/src/command/bench.rs
```diff
@@ -887,44 +887,84 @@ fn quote_ident(value: &str) -> String {
...
+}
+
+fn format_program_and_args(program: &str, args: &[&str]) -> String {
+    let mut parts = Vec::with_capacity(args.len() + 1);
+    parts.push(program);
+    parts.extend_from_slice(args);
+    parts.join(" ")
+}
+
+fn host_name() -> eyre::Result<String> {
```
eeeebbbbrrrr: this doesn't seem like a valuable metric to check.  delete it and any references to it elsewhere in this PR.

pgrx-macros/src/pg_bench.rs
```diff
@@ -0,0 +1,354 @@
...
+                            ::pgrx::pg_sys::MemoryContextSwitchTo(old_context);
+                            ::pgrx::pg_sys::RollbackAndReleaseCurrentSubTransaction();
+                            ::pgrx::pg_sys::MemoryContextSwitchTo(old_context);
+                            ::pgrx::pg_sys::CurrentResourceOwner = old_resource_owner;
+                            Err(error)
+                        }
+                    }
+                }
+            }
+        }
```
eeeebbbbrrrr: these things pollute the top-level namespace of the file in which they're generated, do they not?

wouldn't it be possible to declare all of this in the `fn #run_wrapper_name` below as inner definitions?

eeeebbbbrrrr:
I think this looks pretty good now.

eeeebbbbrrrr:
the CI failure is because it's looking for the new crate `pgrx-bench` on crates.io and, well, that hasn't been published yet.

eeeebbbbrrrr:
This has been working really well so I'm going to merge it and I guess we'll just deal with the one failing CI job until this is actually released and the new crate is published to crates.io

sweatybridge:
I'm interested to try this out. When will it be published to crates.io?

eeeebbbbrrrr:
Soon, I hope. Got another gigantic PR I’m still play testing (#2264).

PR #2264, opened 2026-03-23
one compile, please

eeeebbbbrrrr:
# v18 One Compile Changelog

This branch changes `pgrx` from a two-artifact schema pipeline into a one-compile pipeline.
The extension shared library is now the source of truth for SQL entity metadata, and the
rest of the branch lines up behind that decision: type identity is explicit, schema graph
resolution is stricter, extension-owned and external SQL types are modeled separately, and
the workspace, examples, tests, and docs were all rebuilt around the new flow.

This document is written as an aggregate changelog for the branch, not as a commit diary.
It describes the current end state after all branch commits have landed.

## Executive Summary

- `cargo pgrx schema` now works from the compiled extension shared library instead of
  generating and running a `pgrx_embed` helper binary.
- New extensions and examples are `cdylib`-only. They no longer need `src/bin/pgrx_embed.rs`,
  `[[bin]]`, or the redundant `"lib"` crate type.
- SQL type resolution now uses `TYPE_IDENT`, a qualified Rust-side identity, instead of the
  older `SCHEMA_KEY` model.
- `SqlTranslatable` now separates "what type is this?" from "how should this appear in SQL?".
- The SQL graph now has clearer rules for external types, extension-owned types,
  `extension_sql!` declarations, explicit composites, and duplicate type producers.
- The repo is now a real workspace with an in-tree `cargo-pgrx`, a slimmed-down `pgrx-tests`
  harness, and a new `pgrx-unit-tests` extension crate that exercises the new model directly.
- The branch also adds the design and review trail that explains why these changes exist:
  RFC 0001, review findings, remediation notes, migration notes, and the v18 type-resolution
  docs.

## One-Pass Schema Generation

The biggest user-visible change is that schema generation is now single-pass.

`cargo pgrx schema` no longer builds a helper executable, runs it, and then rebuilds the
extension. It now does one normal extension build at most, locates the compiled shared
library, reads the embedded SQL entity section from that artifact, deserializes the entities,
and builds SQL or DOT output from there.

That changes a few important details:

- `--skip-build` now means "use the artifact that already exists". If that artifact is stale,
  missing, or incompatible, schema generation fails loudly instead of quietly emitting empty
  or misleading SQL.
- The object reader now lives in `cargo-pgrx` and understands the formats it has to inspect,
  including Mach-O universal binaries. This is no longer outsourced to a helper binary.
- Missing embedded schema metadata is a hard error. The tool now treats that as a broken build
  state, not a recoverable "best effort" path.
- Installed artifacts retain the embedded schema section. The branch ended in a state where
  schema metadata is treated as inert runtime data that should stay attached to the shared
  object, not something `cargo-pgrx` strips away after generation.
- Versioned shared-object SQL generation now keys off the extension crate version, so the
  generated SQL stays aligned with the actual versioned library name.

The section names were also shortened across platforms. The canonical names are now `.pgrxsc`
on ELF and PE, and `__DATA,__pgrxsc` on Mach-O. Reads remain backward-compatible with the
older `.pgrx_schema` naming, so newer tools can still consume older artifacts.

## Extension Template And Example Changes

Once schema generation stopped depending on a helper executable, the extension layout could
get simpler.

Newly generated extensions now look like normal `cdylib` extensions:

- no `src/bin/pgrx_embed.rs`
- no `[[bin]]`
- no `crate-type = ["lib", "cdylib"]`

The template now emits `crate-type = ["cdylib"]` and keeps the rest of the manifest focused
on the extension itself.

The examples were updated to match that same model. Across the example tree, the old
`pgrx_embed` binaries are gone and the manifests were normalized to the new crate layout.
That means the examples now serve as current reference material for the one-compile flow,
instead of preserving the old two-step scaffolding.

## Type Identity: SCHEMA_KEY To TYPE_IDENT

The SQL entity graph now resolves concrete Rust types through `TYPE_IDENT`.

This is more than a rename. The old "schema key" concept was replaced with a clearer split
between type identity, type origin, and SQL spelling:

- `TYPE_IDENT` is the canonical Rust-side identity used for graph resolution.
- `TYPE_ORIGIN` says whether the type comes from this extension or from outside it.
- `ARGUMENT_SQL` describes how the type appears in argument position.
- `RETURN_SQL` describes how the type appears in return position.

The identity macro behind this is now qualified with `module_path!()`, so two types with the
same spelled name but different module paths don't collapse onto the same graph key. In other
words, resolution is now keyed by a qualified Rust identity instead of a loosely inferred SQL
name.

The branch also routed concrete type producers and consumers through that model consistently.
Derived SQL entities, manual `SqlTranslatable` impls, and graph lookups all speak the same
identity language now.

## SQL Shape Is Separate From Graph Resolution

Another major branch theme is that SQL spelling and graph resolution are no longer fused
together.

Function and aggregate metadata now carry both:

- optional resolution metadata, used when a type should participate in graph lookup
- SQL-only shape metadata, used to spell the type correctly in emitted SQL

That split matters because wrapper shape and leaf identity are not the same thing.

Wrappers such as `Option<T>`, `Vec<T>`, `Array<T>`, table iterators, and similar constructs
generally forward the leaf type's identity and origin, while the function metadata still keeps
enough SQL shape to emit the correct argument or return form. Resolution follows the concrete
type producer. SQL emission still knows whether the user declared an array, set return, table
return, nullable wrapper, and so on.

This also cleaned up a lot of dead metadata. The branch removed older paths that tried to carry
too much type-origin or identity information in places where it was either redundant or wrong.

## Explicit Composites Are SQL-Only

The branch makes explicit `composite_type!(...)` handling much more precise.

Explicit composites are now modeled as SQL-only declarations. They are not treated as resolvable
Rust type identities. If a function says `composite_type!("Dog")`, the SQL emitter uses that
literal composite name when generating SQL, and the graph does not try to look up some matching
`TYPE_IDENT` for `"Dog"`.

This fixes a real semantic problem in the old model, where explicit composite spellings were too
easy to drag into resolution paths that should only have applied to concrete Rust-defined types.

The same distinction carries through to arrays of explicit composites. The emitted SQL still
comes out correctly, but these uses do not create graph dependencies on derived `PostgresType`
items unless the code actually referenced such a type by identity elsewhere.

## `extension_sql!` Declared Types Now Behave Like First-Class Producers

The branch tightened how `extension_sql!(..., creates = ...)` declarations participate in the
graph.

Declared `Type(...)` and `Enum(...)` entries now keep their `TYPE_IDENT`, and downstream
functions and aggregates can resolve to those declarations just like they resolve to derived
`PostgresType` or `PostgresEnum` entities.

That gives the graph a clean rule:

- one `TYPE_IDENT` can resolve to exactly one producer
- that producer can be a derived type, a derived enum, or an `extension_sql!` declared type/enum

The graph now treats duplicate producers as an error. If two SQL entities claim the same
`TYPE_IDENT`, schema generation fails instead of guessing which one should win.

This work also removed stale metadata from `extension_sql!` declarations. Declared SQL types
keep the identity they need for graph binding, but they no longer carry a separate `type_origin`
there, and functions do not try to smuggle type identity through SQL declaration metadata.

## External Types vs ThisExtension Types

The graph now has much clearer rules for unresolved references.

If a referenced type is marked `TYPE_ORIGIN = External` and there is no in-graph producer for
it, the graph can synthesize an external placeholder node and continue. That keeps built-in or
externally-owned SQL types usable without forcing `pgrx` to pretend it owns them.

If a referenced type is marked `TYPE_ORIGIN = ThisExtension`, the graph no longer falls back to
that placeholder path. A missing producer is a hard error. This is one of the most important
fail-fast changes in the branch: extension-owned references must actually resolve to something
the extension emits.

The `extension_sql!` validation path was tightened to match. `creates = [Type(T)]` and
`creates = [Enum(T)]` are now only accepted for extension-owned, concretely representable SQL
types. External types, skipped types, explicit composites, arrays, and similar unsupported
shapes are rejected up front instead of entering the graph half-specified.

## SQL Emission And Ordering Rules

The graph and emission passes also got more exact about ordering and dependency semantics.

- Derived `PostgresType` SQL is emitted in shell-first order: create the shell, wire up the I/O
  pieces, then materialize the final type definition.
- Within strongly connected components, explicit `requires = [...]` edges now outrank weaker
  type-resolution edges. That matters for SQL-only composites and shell-type ordering.
- Functions and aggregates only ask the graph for a schema prefix when they have resolution
  metadata for a type. SQL-only composites bypass that lookup and emit their declared composite
  name directly.
- Declared `extension_sql!` types participate in dependency ordering like real producers, so
  downstream items can depend on those declarations instead of only on derived Rust entities.

The result is a graph that is both stricter and easier to reason about. More cases are modeled
directly, and fewer cases rely on accidental behavior from over-general metadata.

## Test Harness, Workspace, And CI Rework

This branch also rebuilt the repo layout around the new pipeline.

The root is now an actual workspace that includes:

- `cargo-pgrx`
- the core `pgrx` crates
- `pgrx-tests`
- `pgrx-unit-tests`
- `pgrx-examples/*`

The workspace metadata also points local installs at the in-tree `cargo-pgrx`, which lets the
test harness and CI exercise the branch's `cargo-pgrx` implementation instead of whatever
binary happens to be installed globally.

The test story was split cleanly:

- `pgrx-tests` is now the reusable harness crate
- `pgrx-unit-tests` is the internal test extension crate

`pgrx-tests` now defaults to that smaller harness surface instead of trying to double as the
internal test extension itself.

That split matters because the old setup blurred "helper library" and "thing being tested".
The current layout makes it much easier to run the internal extension against the same schema
generation and graph code that real users hit.

Coverage was expanded around the new behavior:

- `TYPE_IDENT` and `SqlTranslatable` coverage grew substantially
- compile-fail tests now check that manual `SqlTranslatable` impls define `TYPE_IDENT`
- compile-fail tests also verify that wrappers don't hide unsupported non-SQL leaf types
- CI now runs `cargo test --all` across PostgreSQL 13 through 18
- CI includes a beta Rust leg on PostgreSQL 16
- CI also checks `pgrx-unit-tests` in both `cshim` and non-`cshim` modes, runs the UI tests,
  performs an arm64 cross-build check, smoke-tests `cargo pgrx bench`, and exercises schema
  generation against the versioned custom-libname example

This is one of the branch's quieter but more important outcomes. The new invariants are not
just documented. They are exercised in-tree.

## Documentation And Review Trail

This branch carries its own design and review history in the tree.

New or significantly expanded docs in the branch include:

- `rfcs/0001-single-pass-schema-generation.md`
- `rfcs/0001-initial-deep-review.md`
- `rfcs/0001-review-findings.md`
- `rfcs/0001-deep-review-findings.md`
- `rfcs/0002-review-remediation.md`
- `v18.0-MIGRATION.md`
- `v18-TYPE-RESOLUTION.md`

The internal documentation was also updated to describe the linker-section-based schema flow,
and the older "forging SQL from Rust" article was left in place as historical material rather
than current implementation guidance.

That matters for this branch because a lot of the changes here are not obvious from a surface
API diff. The RFC and review trail records the failure cases that motivated the stricter graph
rules and the one-pass artifact design.

## Smaller Cleanup That Still Matters

Not every commit in the branch introduced a new concept, but several follow-up commits materially
improved the final state:

- a lifetime cleanup replaced a broad set of `Box::leak` patterns in schema and SQL graph code
  with borrowed data where possible
- a code-review follow-up trimmed a few rough edges in the object reader, docs, SQL emission,
  and graph code
- tuple iterator naming warnings were silenced in `pgrx`
- repo-local ignore rules were updated for the current working setup
- the final doc cleanup pass tightened wording in the new v18 docs

These are small compared to the core architecture work, but they are part of the branch's final
shape and should be called out as such.

## Practical Migration Notes

For extension authors, the branch leaves a few clear takeaways:

- delete old `pgrx_embed` binaries and `[[bin]]` entries
- use `crate-type = ["cdylib"]`
- treat missing embedded schema metadata as a build problem, not something `cargo pgrx schema`
  should paper over
- on manual `SqlTranslatable` impls, define `TYPE_IDENT`
- don't rely on wrappers to make a non-SQL leaf type acceptable
- use `extension_sql!(..., creates = ...)` only for extension-owned, concrete SQL types or enums
- expect explicit `composite_type!(...)` declarations to stay SQL-only unless there is a real
  concrete type producer elsewhere in the graph

In short, the branch trades some old flexibility for a model that is much more explicit. That is
intentional. The current code prefers exact identity, exact ownership, and hard failures over
implicit guesses.

## Commit Coverage

This changelog covers the branch's current post-rebase commits in aggregate:

- design, review, and doc trail: 6b7c1ff9, 226f7f9a, bd7e04d4, 6cdb7eb1, 074a3b1d
- one-pass schema pipeline and artifact handling: e317229b, 62d16338, 2c39c1f4, bd093d6b
- template and extension layout simplification: 7067708c
- type identity and graph semantics: ab9d9a49, 575322c7, 116da7f0, ff140f50, d3eab0f5,
  d023f577, 1d1ff15e, 30aec2e2, d8b55c7e, 9f665b07
- workspace, harness, and test expansion: a5829b05, 194e986c, b4750d9a, 57fc49cd
- follow-up cleanup and polish: dc6aee0b, e5692a63, 1e2fadab, db180e14
- changelog addition: e1ab0658

If you want the shortest possible summary of the branch, it is this: we now compile the
extension once, read its schema metadata from the shared object, resolve SQL types by explicit
qualified identity, and fail hard when the graph can't prove what owns a type.


eeeebbbbrrrr:
not at all ready to commit.  just wanna see what CI has to say about this, which I bet will be a lot!

eeeebbbbrrrr:
This might be a lot closer to being able to commit.  it's huge tho.  I've spent quite a bit of time reviewing and it and fixing it both with and without my AI buddy.

eeeebbbbrrrr:
CI passed on all platforms, so that's something

.github/workflows/package-test.yaml
cbandy, line 1: Are these package names in a particular order? If not, alphabetical may be easier to maintain.

pgrx-examples/arrays/Cargo.toml
```diff
@@ -32,19 +29,9 @@ pg18 = ["pgrx/pg18", "pgrx-tests/pg18"]
 pg_test = []
 
 [dependencies]
-pgrx = { path = "../../pgrx", default-features = false }
+pgrx = { path = "../../pgrx" }
```
cbandy, line 32: Do all these examples now require `default-features`? If so, should `cshim` stop being a feature that can be disabled?

v18-ONE-COMPILE-CHANGELOG.md
```diff
@@ -0,0 +1,308 @@
...
+- `--skip-build` now means "use the artifact that already exists". If that artifact is stale,
+  missing, or incompatible, schema generation fails loudly instead of quietly emitting empty
+  or misleading SQL.
+- The object reader now lives in `cargo-pgrx` and understands the formats it has to inspect,
+  including Mach-O universal binaries. This is no longer outsourced to a helper binary.
+- Missing embedded schema metadata is a hard error. The tool now treats that as a broken build
+  state, not a recoverable "best effort" path.
+- Installed artifacts retain the embedded schema section. The branch ended in a state where
+  schema metadata is treated as inert runtime data that should stay attached to the shared
+  object, not something `cargo-pgrx` strips away after generation.
```
cbandy, line 49: It sounds like this may change the size of binaries being shipped. If so, please describe whether they're larger, smaller, etc.

🤔 Is this perhaps needed only during "build?" Does `strip` remove these sections? Is that okay?

pgrx-tests/src/framework.rs
```diff
@@ -989,21 +989,42 @@ fn get_cargo_args() -> Vec<String> {
...
+        None => return Ok(None),
+    };
+    let manifest_path = workspace_root.join("cargo-pgrx/Cargo.toml");
+
+    if !manifest_path.is_file() {
+        return Ok(None);
+    }
+
+    let mut command = Command::new(cargo.unwrap_or_else(|| "cargo".into()));
+    command.arg("run").arg("--manifest-path").arg(manifest_path).arg("--").arg("pgrx");
```
cbandy, line 1026: 👍🏻

pgrx-sql-entity-graph/src/metadata/sql_translatable.rs
```diff
@@ -57,9 +58,232 @@ impl SqlMapping {
...
+}
+
+/// Implements `SqlTranslatable` for a type with a fixed external SQL mapping.
+///
+/// This macro uses `pgrx_resolved_type!(T)` for `TYPE_IDENT`, sets
+/// `TYPE_ORIGIN` to `TypeOrigin::External`, and fills in the const SQL metadata
+/// for the common "map this Rust wrapper to an existing SQL type" case.
+///
+/// Write the `unsafe impl SqlTranslatable` by hand when the type is owned by
+/// this extension, or when argument and return SQL need different mappings.
```
cbandy: nit

```suggestion
/// Spell out the `unsafe impl SqlTranslatable` instead when (1) the type is owned by
/// this extension or (2) when its argument and return SQL need different mappings.
```

pgrx-sql-entity-graph/src/metadata/sql_translatable.rs
```diff
@@ -57,9 +58,232 @@ impl SqlMapping {
...
+///
+/// This macro is re-exported by `pgrx`, and also available through
+/// `pgrx::prelude::*`.
```
cbandy: nit

```suggestion
/// This macro is re-exported by `pgrx` and is also available through
/// `pgrx::prelude::*`.
```

pgrx-sql-entity-graph/src/metadata/sql_translatable.rs
```diff
@@ -57,9 +58,232 @@ impl SqlMapping {
...
+        }
+    };
+}
+
 /**
 A value which can be represented in SQL
 
+If you need the common "fixed external SQL type" case, prefer
+`impl_sql_translatable!`. Write the trait impl by hand when the type is owned
+by this extension, or when argument and return SQL differ.
```
cbandy: nit. To me, "argument and return SQL differ" seems to include the `arg_only` case that the macro supports. 🤷🏻 "unusual" is vague enough that I would actually go read the macro doc.

```suggestion
If you need the common "fixed external SQL type" case, prefer
`impl_sql_translatable!`. Spell out this trait impl when (1) the type is owned
by this extension or (2) when the argument or return SQL is unusual.
```

pgrx/src/datetime/time_stamp.rs
```diff
@@ -329,10 +329,11 @@ impl<'de> serde::Deserialize<'de> for Timestamp {
 }
 
 unsafe impl SqlTranslatable for crate::datum::Timestamp {
```
cbandy: Why is this impl on `crate::datum::…` when the others in this file are on `Timestamp` in this module?

pgrx/src/datum/json.rs
```diff
@@ -188,19 +188,19 @@ impl Serialize for JsonString {
...
-    }
+    const TYPE_IDENT: &'static str = crate::pgrx_resolved_type!(Json);
+    const TYPE_ORIGIN: pgrx_sql_entity_graph::metadata::TypeOrigin =
+        pgrx_sql_entity_graph::metadata::TypeOrigin::External;
+    const ARGUMENT_SQL: Result<SqlMappingRef, ArgumentError> = Ok(SqlMappingRef::literal("json"));
+    const RETURN_SQL: Result<ReturnsRef, ReturnsError> =
+        Ok(ReturnsRef::One(SqlMappingRef::literal("json")));
 }
 
 unsafe impl SqlTranslatable for crate::datum::JsonB {
```
cbandy: Here as well. Other traits in this file are impl on `JsonB` of this module, but this one is on `crate::datum::…`. Why's that?

pgrx/src/datum/uuid.rs
```diff
@@ -131,10 +131,10 @@ impl std::fmt::UpperHex for Uuid {
 }
 
 unsafe impl SqlTranslatable for crate::datum::Uuid {
```
cbandy: 📝 datum

pgrx-sql-entity-graph/src/pgrx_sql.rs
```diff
@@ -1417,20 +1647,822 @@ fn make_extern_connection(
...
+    types: &HashMap<PostgresTypeEntity<'a>, NodeIndex>,
+    enums: &HashMap<PostgresEnumEntity<'a>, NodeIndex>,
+) -> bool {
+    find_type_or_enum(type_ident, types, enums)
+        .map(|ty_index| graph.add_edge(ty_index, index, SqlGraphRequires::By))
         .is_some()
 }
+
+#[cfg(test)]
+mod tests {
```
cbandy, line 1824: 👍🏻 🎉

pgrx-sql-entity-graph/src/section.rs
```diff
@@ -0,0 +1,1371 @@
...
+So the current design choice is:
+
+- keep the section format binary while the producer is const-time linker-section
+  emission
+- keep the wire format logic centralized in this module instead of spreading it
+  across macro emitters
+
+If we ever move away from linker-section bytes and toward a normal build-time
+manifest artifact, that would be the right time to re-evaluate a serde-based
+format such as JSON.
```
cbandy, line 41: 📝 These words led me to this post about TLV and serde: https://stackoverflow.com/a/73205819/1608057

rfcs/0001-single-pass-schema-generation.md
```diff
@@ -0,0 +1,906 @@
+# RFC 0001: Single-Pass Schema Generation
+
+## Status
+
+Draft
```
cbandy: Should this change to Accepted or similar during this PR?

rfcs/0001-single-pass-schema-generation.md
```diff
@@ -0,0 +1,906 @@
...
+    const RETURN_SQL: Result<ReturnsRef, ReturnsError>;
+
+    // Runtime methods remain as compatibility shims over the const metadata.
+    fn type_name() -> &'static str { core::any::type_name::<Self>() }
+    fn argument_sql() -> Result<SqlMapping, ArgumentError> {
+        Self::ARGUMENT_SQL.into_runtime()
+    }
+    fn return_sql() -> Result<Returns, ReturnsError> {
+        Self::RETURN_SQL.into_runtime()
+    }
```
cbandy: I saw these removed in many places. Is the compatibility described here still a goal?

rfcs/0001-single-pass-schema-generation.md
```diff
@@ -0,0 +1,906 @@
...
+`SqlMapping` and `Returns`. They use only `&'static str`, booleans, and small
+enums, so proc macros can embed them without executing code or allocating
+`String`s. This preserves the historical `SqlTranslatable` source of truth for
+SQL spellings while making that information available during the single library
+build.
+
+This is a **breaking change**: existing manual `unsafe impl SqlTranslatable`
+blocks must provide compile-time schema metadata, including `TYPE_ORIGIN`, not
+just runtime methods. The migration is still mechanical, but it is larger than
+a one-line change.
```
cbandy: 👍🏻 The migration of builtin types seems straightforward to me in this PR, and the v18 MIGRATION doc provides a variety of examples.

rfcs/0001-single-pass-schema-generation.md
```diff
@@ -0,0 +1,906 @@
...
+  `SqlTranslatable::{TYPE_IDENT, TYPE_ORIGIN, ARGUMENT_SQL, RETURN_SQL}`.
+
+- **Unique static names** are generated by the proc macro using `format_ident!()`
+  (from `proc_macro2`), not by declarative macro name-concatenation. This avoids the
+  nightly-only `${concat(...)}` metavar feature.
+
+- **Platform-conditional section names.** macOS Mach-O requires `"segment,section"`
+  format. The proc macro emits `cfg_attr` to use `"__DATA,__pgrxsc"` on macOS and
+  `".pgrxsc"` on all other platforms. The short names keep the section within the
+  8-byte PE/COFF image-name limit used on Windows.
```
cbandy: This 8-byte limit seems important enough to mention in `pgrx-sql-entity-graph/src/section.rs`.

rfcs/0001-single-pass-schema-generation.md
```diff
@@ -0,0 +1,906 @@
...
+```
+
+The section-reading helpers in `cargo-pgrx` are retained in simplified form. The
+symbol-table scanning, codegen, second build, and binary execution are all deleted.
+
+#### Stripping the section from installed artifacts
+
+This RFC does not require automatic stripping as part of `cargo pgrx install`. The
+embedded schema section is inert at runtime, and install/package-time artifact trimming
+can be layered on later if it proves worthwhile. The important behavior for this RFC is
```
cbandy: This partly answers one of my earlier questions.

🤔 Perhaps a side-effect (or direct benefit) of the TLV encoding is that unstripped binaries are smaller than they would be with another encoding.

> The important behavior for this RFC is that schema generation reads the section from the freshly built shared object and does not require a second build or an executable helper binary.

💯 I agree. 🥇

rfcs/0001-single-pass-schema-generation.md
```diff
@@ -0,0 +1,906 @@
...
+
+- Syntactic wrappers already understood by pgrx's parsing layer (`default!()`,
+  `composite_type!()`, `Option<T>`, `Result<T, E>`, `Vec<T>`, `Array<T>`,
+  `VariadicArray<T>`, `SetOfIterator<T>`, `TableIterator<(...)>`, etc.) are
+  normalized exactly as they are today.
+- For each resulting leaf type, the generated code references
+  `SqlTranslatable::{TYPE_IDENT, TYPE_ORIGIN, ARGUMENT_SQL, RETURN_SQL}`.
+- pgrx's own built-ins (`PgRelation`, `TimeWithTimeZone`, `Range<T>`, `PgBox<T>`,
+  `PgVarlena<T>`, and so on) continue to work by implementing those associated consts
+  in pgrx itself, rather than by being duplicated in proc-macro lookup tables.
```
cbandy: 👍🏻

rfcs/0001-single-pass-schema-generation.md
```diff
@@ -0,0 +1,906 @@
...
+- **No `CARGO_INCREMENTAL=0` needed.** Full incremental compilation support.
+
+## Impact on Extension Authors
+
+### Common case: derive macros only
+
+**Users who use `#[derive(PostgresType)]`, `#[derive(PostgresEnum)]`, `#[pg_extern]`,
+etc. exclusively** — the vast majority of pgrx users:
+
+**No code changes required.**
```
cbandy: AMAZING

rfcs/0001-single-pass-schema-generation.md
```diff
@@ -0,0 +1,906 @@
...
+
+The SQL spelling of a type is not derived from the Rust token and is not guessed from
+`creates = [Type(T)]`. Instead:
+
+- function argument SQL comes from `ARGUMENT_SQL`
+- function return SQL comes from `RETURN_SQL`
+- wrapper behavior (`Result<T, E>`, `Option<T>`, `Vec<T>`, arrays, etc.) comes from the
+  blanket impls' composed const metadata
+
+This preserves cases like a Rust type `HexInt` that must render as SQL `hexint`.
```
cbandy: 📝 `HexInt`, what a handy example!

v18-TYPE-RESOLUTION.md
```diff
@@ -0,0 +1,581 @@
...
+extension_sql!(
+    r#"CREATE TYPE complex;"#,
+    name = "create_complex_shell_type",
+    creates = [Type(Complex)]
+);
+```
+
+pgrx records two things:
+
+- the concrete SQL spelling, taken from `T::argument_sql()`
```
cbandy: Should this be the const, ARGUMENT_SQL?

cbandy (approved):
I am very excited by this! Everything LGTM.

pgrx-examples/arrays/Cargo.toml
eeeebbbbrrrr: It's less of if they **require** `default-features` and just more of "this is easier for us to maintain".

It's still important, unfortunately, that we keep `cshim` as an opt-out feature for various cross-compilation reasons.  But at the end of the day, if none of the cshim items are used by your extension then they'll be dce'd either by the compiler or the LTO pass.

.github/workflows/package-test.yaml
eeeebbbbrrrr: I'm not worried about it either way.

rfcs/0001-single-pass-schema-generation.md
eeeebbbbrrrr: We'll eventually provide a way to strip these things, at least for `cargo pgrx package`.  It just a matter of figuring out how to do that for linux, macos, and windows without reinventing the wheel.  It's on my radar, but towards the bottom.

rfcs/0001-single-pass-schema-generation.md
eeeebbbbrrrr: No, there's no intended compatibility.  Can't anyways as stable Rust doesn't support `const fn` on trait members yet.

I'm also not worried about these plan docs.  I'll end up deleting them -- it's all AI generated planning anyways.

pgrx/src/datum/uuid.rs
eeeebbbbrrrr: All these are just from the AI bot being confused.  I'll clean them up.

cbandy (approved):

cbandy:
I have some extensions that are _only_ `shared_preload_libraries`. For these, this branch fails to `cargo pgrx package` as it cannot find the pgrx schema section.

For example, this very empty module works with `v0.17.0` but not this branch:

```rust
pgrx::pg_module_magic!(name, version);

#[allow(non_snake_case)]
#[pgrx::pg_guard]
pub extern "C-unwind" fn _PG_init() {}
```

Success:

```console
$ cargo install --locked cargo-pgrx@0.17.0
$ cargo pgrx new empty && cat … > empty/src/lib.rs
$ (cd empty && cargo pgrx package)
…
Discovered 0 SQL entities: 0 schemas (0 unique), 0 functions, 0 types, 0 enums, 0 sqls, 0 ords, 0 hashes, 0 aggregates, 0 triggers
…
```

Error:

```console
$ cargo install --locked cargo-pgrx --git 'https://github.com/pgcentralfoundation/pgrx' --branch 'wip-one-compile-please'
$ cargo pgrx new empty && cat … > empty/src/lib.rs
$ # update to git [dependencies]
$ (cd empty && cargo pgrx package)
…
    Finished `release` profile [optimized] target(s) in 55.43s
  Installing extension
     Copying control file to target/release/empty-pg18/usr/share/postgresql/18/extension/empty.control
     Copying shared library to target/release/empty-pg18/usr/lib/postgresql/18/lib/empty.so
Error: 
   0: no embedded pgrx schema section found; expected `.pgrxsc` on ELF/PE or `__DATA,__pgrxsc` on Mach-O. the artifact may have been built with an incompatible pgrx, stripped incorrectly, or selected from the wrong architecture slice

Location:
   cargo-pgrx/src/command/schema.rs:245
…
```

eeeebbbbrrrr:
@cbandy thanks for finding that.  try again as of [70254e4](https://github.com/pgcentralfoundation/pgrx/pull/2264/commits/70254e486ccaf4de66feacc02c1f45a8ea5623ed).  should work now.

cbandy:
Works great!

eeeebbbbrrrr:
Hold on to your hat... I'm gonna merge this.

pmpetit:
Hi, i have the same kind of message using 0.18.0, only on arm64.... not amd64

[here](https://github.com/pmpetit/pglinter/actions/runs/24666932945/job/72126958135)

```log
   Compiling pglinter v2.0.0 (/__w/pglinter/pglinter)
    Finished `release-lto-off` profile [optimized] target(s) in 45.33s
  Installing extension
     Copying control file to target/release-lto-off/pglinter-pg17/home/postgres/.pgrx/17.9/pgrx-install/share/extension/pglinter.control
     Copying shared library to target/release-lto-off/pglinter-pg17/home/postgres/.pgrx/17.9/pgrx-install/lib/pglinter.so
Error: 
   0: no embedded pgrx schema section found; expected `.pgrxsc` on ELF/PE or `__DATA,__pgrxsc` on Mach-O. the artifact may have been built with an incompatible pgrx, stripped incorrectly, or selected from the wrong architecture slice
Location:
   /home/postgres/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cargo-pgrx-0.18.0/src/command/schema.rs:245
   ```

PR #2270, opened 2026-04-14
isolate test framework Postgres instances for parallel execution

eeeebbbbrrrr:
AI coding agents (Claude Code, Codex, etc.) commonly invoke `cargo pgrx test` for multiple tests or crates in parallel. This clobbers because every invocation starts a Postgres instance on the same deterministic port (32200 + pg_major_version) with the same PGDATA directory (target/test-pgdata/<version>/). Two instances can't share a port or a data directory, so the second invocation fails.

Fix both collision vectors:

Port: bind an ephemeral port (port 0) and hold the TcpListener open through the install + initdb window so nothing else can claim it. The listener is dropped immediately before pg_ctl start. The chosen port is injected into PgConfig via a new `test_port_override` field, which flows through all callers of `pg_config.test_port()` (start_pg, client, dropdb, createdb) without changing their call sites.

PGDATA: include the process ID in the directory name (target/test-pgdata/<version>-<pid>/) so every invocation gets its own data directory. The shutdown hook removes it after pg_ctl stop.

### DIFF
--- a/pgrx-tests/src/framework.rs
+++ b/pgrx-tests/src/framework.rs
@@ -43,17 +43,26 @@ static TEST_MUTEX: OnceLock<Mutex<SetupState>> = OnceLock::new();
 /// Set once during test framework initialization, read by every connection.
 static TEST_PORT: OnceLock<u16> = OnceLock::new();
 
-/// Bind an ephemeral port and return the listener that holds it open.
-///
-/// The caller must keep the returned `TcpListener` alive until Postgres
-/// is about to start, then drop it so Postgres can bind the same port.
-/// This prevents another process from claiming the port during the
-/// (potentially long) install + initdb window.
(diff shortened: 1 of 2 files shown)


pgrx-tests/src/framework.rs
```diff
@@ -701,6 +742,9 @@ fn start_pg(loglines: LogLines) -> eyre::Result<String> {
                 String::from_utf8(output.stderr).unwrap()
             );
         }
+
+        // clean up the per-invocation PGDATA directory
+        let _ = std::fs::remove_dir_all(&pgdata);
```
eeeebbbbrrrr, line 756: this is something we didn't used to do, and I've been arguing with myself about this and I think it's actually fine.

I have once or twice in the past manually started a postgres instance and pointed it at the PGDATA dir the test framework created, but probably never three times.  So it's probably better to cleanup after ourselves.

PR #2284, opened 2026-05-04
fix: provide the test stubs for pg_sys extern statics via pg_module_magic! macro

### LINKED ISSUE #2281: typetag::serde trait impls leak Postgres symbols into cargo test / coverage binaries (pgrx 0.18+)
## Summary

This is a follow-up to (and replacement for) #2229, which was filed against pgrx 0.16 and framed around `pgrx_embed`. pgrx 0.18 removed the `pgrx_embed` helper binary, so the original symptom is gone — but the underlying problem with `#[typetag::serde]` and Postgres symbols resurfaces one layer down, in any `cargo test` binary built from a pgrx extension crate.

Filing fresh so the bug report reflects current pgrx reality.

## Environment

- pgrx 0.18+ (no `pgrx_embed`)
- Reproduced on macOS (dyld flat-namespace) and Linux (glibc loader)
- Reproduced under `cargo test`, `cargo pgrx test`, and `cargo llvm-cov`

## Problem

An extension crate (`crate-type = ["cdylib"]`) that uses `#[typetag::serde]` on a trait impl — where the impl's methods call into `pgrx::pg_sys` — fails to load as a unit-test binary:

- **macOS**: `dyld: symbol not found in flat namespace '_BufferBlocks'`
- **Linux**: `undefined symbol: CurrentMemoryContext` / `ErrorContext` from the loader (even after `-Wl,--unresolved-symbols=ignore-all` lets `ld` finish)
- **`cargo llvm-cov`**: coverage instrumentation disables DCE, so the same failure reproduces even on targets where release builds happen to elide the unreachable paths

The failure occurs regardless of whether any `#[test]` actually references the typetag-annotated type. A bare `#[test] fn smoke() {}` elsewhere in the crate is enough to trigger it.

## Why it happens

`#[typetag::serde]` expands to `inventory::submit!`-style "life before main" registration that unconditionally references the trait impl methods. Those methods (in our case) transitively call into `pgrx::pg_sys`, which references extern globals like `CurrentMemoryContext`, `ErrorContext`, and `BufferBlocks`. Those globals are satisfied at runtime by the Postgres process image when the `.so` / `.dylib` is `dlopen`ed — but a `cargo test` binary is a standalone executable that is not loaded into Postgres, so those symbols are unresolved.

Since `cargo test` compiles the whole library into the test binary, and the typetag ctor registration is unconditional, the pg_sys symbols are pulled into the link/load closure whether or not tests exercise the code path. `#[cfg(not(test))]` on the impl mostly works but forces a stub impl to satisfy the upstream typetag-enabled trait, which is awkward.

## Current workaround

Function-pointer indirection initialized in `_PG_init`, so the typetag ctor only references a thin trampoline and the real pg_sys-touching implementation is reachable only at runtime via the installed pointer. Test binaries never call `_PG_init`, so DCE elides the real implementation and the pg_sys references with it.

```rust
type BuildFn = fn(serde_json::Value, u32) -> anyhow::Result<Box<dyn Query>>;
static BUILD_FN: OnceLock<BuildFn> = OnceLock::new();

pub fn init_builder() { BUILD_FN.get_or_init(|| real_impl); }

#[typetag::serde]
impl SomeTrait for MyType {
    fn build(&self, ...) -> ... {
        BUILD_FN.get().expect("call init in _PG_init")(...)
    }
}
```

Works but forces every author of a typetag-serialized extension type to understand link closures, ctor registration, and pgrx's dlopen model.

## What would help from pgrx

Open-ended — a few options that have come up:

1. A pgrx-sanctioned pattern for "this code path should not be reachable from a `cargo test` binary," documented in the README.
2. A way for extension authors to mark an item as "library-only, never link into the unit-test binary" that pgrx's build harness respects.
3. Guidance / tooling that makes pg_sys extern globals weak-linked (or stubbed) under `cfg(test)` so `cargo test` binaries can load even if they never exercise those paths.

## References

- Previous (stale) report: #2229 — will close in favor of this
- Downstream tracking issue: https://github.com/paradedb/paradedb/issues/3715
- Eric's diagnosis in the downstream issue: the mechanism is "Postgres symbols from `pg_sys::` have leaked into the rust test binary"; confirmed, though the leak path is unconditional typetag ctor registration rather than a test directly referencing the annotated type

peterkyle01:
Fixes https://github.com/pgcentralfoundation/pgrx/issues/2281

A pgrx extension is compiled as a `.so` or `.dylib` that gets `dlopen`'d into PostgreSQL at runtime. All the PostgreSQL symbols like `CurrentMemoryContext` and `BufferBlocks` are provided by the running PostgreSQL process so the `.so` doesn't need to define them.

When `cargo test` runs, Rust produces a standalone test executable  which is not loaded into PostgreSQL. Those symbols don't exist anywhere so the linker fails with `undefined symbol`.

Normally `--gc-sections` prunes all unreferenced `pg_sys` extern declarations. But `#[typetag::serde]` uses `inventory::submit!()` to register trait implementations via a "life before main" constructor and that constructor unconditionally references the trait impl methods, which reference `pg_sys` symbols, pulling them into the live-code set and defeating dead code elimination.

In this PR:

The fix adds `#[unsafe(no_mangle)]` stub definitions for the commonly-leaked `pg_sys` extern statics `CurrentMemoryContext`, `error_context_stack`, `PG_exception_stack`, `emit_log_hook`, `Log_error_verbosity`, `Log_line_prefix`, `Log_destination`, `Log_destination_string`, `syslog_sequence_numbers`, `syslog_split_messages` and `BufferBlocks` inside the `pg_magic_func!()` macro, gated behind `#[cfg(test)]`.

Why the macro?

* `#[cfg(test)]` only fires in the crate under test so dependencies like `pgrx-pg-sys` are never compiled with `cfg(test)` by Cargo so placing the stubs in the macro ensures they expand in the extension's own source, where `cfg(test)` actually works.
* Every extension already calls `pg_module_magic!()` 
* The `.so` build never includes the stubs (no `cfg(test)`) so PostgreSQL's real symbols are never shadowed.

### DIFF
--- a/pgrx-tests/src/lib.rs
+++ b/pgrx-tests/src/lib.rs
@@ -16,3 +16,50 @@ mod framework;
 pub use framework::*;
 #[cfg(feature = "proptest")]
 pub mod proptest;
+
+// ── Test stubs ──────────────────────────────────────────────────────────────
+// When a test binary links pgrx-pg-sys, all the extern "C" statics declared in
+// the generated bindings become undefined symbols.  Normally --gc-sections
+// prunes unreachable references, but #[typetag::serde] / inventory::submit!
+// ctors run before main() and pull pg_sys symbols into the live set.
+//
+// Because pgrx-tests is a dev-dependency, it is linked into test binaries but
+// NOT into the extension .so.  These #[unsafe(no_mangle)] statics satisfy the linker
+// without ever shadowing the real PostgreSQL symbols at runtime.
+// ─────────────────────────────────────────────────────────────────────────────
+mod stubs {
+    use pgrx::pg_sys;
+
+    #[unsafe(no_mangle)]
+    pub static mut CurrentMemoryContext: pg_sys::MemoryContext = std::ptr::null_mut();
+
+    #[unsafe(no_mangle)]
+    pub static mut error_context_stack: *mut pg_sys::ErrorContextCallback = std::ptr::null_mut();
+
+    #[unsafe(no_mangle)]
+    pub static mut PG_exception_stack: *mut pg_sys::sigjmp_buf = std::ptr::null_mut();
+
+    #[unsafe(no_mangle)]
+    pub static mut emit_log_hook: pg_sys::emit_log_hook_type = None;
+
+    #[unsafe(no_mangle)]
+    pub static mut Log_error_verbosity: std::ffi::c_int = 0;
+
+    #[unsafe(no_mangle)]
+    pub static mut Log_line_prefix: *mut std::ffi::c_char = std::ptr::null_mut();
+
+    #[unsafe(no_mangle)]
+    pub static mut Log_destination: std::ffi::c_int = 0;
+
+    #[unsafe(no_mangle)]
+    pub static mut Log_destination_string: *mut std::ffi::c_char = std::ptr::null_mut();
+
+    #[unsafe(no_mangle)]
+    pub static mut syslog_sequence_numbers: bool = false;
+
+    #[unsafe(no_mangle)]
+    pub static mut syslog_split_messages: bool = false;
+
+    #[unsafe(no_mangle)]
+    pub static mut BufferBlocks: *mut std::ffi::c_void = std::ptr::null_mut();
+}
--- a/pgrx/src/lib.rs
+++ b/pgrx/src/lib.rs
@@ -354,6 +354,49 @@ macro_rules! pg_magic_func {
             // return the magic
             &MY_MAGIC.0
         }
+
+        // ── Test stubs ──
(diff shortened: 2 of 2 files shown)


eeeebbbbrrrr:
Thanks but this is not the right way to handle this.

If your project has normal rust `#[test]` tests that reference `pg_sys::` symbols, either directly or indirectly, that is a bug with your project, it is not something pgrx should accommodate.  

It is pgrx's job, as much as it can, to not let its users do things they shouldn't do.

peterkyle01:
Okay thanks @eeeebbbbrrrr for the review

PR #2289, opened 2026-05-09
cargo-pgrx: use zip instead of zip-extract

Natureshadow:
While packaging `pgrx` for Debian in order to allow Postgres extensions written with it to be added to `apt.postgresql.org` and Debian itself, we found that `zip-extract` has been deprecated a year ago.

This change trivially replaces it with the `zip` crate recommended by the `zip-extract` maintainers.


eeeebbbbrrrr:
We got a PR for this twice and I merged theirs before seeing yours.  So I'm closing this.  I'll make sure to mention you in the release notes.  :)

cbandy:
wah! same. sorry I missed this!

PR #2291, opened 2026-05-15
cargo-pgrx: build Postgres in parallel during `cargo pgrx init`

philippemnoel:
## Summary

`cargo pgrx init` already takes `-j` / `--jobs` and uses it to size a `jobslot::Client`, which exposes the jobserver auth to child `make` processes via `MAKEFLAGS`. However, neither `make world-bin` nor `make install-world-bin` was given `-j` on its own argv, so GNU make ignored the jobserver and ran serially. Result: `cargo pgrx init` built Postgres single-threaded even on big hosts.

This PR:

- Extracts the existing job-count resolution into `Init::resolved_jobs()` (`--jobs` -> `available_parallelism()` -> `1`)
- Reuses it to seed the jobserver (no behavior change there)
- Passes `-jN` to both `make` invocations so GNU make actually parallelizes

No new flags; the existing `--jobs`/`-j` keeps working and now has an observable effect on the Postgres build, not just the jobserver token pool.

## Test plan

- [x] `cargo check -p cargo-pgrx` passes
- [x] `cargo pgrx init` rebuild is measurably faster on a multi-core host (verified locally by author)
- [x] `cargo pgrx init -j 1` produces a serial build (verified locally by author)

### DIFF
--- a/cargo-pgrx/src/command/init.rs
+++ b/cargo-pgrx/src/command/init.rs
@@ -104,10 +104,7 @@ impl CommandExecute for Init {
     #[tracing::instrument(level = "error", skip(self))]
     fn execute(self) -> eyre::Result<()> {
         self.jobserver
-            .set(
-                jobslot::Client::new(self.resolved_jobs())
-                    .expect("failed to create jobserver"),
-            )
+            .set(jobslot::Client::new(self.resolved_jobs()).expect("failed to create jobserver"))
             .unwrap();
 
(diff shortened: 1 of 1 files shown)


philippemnoel:
The "verify package can build / ubuntu" run will be fixed once https://github.com/pgcentralfoundation/pgrx/pull/2292 is merged

eeeebbbbrrrr (approved):

PR #2294, opened 2026-05-28
Use zip instead of deprecated zip-extract

cbandy:
A small cleanup I noticed while reviewing dependencies.

The zip-extract crate was [deprecated last year](https://crates.io/crates/zip-extract/0.4.1) because the upstream zip crate improved its API.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -478,7 +478,7 @@ dependencies = [
  "url",
  "uuid",
  "which",
- "zip-extract",
+ "zip",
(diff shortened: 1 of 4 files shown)


cbandy:
📝 nightly failing due to:
- https://github.com/bytecodealliance/rustix/issues/1620
- https://github.com/rust-lang/rust/pull/155433

eeeebbbbrrrr (approved):

PR #2298, opened 2026-06-01
refactor remove redundant casts and improve safety in varlena functions

isdaniel:
## Summary

  - Remove redundant identity casts in varlena functions (`ptr as *const pg_sys::varlena` where ptr is already that type, `as
  usize` where return is already usize, `as *mut pg_sys::varlena` where bytea = varlena)
  - Simplify `PgBox::as_ptr()` — use `NonNull::as_ptr()` directly instead of clone + as_mut (NonNull is Copy)
  - Add `is_empty()` methods to `List`, `ListHead`, `PgHeapTuple`, and `NullableContainer` trait (clippy: len-without-is-empty)
  - Add unit tests for `encode_vlen_4b`, `encode_vlen_1b`, `List::Nil`, and `PgBox` pointer semantics

  ## Motivation

  - Identity casts add noise and can mask real type mismatches during refactoring
  - `NonNull::clone()` is semantically misleading (it's Copy, clone is a no-op)
  - Missing `is_empty()` on types with `len()` triggers clippy warnings and forces callers to write `x.len() == 0`

  ## Test plan

  - [x] `cargo check --no-default-features --features pg17` passes
  - [x] `cargo check --no-default-features --features "pg17 cshim"` passes
  - [x] `cargo test --no-default-features --features pg17 -p pgrx --lib` passes

### DIFF
--- a/pgrx/src/datum/varlena.rs
+++ b/pgrx/src/datum/varlena.rs
@@ -397,7 +397,7 @@ pub unsafe fn cbor_decode<'de, T>(input: *mut pg_sys::varlena) -> T
 where
     T: Deserialize<'de>,
 {
-    let varlena = pg_sys::pg_detoast_datum_packed(input as *mut pg_sys::varlena);
+    let varlena = pg_sys::pg_detoast_datum_packed(input);
     let len = varsize_any_exhdr(varlena);
     let data = vardata_any(varlena);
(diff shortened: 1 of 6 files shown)


eeeebbbbrrrr (approved):
Thank you

PR #2299, opened 2026-06-01
Fix docs.rs build by localizing docsrs cfg

eeeebbbbrrrr:
docs.rs was invoking rustdoc with `--cfg docsrs` from package metadata. Cargo applies those rustflags to the whole build graph, not only pgrx/pgrx-pg-sys, so third-party crates also observed cfg(docsrs). bindgen's experimental feature pulled annotate-snippets 0.11.5, whose docsrs cfg enables feature(doc_auto_cfg), removed in rustc 1.92 and rejected by the current docs.rs nightly. That caused pgrx docs build #3444463 to fail before rustdoc reached our crate.

Only `pgrx-pg-sys` needs `cfg(docsrs)`, and only to select the checked-in bindings instead of running bindgen/Postgres discovery. Have its build script emit `cargo:rustc-cfg=docsrs` when `DOCS_RS=1`, plus check-cfg and rerun tracking, and remove global docs.rs rustc/rustdoc args from member manifests.

Also stop enabling bindgen's experimental feature. `pgrx-bindgen` uses the stable builder APIs needed for `wrap_static_fns`, and dropping experimental removes `annotate-snippets/unicode-width` from the lockfile entirely.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -89,16 +89,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
-name = "annotate-snippets"
-version = "0.11.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


eeeebbbbrrrr:
Going to commit this and the next pgrx release ought to work on docs.rs.

Issue #2234, opened 2025-12-27
Add `PBox<'_, Bytea>`


workingjubilee:
This is the most common reason someone wants an implementation of `RetAbi` for Datums.

The reasons why Datums should not simply have RetAbi implemented for them are somewhat complex and start at "well, for one, we have two Datum types".

PR #2301, opened 2026-06-03
feat: add Bytea<'fcx> argument type with zero-copy return

isdaniel:
## Summary

- Add `Bytea<'fcx>` — a lifetime-safe, zero-copy wrapper for bytea arguments that handles detoasting at the boundary
- Integration tests return, Bytea<'_> argument, roundtrip, and edge cases

for #2234


eeeebbbbrrrr:
Can you please elaborate on your motivation behind this?

isdaniel:
Hi @eeeebbbbrrrr thank you for your commnet, here is motivation of this PR:

`bytea` is mapped to `&[u8]` / `Vec<u8>`. While this covers the common case, **the return path is never zero-copy**, even when the bytes already live in a Postgres-allocated varlena, and `PgBox<pg_sys::bytea>` cannot be returned at all.

### 1. Return path would do copy

`IntoDatum for &[u8]` unconditionally performs:

- `pg_sys::palloc(len + VARHDRSZ)` — allocate a fresh varlena in `CurrentMemoryContext`
- `set_varsize_4b(...)` — write the varlena header
- [`std::ptr::copy_nonoverlapping(self.as_ptr(), ..., self.len())` ](https://github.com/pgcentralfoundation/pgrx/blob/ca78cb9070c91c60f8fd381a5981206c88744f61/pgrx/src/datum/into.rs#L371)— memcpy the entire payload

`PgBox<T>` already has zero-copy plumbing (`BoxRet`, `IntoDatum`). However, `#[pg_extern]` macro expansion requires `T:
SqlTranslatable` to emit the SQL declaration, and `pg_sys::varlena` had no such impl — so `PgBox<pg_sys::bytea>` was rejected at compile time even though the runtime path was ready.

### 2. Forwarding an input bytea to another PG C function

`FromDatum for &[u8]` drops the original `*mut varlena` and returns only a detoasted byte slice. To pass the value into another PG C API (`DirectFunctionCall`), we must re-encode through the same palloc + memcpy path above.

- **Before:** `&[u8]` loses the varlena pointer; must re-encode.
- **After:** `Bytea<'_>` keeps the varlena pointer (`as_varlena_ptr()`) while still providing zero-copy `Deref<[u8]>` access, and ties the borrow to the function-call lifetime `'fcx`.

please let me know if you have any idea or further insights, thanks.

eeeebbbbrrrr (changes requested):
I think I'd also like to see a test that exercises a pg_extern function that roundtrips a Bytea<'fcx>, which would be zero-copy:

```rust
#[pg_extern]
fn bytea_roundtrip<'fcx>(input: Bytea<'fcx>) -> Bytea<'fcx> {
   input
}
```

With a #[pg_test] that uses SPI to call that function.


What I don't like about returning `PgBox<pg_sys::bytea>` is that there's nothing that stops the wrapped pointer from being invalid before it's returned from the function.  UAF is a real thing when doing this and there's not really anything the compiler can do to stop you.  PgBox is wildly unsafe, but I think I understand why you'd want it.

isdaniel:
Hi @eeeebbbbrrrr 

Thanks for the suggestion and for laying out the safety concern, I totally agree with the UAF reasoning on `PgBox<pg_sys::bytea>`,
 
I already did some change and modifications from my PR, please let me know if you have any idea or further insights, thanks.

eeeebbbbrrrr:
Does this now give you the abilities you need?

I do understand that `PgBox<pg_sys::bytea>` could be useful.  I just wanted to point out that it can be wildly unsafe because the compiler can't help.

isdaniel:
Hi @eeeebbbbrrrr 

Yes — `Bytea<'fcx>` covers what I need, the goal was true zero-copy on (parameter in and return out) sides, Currently `&[u8]` / `Vec<u8>` returns can't achieve this  `IntoDatum for &[u8]` has to a fresh varlena and `memcpy` the bytes in, which dominates cost for large payloads. `Bytea<'fcx>` hands the original varlena  pointer straight back as a Datum, so the return path is allocation and copy-free.

Since this PR already delivers the zero-copy bytea return path I was after, therefore I don't think it's strictly needed to support `PgBox<pg_sys::bytea>`. Happy to revisit and discuss adding it later if a concrete use case comes up that `Bytea<'fcx>` genuinely can't cover, but for now keeping the surface area limited to the lifetime-safe API seems like the better default.

eeeebbbbrrrr:
Thanks for your contribution, @isdaniel!  merging now.

PR #2308, opened 2026-06-07
implement retry mechanism for cargo pgrx installation in workflows more robust

isdaniel:
Pipeline sometime occurred networking transient issue like below two type issues.

* https://github.com/pgcentralfoundation/pgrx/actions/runs/27069910421/job/79897352358
* https://github.com/pgcentralfoundation/pgrx/actions/runs/27055949833/job/79860452469

implement retry mechanism for cargo pgrx installation in workflows more robust. use https://github.com/nick-fields/retry

### DIFF
--- a/.github/workflows/package-test.yaml
+++ b/.github/workflows/package-test.yaml
@@ -90,7 +90,12 @@ jobs:
           pg_config --version
 
       - name: Install cargo pgrx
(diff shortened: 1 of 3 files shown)


eeeebbbbrrrr:
I appreciate the contribution but I prefer to keep complexity out of CI.  

It's terrible that GH has so many issues but it's a zero-sum game to try and account for all of them.

Issue #2148, opened 2025-10-01
cargo pgrx start --port 5432


daamien:
hey 👋 

In a discussion with the Fedora packager of my extension, we're talking about how to make the RPM maintainer life easier ans it prompted the need for a `--port` option on the  `cargo pgrx start` command. Basically the packager needs to compile the extension against the Postgres build provided by the distribution ( instead of using the one built with`cargo pgrx init` ) and for some arcane pg_regress tests (testing logical replication for instance), it is way easier to make the instance listen on the port defined at compile-time (5432 for most distribution builds). However it seems like `cargo pgrx start` will always make the instance listen on port `28800 + PG_MAJOR_VERSION`...

Maybe I'm missing something, but if I'm correct : Would it be acceptable to add a `--port` option to `cargo pgrx start` ?

I understand that the answer might be « Don't use  `cargo pgrx start` if you didn't build postgres with `cargo pgrx init`» 🙂  but at the same time since there's an ongoing effort to integrate regression tests with `cargo pgrx regress`, maybe providing a bit of flexibility to RPM maintainers would make sense too  ?

PR #2309, opened 2026-06-07
feat(cargo-pgrx): add --port to `start` and `run` 

isdaniel:
## Motivation

Adds `--port <PORT>` to `cargo pgrx start` and `cargo pgrx run`, overriding the default `28800 + PG_MAJOR_VERSION`. Motivating use case (from the #2148 )

## How to use

 ```bash
  # Start pg17 on the distro-default port instead of 28817
  cargo pgrx start pg17 --port 5432

  # Run an extension and connect via psql on a custom port
  cargo pgrx run pg17  --port 5432

 # Stop / status are unchanged
  cargo pgrx stop pg17
```

### DIFF
--- a/cargo-pgrx/src/command/run.rs
+++ b/cargo-pgrx/src/command/run.rs
@@ -56,6 +56,9 @@ pub(crate) struct Run {
     install_only: bool,
     #[clap(long)]
     valgrind: bool,
+    /// Port to listen on (overrides the default of 28800 + PG_MAJOR_VERSION)
+    #[clap(long)]
+    port: Option<u16>,
 }
 
 impl From<&Regress> for Run {
@@ -73,6 +76,7 @@ impl From<&Regress> for Run {
             pgcli: false,
             install_only: false,
             valgrind: regress.valgrind,
+            port: None,
         }
     }
 }
@@ -106,6 +110,10 @@ impl Run {
             Some(&mut self.features),
             true,
         )?;
+        let pg_config = match self.port {
+            Some(port) => pg_config.with_port_override(port),
+            None => pg_config,
+        };
 
         let dbname = match &self.dbname {
             Some(dbname) => dbname.clone(),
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by eeeebbbbrrrr** [comment]
I'm not sure how I feel about this one.  It kinda violates the principle that the various `cargo-pgrx` subcommands "just work".  It feels wrong to me that `cargo pgrx run`, etc just end up being pass-throughs to some random postgres server.

@daamien the bug referenced here is originally from you, what do you think about this?  I know that bug is almost a year old now.  😬 

One thought I just had is what if we enhanced `~/.pgrx/config.toml` to allow either a port number per postgres version *or* perhaps a full database connstr?

The thing that I want to make sure is that when someone runs `cargo pgrx run` that not only are they getting the extension compiled/installed to the right place but are also connecting to that same database.  Just adding `--port` as an argument doesn't get us that.

Also, we do have `cargo pgrx init --baseport` already.  I use that from time-to-time with my clankers and worktrees and it's pretty helpful.

**Comment by daamien** [comment]
Hi @isdaniel , @eeeebbbbrrrr 

First of all thanks a lot for looking into this ! 

There is sttill the need for changing the listening port of an instance, although I woudl'nt qualify this as a bug. More a like an arcane feature request :)

I was not aware of the `--base-port` , it may be sufficient for the need 

I haven't tested every thing but doing somthing like this should be enough...

``` 
# 5432 - 15 = 5417
cargo pgrx init --pg15=/usr/lib/postgresql/15/bin/pg_config --base-port=5417

cargo pgrx start pg15
    Starting Postgres v15 on port 5432
```

daamien:
> I haven't tested every thing but doing somthing like this should be enough...
> 
> ```
> # 5432 - 15 = 5417
> cargo pgrx init --pg15=/usr/lib/postgresql/15/bin/pg_config --base-port=5417
> 
> cargo pgrx start pg15
>     Starting Postgres v15 on port 5432
> ```

To answer myself  : it seems like the `--base-port` option is broken

https://github.com/pgcentralfoundation/pgrx/issues/2339

isdaniel:
The `--base-port` parameter can satisfy my scenario and work good to me, I think we don't need current PR so far

PR #2311, opened 2026-06-07
Configure dependabot to update GitHub Actions

cbandy:
Workflow runs are currently annotated with the following warning:

    Node.js 20 actions are deprecated.

Rather than figure this out once manually, let GitHub's tools handle it on a regular schedule.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,24 @@
+# documentation: https://docs.github.com/code-security/dependabot/dependabot-version-updates
+# schema documentation: https://docs.github.com/code-security/dependabot/working-with-dependabot/dependabot-options-reference
+# $schema: https://json.schemastore.org/dependabot-2.0.json
+---
+version: 2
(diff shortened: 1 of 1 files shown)


cbandy:
You may also need to enable "Dependabot version updates" in [security settings](https://github.com/pgcentralfoundation/pgrx/settings/security_analysis). When I did that on my fork, Dependabot opened this PR: https://github.com/cbandy/pgrx/pull/3

eeeebbbbrrrr:
interesting.  I'm currently running the repo's `./upgrade-deps.sh` script and will PR that.

I am not sure I want dependabot nagging?  Let me think on this.

eeeebbbbrrrr:
I actually don't have permissions to modify the repo settings, so let me talk to the man who does, then sure, we can merge this.

cbandy:
I believe by having the config file, Dependabot will only evaluate what's configured. Just GitHub Actions here.

PR #2312, opened 2026-06-07
upgrade dependencies and rust edition to 2024

eeeebbbbrrrr:
This upgrades all our dependencies using our `./upgrade-deps.sh` script.

It also makes sure the whole project is on `edition = "2024"`.  This wasn't super clean, apparently.

It also fixes a bug with the pgthread example where it kept stale compiled artifacts around causing linker errors.


pgrx-examples/numeric/Cargo.toml
```diff
@@ -31,7 +31,7 @@ pg_test = []
 
 [dependencies]
 pgrx = { path = "../../pgrx" }
-rand = "0.10.1"
+rand = "0.8.6"
```
eeeebbbbrrrr: where I'm from, 10 > 8.  

Strange.

cbandy: Maybe to align with the _very_ old `proptest` it selected?

Issue #2316, opened 2026-06-08
#[pg_test(expected = ...)] corrupts raw-string literals and panics on malformed input


isdaniel:
## Summary

The #[pg_test] attribute parses its expected = "..." argument with a hand-rolled TokenTree walker instead of syn. This causes bug:

Raw-string corruption — expected values written as raw strings (needed whenever the expected Postgres error message contains a ") are silently mangled, so the test never matches the real error.

## pre-prepare

```
/// Raises an error whose message contains a literal double-quote: foo "bar"
#[pg_extern]
fn raise_quoted_error() -> &'static str {
    error!(r#"foo "bar""#);
}
```

  Bug — raw-string corruption
```
#[pg_test(error = r#"foo "bar""#)]
fn bug1_raw_string_expected_is_corrupted() {
	Spi::get_one::<&str>("SELECT raise_quoted_error()").unwrap();
}
```

Result:

```
Client Error:
foo "bar"
postgres location: lib.rs:8
rust location: <unknown>


note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

failures:
    tests::pg_bug1_raw_string_expected_is_corrupted
```

The real error message from `raise_quoted_error()` is foo "bar", so the test fails with a mismatch even though the author wrote the correct expected string.

Environment

  - pgrx: 0.18.1
  - OS: Linux

isdaniel:
fix via #2317

PR #2317, opened 2026-06-08
refactor: unify #[pg_extern] attribute parsing via syn

isdaniel:
Motivation

* Collapses three diverging code paths that handled #[pg_extern] / #[pg_test] attributes into a single syn-based parser.
* Fixes latent bug in #[pg_test] argument parsing exposed by raw and malformed string literals.

Bugs fixed

Raw-string corruption in #[pg_test(expected = ...)] — the old walker did literal.to_string()[1..len-1], which strips one byte each side. For r#"foo "bar""# it produced #"foo "bar""# (delimiters leaked into the value), making the runtime error-message comparison silently mismatch.

for #2316

### DIFF
--- a/pgrx-examples/errors/src/lib.rs
+++ b/pgrx-examples/errors/src/lib.rs
@@ -72,7 +72,7 @@ mod tests {
     use pgrx::prelude::*;
 
     /// Raises an error whose message contains a literal double-quote: foo "bar"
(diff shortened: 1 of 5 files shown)


eeeebbbbrrrr (approved):
good cleanup, thanks.

PR #2319, opened 2026-06-10
feat: buffer with safe RAII wrapper for Postgres BufferManager

isdaniel:
## Motivation

Raw `pg_sys::ReadBuffer` / `LockBuffer` / `ReleaseBuffer` FFI. Every early return, or panic path must manually pair its own unpin / unlock — miss one and you get a pin leak or UB. Postgres' `AtEOXact_Buffers` cleans up after the fact; 

- **Fewer production incidents.** Pin leaks, double-locks, and use-after-unlock , three of the most common BufferManager pitfalls, become impossible in safe Rust.
- **Lower onboarding cost.** Extension authors get a wrapper that matches the existing `PgRelation` shape: `unsafe` only at construction, everything else borrow-checked.  No need to learn BufferManager pairing rules to be productive.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2083,6 +2083,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "page_inspect"
+version = "0.0.0"
+dependencies = [
+ "pgrx",
+ "pgrx-tests",
+]
+
+[[package]]
 name = "page_size"
 version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


eeeebbbbrrrr:
Hate to do this to you, but I don't think this is something that pgrx needs.

Postgres' buffer manager is quite complex and covers a very large surface area.  I actually don't believe it's reasonable for pgrx to try and make any of that safe.

The reason is that the way one uses Buffers and Pages and such is going to be different from use-case to use-case.  And while maybe it pushes the act of safely wrapping these things downstream to individual extension developers, I think that's a better approach because then they can decide for themselves what they need to be safe and what they don't and how those decisions impact their design. 

Even if I thought there is a single safe API that we could create around the buffer manager, I don't know what it looks like.

With my sincere appreciation, I'm gong to close this PR.

isdaniel:
Hi @eeeebbbbrrrr noted, thank for your feedback!

PR #2323, opened 2026-06-15
Add #[pg_guc_hook] macro for simple GUC hooks

cbandy:
This adds a `#[pg_guc_hook]` attribute macro that wraps a Rust function to do the unsafe work of converting to and from `GucValue` types. With this macro, a check hook can return a `GucCheckError` type to populate the error code, detail, and hint when rejecting a value.
    
The signatures of `GucRegistry::define_*_guc_with_hooks` do not change.

### DIFF
--- a/pgrx-macros/src/lib.rs
+++ b/pgrx-macros/src/lib.rs
@@ -1430,6 +1430,197 @@ pub fn pgrx(_attr: TokenStream, item: TokenStream) -> TokenStream {
 }
 
 /**
+Declare a function as a GUC hook. This takes one argument: `show`, `check`, or `assign`.
+
+The first parameter of `check` and `assign` hooks must implement `pgrx::guc::GucValue`.
+
+Examples:
+```rust,ignore
+#[pg_guc_hook(show)]
+fn my_show_hook() -> String {
+    "CUSTOM_VALUE".to_string()
+}
+
+#[pg_guc_hook(check)]
+fn my_check_hook(newval: i32) -> Result<(), GucCheckError> {
+    // accept or reject newval
+}
+
+#[pg_guc_hook(assign)]
+fn my_assign_hook(newval: i32) {
+    // do more now that every change is accepted
+}
+```
+
+# Check Hooks
+
+This macro adapts check hooks of multiple forms.
+
+The simplest of these takes a single argument and returns a `bool`.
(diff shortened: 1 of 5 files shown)


pgrx-macros/src/lib.rs
```diff
@@ -1429,6 +1429,136 @@ pub fn pgrx(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
+
+Examples:
+```rust,ignore
+#[pg_guc_hook(show)]
+fn my_show_hook() -> String {
+    "CUSTOM_VALUE".to_string()
+}
+
+#[pg_guc_hook(check)]
+fn my_check_hook(newval: i32, source: pg_sys::GucSource::Type) -> Result<(), GucCheckError> {
```
cbandy: 🤔 I've never used this second argument. It would be nice if it were optional.

cbandy: Done.

pgrx-macros/src/lib.rs
cbandy, line 1: 📝 I have almost zero experience writing macros.

pgrx-macros/src/lib.rs
```diff
@@ -1429,6 +1429,151 @@ pub fn pgrx(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
+                    source: ::pgrx::pg_sys::GucSource::Type,
+                ) -> bool {
+                    #[inline(always)]
+                    #inner_func
+
+                    debug_assert!(!newval.is_null());
+                    let value = unsafe { <#arg_type as ::pgrx::guc::GucValue>::from_raw(*newval) };
+                    match #invoke_inner {
+                        Ok(()) => true,
+                        Err(err) => {
```
cbandy: I'm going to see if I can detect and allow the user's check hook to return bool. So vaguely something like:

```rust
fn check(next: T) -> bool; // direct return to postgres
fn check(next: T) -> Result<>; // current approach; Ok() => true, Err() => messages + false
```

cbandy: Done.

pgrx-macros/src/lib.rs
eeeebbbbrrrr: can't say that now!

eeeebbbbrrrr (approved):

eeeebbbbrrrr:
Thanks and sorry for the delay.

PR #2337, opened 2026-06-23
fix(pgrx-tests): respect configured PostgreSQL GUCs

### LINKED ISSUE #2151: Setting `log_min_messages` in postgresql_conf_options does not work
Hi,

setting `log_min_messages` via `postgresql_conf_options` does not work:

```rust
pub fn postgresql_conf_options() -> Vec<&'static str> {
    vec!["log_min_messages = debug1"]
}
```

Is it maybe overshadowed by the [`client()`](https://github.com/pgcentralfoundation/pgrx/blob/7efdd7bdf5cdce39c8d63a18da6f86de8e30ab6f/pgrx-tests/src/framework.rs#L119) function running

```rust
query_wrapper(Some("SET log_min_messages TO 'INFO';".to_string()), None, |query, _| {
    client.simple_query(query.unwrap().as_str())
})
```

always before a test is run? Is there a way to still set the option for a test?

0xPoe:
close https://github.com/pgcentralfoundation/pgrx/issues/2151

pgrx-tests sets a few session-level logging defaults when creating a client. In this PR, I have baked pgrx's defaults into `postgresql.conf`. Now they can be overridden by the user's `postgresql_conf_options`.

I just added a simple unit test instead of an integration test, as there is no existing integration test suite in the `pgrx-tests` crate.

### DIFF
--- a/pgrx-tests/src/framework.rs
+++ b/pgrx-tests/src/framework.rs
@@ -314,21 +314,6 @@ pub fn client() -> eyre::Result<(postgres::Client, String)> {
         None => Err(eyre!("Failed to obtain a client Session ID from Postgres"))?,
     };
 
-    query_wrapper(Some("SET log_min_messages TO 'INFO';".to_string()), None, |query, _| {
-        client.simple_query(query.unwrap().as_str())
-    })
-    .wrap_err("Postgres Client setup failed to SET log_min_messages TO 'INFO'")?;
-
-    query_wrapper(Some("SET log_min_duration_statement TO 1000;".to_string()), None, |query, _| {
-        client.simple_query(query.unwrap().as_str())
-    })
-    .wrap_err("Postgres Client setup failed to SET log_min_duration_statement TO 1000;")?;
-
-    query_wrapper(Some("SET log_statement TO 'all';".to_string()), None, |query, _| {
-        client.simple_query(query.unwrap().as_str())
-    })
-    .wrap_err("Postgres Client setup failed to SET log_statement TO 'all';")?;
-
     Ok((client, session_id))
 }
 
@@ -556,17 +541,7 @@ fn modify_postgresql_conf(
     socket_dir: PathBuf,
     postgresql_conf: Vec<&'static str>,
 ) -> eyre::Result<()> {
-    let mut contents = String::new();
-
-    contents.push_str("log_line_prefix='[%m] [%p] [%c]: '\n");
-    contents.push_str(&format!(
-        "unix_socket_directories = '{}'\n",
-        socket_dir.display().to_string().replace("\\", "\\\\")
-    ));
-    for setting in postgresql_conf {
-        contents.push_str(&format!("{setting}\n"));
-    }
-
+    let contents = postgresql_conf_contents(&socket_dir, &postgresql_conf);
     let postgresql_auto_conf = pgdata.join("postgresql.auto.conf");
 
     if let Some(runas) = get_runas() {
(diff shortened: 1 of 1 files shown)


pgrx-tests/src/framework.rs
```diff
@@ -314,22 +314,72 @@ pub fn client() -> eyre::Result<(postgres::Client, String)> {
...
-    .wrap_err("Postgres Client setup failed to SET log_min_duration_statement TO 1000;")?;
+    .wrap_err(error_message)?;
 
-    query_wrapper(Some("SET log_statement TO 'all';".to_string()), None, |query, _| {
-        client.simple_query(query.unwrap().as_str())
+    Ok(())
+}
+
+// Returns whether pg_settings reports the GUC source as Postgres' built-in default.
+fn client_guc_uses_default_source(
```
0xPoe: Not sure it’s worth making this query run only once for all settings.

0xPoe (commented):
🔢 Self-check (PR reviewed by myself and ready for feedback)

- [x] Code compiles successfully

- [x] Unit tests added

- [x] No AI-generated elegant nonsense in PR.

- [x] Comments added where necessary

- [x] PR title and description updated

- [x] PR size is reasonable

eeeebbbbrrrr:
Rather than issuing SETs ourselves I think we should bake pgrx's defaults into "postgresql.conf" and allow what an extension specifies in its `fn postgresql_conf_options()` to override our defaults.

Thoughts?  Wanna rework this?

0xPoe:
> Rather than issuing SETs ourselves I think we should bake pgrx's defaults into "postgresql.conf" and allow what an extension specifies in its `fn postgresql_conf_options()` to override our defaults.
> 
> Thoughts? Wanna rework this?

Makes sense. I'll rework this by moving the pgrx logging defaults into the generated Postgres config.

0xPoe:
I have tested locally:
1. Setup a new pgrx project:
```sh
WORK="$(mktemp -d)"
cd "$WORK"

cargo install --git "https://github.com/0xPoe/pgrx.git" --branch "poe-patch-guc" cargo-pgrx --root "$WORK/.tools"

"$WORK/.tools/bin/cargo-pgrx" pgrx new pgrx_issue2151_check
cd pgrx_issue2151_check
```
2. Update the Cargo.toml to use this PR:
```toml
[patch.crates-io]
pgrx = { git = "https://github.com/0xPoe/pgrx.git", branch = "poe-patch-guc" }
pgrx-tests = { git = "https://github.com/0xPoe/pgrx.git", branch = "poe-patch-guc" }
```
3. Replace `src/lib.rs` with this full file:
```sh
cat > src/lib.rs <<'EOF'
use pgrx::prelude::*;

::pgrx::pg_module_magic!(name, version);

#[pg_extern]
fn hello_pgrx_issue2151_check() -> &'static str {
    "Hello, pgrx_issue2151_check"
}

#[cfg(any(test, feature = "pg_test"))]
#[pg_schema]
mod tests {
    #[pgrx::pg_test]
    fn postgresql_conf_options_controls_log_min_messages() {
        let value = pgrx::Spi::get_one::<String>("SELECT current_setting('log_min_messages');")
            .expect("SPI should succeed")
            .expect("current_setting should return a value");

        assert_eq!(value, "debug1");
    }
}

/// This module is required by `cargo pgrx test` invocations.
/// It must be visible at the root of your extension crate.
#[cfg(test)]
pub mod pg_test {
    pub fn setup(_options: Vec<&str>) {}

    #[must_use]
    pub fn postgresql_conf_options() -> Vec<&'static str> {
        vec!["log_min_messages = debug1"]
    }
}
EOF
```
5. Run the test:
```sh
CARGO_PGRX_TEST_PGDATA="$WORK/pgdata" \
cargo test --no-default-features --features pg18 --lib \
postgresql_conf_options_controls_log_min_messages -- --nocapture
```
6. Test result:
```sh
test tests::pg_postgresql_conf_options_controls_log_min_messages ... ok
```

pgrx-tests/src/framework.rs
```diff
@@ -595,6 +570,29 @@ fn modify_postgresql_conf(
...
+    contents.push_str("log_line_prefix='[%m] [%p] [%c]: '\n");
+    contents.push_str(&format!(
+        "unix_socket_directories = '{}'\n",
+        socket_dir.display().to_string().replace("\\", "\\\\")
+    ));
+    // PostgreSQL applies repeated settings in file order, so put pgrx's
+    // defaults first and let extension options override them below.
+    for setting in PGRX_POSTGRESQL_CONF_DEFAULTS.iter().chain(postgresql_conf) {
+        contents.push_str(&format!("{setting}\n"));
+    }
```
0xPoe, line 591: Technically, we could parse the setting to avoid duplicating the same config from the default and user override values. However, I’m not sure it’s worth the extra complexity, since PostgreSQL can handle this correctly by applying the settings in order.

0xPoe (commented):
🔢 Self-check (PR reviewed by myself and ready for feedback)

pgrx-tests/src/framework.rs
eeeebbbbrrrr: yeah, I don't think we should silently skip work the user/code asked us to do, even if it's duplicate work.  I think you're right to leave this as-is.

eeeebbbbrrrr (approved):

PR #2341, opened 2026-06-25
Add utils/datum.h bindings + generic_agg example

isdaniel:
## Motivation

These are the canonical, type-length-aware way to copy/compare a `Datum` when the concrete type is only known at runtime (`anyelement`, polymorphic aggregates, generic state). Today an extension author has to hand-redeclare them `extern "C"` — easy to get subtly wrong (`typLen` is C `int`, not `int16`) and missing pgrx's `#[pg_guard]`, so a `palloc` OOM longjmp becomes UB.

## What this does

- Add `#include "utils/datum.h"` to all seven wrappers (`pg13`–`pg19`).
- Add `pgrx-examples/generic_agg`: a polymorphic `count_changes(anyelement)` aggregate that needs `datumCopy` (keep a by-ref value alive across transition calls) + `datumIsEqual` (byte-compare type-erased values), with `#[pg_test]`s covering text/numeric (by-ref), int4 (by-value), NULL, and empty input.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1187,6 +1187,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "generic_agg"
+version = "0.0.0"
+dependencies = [
+ "pgrx",
+ "pgrx-tests",
+]
+
+[[package]]
 name = "getrandom"
 version = "0.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


eeeebbbbrrrr:
I put up the same fix for the flaky tests in #2342, so if you'll remove it from this PR I'd appreciate it.

isdaniel:
Hi @eeeebbbbrrrr noted, I already removed that commit from my branch, and also change the PR comment as well

PR #2346, opened 2026-07-01
Make &FlatArray<Text> serializable 

isdaniel:
## Motivation

The borrowed flat-array API supported scalars and CStr, but text arrays had a long-stubbed `&FlatArray<&str>` placeholder unsound, since &str excludes the varlena header and breaks iteration stride. No way to serialize a borrowed `text[]`.

## Changes

- Add Text varlena element: strides by full varlena size (varsize_any), mirroring the proven CStr element. 
- Enable borrow_serde_serialize_array fn + its pg_test.
-  rustc caps that hint list at 8 entries then prints and $N others, and the list is alphabetically sorted. Inserting `Text` (alphabetically between Point and Time) pushed the window forward by one, so `TimestampWithTimeZone` (previously 8th) dropped off into and $N others. The original fix only added Text (9 lines) without  removing TimestampWithTimeZone, so it still didn't match the compiler's truncated 8-line output.

### DIFF
--- a/pgrx/src/datum/borrow.rs
+++ b/pgrx/src/datum/borrow.rs
@@ -146,10 +146,12 @@ unsafe impl Element for ffi::CStr {
 pub struct Text([u8]);
 
 impl Text {
-    /// The string contents (UTF-8), excluding the varlena header.
+    /// The string contents, excluding the varlena header.
+    ///
+    /// Uses pgrx's encoding-aware conversion: when the database encoding isn't guaranteed UTF-8 (e.g. `SQL_ASCII`), the bytes are validated and this panics on non-UTF-8 data — same behavior as reading a `text` datum as `&str`.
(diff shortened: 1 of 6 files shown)


pgrx/src/datum/borrow.rs
```diff
@@ -140,3 +140,48 @@ unsafe impl Element for ffi::CStr {
         unsafe { ffi::CStr::from_ptr(char_ptr) }
     }
 }
+
+/// A borrowed Postgres `text` varlena, usable as a `FlatArray` element.
+#[repr(transparent)]
+pub struct Text([u8]);
```
eeeebbbbrrrr, line 146: Without a lifetime this can lead to UB.  I think this design needs a lot more thought than I'm able to give it right now.

eeeebbbbrrrr (changes requested):

PR #2347, opened 2026-07-03
Safe casting of PgNode values with bindgen tag discovery

cbandy:
Postgres implements various node types to represent parse, plan, and executor trees. It uses tagged polymorphism to emulate object-oriented inheritance in C: every node struct begins with a `NodeTag` field that identifies its type at runtime. Pointers to these nodes are accessed generically as `Node *` and queried using the `IsA` macro. The hierarchy includes intermediate base types, such as `Expr`, which are inherited by concrete types like `Const` and `Var`.
    
This PR introduces type-safe node casting (both upcasting and downcasting) via the `PgNode` trait. Safe casts are statically generated with the trait implementations emitted by `pgrx-bindgen`. Each node type may be cast to any of its "parent" types, all the way up to `Node`.
    
The `PgNode` trait is extended with:
    
- A constant `CAST_TAGS` slice containing the valid tags for the type and all of its descendants. This is efficient because the hierarchy is shallow and wide. The vast majority of types have four or fewer tags in `CAST_TAGS`.
- A `try_cast` associated function and a `try_as` method check `CAST_TAGS` before performing an unsafe pointer cast. The lookup is a simple linear scan and relies on compiler and CPU optimizations.
- A `node_tag` method that returns the tag and an `is_a` method that behaves like the `IsA` macro.

Postgres 13 and 14 use the `Value` struct to represent multiple possible value types, including integer, float, and string. Each type has its own `NodeTag`, but they can't be automatically associated with the struct using the information in the bindgen bindings. This hard-codes the associations just before emitting all the `PgNode` implementations.

Postgres 15 introduced separate structs for each value type. These fit nicely into the automatic struct-to-tag scheme, but we weren't handling the `ValUnion` union Postgres uses to represent these multiple types in a single `Node`. This extends the detection of parent-child types to implement `PgNode` for union nodes.

See: pgcentralfoundation/pgrx#1048

### DIFF
--- a/pgrx-pg-sys/src/node.rs
+++ b/pgrx-pg-sys/src/node.rs
@@ -129,4 +129,70 @@ mod tests {
         let node = expr.try_as::<crate::Node>().expect("upcast to Node should succeed");
         assert_eq!(node.node_tag(), crate::NodeTag::T_AlternativeSubPlan, "tag remains");
     }
+
+    // In versions prior to pg15, the `Value` struct is used with tags T_Null, T_Integer, T_Float,
+    // T_String, and T_BitString.
+    #[cfg(any(feature = "pg13", feature = "pg14"))]
+    #[test]
+    fn value_struct_cast() {
+        let value =
+            crate::Value { type_: crate::NodeTag::T_Integer, val: crate::ValUnion { ival: 42 } };
+
+        let node = value.as_node(); // infallible upcast
+        assert_eq!(node.node_tag(), crate::NodeTag::T_Integer, "tag remains");
+
+        let next = node.try_as::<crate::Value>().expect("downcast to self should succeed");
+        assert_eq!(next.node_tag(), crate::NodeTag::T_Integer, "tag remains");
+        assert_eq!(unsafe { next.val.ival }, 42);
+
+        for tag in &[
+            crate::NodeTag::T_Float,
+            crate::NodeTag::T_String,
+            crate::NodeTag::T_BitString,
+            crate::NodeTag::T_Null,
+        ] {
+            let value = crate::Value {
+                type_: *tag,
+                val: crate::ValUnion { str_: c"something".as_ptr() as *mut _ },
+            };
+
+            let node = value.as_node(); // infallible upcast
+            assert_eq!(node.node_tag(), *tag, "tag remains");
+
+            let next = node.try_as::<crate::Value>().expect("downcast to self should succeed");
+            assert_eq!(next.node_tag(), *tag, "tag remains");
+            assert_eq!(unsafe { next.val.str_ }, c"something".as_ptr() as *mut _);
+        }
+    }
+
+    // Since pg15, Boolean, Integer, Float, String, and BitString are separate node structs.
+    // `ValUnion` is a single node that can hold any one of these types. That union should implement
+    // PgNode so we can cast to one of the above structs.
+    #[cfg(any(
+        feature = "pg15",
+        feature = "pg16",
+        feature = "pg17",
+        feature = "pg18",
+        feature = "pg19"
+    ))]
+    #[test]
+    fn value_union_cast() {
+        let vu = crate::ValUnion {
+            sval: crate::String {
+                type_: crate::NodeTag::T_String,
+                sval: c"something".as_ptr() as *mut _,
+            },
+        };
+
+        let node = vu.try_as::<crate::Node>().expect("upcast to Node should succeed");
+        let next = node.try_as::<crate::String>().expect("downcast to self should succeed");
+
+        assert_eq!(next.type_, crate::NodeTag::T_String);
+        assert_eq!(next.sval, c"something".as_ptr() as *mut _);
+
+        assert!(
+            node.try_as::<crate::ValUnion>().is_none(),
+            "downcast to untagged union should fail"
+        );
+    }
 }
--- a/pgrx-pg-sys/src/node.rs
+++ b/pgrx-pg-sys/src/node.rs
@@ -2,7 +2,29 @@ use core::ffi::CStr;
 use core::ptr::NonNull;
 
 /// A trait applied to all Postgres `pg_sys::Node` types and subtypes
-pub trait PgNode: crate::seal::Sealed {
+pub trait PgNode: crate::seal::Sealed + Sized {
+    /// [`PgNode`] values that have one of these tags can safely be cast to `Self`
+    const CAST_TAGS: &'static [crate::NodeTag];
+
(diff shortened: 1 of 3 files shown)


cbandy:
The first commit adds to `impl_pg_node` in `pgrx-bindgen`, mostly independently of the existing implementation. The second commit is more invasive and rearranges the function to find nodes and accumulate tags simultaneously. Both perform similarly, so we should use the one that seems easier to read/maintain.

I had an agent compare all three commits using `hyperfine`:

### Benchmark Results: Postgres Node Hierarchy Traversal

Below is the comparative performance benchmark measuring compilation and binding generation for `pgrx-pg-sys` with the release-time code path:
```bash
PGRX_PG_SYS_GENERATE_BINDINGS_FOR_RELEASE=1 cargo build -p pgrx-pg-sys --no-default-features -F pg18
```

| Metric | `develop` (No Feature) | `HEAD^` (Original Cast Feature) | `HEAD` (Refactored Cast Feature) |
| :--- | :--- | :--- | :--- |
| **All Runs (Mean)** | 16.867 s ± 1.455 s | 17.260 s ± 1.593 s | 17.155 s ± 1.627 s |
| **Warm Runs (Mean)** ¹ | **16.221 s** | **16.551 s** | **16.431 s** |
| **Minimum Build Time** | 16.030 s | 16.425 s | 16.287 s |
| **Max Memory Usage** | **1.72 GB** (`1,716,551,680 B`) | **1.78 GB** (`1,784,430,592 B`) | **1.74 GB** (`1,740,562,432 B`) |

¹ *Warm Runs Mean ignores the first run to account for filesystem cache warm-up.*

#### Key Takeaways
* **Build Time Overhead:** 
  * The original implementation (`HEAD^`) added **~330ms** to build times (a ~2.0% compile-time increase over `develop`) because of complex recursive DFS traversals and syntax item filtering.
  * The refactored version (`HEAD`) reduces the warm-build mean to **16.431s**, recovering **~120ms** (or **~36%**) of the added overhead.
* **Memory Usage Overhead:** 
  * The original implementation (`HEAD^`) added **~68 MB** of memory overhead (a ~4.0% increase over `develop`) during compilation.
  * The refactoring in `HEAD` (which eliminates the intermediate `node_set` and `node_map` structures and optimizes name mapping lookups) brings the memory overhead down to **~24 MB** over develop, reclaiming **~44 MB** (or **~65%**) of the memory overhead introduced in `HEAD^`.

pgrx-pg-sys/src/include/pg13.rs
cbandy, line 1: 1. These files generated quite differently on my system, so I only included the changes to PgNode impl here.
2. I also don't have PG 19 set up yet, so that is missing from the PR.

Should this PR include these changes, or perhaps are they regenerated only when it is time to tag?

pgrx-pg-sys/src/node.rs
```diff
@@ -45,3 +85,48 @@ pub(crate) unsafe fn display_node_impl(node: NonNull<crate::Node>) -> String {
...
+    use super::PgNode;
+
+    #[test]
+    fn cast_roundtrip() {
+        let rt = crate::RangeTblRef { type_: crate::NodeTag::T_RangeTblRef, rtindex: 9 };
+
+        let node = rt.try_as::<crate::Node>().expect("upcast to Node should succeed");
+        let next = node.try_as::<crate::RangeTblRef>().expect("downcast to self should succeed");
+
+        assert_eq!(next.type_, crate::NodeTag::T_RangeTblRef);
```
cbandy, line 100: 🤔 Perhaps this should be `node_tag()` just because.

cbandy:
I'll update this to include the `Value` type.

cbandy:
This is ready! The `diff --stat` is huge because I included the changes to the generated files to aide in review:

```
git diff --stat -- pgrx-pg-sys/src/include
 7 files changed, 11323 insertions(+), 3191 deletions(-)
```

The code changes are small-ish:
```
git diff --stat -- pgrx-bindgen pgrx-pg-sys/src/node.rs
 2 files changed, 404 insertions(+), 153 deletions(-)
```

pgrx-pg-sys/src/include/pg13.rs
eeeebbbbrrrr: they're generated when we cut a release, so yeah, please remove them from the PR.  Thanks for asking!

pgrx-pg-sys/src/node.rs
```diff
@@ -2,7 +2,29 @@ use core::ffi::CStr;
...
+        // This uses `slice::contains` to perform a linear search because the vast majority (>95%)
+        // of PgNode have four or fewer tags in `CAST_TAGS`. For small contiguous slices of ints,
+        // a linear scan is faster than structured lookups (like binary search or hash sets) due
+        // to cache locality, register/instruction efficiency, and compiler autovectorization.
+        if Self::CAST_TAGS.contains(&node.node_tag()) {
+            Some(unsafe { &*core::ptr::from_ref(node).cast() })
+        } else {
+            None
+        }
+    }
```
eeeebbbbrrrr, line 20: I feel like this ought to allow casing to `pg_sys::Node` too, yeah?  I don't see a need to include that in every `CONST_TAGS` definition, but hardcoding it here would be alright as an `else if` condition

eeeebbbbrrrr (changes requested):

pgrx-pg-sys/src/node.rs
cbandy: Yep, done. Rather than include every tag on `Node`, the code generation overrides this function for `Node` so every node can cast to it:  [build.rs#L600-L610](https://github.com/pgcentralfoundation/pgrx/pull/2347/changes#diff-0b05d9484b65da7e00c97b85b4d484f6e993a5858906fbeddec86988f088dd1cR600-R610)

```rust
#[inline]
fn try_cast<T: pg_sys::PgNode>(node: &T) -> Option<&Self> {
    Some(node.as_node())
}
```

cbandy: This function and the `try_as` method kinda swap arguments and types:

```rust
pg_sys::Expr::try_cast(&some_pg_node) -> Option<&pg_sys::Expr>
some_pg_node.try_as::<pg_sys::Expr>() -> Option<&pg_sys::Expr>
```

```rust
pg_sys::Node::try_cast(&some_pg_node) -> Option<&pg_sys::Node>
some_pg_node.try_as::<pg_sys::Node>() -> Option<&pg_sys::Node>
some_pg_node.as_node() -> &pg_sys::Node
```

eeeebbbbrrrr (approved):

PR #2354, opened 2026-07-13
Fix aggregate deserial function signature

### LINKED ISSUE #2351: BUG:  [pg_aggregate] with parallel aggregation and internal types generates wrong functions
When creating an aggregate using the `pg_aggregate` macro with `ParallelOption::Safe` the deserial function has the wrong signature.

From: https://www.postgresql.org/docs/current/sql-createaggregate.html

> deserialfunc
>    Deserialize a previously serialized aggregate state back into state_data_type. This function must take two arguments of types bytea and internal, and produce a result of type internal. (Note: the second, internal argument is unused, but is required for type safety reasons.)

Generated is a function with a first argument of `Self::State`/`internal` though.

See the (AI generated) reproduction at https://github.com/ibotty/pgrx-internal-parallel-aggregate-bug-repro

eeeebbbbrrrr:
## Summary

- generate aggregate deserial functions with PostgreSQL's required `(bytea, internal) -> internal` signature
- update `Aggregate::deserial` to mirror that contract without a spurious current-state argument or `PgBox` conversion
- add macro-level and live PostgreSQL regression coverage for a parallel aggregate with `internal` state

Fixes #2351

## Testing

- `cargo test -p pgrx-sql-entity-graph`
- `cargo pgrx test pg18 aggregate_parallel_internal_deserial`
- `cargo check -p pgrx --no-default-features --features pg18`
- `cargo check -p pgrx-unit-tests --no-default-features --features pg18,pg_test`
- `cargo check --manifest-path pgrx-examples/aggregate/Cargo.toml --no-default-features --features pg18`
- `cargo clippy -p pgrx-sql-entity-graph -- -D warnings`

### DIFF
--- a/pgrx-examples/aggregate/src/lib.rs
+++ b/pgrx-examples/aggregate/src/lib.rs
@@ -120,7 +120,7 @@ impl Aggregate<IntegerAvgState> for IntegerAvgState {
     //     unimplemented!()
     // }
 
-    // fn deserial(current: Self::State, _buf: Vec<u8>, _internal: PgBox<Self::State>, _fcinfo: pgrx::pg_sys::FunctionCallInfo) -> PgBox<Self::State> {
+    // fn deserial(_buf: Vec<u8>, _internal: pgrx::Internal, _fcinfo: pgrx::pg_sys::FunctionCallInfo) -> pgrx::Internal {
     //     unimplemented!()
     // }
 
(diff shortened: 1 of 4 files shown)


eeeebbbbrrrr:
I am not sure that the attempt here is actually correct or in the spirit of pgrx.  Having a function signature that takes `Internal` as an argument I believe is incorrect for this API.  I need to study this more.

PR #2358, opened 2026-07-15
implement DatumPass, Element and Scalar for Uuid

### LINKED ISSUE #2357: Uuid could support Scalar, DatumPass and Element
For an extension I wanted `Array`'s `to_slice` for `Uuid` arrays.  Unfortunately that did not work, because `Uuid` is not `Scalar`.  I figured it should and asked the Opus 4.8 which agreed (however much that means).

ibotty:
fixes #2357

### DIFF
--- a/pgrx/src/array.rs
+++ b/pgrx/src/array.rs
@@ -438,3 +438,6 @@ unsafe impl Scalar for i32 {
 unsafe impl Scalar for i64 {
     const OID: pg_sys::Oid = pg_sys::INT8OID;
 }
+unsafe impl Scalar for crate::datum::Uuid {
+    const OID: pg_sys::Oid = pg_sys::UUIDOID;
+}
--- a/pgrx/src/datum/borrow.rs
+++ b/pgrx/src/datum/borrow.rs
@@ -119,7 +119,8 @@ macro_rules! impl_borrow_fixed_len {
 impl_borrow_fixed_len! {
     i8, i16, i32, i64, bool, f32, f64,
     pg_sys::Oid, pg_sys::Point,
-    Date, Time, TimeWithTimeZone, Timestamp, TimestampWithTimeZone
+    Date, Time, TimeWithTimeZone, Timestamp, TimestampWithTimeZone,
+    Uuid
 }
 
 /// It is rare to pass CStr via Datums, but not unheard of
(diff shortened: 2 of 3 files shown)


pgrx/src/datum/borrow.rs
```diff
@@ -119,7 +119,11 @@ macro_rules! impl_borrow_fixed_len {
 impl_borrow_fixed_len! {
     i8, i16, i32, i64, bool, f32, f64,
     pg_sys::Oid, pg_sys::Point,
-    Date, Time, TimeWithTimeZone, Timestamp, TimestampWithTimeZone
+    Date, Time, TimeWithTimeZone, Timestamp, TimestampWithTimeZone,
+    // Uuid is a fixed 16-byte, char-aligned, by-ref type stored inline in arrays with no
+    // header and no byte-swapping, so the macro's size>Datum big-endian branch (ptr.cast()) is
+    // correct for it.
```
eeeebbbbrrrr: this editorialization is unnecessary

pgrx/src/array.rs
```diff
@@ -438,3 +438,9 @@ unsafe impl Scalar for i32 {
 unsafe impl Scalar for i64 {
     const OID: pg_sys::Oid = pg_sys::INT8OID;
 }
+// Uuid is a fixed 16-byte `[u8; 16]` (no padding, all bitpatterns valid). Postgres stores `uuid`
+// inline in arrays as packed, char-aligned raw bytes with no per-element header and no endian
+// byte-swapping, so `&[Uuid]` matches the array data layout on both LE and BE.
```
eeeebbbbrrrr: so is this.

eeeebbbbrrrr (approved):

eeeebbbbrrrr (changes requested):
please see my comments

pgrx/src/datum/borrow.rs
ibotty: Ack.  Will remove both.

ibotty:
I *think* the change is semantically correct, but I am no expert in PostgreSQL.  I mostly went from Uuid (in PostgreSQL) is basically a i128 (and I am using it as that at times) to this should definitely work.  I only briefly looked into the PostgreSQL code myself.

eeeebbbbrrrr (approved):

Issue #2194, opened 2025-11-02
Minimize non-"meta" contents of `pgrx::datum`


workingjubilee:
Currently, in `pgrx::datum`, there are two major kinds of items:

1. Items describing [basic Postgres datatypes](https://www.postgresql.org/docs/current/datatype.html) or functions on those types
2. Items describing interactions with the notion of a `Datum` itself (whether the "raw" `pg_sys::Datum` or the lifetime-bearing `pgrx::datum::Datum`)

I find all of them being together to be slightly confusing. It poses two issues, one conceptual and one simply organizational.
- The conceptual issue: There are actually several cases where major interactions with a `Datum` are already not in that location because it would be strange to put them there. For instance, the current way HeapTuples work is entirely in [heap_tuple.rs](https://github.com/pgcentralfoundation/pgrx/blob/develop/pgrx/src/heap_tuple.rs) but... often, these are just another kind of datum! Certainly when a heap tuple is used to describe a composite type, for instance. The fact that something has a "datum representation" is just a side fact of how Postgres works.
- The organizational issue: When dealing with tightly-related code, it isn't necessarily organized "closely together". Or rather, it's also organized closely together with unrelated code, making it hard to tell what is intimately tied to what. Code that spreads over multiple related types (e.g. datetime types) is instead in a shared module that is visible to a bunch of other stuff. Or it's instead in an arbitrary module, instead of a shared one! It's a small impedance, but I struggle slightly more when I have to sift through sibling modules to find the right one. And the story of pgrx *is* essentially "with enough papercuts, you can completely stall work", which is what happened to pgrx-sql-entity-graph.

I think that the entire first kind should be reorganized into `pgrx::datetime`, `pgrx::num`, etc. and eventually have their alternate paths deprecated. They should remain exposed in `pgrx::prelude`, so this won't affect people using them via that, only those interacting with the "absolute" paths into `pgrx`. This also will have the benefit of shortening some of the names in error messages, mangled symbols, etc.

- [x] datetime: https://github.com/pgcentralfoundation/pgrx/pull/2199
- [ ] numeric
- [ ] json
- [ ] uuid
- [ ] net

This will leave `pgrx::datum` for everything "meta" about datums.

PR #2359, opened 2026-07-17
Minimize non-"meta" contents of `pgrx::datum`: uuid, json, net, num

ChronicallyJD:
## Summary

Continues #2194 by moving the remaining basic-datatype items out of `pgrx::datum` into dedicated top-level modules, following the `datetime` precedent from #2199. This covers the outstanding checklist items:

- `uuid` → `pgrx::uuid`
- `json` → `pgrx::json`
- `net` (inet) → `pgrx::net`
- `numeric` → `pgrx::num` (with its support code at `pgrx::num::support`)

Each type group is a separate commit.

Everything remains accessible from the previous paths via re-export, so this is purely organizational with no end-user impact except new path options. In particular, the module paths `pgrx::datum::numeric` and `pgrx::datum::numeric_support` are preserved, along with the re-exports flowing through `pgrx::prelude` and the crate root (`pgrx::Uuid`, `pgrx::Inet`, `pgrx::Json`, `pgrx::JsonB`, `pgrx::AnyNumeric`, `pgrx::Numeric`, `pgrx::numeric`).

`pgrx::num` uses the same `num.rs` + `num/` layout that `datetime.rs` + `datetime/` uses.

## Note on clippy

Moving code out of `datum` drops the module-level `#![allow(...)]` on `datum/mod.rs`, so the numeric move surfaces two `clippy::unnecessary_fallible_conversions` warnings (tracked in #1525). This matches the `datetime` migration exactly: `datetime.rs` carries no such allows and already surfaces the same lint on `develop`. No new allows were added, to stay consistent with that precedent.

## Testing

Against PostgreSQL 18:

- `cargo test --all --features "pg18 pg_test cshim proptest" --no-default-features` — all suites pass (977 tests, 0 failures, including the 603-test `pgrx-tests` integration suite).
- `cargo fmt --all --check` — clean.
- `cargo clippy -p pgrx-sql-entity-graph -- -Dwarnings` — clean.

Part of #2194.


OffgridwithJD:
Note: the failing \`verify package can build\` check on this PR is a pre-existing issue on \`develop\`, unrelated to this change. That workflow builds with nightly, which promotes the \`semicolon_in_expressions_from_macros\` lint (rust-lang/rust#79813) to a hard error, tripping some \`bail!(...)\` calls used in expression position. This PR does not touch that code. Fixed separately in #2360.

PR #2361, opened 2026-07-30
Restore ERROR detail preservation through error handling

eeeebbbbrrrr:
#2262 was intended to smuggle Rust backtraces into Postgres' ERROR detail line, but the implementation had the unintended side effect of wiping out user/postgres-provided ERROR detail lines, including those generated from pl/pgsql statements.

PR brings our use of `pg_re_throw()` back, preserving the actual Postgres ERROR as given, doesn't sneak a backtrace into caught Postgres ERRORs, adds a test around this (now known to be expected) behavior, and tidies up a few comments.


eeeebbbbrrrr:
no good deed goes unpunished

PR #2368, opened 2026-08-31
Avoid unsafe intermediate slice construction in varlena functions

SergeiPatiakin:
`unsafe __IncompleteArrayField<T>::as_slice` calls `unsafe
::core::slice::from_raw_parts`, which [is UB](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#safety) if the slice memory range is invalid.

But the length we pass in the `varlena_4b*` cases is `varsize_1b(ptr)` which  is a garbage value
potentially larger than the real length. Therefore the slice memory range can indeed be invalid.

I don't think this is a practical concern today because we immediately do `.as_ptr` on
the intermediate slice, but still looks like technically UB.


0xPoe (approved):
Although it doesn’t cause any issues now, it’s reasonable to fix it.

PR #2369, opened 2026-09-04
fix: reduce the amount of useless allocations for errors on hot paths

Ddystopia:
Hi, I noticed a bunch of strange allocations on my hot path and it turned out to be some `.ok_or(SpiError::CursorNotFound(name.to_string()))?` at `pgrx/src/spi/client.rs:159`, which was immediately dropped. I became upset and told claude to find all similar places there and make them like they should be, in `ok_or_else` and similar. I reviewed the code and I believe it is fine - and the changes are trivial.

### DIFF
--- a/cargo-pgrx/src/command/bench.rs
+++ b/cargo-pgrx/src/command/bench.rs
@@ -110,7 +110,7 @@ impl CommandExecute for Bench {
         )?;
 
         let extname = get_property(&package_manifest_path, "extname")?
-            .ok_or(eyre!("could not determine extension name"))?;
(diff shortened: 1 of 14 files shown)


0xPoe (commented):
Maybe we should consider enabling the [`or_fun_call`](https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call) lint if we accept these changes?

Ddystopia:
It depends on what exactly you want. There were many cases of `ok_or(something_static)`, or similar. If it will get flagged, I am concerned about developer experience, as the code will become uglier.

PR #2377, opened 2026-09-13
include more headers

usamoi:
This PR makes pgrx include more headers. The ultimate goal is to include all headers to avoid the awkward situation where pgrx is missing the header files required by extensions.

Now all `./pgrx-pg-sys/include/pg*.h` are generated by

```shell
# ./pgrx-pg-sys/generate.sh $(pg_config --pkgincludedir)
./pgrx-pg-sys/generate.sh ~/.pgrx/13.23/pgrx-install/include/postgresql > ./pgrx-pg-sys/include/pg13.h
./pgrx-pg-sys/generate.sh ~/.pgrx/14.23/pgrx-install/include/postgresql > ./pgrx-pg-sys/include/pg14.h
./pgrx-pg-sys/generate.sh ~/.pgrx/15.18/pgrx-install/include/postgresql > ./pgrx-pg-sys/include/pg15.h
./pgrx-pg-sys/generate.sh ~/.pgrx/16.14/pgrx-install/include/postgresql > ./pgrx-pg-sys/include/pg16.h
./pgrx-pg-sys/generate.sh ~/.pgrx/17.10/pgrx-install/include/postgresql > ./pgrx-pg-sys/include/pg17.h
./pgrx-pg-sys/generate.sh ~/.pgrx/18.4/pgrx-install/include/postgresql > ./pgrx-pg-sys/include/pg18.h
./pgrx-pg-sys/generate.sh ~/.pgrx/19beta2/pgrx-install/include/postgresql > ./pgrx-pg-sys/include/pg19.h
```

fix macos ci only: #2378


usamoi:
https://github.com/postgres/postgres/commit/8ec569479fc28ddd634a13dc100b36352ec3a3c2

It seems that PostgreSQL 14 and earlier versions do not mark all variables as `dllimport`, which means these symbols cannot be used in Windows extensions.

Probably we can temporarily exclude these problematic headers. Once support for PostgreSQL 14 is dropped by pgrx, we can add them back.

Edit: It seems that too many problematic headers are already being included. Since we cannot remove the problematic headers that are already in use, we could instead remove only the `static inline` functions that cause compilation failures.

PR #2378, opened 2026-09-13
fix macos ci

usamoi:


usamoi:
https://github.com/postgres/postgres/commit/8ec569479fc28ddd634a13dc100b36352ec3a3c2

It seems that PostgreSQL 14 and earlier versions do not mark all variables as `dllimport`, which means these symbols cannot be used in Windows extensions.

Probably we can temporarily exclude the headers that have this issue. Once support for PostgreSQL 14 is dropped by pgrx, we can add them back.

PR #2380, opened 2026-09-16
Add support for a fragmented SQL release process.

### LINKED ISSUE #2379: Feature Request: Fragment-based extension migration & upgrade management for development and release workflows
### Summary

We propose adding first-class support in `cargo-pgrx` for managing unreleased SQL migration fragments.

In active extension repositories with stable release branches (e.g. maintaining `main` for upcoming minor releases while backporting bugfixes to `0.25.x`), managing SQL upgrade scripts in monolithic files (`extension--<prev>--<target>.sql`) presents a few practical challenges:
1. Branch decoupling & cherry-picking: When authoring a change on `main`, it is not always clear which release version will ship first. Monolithic scripts couple SQL changes to specific version-range filenames, which means cherry-picking a change to a stable release branch typically requires manually adjusting or splitting migration scripts across branches.
2. Local dev iteration with fragments: While `cargo pgrx install` copies pre-existing version-to-version scripts (`sql/extension--<old>--<new>.sql`), it does not assemble unreleased fragment files. Without an ephemeral upgrade script in Postgres's extension directory during development, developers testing in-flight changes against existing databases often resort to dropping and recreating the extension, which resets test data and requires rebuilding dependent indexes.

Taking inspiration from ORM migration frameworks (such as Django, Alembic, or Flyway), where schema modifications are authored as discrete migration units declaring prerequisite dependencies, a fragment-based workflow addresses these challenges:
- Dedicated, version-agnostic fragments: Each change introduces a standalone fragment file (e.g. `sql/unreleased/<PR>.<description>.sql`). The fragment cherry-picks cleanly between branches as an isolated unit and is assembled into the correct versioned upgrade script when that specific branch cuts a release.
- Ephemeral dev assembly: `cargo pgrx install` automatically assembles unreleased fragments directly into Postgres's `$SHAREDIR/extension/` and updates `default_version` in the installed `.control` file. Developers can run `ALTER EXTENSION ... UPDATE;` immediately without modifying the git working tree or resetting existing test databases.
- Release assembly: A dedicated migration subcommand (`cargo pgrx migrate assemble`) topologically sorts fragments based on explicit dependencies (`-- depends-on:`), concatenates them into the final versioned upgrade script, and removes the consumed fragments.

---

### Motivation & Problem Description

Currently, `cargo-pgrx` handles extension installation in [cargo-pgrx/src/command/install.rs](https://github.com/pgcentralfoundation/pgrx/blob/70383e8840428d0859a85011709422a57864aa9a/cargo-pgrx/src/command/install.rs#L418-L439) by:
1. Generating the full base schema `extension--<version>.sql` and placing it in `$SHAREDIR/extension/`.
2. Copying pre-existing version upgrade files strictly matching `^extension--.+--.+\.sql$` from `<crate>/sql/` into `$SHAREDIR/extension/`.

In projects with concurrent pull requests or multiple release branches, this model has a few areas that can be improved:

#### 1. Decoupling Changes for Stable Branch Backports & Cherry-Picking

In extensions with active stable branches (e.g. fixing a bug on `main` that must also be backported to `0.25.x`):
* On `main`, the next release might be `0.26.0` (target file: `extension--0.25.0--0.26.0.sql`).
* On the stable branch, the next release might be `0.25.5` (target file: `extension--0.25.4--0.25.5.sql`).
* When authoring a commit on `main`, developers cannot predict with certainty which patch release will receive the backport, or whether the change will only ever ship in the next minor release.
* Committing SQL directly to a monolithic version-range file can complicate `git cherry-pick` operations or require manual version adjustments across branches.

With standalone fragments, each change lives in its own file (`sql/unreleased/<PR>.<description>.sql`). The fragment can be cherry-picked cleanly between branches without file conflicts, and each branch independently resolves and assembles the fragments present on that branch at release time.

#### 2. Local Development Iteration with Unreleased Fragments

When developers work with monolithic files in `sql/`, `cargo pgrx install` copies those files over, and `ALTER EXTENSION ... UPDATE;` works as expected. However:
* Testing changes locally against an existing database requires adding SQL to a version-coupled script in `sql/` (e.g. `extension--0.25.0--0.26.0.sql`), introducing version assumptions in the working copy that may need to be adjusted later when cherry-picking or rebasing.
* If an extension instead organizes pending changes into unreleased fragments (`sql/unreleased/*.sql`), `cargo pgrx install` ignores them because they do not match `^extension--.+--.+\.sql$`.
* Without an assembled upgrade script in `$SHAREDIR/extension/`, `ALTER EXTENSION ... UPDATE;` is a no-op because Postgres does not find an upgrade path between the installed version and the target version.
* The common fallback is to run `DROP EXTENSION ... CASCADE; CREATE EXTENSION ...;`. For extensions backed by substantial datasets or complex indexes (such as BM25 search indexes, vector indices, or benchmark suites like ParadeDB or TxPipe's Mumak in [#1960](https://github.com/pgcentralfoundation/pgrx/issues/1960)), recreating the extension drops dependent objects and requires re-indexing from scratch, which slows down local iteration.

#### 3. Merge Conflicts on Monolithic Scripts

When multiple contributors work on separate features concurrently, having every PR edit the same monolithic upgrade script (`extension--0.25.0--0.26.0.sql` or `unreleased.sql`) can cause merge conflicts upon rebase or merge.

---

#### 1. Fragment Directory & File Convention

PRs that introduce DDL or schema adjustments place standalone migration fragments in:
```
<crate>/sql/unreleased/<ID>.<description>.sql
```
For example:
* `sql/unreleased/6221.inline_row_evaluation.sql`
* `sql/unreleased/6245.add_bm25_similarity_operator.sql`

The `<ID>` typically corresponds to a PR or issue number (numeric) or a unique slug. 

The unreleased directory can be configured in `Cargo.toml` under `[package.metadata.pgrx]`:
```toml
[package.metadata.pgrx]
unreleased-sql-dir = "sql/unreleased" # default
```

#### 2. Explicit Dependency Tracking & Topological Sorting

When one migration fragment depends on objects introduced by another in-flight fragment, it declares dependencies via a standardized comment header:
```sql
-- depends-on: 6099, 6120
```
or
```sql
/* depends-on: 6099, 6120 */
```

`cargo-pgrx` parses these declarations and builds a dependency graph:
* Topological Sort: Fragments are ordered so dependencies run before dependents.
* Deterministic Tie-Breaking: Independent fragments are sorted deterministically (e.g. by `(ID, filename)`).
* Cycle & Missing Dependency Detection: Cycles or missing dependencies are reported with clear diagnostics. On release branches, this also allows CI to verify that any prerequisite fragments were backported along with the dependent fragment.

This mechanism takes direct inspiration from ORM migration systems (such as Django migrations or Alembic), where migration files form a directed acyclic graph (DAG) via explicit dependency declarations rather than relying on brittle global sequence numbers. In the context of PostgreSQL extensions, `cargo-pgrx` adapts this model by resolving the DAG and compiling it into the versioned `<ext>--<prev>--<target>.sql` file expected by Postgres.

#### 3. Development Workflow: Ephemeral Assembly during `cargo pgrx install`

When running `cargo pgrx install` (or `cargo pgrx run`):
1. Normal compilation and installation proceed (building the shared library into `$PKGLIBDIR`, generating the base schema, and copying pre-existing released upgrade scripts).
2. If `sql/unreleased/` contains `.sql` fragments:
   - Resolve Previous Version: Detect the highest target version from existing `sql/<ext>--*--*.sql` files (or fallback to the base version in `Cargo.toml`).
   - Compute Ephemeral Target Version: Derive an ephemeral target version (e.g. bumping the minor/patch version, such as `0.25.9` -> `0.26.0-dev` or `0.26.0`).
   - Assemble Ephemeral Script: Topologically sort the unreleased fragments and assemble them directly into `$SHAREDIR/extension/<ext>--<prev>--<target>.sql`.
     - Includes standard safety header:
       ```sql
       \echo Use "ALTER EXTENSION <ext> UPDATE TO '<target>'" to load this file. \quit
       ```
     - Includes fragment delimiter banners for clear SQL debugging.
   - Update Installed Control File: Update `default_version` in `$SHAREDIR/extension/<ext>.control` to match the target version.

Crucially, this happens entirely within Postgres's `$SHAREDIR/extension/`:
* The repository working copy and git status remain completely clean (no dirty untracked files).
* `cargo pgrx install` requires no live database connection.
* Developers can run `ALTER EXTENSION <ext> UPDATE;` in their active test database to apply in-flight changes without dropping tables or rebuilding indexes.

CLI Flags:
* `--assemble-unreleased`: Defaults to `true` when unreleased fragments exist.
* `--no-assemble-unreleased`: Disables ephemeral script assembly.

#### 4. Release Workflow: `cargo pgrx migrate assemble`

At release time, maintainers freeze the unreleased fragments into a permanent versioned upgrade script:

```bash
cargo pgrx migrate assemble [TARGET_VERSION] [OPTIONS]
```

Options for `assemble`:
* `TARGET_VERSION`: The new release version (e.g. `0.26.0`). If omitted, reads `package.version` from `Cargo.toml`.
* `--prev-version <VERSION>`: The previous release version (e.g. `0.25.9`). If omitted, automatically resolves the latest target version from existing upgrade scripts in `sql/`.
* `--consume`: (Default for releases) Deletes the assembled fragments from `sql/unreleased/` once concatenated into `sql/<ext>--<prev>--<target>.sql`.
* `--preserve-fragments`: Preserves fragments in `sql/unreleased/` (useful for release candidate / beta builds).
* `--output-dir <DIR>`: Output directory for the assembled script (defaults to `<crate>/sql/`).

Command Placement:
We propose a top-level `migrate` subcommand rather than nesting under `schema` or `package`:
* `cargo pgrx schema` currently accepts positional arguments (`cargo pgrx schema [PG_VERSION] [ENTITY_NAME...]`) for AST schema generation. Nesting migration assembly here (`cargo pgrx schema assemble`) creates clap argument ambiguity with entity names and conflates AST code generation with version upgrade lifecycle management.
* `cargo pgrx package` is focused on compiling and staging binary distribution packages in `target/`. Mutating repository sources (deleting fragments and writing release SQL) within `package` would break its idempotency and cause conflicts during multi-version matrix builds (`pg13`..`pg17`).
* A dedicated `migrate` command creates a clean, discoverable home for upgrade workflows, leaving room for future subcommands like `cargo pgrx migrate lint`. (If maintainers prefer grouping under an existing command, `cargo pgrx schema assemble` is an alternative).

#### 5. Role of `cargo pgrx package` as Downstream Consumer

`cargo pgrx package` delegates directly to `install_extension(...)` to copy extension artifacts into the packaging staging directory (`target/release/extname-pgXX/share/extension/`). It should consume this workflow in two ways:

1. Release Validation (Default):
   When running `cargo pgrx package` to build distribution artifacts, `package` checks whether `sql/unreleased/` contains unreleased fragments. If found, it warns or errors:
   `error: found unreleased SQL fragments in sql/unreleased/. Run cargo pgrx migrate assemble before packaging for release.`
   This prevents accidental releases where a version bump was committed but in-flight migration fragments were forgotten.

2. Hermetic Package Assembly:
   For CI or nightly pipelines that build release tarballs/packages without committing the assembled script back to git first, `cargo pgrx package --assemble-unreleased` can assemble the unreleased fragments directly into the staging package directory (`out_dir`), matching what `install` does for `$SHAREDIR`.

---

#### 1. Pure Schema-Diffing (Issue [#2375](https://github.com/pgcentralfoundation/pgrx/issues/2375))

Issue [#2375](https://github.com/pgcentralfoundation/pgrx/issues/2375) discusses integrating schema-diffing tools (such as `pg-schema-diff`) directly into `cargo-pgrx` to automatically generate `extension--<old>--<new>.sql` upgrade scripts.

In fact, ParadeDB already uses `pg-schema-diff` in CI today: CI diffs the base branch schema against the pull request schema to generate a suggested SQL diff, and validates that the pull request contains those changes.

However, relying purely on release-time schema diffing without a fragment workflow has a few limitations:

1. Destination for in-flight changes and cherry-picking:
   Even when a tool like `pg-schema-diff` generates the SQL diff between a pull request and its base branch, that SQL needs a place to live in the repository. If it is committed directly to a version-named monolithic file (e.g. `extension--0.25.0--0.26.0.sql`), cherry-picking that commit to a stable release branch (targeting `0.25.5`) requires manually moving the SQL into a different version-range file. Fragments provide a version-agnostic container for the diff.
2. Local development iteration without branch comparisons:
   Generating a schema diff requires a clean baseline schema from a base branch (which CI can generate by checking out the base branch). During offline or iterative local development, checking out separate branches to diff against is less convenient than having `cargo pgrx install` assemble unreleased fragments directly into `$SHAREDIR/extension/`.
3. Destructive ambiguity and ordering:
   Declarative diffs cannot always deduce developer intent for renames or signature changes. For example, renaming a function or changing an argument type can cause a declarative diff to emit `DROP FUNCTION old(...)` followed by `CREATE FUNCTION new(...)`. In Postgres, if user views, indexes, or privileges depend on `old(...)`, dropping the function can fail or unexpectedly drop dependent user objects. Explicit fragments allow developers to review the diff generated by tooling and adjust it (e.g., adding backward-compatible wrapper functions or ordering dependencies).
4. Non-DDL and extension-specific commands:
   While most Postgres extension upgrades consist of function and type DDL, extensions that maintain internal tables (such as metadata catalogs or sequence caches) sometimes require custom commands (e.g., `SELECT pg_catalog.pg_extension_config_dump(...)`, custom `GRANT`s, or metadata backfills) that declarative catalog diffs do not generate.

How Fragments and Schema-Diffing Work Together:
Rather than competing alternatives, fragments and schema-diffing complement each other naturally:
- Tooling like `pg-schema-diff` can generate the suggested SQL diff for a pull request.
- The fragment file provides the version-agnostic container that can be reviewed, committed, and cherry-picked cleanly between branches.
- `cargo pgrx migrate assemble` bundles those fragments into the versioned script at release time.

#### 2. Monolithic `sql/unreleased.sql`

Another alternative is maintaining a single `sql/unreleased.sql` file on `main`.
* Drawback: This reintroduces cherry-picking friction. When backporting a bugfix to a stable release branch, cherry-picking edits to a single shared `unreleased.sql` file causes merge conflicts across branches and between concurrent pull requests on `main`.
* Standalone fragments allow changes to be cherry-picked cleanly as discrete units.

#### 3. Ad-hoc Entity Extraction (`cargo pgrx schema <entity>`, PR [#2293](https://github.com/pgcentralfoundation/pgrx/pull/2293))

PR [#2293](https://github.com/pgcentralfoundation/pgrx/pull/2293) introduced `cargo pgrx schema <entity_name>` to emit SQL with `ALTER EXTENSION ... ADD ...`.
* Drawback: This was designed as an ad-hoc debugging tool for individual functions. It requires the developer to manually determine every modified entity, pipe SQL into `psql`, and does not handle ordered multi-object migrations, data adjustments, or automated release script generation.

#### 4. AST / Proc-Macro Snapshot Diffing (PR [#2197](https://github.com/pgcentralfoundation/pgrx/pull/2197))

PR [#2197](https://github.com/pgcentralfoundation/pgrx/pull/2197) explored generating upgrade scripts by comparing serialized snapshot graphs in `pgrx-sql-entity-graph`.
* Drawback: This approach involved maintaining snapshot serialization types and diffing logic within the proc-macro and SQL graph infrastructure, which proved challenging to maintain alongside evolving AST features. Managing migrations at the SQL fragment level keeps `cargo-pgrx` simpler and decoupled from compiler AST internals.

---

### Reference Implementation

This workflow is currently used in production in ParadeDB (for `pg_search`):
* Migration fragments, topological sorting, dependency validation, and branch linting are run across hundreds of PRs.
* The ephemeral `$SHAREDIR` installation wrapper allows developers to pull branches, run `cargo pgrx install`, and test changes against existing databases using `ALTER EXTENSION pg_search UPDATE;` without recreating the extension or rebuilding indexes.

stuhood:
Fixes https://github.com/pgcentralfoundation/pgrx/issues/2379. See the issue for background, design rationale, and complete workflow details.

### Summary

Introduces first-class support for modular SQL migration fragments in `cargo-pgrx` to eliminate merge conflicts on monolithic upgrade scripts, simplify cross-branch cherry-picking, and automate extension upgrade testing.

### Key Capabilities

* In-flight migration fragments (`sql/unreleased/`): Developers commit discrete SQL fragments (`<PR>.<slug>.sql`) declaring prerequisites (`-- depends-on:`). Fragments are topologically sorted during assembly.
* Ephemeral local development: `cargo pgrx install`, `run`, and `test` assemble fragments into an ephemeral upgrade script and matching base schema directly within Postgres's extension directory, updating the installed `.control` file without dirtying the Git working tree.
* Release assembly (`cargo pgrx migrate assemble`): Consolidates unreleased fragments into a permanent release upgrade script (`ext--<old>--<new>.sql`), deletes consumed fragments, and optionally updates `default_version` in the `.control` file (`--update-control`).
* Packaging protection: `cargo pgrx package` fails if unreleased fragments are present unless `--assemble-unreleased` is passed, ensuring development fragments do not leak into release distributions.
* CI and pre-commit linting (`cargo pgrx migrate check` / `lint`): Validates fragment DAGs for cycles, verifies that modifications to unreleased objects declare prerequisites, checks for mixed released/unreleased objects (`--deny-mixed-objects`), and supports git diff scoping (`--base <REF>`).
* Plan introspection (`cargo pgrx migrate info`): Previews predecessor/target versions and ordered fragment plans in text or JSON.

### Documentation & Scaffolding

* Added migration guide in `docs/src/extension/migrations.md` and documented CLI commands in `cargo-pgrx/README.md`.
* `cargo pgrx new` now scaffolds `sql/unreleased/` with a `.gitkeep`.


comment-issue-2375.md
```diff
@@ -0,0 +1,13 @@
+To add more context from the ParadeDB side: we have been working to improve extension upgrade and migration workflows for a while.
```
mdashti, line 1: These three markdown files at the repo root (`comment-issue-2375.md`, `feature-request-sql-migration-fragments.md`, `migration-target-and-base-schema-fixes.md`) look like working notes that slipped into the commit. The content is already in #2379 and in `docs/src/extension/migrations.md`.

cargo-pgrx/src/command/migrate/lint.rs
```diff
@@ -0,0 +1,516 @@
...
+    let mut changed = Vec::new();
+    let unreleased_canonical =
+        unreleased_dir.canonicalize().unwrap_or_else(|_| unreleased_dir.to_path_buf());
+
+    for line in stdout.lines() {
+        let trimmed = line.trim();
+        if trimmed.is_empty() || !trimmed.ends_with(".sql") {
+            continue;
+        }
+        let full_path = repo_root.join(trimmed);
```
mdashti, line 199: `git diff --name-only` prints paths relative to the repository root, whatever the cwd is. But `repo_root` here is the crate directory (`Check::execute` passes `crate_dir`). For any crate that isn't at the git root, `full_path` becomes `<crate>/<crate>/sql/unreleased/...`, nothing matches, and `--base` lints nothing while reporting success. I hit this with `pg_search` inside the ParadeDB workspace: a fragment that re-creates a function from another unreleased fragment fails without `--base` and passes with `--base HEAD~1`. `git rev-parse --show-toplevel` (or `git diff --relative` run from the crate dir) would fix it. The e2e test in `cli_migrate.rs` never initialises a git repo, so a nested-crate case there would catch this.

cargo-pgrx/src/command/migrate/version.rs
```diff
@@ -0,0 +1,236 @@
...
+use eyre::{Context, eyre};
+use regex::Regex;
+use semver::Version;
+use std::fs;
+use std::path::Path;
+
+pub fn clean_version_string(ver: &str) -> &str {
+    let ver = ver.trim();
+    let ver = ver.strip_prefix('v').unwrap_or(ver);
+    ver.split('-').next().unwrap_or(ver)
```
mdashti, line 15: Dropping everything after `-` makes pre-release targets impossible. `assemble 0.26.0-rc.1` writes `ext--0.25.9--0.26.0.sql` with `UPDATE TO '0.26.0'` inside, while the crate (and so `default_version`) says `0.26.0-rc.1`. Postgres compares extension versions as opaque strings, so the rc has no upgrade path to itself. `semver::Version` already parses and orders pre-releases (`0.26.0-rc.1 < 0.26.0`). I think only the `v` prefix needs stripping here, and `get_existing_sql_targets` would then keep the exact target string for the filename.

cargo-pgrx/src/command/migrate/mod.rs
```diff
@@ -0,0 +1,418 @@
...
+        )),
+        UnreleasedFragmentMode::AssembleEphemeral | UnreleasedFragmentMode::PackageAssemble => {
+            let crate_dir = package_manifest_path.parent().unwrap_or_else(|| Path::new("."));
+            let sql_dir = crate_dir.join("sql");
+
+            let existing_targets = version::get_existing_sql_targets(&sql_dir, extname);
+            let manifest_ver_str = package_manifest.package_version().ok();
+            let manifest_ver = manifest_ver_str.as_deref().and_then(version::parse_version_lossy);
+
+            let (prev_ver, target_ver) = if let Some(latest_target) =
```
mdashti, line 103: This assembles whenever fragments exist, even if `sql/` already has a script targeting the manifest version (fragments kept with `--preserve-fragments` for an rc, or a CI job that ran `migrate assemble` first). The same fragments then sit twice on the `ALTER EXTENSION ... UPDATE` path. In ParadeDB's upgrade job, `ext--0.25.9--999.0.0.sql` is copied, then `Assembling unreleased fragments for 999.0.0 -> 999.0.1` follows and `default_version` is bumped to `999.0.1`. Any plain `CREATE FUNCTION` in a fragment fails on the second pass. Shouldn't this skip (or at least warn) when `<ext>--*--<manifest_version>.sql` already exists?

One more limit, maybe for the docs: the ephemeral target is always `latest + 1`. Once a dev database is on `0.25.10`, every fragment that lands later still assembles as `0.25.9 -> 0.25.10`, and `ALTER EXTENSION ... UPDATE` becomes a no-op for that database.

cargo-pgrx/src/command/package.rs
```diff
@@ -86,6 +89,12 @@ impl Package {
             build_base_path(&pg_config, &package_manifest_path, &profile, self.target.as_deref())?
         };
 
+        let unreleased_mode = if self.assemble_unreleased {
```
mdashti, line 92: `install` has `--no-assemble-unreleased`, but `package` only offers `Disallow` or assemble. When fragments are preserved on purpose (rc builds), neither fits: the default errors, and `--assemble-unreleased` stacks them on top of the already assembled script. Is it worth a skip mode here too? For projects where `sql/unreleased/` is never empty on `main` (ParadeDB), the `Disallow` default also turns every `cargo pgrx package` on a dev checkout into an error. A warning may be the friendlier default.

cargo-pgrx/src/command/migrate/lint.rs
```diff
@@ -0,0 +1,516 @@
...
+}
+
+/// Catalogs all objects created or replaced in unreleased migration fragments.
+pub fn catalog_unreleased_objects(fragments: &[MigrationFragment]) -> BTreeMap<String, FragmentId> {
+    let mut catalog = BTreeMap::new();
+    for fragment in fragments {
+        for stmt in extract_statements_and_objects(&fragment.sql_content) {
+            let v = stmt.verb.to_uppercase();
+            if v.contains("CREATE") || v.contains("REPLACE") {
+                catalog.entry(stmt.signature).or_insert_with(|| fragment.id.clone());
```
mdashti, line 155: `raw_fragments` comes straight from `read_dir`, so the owner recorded here is whichever file the filesystem lists first. Two independent fragments that both `CREATE OR REPLACE` the same released function get flagged as depending on each other, and the direction changes between machines. Sorting by `(id, filename)` first (the tiebreaker the topological sort already uses), and only attributing an object to fragments that sort earlier, would make this deterministic.

cargo-pgrx/src/command/migrate/assemble.rs
```diff
@@ -0,0 +1,523 @@
...
+            return Ok(());
+        }
+
+        let retention = if self.preserve_fragments {
+            FragmentRetention::Preserve
+        } else {
+            FragmentRetention::Consume
+        };
+
+        let allow_empty = self.target_version.is_some() || self.allow_empty;
```
mdashti, line 293: Passing a target version silently implies `--allow-empty`, so the flag only matters when the version is omitted. Is that intended? The README describes `--allow-empty` as the opt-in for a stub script. A release run that forgot to cherry-pick its fragments would rather fail than ship an empty upgrade.

cargo-pgrx/src/command/migrate/assemble.rs
```diff
@@ -0,0 +1,523 @@
...
+                manifest_version: ManifestVersionStr(manifest_pkg_ver.as_deref()),
+            },
+            retention,
+            allow_empty,
+            control_update,
+        };
+
+        assemble_release_upgrade_script(params)?;
+
+        if self.json {
```
mdashti, line 317: This rebuilds the plan after the fragments were consumed, so `--json` reports `is_empty: true` and an empty `fragments` list for the run that just assembled them. Building the plan once before `assemble_release_upgrade_script` and printing it here would fix that.

cargo-pgrx/src/command/migrate/version.rs
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    if let Some(manifest_ver) = manifest_version {
+        let cleaned = clean_version_string(manifest_ver);
+        if let Some(v) = parse_version_lossy(cleaned)
+            && &v < target_version
+        {
+            return Ok(v.to_string());
+        }
+        return Ok(cleaned.to_string());
```
mdashti, line 92: If the manifest version isn't below the target, this still returns it as the predecessor. So `assemble 0.25.0` in a crate at `0.26.0` with an empty `sql/` yields `ext--0.26.0--0.25.0.sql`. The error just below seems the safer way to handle that case.

cargo-pgrx/src/command/run.rs
```diff
@@ -53,9 +53,12 @@ pub(crate) struct Run {
     pgcli: bool,
     /// Install without running
     #[clap(long)]
-    install_only: bool,
+    pub(crate) install_only: bool,
```
mdashti, line 56: Do these need `pub(crate)` now? Nothing outside this module accesses them.

cargo-pgrx/src/command/new.rs
```diff
@@ -69,6 +69,7 @@ fn create_directory_structure(root: PathBuf) -> Result<(), std::io::Error> {
     std::fs::create_dir_all(root.join("tests").join("pg_regress").join("expected"))?;
     std::fs::create_dir_all(root.join("tests").join("pg_regress").join("sql"))?;
     std::fs::create_dir_all(root.join("sql"))?;
+    std::fs::create_dir_all(root.join("sql").join("unreleased"))?;
```
mdashti, line 72: The PR description mentions a `.gitkeep`, but this only creates the directory, which git won't track. Either drop the mention or write the file.

docs/src/extension/migrations.md
```diff
@@ -0,0 +1,136 @@
...
+
+During development (`cargo pgrx install`, `cargo pgrx run`, `cargo pgrx test`), `pgrx` automatically compiles and installs unreleased migration fragments ephemerally:
+
+1. Resolves the latest released target version in `sql/` (or package version in `Cargo.toml`).
+2. Derives an ephemeral target version bump (e.g. `0.25.0` -> `0.25.1`).
+3. Assembles unreleased fragments topologically directly into the destination extension directory as `my_ext--0.25.0--0.25.1.sql`.
+4. Copies the compiled base schema to `my_ext--0.25.1.sql` so `CREATE EXTENSION` installs directly at the new version without traversing upgrade paths.
+5. Updates `default_version` in the installed `my_ext.control` file.
+
+Your Git working tree remains completely clean while testing migrations locally. To disable ephemeral assembly during install or run, pass `--no-assemble-unreleased`.
```
mdashti, line 61: What happens to a dev database once the fragments ship? It sits on the ephemeral `0.25.10`, the next checkout has no fragments, and `sql/` only has `0.25.9--0.26.0`. So there is no path from `0.25.10` anywhere, and the DB needs `DROP EXTENSION`. Fine for a scratch DB, but it's the opposite of the promise in this paragraph, so I'd mention it here.

mdashti (commented):
I reviewed this together with paradedb/paradedb#6370, which pins the `paradedb/pgrx` branch (this PR plus `cargo fmt`). I ran the commands against `pg_search`, a crate nested in a workspace, and hit a few things the tests (all single-crate, no git) don't cover. Comments are inlined below.

